PR #162, opened 2021-07-06
docs: fix webhook response format

severo:


julien-c:
arg this was actually intended, I'm deprecating the previous array-contained format (as the array always one 1 element anyways)

julien-c:
mind reverting?

severo:
hmmm, no problem for me, but... my poor https://github.com/huggingface/tensorboard-launcher will break again if you change the API response. When do you think switching to the new format?

julien-c:
this is already deployed, but I preserved backward compatibility for you + other current subscribers.

see [this commit](https://github.com/huggingface/moon-landing/pull/871)

PR #164, opened 2021-07-06
Add dataset_info and list_datasets

lhoestq:
Hi !
I just added `dataset_info()` and `list_datasets()` to the `HfAPI`
It uses the endpoint at /api/datasets

Though there are a few differences that we may need to address compared to the /api/models endpoint:
- /api/datasets doesn't return the `sha`
- /api/datasets does return `branch: main` even if the specified revision is on another branch 
- /api/datasets doesn't support filtering afaik

TODO:
- [ ] update the test on dataset_info as soon as `sha` is available, or use something else


src/huggingface_hub/hf_api.py
```diff
@@ -277,6 +341,29 @@ def model_list(self) -> List[ModelInfo]:
...
+        self,
+        full: Optional[bool] = None,
+    ) -> List[DatasetInfo]:
+        """
+        Get the public list of all the datasets on huggingface.co
+
+        Args:
+            full (:obj:`bool`, `optional`):
+                Whether to fetch all dataset data, including the `lastModified` and the `card_data`.
+                This is set to `True` by default when using a filter.
```
LysandreJik, line 354:
```suggestion
```

LysandreJik (approved):
Great, thank you @lhoestq!

There are quite a few features missing on the endpoint compared to the list models utility function: `filter`, that you mention, but also `sort`, `direction`, and `limit`.

I think the `filter` one is especially important as it allows super easy management directly in `hfh`.

No opposition to merge this now and revisit once the endpoint has evolved.

docs/hub/endpoints.md
```diff
@@ -13,7 +13,9 @@ We have open endpoints that you can use to retrieve information from the Hub as
...
+| /api/datasets/{repo_id} <br/> /api/datasets/{repo_id}/revision/{revision} <br/>  GET | Get all information for a specific dataset.  <br/>- `full`: Whether to fetch most dataset data, such as all tags, the files, etc.                                                                                                                                                                                                                                                                                                                                             | `dataset_info(repo_id, revision)`        | ```headers = { "authorization" :  "Bearer $token", "full" : "full"  }```                                                                                     |
```
julien-c, line 18: i would group those two and not interspede them like this

julien-c (commented):
Tagging @elishowk (and possibly @SBrandeis) for visibility, if they want to implement `filter` on the server side.

> I think the filter one is especially important as it allows super easy management directly in hfh.

Note that unlike models, the filtering/etc for datasets will be done in memory anyways on the server-side, so it's not super different from just implementing those features client-side (which you could do in this PR)

Just the bandwidth/latency of the http request will be different, which doesn't seem like a huge deal until we have 10_000 datasets.

If however we want to implement this server-side for consistency, I'm fine with that as well

lhoestq:
Once `sha` is available in the response of the API, this should fix the CI

lhoestq:
Merging this one :)
I'll open a new PR to include the filter/sort/direction/limit parameters to the `list_datasets()` function (some of these params are implemented in PRs that are still under reviews on the backend)

PR #166, opened 2021-07-07
Widget image captioning

mishig25:
Widget side of Image Captioning pipeline

input/request: identical to ImageClassificationWidget
`const requestBody = { file }; // uploaded image`

output/response: identical to TextGenerationWidget
`[ { "generated_text": "..." } ]`

Screenshots:
Case 1: Initial state             |  Case 2: Success 
:-------------------------:|:-------------------------:
Screenshot from 2021-06-20 08-13-40 | Screenshot from 2021-06-20 08-19-19

Case 3: Fail (wrong file type)  
:-------------------------:
Screenshot from 2021-06-20 08-21-28

Creating this PR as @osanseviero mentioned that there will be image captioning models with the Flax event.


widgets/src/lib/InferenceWidget/widgets/ImageToTextWidget/ImageToTextWidget.svelte
```diff
@@ -0,0 +1,99 @@
...
+			error = res.error;
+		}
+	}
+
+	function parseOutput(body: unknown): string {
+		if (Array.isArray(body) && body.length) {
+			const firstEntry = body[0];
+			return (
+				firstEntry["generated_text"] ?? // text-generation + text2text-generation + image-to-text
+				firstEntry["translation_text"] ?? // translation
```
Pierrci, line 74: Since we know here that we're in the case of `image-to-text`, why not just check for `firstEntry["generated_text"]`? I think that the behavior in `TextGenerationWidget` is because it can handle different tasks, which is not the case here? (cc @beurkinger, I might be wrong!)

mishig25: done 👍

adrinjalali:
Is this getting in? Last update was in July 2021.

osanseviero:
Closing this as it's outdated/wrong repo now

PR #170, opened 2021-07-08
scikit rebased

Narsil:


Pierrci:
It seems it's missing the `.gitmodules` file at the root for the `sentence-similarity` submodule located in `api-inference-similarity/src` @Narsil (getting errors on `git pull` with submodules enabled)

julien-c:
i think it's not meant to be a submodule, just an extraneous file/pointer. I'll let @Narsil confirm

Pierrci:
Yes indeed, but in any case, there is a problem that needs to be fixed I think (same errors w/ the CLI, it's not specific to gitkraken)

Screen Shot 2021-07-09 at 10 22 52 AM

julien-c:
or just delete this on master @Pierrci – i don't think it can break anything (as it's not even _really_ there)

Narsil:
MIssed the notifications on this one, yes we can remove !

PR #171, opened 2021-07-08
Add non-blocking git_push

### LINKED ISSUE #169: Non-blocking `git_push`
On local machine, the push can take a while, but usually there is no need to wait for it to complete, it would be great to keep training while the `git-push` uploads model in background. 
This can be done by replacing `subprocess.run` with `subprocess.Popen`.
 https://github.com/huggingface/huggingface_hub/blob/0d185de837846859668a0ee5b1aa5ef71a4f9ac4/src/huggingface_hub/repository.py#L431

cccntu:
Fixes #169 

@LysandreJik

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -428,24 +428,33 @@ class Repository:
             else:
                 raise EnvironmentError(exc.stdout)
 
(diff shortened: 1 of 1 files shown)


LysandreJik:
This was fixed by #315!

PR #176, opened 2021-07-08
Fixing keras outputs  (widgets was ignoring because of type mismatch, now testing for it)

Narsil:


osanseviero (approved):
Looks good, let's also change https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/docker_images/timm/tests/test_api_image_classification.py so this does not slip away.

osanseviero:
And thanks for sending the PR :)

PR #177, opened 2021-07-08
Auto track with LFS

LysandreJik:
This PR adds auto-tracking for LFS files. It automatically tracks files that are larger than 10MB and that are not currently being tracked.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -406,6 +406,9 @@ class Repository:
         except subprocess.CalledProcessError as exc:
             raise EnvironmentError(exc.stderr)
 
+        if len(git_status) == 0:
+            return []
+
         # Receives a status like the following
         #  D .gitignore
         #  D new_file.json
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -621,12 +621,10 @@ class Repository:
         return self.git_push()
 
     @contextmanager
-    def commit(
-        self,
-        commit_message: str,
-    ):
+    def commit(self, commit_message: str, track_large_files: bool = True):
         """
-        Context manager utility to handle committing to a repository.
+        Context manager utility to handle committing to a repository. This automatically tracks large files (>10Mb)
+        with git-lfs. Set the `track_large_files` argument to `False` if you wish to ignore that behavior.
 
         Examples:
 
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -15,7 +15,7 @@ from .lfs import LFS_MULTIPART_UPLOAD_COMMAND
 logger = logging.getLogger(__name__)
 
 
-def is_git_repo(folder: Union[str, Path]):
+def is_git_repo(folder: Union[str, Path]) -> bool:
     """
     Check if the folder is the root of a git repository
     """
@@ -26,7 +26,7 @@ def is_git_repo(folder: Union[str, Path]):
     return folder_exists and git_branch.returncode == 0
 
 
-def is_local_clone(folder: Union[str, Path], remote_url: str):
+def is_local_clone(folder: Union[str, Path], remote_url: str) -> bool:
     """
     Check if the folder is the a local clone of the remote_url
     """
@@ -48,11 +48,12 @@ def is_local_clone(folder: Union[str, Path], remote_url: str):
     return remote_url in remotes
 
 
-def is_tracked_with_lfs(filename: str):
+def is_tracked_with_lfs(filename: Union[str, Path]) -> bool:
     """
     Check if the file passed is tracked with git-lfs.
     """
     folder = Path(filename).parent
+    filename = Path(filename).name
 
     try:
         p = subprocess.run(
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/repository.py
```diff
@@ -393,10 +428,40 @@ def git_pull(self, rebase: Optional[bool] = False):
...
             raise EnvironmentError(exc.stderr)
 
-    def git_add(self, pattern="."):
+    def auto_track_large_files(self, pattern="."):
+        """
+        Automatically track large files with git-lfs
+        """
+        try:
+            p = subprocess.run(
+                ["git", "ls-files", "-mo", pattern],
```
julien-c, line 437: in my tests this also lists files that are gitignored

src/huggingface_hub/repository.py
```diff
@@ -48,6 +48,41 @@ def is_local_clone(folder: Union[str, Path], remote_url: str):
...
+        return False
+
+    found_lfs_tag = {"diff": False, "merge": False, "filter": False}
+
+    for attribute in attributes.split("\n"):
+        for tag in found_lfs_tag.keys():
+            if tag in attribute and "lfs" in attribute:
+                found_lfs_tag[tag] = True
+
+    return all(found_lfs_tag.values())
```
julien-c, line 83: if youre going through the hassle of parsing the output of `git check-attr` you can also just read `.gitattributes` and select all lines that end with `filter=lfs diff=lfs merge=lfs -text`

src/huggingface_hub/repository.py
julien-c: not sure if there's a workaround

src/huggingface_hub/repository.py
```diff
@@ -393,10 +428,40 @@ def git_pull(self, rebase: Optional[bool] = False):
...
+            )
+            files_to_be_staged = p.stdout.strip().split("\n")
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        for filename in files_to_be_staged:
+            path_to_file = os.path.join(os.getcwd(), self.local_dir, filename)
+            size_in_mb = os.path.getsize(path_to_file) / (1024 * 1024)
+
+            if size_in_mb >= 10 and not is_tracked_with_lfs(path_to_file):
```
julien-c, line 452: would be better (though way harder) to do a first pass to collect all files, then factor them into file extensions patterns... Might be overkill though. @Pierrci ?

julien-c (commented):
wow that was fast :)

src/huggingface_hub/repository.py
LysandreJik: That sounds a bit more complicated though, because I'd have to parse all wildcards to see if it contains the filename ... and then I'd have to do the same parsing; that sounds more complex than what's implemented here

julien-c: oh yes ok

src/huggingface_hub/repository.py
LysandreJik: Oh, exceptional catch! Actually, this is somewhat of a workaround:
```py
git ls-files -mo -X .gitignore 
```
But it doesn't look for nested `.gitignore`s.

However I'm not sure if this is actually important, as these files won't get `git add`ed so they won't end up staged?

julien-c: probably not very important (other than increasing the size of the .gitattributes file but i guess that's not a huge deal – unless it gets to > 10MB)

LysandreJik: If it's important I'm sure I can find a workaround - let me know if it's high priority

julien-c: no I don't think it's a big deal at all

src/huggingface_hub/repository.py
Pierrci: I feel like it would be overkill yes, who really looks at the content of a `.gitattributes` anyway? 😅

julien-c: just me probably

julien-c: but for instance if you rename a file (w/o pattern) it won't be lfs-tracked anymore

Pierrci: yes, maybe remove obsolete files from the list when renamed/removed @LysandreJik? 😅 (even though I feel like 99.9999% of the time, keeping them will never present a problem)

(in case of rename, I think the current code will correctly catch the new name and add it)

LysandreJik:
I should have applied your comments @julien-c, @Pierrci, thank you for the review!

I'm not super super satisfied of the API that removes items from the `.gitattributes` file, but I'm having a hard time coming up with a better one. Right now it works only if the file removal is staged. If users don't do `git add <pattern>` beforehand, then I'm not detecting that the file is removed, and I'm not removing it from the `.gitattributes`.

Let me know if this API is what you had in mind or if you were thinking of something different. Feel free to take a look at the tests, too, they each cover specific use-cases and are the best way to see how it works in practice.

julien-c (commented):
I will play a little bit more when i get a chance, but it's not blocking – looks great from a cursory glance!

thomwolf:
Would be happy to mention it for the book if you think the API is rather stable :)

Current use case is uploading large dataset files on the hub (>250MB .json.gz files so untracked by default)

LysandreJik:
Will do a final pass on the API today and merge!

LysandreJik:
Played a bit more with it and I'm satisfied with the API. Internals may evolve, but I'd say you can go ahead and mention it in the book @thomwolf, as the user-facing API is likely to be set!

PR #184, opened 2021-07-09
Fixing audio-to-audio validation.

Narsil:
`src` does NOT exist on the incoming payload.

`content-type`, `blob` and `label` are the keys.
`src` is only added later as a convenience to display.

Example of usable model: https://huggingface.co/julien-c/DPRNNTasNet-ks16_WHAM_sepclean

### DIFF
--- a/widgets/src/lib/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
+++ b/widgets/src/lib/InferenceWidget/widgets/AudioToAudioWidget/AudioToAudioWidget.svelte
@@ -34,7 +34,7 @@
 		blob: string;
 		label: string;
 		src?: string;
-        content-type: string;
+        "content-type": string;
 	}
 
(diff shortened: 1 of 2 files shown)


osanseviero:
Could you please also add the Asteroid model to https://github.com/huggingface/huggingface_hub/blob/main/widgets/src/routes/index.svelte so we can confirm it works properly? 

I tried the audio to audio models launched by GitHub actions 
https://60e85b3703eadf1975749dcb--huggingface-widgets.netlify.app/

* speechbrain/mtl-mimic-voicebank works fine
* speechbrain/sepformer-wham gets stuck.

osanseviero:
Thanks! It works and that source separation model is quite good actually

Narsil:
I will wait for @mishig25 eyes before merging !

mishig25 (approved):
lgtm

osanseviero (approved):

PR #202, opened 2021-07-16
Packaging & publishing to npm

mishig25:
Find details on [huggingface/moon-landing/872](https://github.com/huggingface/moon-landing/pull/872)


Pierrci:
Did you close this accidentally @mishig25 or was it intended?

mishig25:
@Pierrci  Yeah it was intended. I think changes I've made to `tsconfig` is causing `svelte-check` to fail

mishig25:
@Pierrci Please let me know if you think this approach is acceptable.
#### Why we can't use a package generated by `svelte-kit package`?
`svelte-kit package` expects the consumer (in this case, `moon-landing`) to be using `svelte-kit`, which directly consumes third party component sources (rather than bundled files).
However, on `moon-landing`, we don't use `svelte-kit`. There are 2 ways a component gets compiled on `moon-landing`: `build-svelte.mjs` on `server` and [rollup](https://github.com/huggingface/moon-landing/blob/master/front/rollup.config.js) on `front` (`rollup` can consume svelte component sources directly, so we can use svelte-kit package for the `front` (not `server`) see [here](https://github.com/sveltejs/rollup-plugin-svelte#pkgsvelte)).
#### As a intermediary solution
I've copied [moon-landing/server/build-svelte.mjs](https://github.com/huggingface/moon-landing/blob/master/server/build-svelte.mjs) into [hub/widgets/build.js](https://github.com/huggingface/huggingface_hub/blob/f143c737981f68e218ee46b50af1eee08b5213fa/widgets/build.js). In `hub/widgets/package` folder: 
1. a component source code is copied here from hub/widgets/package` (with `svelte-kit package` for `moon-landing/front`) 
2. a component gets compiled here with [hub/widgets/build.js](https://github.com/huggingface/moon-landing/blob/master/server/build-svelte.mjs) (for `moon-landing/server`) 

And [npm package](https://www.npmjs.com/package/hf-widgets) is published.

I think this approach is quite "hacky". As we add mode svelte components, we should definitely switch to what @beurkinger suggested.

julien-c:
nit, but the npm package name (if we go that way) would be better named `huggingface-widgets` for consistency w/ other library etc. naming

Pierrci:
I finally had a look at it, and IMO it looks very promising @mishig25 :)

> I think changes I've made to tsconfig is causing svelte-check to fail

I bumped the version of `svelte-check` to the last one locally (2.2.3) and it seems to be working fine, it was probably linked to some bug :)

> I think this approach is quite "hacky". As we add more svelte components, we should definitely switch to what @beurkinger suggested.

I agree the approach can be considered as "hacky", but I'm not sure why adding more components would make it worse than it is now?

More generally, I think we could refactor a bit the js structure in this repo w/ a more generic `js` (or similar) top-level directory instead of `widgets`, move `interfaces` in `lib` and publish a more generic `huggingface`(`-hub`?) package exposing both the widgets through `InferenceWidget` and what would be in `lib/interfaces`. (+ `ModelPipelineIcon` and `ModelPipelineTag` since we also use them in moon-landing). wdyt @julien-c?

This way we can import all the js code we need from `huggingface_hub` in a unified way (and not using the npm package in some places and weird paths pointing to the submodule in others).

mishig25:
I made an ambiguous statement above. I meant: we should switch to rollup if/when we add third-aprty svelte component libraries.
> but I'm not sure why adding more components would make it worse than it is now?

julien-c:
> More generally, I think we could refactor a bit the js structure in this repo w/ a more generic js (or similar) top-level directory instead of widgets, move interfaces in lib and publish a more generic huggingface(-hub?) package exposing both the widgets through InferenceWidget and what would be in lib/interfaces. (+ ModelPipelineIcon and ModelPipelineTag since we also use them in moon-landing). wdyt @julien-c?

No strong opinion (I think git-submodules are fine for non-Svelte files), I'll leave that up to you guys.

Pierrci:
> No strong opinion (I think git-submodules are fine for non-Svelte files), I'll leave that up to you guys.

Yes, we can leave the hub doc outside of the package 😄. But moving `interfaces` would also simplify both their use inside the widgets and the package's build (@mishig25 had to add a specific build step because of their current path). What do you think of the idea @mishig25?

@LysandreJik, what do you think would be the best way to regroup all the js code (`widgets` + `interfaces` - if we go this way) without interfering too much with all the pythonic stuff? something inside `src`? next to it?

julien-c:
What about `interfaces/` and nested inside `interfaces/widgets/`?

LysandreJik:
I'd rather we don't move it inside `src` as that's the current location of the python source code and is what we use in the [`setup.py` to create the wheels](https://github.com/huggingface/huggingface_hub/blob/main/setup.py#L55).

Moving widgets under `interfaces/widgets` is ok for me. Moving both folders under an unambiguous folder `frontend` or `hub` is also good for me if you'd like to keep the same structure.

mishig25:
> Moving both folders under an unambiguous folder `frontend`
sounds great to me.  
In that case, @Pierrci, could we just add `huggingface_hub/frontend` dir to tsconfig files & `build-svelte.mjs` script and not use rollup at the moment? And use tsconfig#paths so that we don't use really long import statements like `import huggingface_hub/frontend/widgets/src/lib/InferenceWidget`?

julien-c:
I'll reiterate that I'd prefer to nest widgets under `interfaces/` given that we already have a `interfaces/` directory (and its contents does not need to be moved unless I'm mistaken)

mishig25:
got it 👍
then, I'll move `widgets` inside `interfaces` as `interfaces/widgets/`

Pierrci:
What benefits do you see @julien-c in keeping `interfaces` and move `widgets` in it @julien-c?

Moving `interface` inside `widgets/lib` would simplify a lot the build setup (from my tests we can remove the second half of the steps - maybe more), but the opposite doesn't change anything, except a few relative paths that lose a `/interface` in them...

julien-c:
Ok ok, I haven't tested the whole tooling so let's go with what @Pierrci preconizes. 🙏

mishig25:
@Pierrci in terms for publishing the widgets, what should be the next step:
Option A: use the build script plus svelte-kit package method suggested by this PR
Option B: something else, as referred:
> (from my tests we can remove the second half of the steps - maybe more

Pierrci:
@mishig25 option A, but following #227 we can simplify what's done in this PR and remove a few steps from the `build.js` file since you'll find that moving `interfaces` inside `lib/` makes the `svelte-kit package` command runs more as one would expect (files aren't outputted in 2 different dirs inside `package/` anymore for example).

Ideally we would end up exposing `InferenceWidget`, `ModelPipelineTag`, `ModelPipelineTag` and `WidgetHeader` from the "root" of the package, and everything that's in `interfaces` from `huggingface-widgets/interfaces`? You'll probably need to create one root `index.ts` and another one in `interfaces` (they talk about creating an `index.ts` or `index.svelte` in https://kit.svelte.dev/docs#packaging, but it's a pretty common way of packaging anyway).

widgets/package.json
```diff
@@ -1,10 +1,44 @@
...
+		"./WidgetHeader.svelte": "./package/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte",
+		"./WidgetHeader.svelte.js": "./package/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte.js",
+		"./interfaces/DefaultWidget": "./package/interfaces/DefaultWidget.js",
+		"./interfaces/DefaultWidget.js": "./package/interfaces/DefaultWidget.js",
+		"./interfaces/Language": "./package/interfaces/Language.js",
+		"./interfaces/Language.js": "./package/interfaces/Language.js",
+		"./interfaces/Libraries": "./package/interfaces/Libraries.js",
+		"./interfaces/Libraries.js": "./package/interfaces/Libraries.js",
+		"./interfaces/Types": "./package/interfaces/Types.js",
+		"./interfaces/Types.js": "./package/interfaces/Types.js"
```
mishig25, line 24: I'm exporting `components` (InferenceWidget, ModelPipelineTag, ...) & `interfaces` (Types, Libraries, ...) in  `package.json` rather than creating `index.ts` files. @Pierrci please let me know what you think about this point

widgets/build.js
```diff
@@ -0,0 +1,164 @@
...
+	for (const fname of filenames) {
+		await compileFile(fname, srcDir, distDir);
+	}
+}
+
+/*
+Apply path alises defined in tsconfig.json
+Currently, only one path alias is defined: $lib -> src/lib
+*/
+async function applyPathAliases(distDir) {
```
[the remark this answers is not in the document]
Pierrci: It seems the sveltekit issue just got given a p1 priority, so maybe we can stick with this workaround for now and hope for it to be solved soon so we can get rid of that.

mishig25:
@Pierrci
Since [internal](https://github.com/huggingface/moon-landing/pull/1160) was merged, I've advanced this PR.
To remind:

1. This PR published widgets as a npm package [https://www.npmjs.com/package/huggingface-widgets](https://www.npmjs.com/package/huggingface-widgets)
2. Removed the `$lib` usage since [https://github.com/sveltejs/kit/issues/1950](https://github.com/sveltejs/kit/issues/1950) is still open. (You mentioned about it [here](https://github.com/huggingface/huggingface_hub/pull/202#discussion_r681025956))
3. Files that are being export are:

```
".": "./InferenceWidget/InferenceWidget.svelte",
"./WidgetHeader.svelte": "./InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte",
"./ModelPipelineIcon.svelte": "./ModelPipelineIcon/ModelPipelineIcon.svelte",
"./ModelPipelineTag.svelte": "./ModelPipelineTag/ModelPipelineTag.svelte",
"./inferenceSnippets": "./inferenceSnippets/index.js",
"./inferenceSnippets/inputs": "./inferenceSnippets/inputs.js",
"./interfaces/DefaultWidget": "./interfaces/DefaultWidget.js",
"./interfaces/Language": "./interfaces/Language.js",
"./interfaces/Libraries": "./interfaces/Libraries.js",
"./interfaces/Types": "./interfaces/Types.js"

```

And it gets consumed as:

```
import InferenceWidget from "huggingface-widgets";
import ModelPipelineTag from "huggingface-widgets/ModelPipelineTag.svelte";
import { Language } from "huggingface-widgets/interfaces/Language";
import type { PipelineType } from "huggingface-widgets/interfaces/Types";
...

```

4. Checkout its companion PR internally [here](https://github.com/huggingface/moon-landing/pull/1172)

widgets/package.json
```diff
@@ -15,18 +28,37 @@
...
+		".": "./InferenceWidget/InferenceWidget.svelte",
+		"./WidgetHeader.svelte": "./InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte",
+		"./ModelPipelineIcon.svelte": "./ModelPipelineIcon/ModelPipelineIcon.svelte",
+		"./ModelPipelineTag.svelte": "./ModelPipelineTag/ModelPipelineTag.svelte",
+		"./inferenceSnippets": "./inferenceSnippets/index.js",
+		"./inferenceSnippets/inputs": "./inferenceSnippets/inputs.js",
+		"./interfaces/DefaultWidget": "./interfaces/DefaultWidget.js",
+		"./interfaces/Language": "./interfaces/Language.js",
+		"./interfaces/Libraries": "./interfaces/Libraries.js",
+		"./interfaces/Types": "./interfaces/Types.js"
```
Pierrci, line 61: Can we have an `interfaces/index.js` that re-exports those, so that we can do `import { DefaultWidget, Language, ... } from "huggingface-widgets/interfaces"`?

widgets/package.json
```diff
@@ -1,10 +1,23 @@
 {
 	"name": "huggingface-widgets",
-	"version": "0.0.1",
+	"version": "0.15.0",
+	"type": "module",
+	"main": "./InferenceWidget/InferenceWidget.svelte",
```
Pierrci, line 5: `main` is supposed to be something in plain js I think (that's why you can add a specific `svelte` entry for the "non-compiled" version), but for now, we can leave it as is - and add a "compiled" version later if there is a demand for it.

widgets/package.json
```diff
@@ -1,10 +1,23 @@
 {
 	"name": "huggingface-widgets",
-	"version": "0.0.1",
+	"version": "0.15.0",
```
Pierrci, line 3: What about bumping the version to `1.0.0` when we release for good? :)

Pierrci (commented):
We're almost there! :)

widgets/package.json
mishig25: Does it make sense to do so? because, the use case has always been:
`import { x, y, z } from Interfaces or Language`,
not `import Interfaces`

Pierrci: Yes, you're right actually! I think I'm feeling the effects of the jetlag x)

Issue #186, opened 2021-07-12
Better error message when cloning without token


LysandreJik:
When cloning a repository without the `use_auth_token` parameter, the error is unintuitive:

```py
from huggingface_hub import Repository
repo = Repository("hubert-base-ls960", clone_from="facebook/hubert-base-ls960")
```
```
OSError: fatal: repository 'facebook/hubert-base-ls960' does not exist
```

When the `use_auth_token` parameter is omitted, the URL isn't correctly reconstructed.

PR #203, opened 2021-07-16
No need for token to understand HF urls

LysandreJik:
This patches an issue where the token was necessary to understand links to the Hugging Face Hub, which was involuntary behavior.

Fixes https://github.com/huggingface/huggingface_hub/issues/186

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -225,12 +225,6 @@ class Repository:
 
         repo_type, namespace, repo_id = repo_type_and_id_from_hf_id(repo_url)
 
-        if namespace is None and token is None:
-            raise ValueError(
-                f"Cannot clone from URL {repo_url}. If this repository lives under your namespace, make sure to "
-                "pass your authentication token with the `use_auth_token` parameter."
-            )
-
         if repo_type is not None:
             self.repo_type = repo_type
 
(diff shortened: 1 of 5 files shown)


src/huggingface_hub/repository.py
```diff
@@ -265,7 +275,7 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
 
             # checks if repository is initialized in a empty repository or in one with files
             if len(os.listdir(self.local_dir)) == 0:
-                logger.debug(f"Cloning {clean_repo_url} into local empty directory.")
+                logger.warning(f"Cloning {clean_repo_url} into local empty directory.")
```
LysandreJik, line 278: I'm also taking the opportunity to elevate this log's importance, as without it the user might wonder why the Repo instantiation is taking so long. 

Can ignore this in favor of a slight rework of logs as proposed by @thomwolf

src/huggingface_hub/repository.py
```diff
@@ -223,23 +223,30 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
         token = use_auth_token if use_auth_token is not None else self.huggingface_token
         api = HfApi()
 
+        repo_type, namespace, repo_id = repo_type_and_id_from_hf_id(repo_url)
+
+        if namespace is None and token is None:
```
LysandreJik, line 228: This isn't robust to the repositories that live at the root, so I will think a bit more about this.

LysandreJik:
This PR removes the ability to clone from a repository using only its model ID, guessing its namespace in return. This is unfortunately due to the models that live at the root (`bert-base-cased`, `gpt2`).

julien-c:
I didn't read thoroughly but I don't think we care about `Repository` being compatible with top-level (legacy) models, no?

LysandreJik:
Merging with offline approval: this is the most rigorous way to handle things.

Issue #187, opened 2021-07-13
Repository `use_auth_token` should default to `True` when in a HF-hub specific workflow


LysandreJik:
The `Repository` class is thought out as a wrapper to git/git-lfs, and isn't specific to the Hugging Face Hub. It's a useful helper to manage offline repositories.

However, the recent PRs https://github.com/huggingface/huggingface_hub/pull/132, https://github.com/huggingface/huggingface_hub/pull/143, https://github.com/huggingface/huggingface_hub/pull/150 and https://github.com/huggingface/huggingface_hub/pull/151 allow cloning repositories from the hub using only `<namespace>/<model_id>` or even `<model_id>` as identifiers, which is hf-hub specific.

@thomwolf rightfully asks why the `use_auth_token` parameter isn't `True` by default, and I tend to agree that for hf-specific workflows this should be the default as it would prevent a lot of pain points when cloning a repository without authentication.

julien-c:
I have no objection to setting `use_auth_token` to True if it helps onboard users not very familiar with git.

Just a note that git authentication is usually handled by the user themselves on the system level, using e.g. a credential helper. For instance when cloning from, or pushing to GitHub, I suspect most people don't have to re-type their password or token each time.

We might want to add some doc about how to do this, or to expose helpers that help do this.

It might be semi-orthogonal to this issue though.

PR #204, opened 2021-07-16
Set use_auth_token to True by default

LysandreJik:
This sets the `use_auth_token` parameter to `True` by default for the `Repository` instantiation.

If users are not logged in, then it will behave similarly to the previous default, or if the `use_auth_token` parameter was set to `False`: the `HfFolder().get_token()` returns `None`, and the `Repository` behaves as unauthenticated.

For logged-in users, however, this should dramatically reduce the user errors linked to cloning a repository without specifying the authentication token, in turn being unable to push to that repository with an unhelpful error.

Fix https://github.com/huggingface/huggingface_hub/issues/187

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -120,7 +120,7 @@ class Repository:
         local_dir: str,
         clone_from: Optional[str] = None,
         repo_type: Optional[str] = None,
-        use_auth_token: Union[bool, str, None] = None,
+        use_auth_token: Union[bool, str] = True,
         git_user: Optional[str] = None,
         git_email: Optional[str] = None,
     ):


julien-c:
> For logged-in users, however, this should dramatically reduce the user errors linked to cloning a repository without specifying the authentication token, in turn being unable to push to that repository with an unhelpful error.

I think it 'd be an important fix to provide a helpful error in that case, no?

LysandreJik:
Yes, it's this error: https://github.com/huggingface/transformers/issues/12572

julien-c:
we can catch it in `Repository` and provide a more understandable error that they need to authenticate?

julien-c (approved):

LysandreJik:
Agreed with @julien-c to merge this while keeping in mind that this isn't ideal either, as it saves the credentials in the remotes of the repository. I'm working on a rework involving `git credential` in order to not rely on the HF token anymore to pull/push from/to the hub.

PR #208, opened 2021-07-18
Adding  protein_folding widget.

Narsil:
output

The goal is to add a few proteins already ran by @Rocketknight1 , and for unknown proteins, instead of running any inference, we would redirect to a voting website for next proteins to be run.

https://poll.ly/#/PwjxBxj2 for instance.

### DIFF
--- a/widgets/package.json
+++ b/widgets/package.json
@@ -28,5 +28,8 @@
 		"tslib": "^2.0.0",
 		"typescript": "^4.0.0"
 	},
-	"type": "module"
+	"type": "module",
+	"dependencies": {
+		"ngl": "^0.10.4"
+	}
 }
--- a/widgets/src/lib/InferenceWidget/shared/WidgetProteinViewer/WidgetProteinViewer.svelte
+++ b/widgets/src/lib/InferenceWidget/shared/WidgetProteinViewer/WidgetProteinViewer.svelte
@@ -1,0 +1,34 @@
+<script>
+	import { onMount } from "svelte";
+	export let autoplay = false;
+	export let classNames = "";
+	export let controls = true;
+	export let label = "";
+	export let src: string;
+	export let pdb_src: string;
+	// create a `stage` object
+
+	onMount(async () => {
+		const NGL = await import("ngl");
+		var stage = new NGL.Stage("viewport");
+		// load a PDB structure and consume the returned `Promise`
+		pdb_src = src;
+		stage.loadFile(src).then(function (component) {
+			// add a "cartoon" representation to the structure component
+			component.addRepresentation("cartoon");
+			// provide a "good" view of the structure
+			component.autoView();
+		});
+	});
+</script>
+
+<div class={classNames}>
+	{#if $$slots.default}
+		<slot />
+	{:else if label.length}
+		<div class="mb-1.5 text-sm text-gray-500 truncate">{label}</div>
+	{/if}
+	<!-- svelte-ignore a11y-media-has-caption -->
+	<div id="viewport" style="height:400px;width:400px" />
+	<a href={pdb_src} class="btn-widget">Download pdb file</a>
+</div>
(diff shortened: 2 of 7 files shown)


widgets/package.json
```diff
@@ -28,5 +28,8 @@
 		"tslib": "^2.0.0",
 		"typescript": "^4.0.0"
 	},
-	"type": "module"
+	"type": "module",
+	"dependencies": {
+		"ngl": "^0.10.4"
```
julien-c, line 33: what's the bundle size when adding this client side?

widgets/src/lib/InferenceWidget/shared/WidgetProteinViewer/WidgetProteinViewer.svelte
```diff
@@ -0,0 +1,32 @@
...
+	export let src: string;
+	export let pdb_src: string = "";
+	// create a `stage` object
+
+	onMount(async () => {
+		const NGL = await import("ngl");
+		var stage = new NGL.Stage("viewport");
+		// load a PDB structure and consume the returned `Promise`
+		pdb_src = src;
+		stage.loadFile(src).then(function (component) {
```
julien-c, line 14: you can probably flatten this `then` into an await

widgets/src/lib/InferenceWidget/widgets/ProteinFoldingWidget/ProteinFoldingWidget.svelte
```diff
@@ -0,0 +1,144 @@
...
+				onClick={() => {
+					getOutput();
+				}}
+			/>
+		</form>
+	</svelte:fragment>
+	<svelte:fragment slot="bottom">
+		<WidgetProteinViewer
+			classNames="mt-4"
+			src="http://localhost:8000/ranked_0.pdb"
```
julien-c, line 138: src seems bogus

widgets/package.json
julien-c: according to their repo, >1MB so i don't think we would want to embed that on every page

julien-c (commented):
I think this might fit better inside a Space, no? (I think using Gradio has a way to inject completely custom HTML/JS)

widgets/src/lib/InferenceWidget/widgets/ProteinFoldingWidget/ProteinFoldingWidget.svelte
Narsil: It's a way to test locally without having to use the real API.

widgets/package.json
Narsil: How can we make it separate ?

Isn't the svelte compiler able to make such decisions already based on `import` statements ?

Narsil:
> I think this might fit better inside a Space, no? (I think using Gradio has a way to inject completely custom HTML/JS)

Hm I don't think it's simple to add JS within GRADIO, as I have no way to know the names of the html tags.

I could dump raw html with gradio (I mean everything), but I am unsure of the benefits there. hosting simple html page is probably better (lazy JS loading etc..)

Here the benefit is that we can reuse the radio buttons to suggest existing molecules to user (which we would put in cache) so that we could keep the experience of instant results.

For unknown molecules the idea is to refer them to another place (voting website for instance) where maybe they could share the most interesting ones we should run (which we would in TURN add to the cache).

Narsil:
Dropping this (stale)

julien-c:
Note @Narsil that we now have a way for power users to just drop a `static` webpage in a Space now (and HF users can drop any JS they want inside it)

cc @cbensimon and @osanseviero

Narsil:
Good to know, I'll drop it anyway, because of the obvious hard work to make everything work in inference + the open sourcing of many molecules from deepmind:

https://www.theguardian.com/technology/2021/jul/22/ai-firm-deepmind-puts-database-building-blocks-life-human-protein-structure-free-online

Still could be interesting for demo purposes when more appropriate. At least prep work exists.

julien-c:
yes my note was more general, for future similar integrations

PR #209, opened 2021-07-19
Throw catch error when output is invalid

### LINKED ISSUE #182: Give user a feedback when API returns something unepected with 200
When API returns something unepected with 200, currently no feedback is given to a user (i.e. no error message).
Necessary updates: when API returns something unepected with 200,
* set `error` message with "Invalid/unexpected ..."
* set `outputJson` with the API output

mishig25:
Closes #182 

Throw an error when: [response.ok](https://github.com/huggingface/huggingface_hub/blob/730f7e92e338fba489eb674336d48e2c09fbe525/widgets/src/lib/InferenceWidget/shared/helpers.ts#L111) is `true` (response_status_code == 200) && `isValidOutput` is `false` ([example](https://github.com/huggingface/huggingface_hub/blob/730f7e92e338fba489eb674336d48e2c09fbe525/widgets/src/lib/InferenceWidget/widgets/ImageClassificationWidget/ImageClassificationWidget.svelte#L79)). Catch an error and display error message in WidgetInfo.

This error message would be particularly useful when implement a adding a new pipeline or making a change in Inference API

Screenshot 2021-07-19 at 12 00 16


julien-c:
This error message is not super clear for the end user: shouldn't we prepend the error message with something like

```
API Implementation Error: Invalid output: ...
```

Pierrci:
@beurkinger can you review please? you are the most knowledgeable person when it comes to widgets I think :)

julien-c (approved):
lgtm

Issue #168, opened 2021-07-08
Add `--filename` support for `git lfs track`


cccntu:
We need this to reliably track individual file with their file names.

Here in the doc:
https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-track.1.ronn

> `git lfs track --filename "project [1].psd"`

relevant issue #138 (identify large files and track by their file names.)

julien-c:
Should be fairly easy to add (though the number of cases where it's needed, vs. a pattern, is fairly small – need to have a `{*, ?, [, ]}` in your filename)

cccntu:
Yes, I just think it's safer when I read the man page.

PR #212, opened 2021-07-19
Add `filename` option to `lfs_track`

LysandreJik:
Adds the `filename` parameter which can be used to set all the `patterns` passed to `lfs_track` as filenames rather than globs.

Closes https://github.com/huggingface/huggingface_hub/issues/168

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -444,7 +444,7 @@ class Repository:
 
         Setting the `filename` argument to `True` will treat the arguments as literal filenames,
         not as patterns. Any special glob characters in the filename will be escaped when
-        writing the .gitattributes file.
+        writing to the .gitattributes file.
         """
         if isinstance(patterns, str):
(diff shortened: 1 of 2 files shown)


julien-c (approved):

src/huggingface_hub/repository.py
```diff
@@ -438,16 +438,21 @@ def list_deleted_files(self) -> List[str]:
...
         return deleted_files
 
-    def lfs_track(self, patterns: Union[str, List[str]]):
+    def lfs_track(self, patterns: Union[str, List[str]], filename: bool = False):
         """
         Tell git-lfs to track those files.
+
+        Setting the `filename` argument to `True` will treat the arguments as literal filenames,
+        not as patterns. Any special glob characters in the filename will be escaped when
+        writing the .gitattributes file.
```
julien-c, line 447:
```suggestion
        writing to the .gitattributes file.
```

PR #225, opened 2021-07-27
Adding SD for superb (speech-classification).

Narsil:
- Adding a normalizers to help reformat the tensor into a list of classes
- Test for said normalizer (depends on torch, but it should be overall optional)
- Add the actual pipeline within superb.
- Will need update to `api-inference-community`.

### DIFF
--- a/api-inference-community/api_inference_community/normalizers.py
+++ b/api-inference-community/api_inference_community/normalizers.py
@@ -2,7 +2,8 @@
 Helper classes to modify pipeline outputs from tensors to expected pipeline output
 """
 
-from typing import List, Dict, Union, TYPE_CHECKING
+from typing import TYPE_CHECKING, Dict, List, Union
+
 
 Classes = Dict[str, Union[str, float]]
 
--- a/api-inference-community/api_inference_community/normalizers.py
+++ b/api-inference-community/api_inference_community/normalizers.py
@@ -15,12 +15,12 @@ if TYPE_CHECKING:
 
 
 def speaker_diarization_normalize(
-    tensor: "torch.Tensor", sampling_rate: int, classe_names: List[str]
+    tensor: "torch.Tensor", sampling_rate: int, classnames: List[str]
 ) -> List[Classes]:
     N = tensor.shape[1]
-    if len(classe_names) != N:
+    if len(classnames) != N:
         raise ValueError(
-            f"There is a mismatch between classe_names ({len(classe_names)}) and number of speakers ({N})"
+            f"There is a mismatch between classnames ({len(classnames)}) and number of speakers ({N})"
         )
     classes = []
     for i in range(N):
@@ -30,7 +30,7 @@ def speaker_diarization_normalize(
             if v == 1:
                 classes.append(
                     {
-                        "class": classe_names[i],
+                        "class": classnames[i],
                         "start": offset / sampling_rate,
                         "end": (offset + c.item()) / sampling_rate,
                     }
--- a/api-inference-community/api_inference_community/normalizers.py
+++ b/api-inference-community/api_inference_community/normalizers.py
@@ -1,0 +1,40 @@
+"""
+Helper classes to modify pipeline outputs from tensors to expected pipeline output
+"""
+
+from typing import List, Dict, Union, TYPE_CHECKING
+
+Classes = Dict[str, Union[str, float]]
+
+if TYPE_CHECKING:
+    try:
+        import torch
+    except Exception:
+        pass
+
+
+def speaker_diarization_normalize(
+    tensor: "torch.Tensor", sampling_rate: int, classe_names: List[str]
+) -> List[Classes]:
+    N = tensor.shape[1]
+    if len(classe_names) != N:
+        raise ValueError(
+            f"There is a mismatch between classe_names ({len(classe_names)}) and number of speakers ({N})"
+        )
+    classes = []
+    for i in range(N):
+        values, counts = tensor[:, i].unique_consecutive(return_counts=True)
+        offset = 0
+        for v, c in zip(values, counts):
+            if v == 1:
+                classes.append(
+                    {
+                        "class": classe_names[i],
+                        "start": offset / sampling_rate,
+                        "end": (offset + c.item()) / sampling_rate,
+                    }
+                )
+            offset += c.item()
+
+    classes = sorted(classes, key=lambda x: x["start"])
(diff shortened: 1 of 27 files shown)


api-inference-community/api_inference_community/normalizers.py
```diff
@@ -0,0 +1,41 @@
...
+
+if TYPE_CHECKING:
+    try:
+        import torch
+    except Exception:
+        pass
+
+
+def speaker_diarization_normalize(
+    tensor: "torch.Tensor", sampling_rate: int, classe_names: List[str]
```
osanseviero, line 18: nit: "classes_names" and same below

api-inference-community/docker_images/superb/app/pipelines/speech_classification.py
```diff
@@ -0,0 +1,57 @@
+import os
```
osanseviero, line 1: Few things
* Please add `speech-classification` to `common` as well.
* Could we also add a test with Lewis repo?

api-inference-community/docker_images/superb/app/pipelines/speech_classification.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __call__(self, inputs: np.array) -> List[Dict[str, Union[str, float]]]:
+        """
+        Args:
+            inputs (:obj:`np.array`):
+                The raw waveform of audio received. By default at 16KHz.
+                Check `app.validation` if a different sample rate is required
+                or if it depends on the model
+        Return:
+            A :obj:`dict`:. The object return should be liked {"text": "XXX"} containing
```
osanseviero, line 47: Please update the return type, this is still from ASR

osanseviero (commented):
Overall looks great! Just a couple of comments

api-inference-community/api_inference_community/normalizers.py
Narsil: I really would like to avoid double ss in names, it's very confusing. 
-1 "s" at the end of a variables -> helps understand it's a list
- otherwise plurals are to be avoided IMO as it's very inconsistent for non natives (even for natives sometimes)

`class_names` or `classnames` if you prefer.

api-inference-community/docker_images/superb/app/pipelines/speech_classification.py
Narsil: Sure !

api-inference-community/docker_images/superb/app/pipelines/speech_classification.py
Narsil: Done.

api-inference-community/docker_images/superb/app/pipelines/speech_classification.py
Narsil: Done.

api-inference-community/api_inference_community/normalizers.py
Narsil: I opted for `classnames` right now (classe is french so it was wrong anyway)

api-inference-community/docker_images/common/app/pipelines/speech_classification.py
```diff
@@ -0,0 +1,34 @@
...
+        )
+
+    def __call__(self, inputs: np.array) -> Dict[str, str]:
+        """
+        Args:
+            inputs (:obj:`np.array`):
+                The raw waveform of audio received. By default at self.sampling_rate, otherwise 16KHz.
+        Return:
+            A :obj:`list`:. Each item in the list is like {"class": "XXX", "start": float, "end": float}
+            "class" is the associated class of the audio segment, "start" and "end" are markers expressed in seconds
```
osanseviero, line 27: Nit: maybe we should also mention that they can overlap (i.e. two speakers can speak at same time)

osanseviero (approved):
Thanks a lot! Approving this since it looks good IMO.

Last thing I think we should add are unit tests (under `superb/tests`)

api-inference-community/docker_images/common/app/pipelines/speech_classification.py
Narsil: If you feel that it is necessary, I'll add it (it's not necessarily speakers btw ^_^)

osanseviero: Yes, speakers was just an example, it could be multiple classes in the same timeframe :)

api-inference-community/docker_images/common/app/pipelines/speech_segmentation.py
```diff
@@ -0,0 +1,34 @@
...
+    def __call__(self, inputs: np.array) -> Dict[str, str]:
+        """
+        Args:
+            inputs (:obj:`np.array`):
+                The raw waveform of audio received. By default at self.sampling_rate, otherwise 16KHz.
+        Return:
+            A :obj:`list`:. Each item in the list is like {"class": "XXX", "start": float, "end": float}
+            "class" is the associated class of the audio segment, "start" and "end" are markers expressed in seconds
+            within the audio file.
+        """
```
osanseviero, line 29:
```suggestion
    def __call__(self, inputs: np.array) -> List[Dict[str, Union[str, float]]]:
        """
        Args:
            inputs (:obj:`np.array`):
                The raw waveform of audio received. By default at self.sampling_rate, otherwise 16KHz.
        Return:
            A :obj:`list`:. Each item in the list is like {"class": "XXX", "start": float, "end": float}
            "class" is the associated class of the audio segment, "start" and "end" are markers expressed in seconds
            within the audio file. Segments can overlap in any way.
        """
```

api-inference-community/tests/test_dockers.py
```diff
@@ -223,6 +229,7 @@ def framework_docker_test(self, framework: str, task: str, model_id: str):
             self.assertEqual(response.content, b'{"ok":"ok"}')
 
             response = httpx.post(url, data=b"This is a test", timeout=timeout)
+            print(response.content)
```
osanseviero, line 232: Nit: remove?

osanseviero (approved):
Looks great!

api-inference-community/tests/test_dockers.py
Narsil:
```suggestion
```

Narsil: Done.

PR #227, opened 2021-07-27
Mv `interfaces` -> `widgets/lib/interfaces`

mishig25:
As discussed in #202, moving `interfaces` folder to `widgets/lib/interfaces`

What's been done:
1. Move all files from `interfaces`  to `widgets/lib/interfaces` & delete original `interfaces` folder
2. Update `workflows` for Github actions, updating `widgets/lib/interfaces` path
3. Update `readmes`, updating `widgets/lib/interfaces` path


widgets/src/lib/InferenceWidget/InferenceWidget.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
 	import type { SvelteComponent } from "svelte";
-	import type { PipelineType } from "../../../../interfaces/Types";
+	import type { PipelineType } from "../interfaces/Types";
```
Pierrci, line 3: I think we can use `$lib/interfaces/...` everywhere now in the widgets instead no?

mishig25: Updated all the imports to use `$lib/interfaces/...` 👍

Pierrci (approved):
Thanks @mishig25!

Can you open a small PR in moon-landing to reflect the changes in paths once you merge this one? (you'll also need to update the `huggingface_hub` submodule)

I know this will be rendered obsolete soon once we switch to package import but until then... :)

PR #230, opened 2021-07-27
Add keras mixin

nateraw:
Adds mixin for keras

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -49,7 +49,7 @@ except importlib_metadata.PackageNotFoundError:
 
 _tf_version = "N/A"
 _tf_available = False
-for package_name in ['tensorflow', 'tensorflow-cpu', 'tensorflow-gpu']:
+for package_name in ["tensorflow", "tensorflow-cpu", "tensorflow-gpu"]:
     try:
         _tf_version = importlib_metadata.version(package_name)
         _tf_available = True
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -49,7 +49,19 @@ except importlib_metadata.PackageNotFoundError:
 
 _tf_version = "N/A"
 _tf_available = False
-for package_name in ["tensorflow", "tensorflow-cpu", "tensorflow-gpu"]:
+_tf_candidates = (
+    "tensorflow",
+    "tensorflow-cpu",
+    "tensorflow-gpu",
+    "tf-nightly",
+    "tf-nightly-cpu",
+    "tf-nightly-gpu",
+    "intel-tensorflow",
+    "intel-tensorflow-avx512",
+    "tensorflow-rocm",
+    "tensorflow-macos",
+)
+for package_name in _tf_candidates:
     try:
         _tf_version = importlib_metadata.version(package_name)
         _tf_available = True
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -49,11 +49,13 @@ except importlib_metadata.PackageNotFoundError:
 
 _tf_version = "N/A"
 _tf_available = False
-try:
-    _tf_version = importlib_metadata.version("tensorflow")
-    _tf_available = True
-except importlib_metadata.PackageNotFoundError:
-    pass
+for package_name in ['tensorflow', 'tensorflow-cpu', 'tensorflow-gpu']:
+    try:
+        _tf_version = importlib_metadata.version(package_name)
+        _tf_available = True
+        break
+    except importlib_metadata.PackageNotFoundError:
+        pass
 
 
 def is_torch_available():
--- a/src/huggingface_hub/hub_mixin.py
+++ b/src/huggingface_hub/hub_mixin.py
@@ -58,12 +58,12 @@ class ModelHubMixin:
                 json.dump(config, f)
 
         # saving model weights/files
-        self._save_pretrained(save_directory)
+        self._save_pretrained(save_directory, **kwargs)
 
         if push_to_hub:
             return self.push_to_hub(save_directory, **kwargs)
 
-    def _save_pretrained(self, save_directory):
+    def _save_pretrained(self, save_directory, **kwargs):
         """
         Overwrite this method in subclass to define how to save your model.
         """
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -17,6 +17,12 @@ class KerasModelHubMixin(ModelHubMixin):
         """
         Mix this class with your keras-model class for ease process of saving & loading from huggingface-hub
 
+        NOTE - Dummy Inputs are required to save/load models using this mixin. When saving, you are required to either:
+
+            1. Assign an attribute to your class, self.dummy_inputs, that defines inputs to be passed to the model's call
+            function to build the model.
+            2. Pass the dummy_inputs kwarg to save_pretrained. We will save this along with the model (as if it were an attribute).
+
         Example::
 
             >>> from huggingface_hub import KerasModelHubMixin
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -1,0 +1,70 @@
+import logging
+import os
+from pathlib import Path
+
+from huggingface_hub import ModelHubMixin, hf_hub_download
+
+
+logger = logging.getLogger(__name__)
+
+
+class KerasModelHubMixin(ModelHubMixin):
+
+    _CONFIG_NAME = 'config.json'
+    _WEIGHTS_NAME = 'tf_model.h5'
+
+    def _save_pretrained(self, save_directory, dummy_inputs=None, **kwargs):
+
+        dummy_inputs = dummy_inputs if dummy_inputs is not None else getattr(self, 'dummy_inputs', None)
+
+        if dummy_inputs is None:
+            raise RuntimeError("You must either provide dummy inputs or have them assigned as an attribute of this model")
+
+        _ = self(dummy_inputs, training=False)
+
+        save_directory = Path(save_directory)
+        model_file = save_directory / self._WEIGHTS_NAME
+        self.save_weights(model_file)
+        logger.info(f"Model weights saved in {model_file}")
+
+    @classmethod
+    def _from_pretrained(
+        cls,
+        model_id,
+        revision,
+        cache_dir,
+        force_download,
+        proxies,
+        resume_download,
+        local_files_only,
+        use_auth_token,
+        by_name=False,
+        **model_kwargs,
+    ):
+        if os.path.isdir(model_id):
+            print("Loading weights from local directory")
+            model_file = os.path.join(model_id, cls._WEIGHTS_NAME)
+        else:
+            model_file = hf_hub_download(
+                repo_id=model_id,
+                filename=cls._WEIGHTS_NAME,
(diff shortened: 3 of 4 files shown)


Rocketknight1:
One question! What's the connection between this and the existing `PushToHubMixin` [here](https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L2025)? Are we deprecating that one in favour of this?

nateraw:
> One question! What's the connection between this and the existing `PushToHubMixin` [here](https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L2025)? Are we deprecating that one in favour of this?

@Rocketknight1 That's a hard-coded integration w/ transformers, AFAIK. In the future, it would probably be nice to use these Mixins directly instead and deprecate `PushToHubMixin` and other hub-related functions in transformers, but since these are changing a bit more rapidly, its probably for the best to leave the `PushToHubMixin` as-is for now.

tests/test_keras_mixin.py
```diff
@@ -0,0 +1,153 @@
...
+            f"{WORKING_REPO_DIR}/{REPO_NAME}-FROM_PRETRAINED",
+            config={"num": 10, "act": "gelu_fast"},
+        )
+
+        model = DummyModel.from_pretrained(
+            f"{WORKING_REPO_DIR}/{REPO_NAME}-FROM_PRETRAINED"
+        )
+        self.assertDictEqual(model.config, {"num": 10, "act": "gelu_fast"})
+
+    # def test_push_to_hub(self):
```
nateraw, line 137: Oops! Commented this out so I wasn't flooding the staging endpoint while working. Uncomment before merging.

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+        dummy_inputs = (
+            dummy_inputs
+            if dummy_inputs is not None
+            else getattr(self, "dummy_inputs", None)
+        )
+
+        if dummy_inputs is None:
+            raise RuntimeError(
+                "You must either provide dummy inputs or have them assigned as an attribute of this model"
+            )
```
osanseviero, line 52: From a user perspective I'm not sure `dummy_inputs` is very clear. Should we explain it in a docstring?

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+        assert (
+            hasattr(model, "dummy_inputs") and model.dummy_inputs is not None
+        ), "Model must have a dummy_inputs attribute"
+
+        _ = model(model.dummy_inputs, training=False)
+
+        model.load_weights(model_file, by_name=by_name)
+
+        _ = model(model.dummy_inputs, training=False)
+
```
osanseviero, line 102: Is running the model twice intended?

osanseviero (commented):
Overall looks good, just a couple of questions

src/huggingface_hub/keras_mixin.py
nateraw: Can do!

src/huggingface_hub/keras_mixin.py
nateraw: I copied this from `transformers`. I'm guessing its to first built the model, then after loading weights you sanity check the dummy inputs still work. No clue if wrong to do 😅 . Whatcha think, @Rocketknight1 ?

Rocketknight1: Keras lazily builds models when they first get inputs, yeah. I feel like after you build it and call `load_weights()` it shouldn't need to be built again, but I'm happy to assume it's there for a reason and leave it, lol

osanseviero: Yes, I'm sure it's there for a good reason :)

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,103 @@
...
+logger = logging.getLogger(__name__)
+
+
+class KerasModelHubMixin(ModelHubMixin):
+
+    _CONFIG_NAME = "config.json"
+    _WEIGHTS_NAME = "tf_model.h5"
+
+    def __init__(self, *args, **kwargs):
```
Rocketknight1, line 16: I assume the actual content of the `__init__` method is coming later, even if it's just a `super()` call!

Rocketknight1 (changes requested):

src/huggingface_hub/keras_mixin.py
nateraw: Resolved offline - no super call needed here since its a mixin 😄

src/huggingface_hub/keras_mixin.py
nateraw: Added a docstring in `__init__`, since I think that makes the most sense (as it'll be visible to user).

src/huggingface_hub/file_download.py
```diff
@@ -49,11 +49,13 @@
 
 _tf_version = "N/A"
 _tf_available = False
-try:
-    _tf_version = importlib_metadata.version("tensorflow")
-    _tf_available = True
-except importlib_metadata.PackageNotFoundError:
-    pass
+for package_name in ["tensorflow", "tensorflow-cpu", "tensorflow-gpu"]:
```
LysandreJik, line 52: Might as well rely on `transformers` and copy/paste all the packages we've faced up to now :smile:

https://github.com/huggingface/transformers/blob/fbf468b0573baddb1b9d1bb088a8b6d5c9303a7e/src/transformers/file_utils.py#L83-L93

src/huggingface_hub/hub_mixin.py
```diff
@@ -58,12 +58,12 @@ def save_pretrained(
                 json.dump(config, f)
 
         # saving model weights/files
-        self._save_pretrained(save_directory)
+        self._save_pretrained(save_directory, **kwargs)
```
LysandreJik, line 61: The `kwargs` are advertised as the kwargs to be passed to `push_to_hub` yet they're passed to `_save_pretrained`. I'd argue the kwargs to pass to each method ought to be different!

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+    def _from_pretrained(
+        cls,
+        model_id,
+        revision,
+        cache_dir,
+        force_download,
+        proxies,
+        resume_download,
+        local_files_only,
+        use_auth_token,
```
LysandreJik, line 77: My intuition tells me all of these should be optional kwargs, even if they're only supposed to be called internally

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+        by_name=False,
+        **model_kwargs,
+    ):
+        if os.path.isdir(model_id):
+            print("Loading weights from local directory")
```
LysandreJik, line 82:
```suggestion
            logger.info("Loading weights from local directory")
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+                resume_download=resume_download,
+                use_auth_token=use_auth_token,
+                local_files_only=local_files_only,
+            )
+
+        model = cls(**model_kwargs)
+
+        assert (
+            hasattr(model, "dummy_inputs") and model.dummy_inputs is not None
+        ), "Model must have a dummy_inputs attribute"
```
LysandreJik, line 101: I would raise a real `ValueError` here, as assertions can be disabled

LysandreJik (approved):
Solid, I like what you've done with it! And the testing suite is very nice, too. LGTM!

src/huggingface_hub/file_download.py
osanseviero: That's a lot of `tensorflow` 😄

tests/test_keras_mixin.py
```diff
@@ -0,0 +1,153 @@
...
+
+if is_tf_available():
+    import tensorflow as tf
+
+
+def require_tf(test_case):
+    """
+    Decorator marking a test that requires PyTorch.
+
+    These tests are skipped when PyTorch isn't installed.
```
osanseviero, line 29:
```suggestion
    Decorator marking a test that requires TensorFlow.

    These tests are skipped when TensorFlow isn't installed.
```

tests/test_keras_mixin.py
```diff
@@ -0,0 +1,153 @@
...
+    DummyModel = None
+
+
+@require_tf
+class HubMixingCommonTest(unittest.TestCase):
+    _api = HfApi(endpoint=ENDPOINT_STAGING)
+
+
+@require_tf
+class HubMixingTest(HubMixingCommonTest):
```
osanseviero, line 62:
```suggestion
class HubMixinCommonTest(unittest.TestCase):
    _api = HfApi(endpoint=ENDPOINT_STAGING)


@require_tf
class HubMixinTest(HubMixinCommonTest):
```

tests/test_keras_mixin.py
```diff
@@ -0,0 +1,153 @@
...
+            return self.l1(x)
+
+
+else:
+    DummyModel = None
+
+
+@require_tf
+class HubMixingCommonTest(unittest.TestCase):
+    _api = HfApi(endpoint=ENDPOINT_STAGING)
```
osanseviero, line 58: Is this super class really needed? We only have one test class so maybe the API wrapper should be instantiated in setUpClass?

osanseviero (approved):
This looks great! Thanks a lot

osanseviero:
Just a parenthesis discussion for a potential follow-up PR. Given that TF 2 suggested way of saving models is through SavedModels, should we also consider this use case? It's the recommended file format that is portable with TF Lite, TF Serving, etc.

Pros
* Much, much easier to use and save (easier to work with custom code, etc).
* Easy to connect with other tools of TF ecosystem (TensorFlow Serving for example)

Cons
* As discussed in the email, this was needed in `transformers` to only reload a part of a model from a checkpoint pre-trained with another architecture, which was only possible with the `.h5` format.
* Probably other things I'm not aware of.

@nateraw, does the existing solution work with custom objects such as custom layers? Or will this only work if loaded with the same class? If not, I guess we can't use the created repos out of the box in the Inference API.

src/huggingface_hub/keras_mixin.py
nateraw: Done!

tests/test_keras_mixin.py
nateraw: Just following what was done in `test_hubmixin.py`, so leaving this for now (I'd want to change both at once). Not mad at this update tho haha 😉

tests/test_keras_mixin.py
nateraw: 🔥 thank you sir!!!

tests/test_keras_mixin.py
nateraw: You're probably right. I didn't want to mess with it too much, though. again, just following the previous file. I did find the superclass useful (if I remember right) when I played around with parametrizing the tests so we could run them over TF/PyTorch/etc.

nateraw:
@osanseviero I'm not sure if this supports custom layers/objects 😅 . I think we definitely want to use savedmodel going forward. This is a good first effort, but the next PR should be to use SavedModel. Didn't want to mess with this one too much just to keep things similar to whats available elsewhere in the HF ecosystem.

Rocketknight1 (approved):

PR #159, opened 2021-07-05
add _from_pretrained hook

nateraw:
Resolves #149 .

PR #231, opened 2021-07-27
Use hf_hub_download in mixins

nateraw:
Use `hf_hub_download` instead of combination of `cached_download` and `hf_hub_url`. Mentioned in #159


osanseviero (approved):
Looks good!

PR #236, opened 2021-07-28
Remove trailing new line character from token

albertvillanova:
Python `f.read()` returns string with a trailing new line character:
```
"my-token-read-from-file\n"
```

This PR removes trailing new line character:
```python
"my-token-read-from-file"
```

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -732,7 +732,7 @@ class HfFolder:
         """
         try:
             with open(cls.path_token, "r") as f:
-                return f.read()
+                return f.read().strip()
         except FileNotFoundError:
             pass
 


julien-c:
We are going to move away from this way of storing the tokens soon-ish anyways, so i wouldn't make this change (fyi @LysandreJik)

albertvillanova:
OK @julien-c, I understand. 

But I'm wondering how users have accessed their private repos until now, because this bug returns a wrong token which raises a 404 client error.

julien-c:
hmm, i don't understand. The token is written **without** a trailing newline a few lines of code above, if I'm not mistaken.

Besides the feature is used by all authenticated calls in the library so if it was broken we'd know it no?

albertvillanova:
Sorry, you are right @julien-c.

It was my text editor (vim) which added a trailing new line character after opening (and saving-exit) the file. I've read that Gedit does the same.

PR #238, opened 2021-07-29
Error message when mic access isn't given

mishig25:
Screenshot 2021-07-29 at 10 12 16

Give feedback (i.e. error message) when a user: 
1. Blocks HF to access their microphone
2. AND tries to use voice recording on audio widgets


widgets/src/lib/InferenceWidget/shared/WidgetRecorder/WidgetRecorder.svelte
```diff
@@ -21,13 +22,19 @@
...
+			if (isRecording) {
+				await recorder.start();
+				onRecordStart();
+			} else {
+				const blob = await recorder.stopRecording();
+				onRecordStop(blob);
+			}
+		} catch (e) {
+			isRecording = false;
+			error = "Please allow 🤗 to access your microphone";
```
Pierrci, line 37: Maybe check the kind of error here to display a more precise error?

From https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia: "If the user denies permission, or matching media is not available, then the promise is rejected with `NotAllowedError` or `NotFoundError` respectively."

widgets/src/lib/InferenceWidget/widgets/AutomaticSpeechRecognitionWidget/AutomaticSpeechRecognitionWidget.svelte
```diff
@@ -141,6 +141,7 @@
 					classNames="mt-1.5"
 					{onRecordStart}
 					onRecordStop={onSelectFile}
+					bind:error
```
beurkinger, line 144: Really not a fan of binding to components internal values. 

Event the Svelte doc warns against it :
```
Use component bindings sparingly. It can be difficult to track the flow of data around your application if you have too many of them, especially if there is no 'single source of truth'.
```
cf https://svelte.dev/tutorial/component-bindings

I would much rather have an explicit `onError` function prop.

beurkinger (changes requested):

mishig25:
@beurkinger @Pierrci please feel free to re-review it when you have a chance
1. No more binding errors. Instead passing prop (a function)
2. Handling different types of microphone errors with different error messages (previously it was only one generic error message)

widgets/src/lib/InferenceWidget/shared/WidgetRecorder/WidgetRecorder.svelte
```diff
@@ -21,13 +22,31 @@
...
 	async function onClick() {
-		isRecording = !isRecording;
-		if (isRecording) {
-			recorder.start();
-			onRecordStart();
-		} else {
-			const blob = await recorder.stopRecording();
-			onRecordStop(blob);
+		try {
+			isRecording = !isRecording;
```
beurkinger, line 26: Little nitpick: I would personally put this line just above the `try {`, as initial setting the recording state is not dependent on the success or failure of the operation down the line - but I guess that doesn't matter that much :D

beurkinger (approved):

widgets/src/lib/InferenceWidget/shared/WidgetRecorder/WidgetRecorder.svelte
```diff
@@ -21,13 +22,31 @@
...
+				onRecordStart();
+			} else {
+				const blob = await recorder.stopRecording();
+				onRecordStop(blob);
+			}
+		} catch (e) {
+			isRecording = false;
+			switch (e.name) {
+				case "NotAllowedError": {
+					onError("Please allow 🤗 to access your microphone");
```
Pierrci, line 38:
```suggestion
					onError("Please allow access to your microphone");
```

widgets/src/lib/InferenceWidget/shared/WidgetRecorder/WidgetRecorder.svelte
```diff
@@ -21,13 +22,31 @@
...
+					break;
+				}
+				case "NotFoundError": {
+					onError("🤗 couldn't find microphone on your device");
```
Pierrci, line 42:
```suggestion
					onError("No microphone found on your device");
```

Pierrci (approved):

PR #242, opened 2021-07-30
Hotfix2.

Narsil:
This is actually deployed.

### DIFF
--- a/api-inference-community/api_inference_community/batch.py
+++ b/api-inference-community/api_inference_community/batch.py
@@ -16,7 +16,6 @@ def iterate_slow(pipe, dataset, f, task: str):
             assert isinstance(
                 item, bytes
             ), f"Batching cannot validate received {type(item)} but expected (str, bytes)"
-            print(item)
             inputs, parameters = normalize_payload(
                 item,
                 task,
--- a/api-inference-community/api_inference_community/batch.py
+++ b/api-inference-community/api_inference_community/batch.py
@@ -1,9 +1,11 @@
 import io
 import json
+import os
 
 import datasets
 import tqdm
 from huggingface_hub import HfApi
+from api_inference_community.validation import normalize_payload
 
 
 def iterate(pipe, dataset, f):
@@ -14,7 +16,18 @@ def iterate(pipe, dataset, f):
 def iterate_slow(pipe, dataset, f):
     for i, item in enumerate(tqdm.tqdm(dataset)):
         try:
-            result = pipe(item)
+            if isinstance(item, str):
+                # Can be filename
+                item = item.encode("utf-8")
+            assert isinstance(
+                item, bytes
+            ), f"Batching cannot validate received {type(item)} but expected (str, bytes)"
+            inputs, parameters = normalize_payload(
+                item,
+                os.getenv("TASK"),
+                sampling_rate=getattr(pipe, "sampling_rate", None),
+            )
+            result = pipe(inputs, **parameters)
         except Exception as e:
             result = {"error": str(e)}
 
--- a/api-inference-community/api_inference_community/batch.py
+++ b/api-inference-community/api_inference_community/batch.py
@@ -8,12 +8,7 @@ from api_inference_community.validation import normalize_payload
 from huggingface_hub import HfApi
 
 
-def iterate(pipe, dataset, f):
-    for i, result in enumerate(tqdm.tqdm(pipe.iter(dataset))):
-        write_result(result, f)
-
-
-def iterate_slow(pipe, dataset, f):
+def iterate_slow(pipe, dataset, f, task: str):
     for i, item in enumerate(tqdm.tqdm(dataset)):
         try:
             if isinstance(item, str):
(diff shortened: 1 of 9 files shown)


api-inference-community/api_inference_community/batch.py
```diff
@@ -3,19 +3,28 @@
...
     for i, item in enumerate(tqdm.tqdm(dataset)):
         try:
-            result = pipe(item)
+            if isinstance(item, str):
+                # Can be filename
+                item = item.encode("utf-8")
+            assert isinstance(
+                item, bytes
+            ), f"Batching cannot validate received {type(item)} but expected (str, bytes)"
+            print(item)
```
osanseviero, line 19: nit: remove print statement?

api-inference-community/api_inference_community/batch.py
```diff
@@ -3,19 +3,28 @@
...
 from huggingface_hub import HfApi
 
 
-def iterate(pipe, dataset, f):
-    for i, result in enumerate(tqdm.tqdm(pipe.iter(dataset))):
-        write_result(result, f)
-
-
-def iterate_slow(pipe, dataset, f):
+def iterate_slow(pipe, dataset, f, task: str):
```
osanseviero, line 10: Should this be renamed to just `iterate` now?

api-inference-community/manage.py
```diff
@@ -39,6 +42,84 @@ def create_docker(name: str) -> str:
...
+        for task_template in task_templates:
+            if task_template.task == task:
+                args.dataset_column = task_template.audio_file_path_column
+                print(f"Inferred dataset_column {args.dataset_column}")
+    return (
+        args.dataset_name,
+        args.dataset_config,
+        args.dataset_split,
+        args.dataset_column,
+    )
```
osanseviero, line 74: No strong opinion, but consider using dataclasses instead of tuples of 4 elements. That way we can have better typings and make the program more readable in my opinion

osanseviero (approved):
Looks great! I just left a couple of minor comments.

api-inference-community/api_inference_community/validation.py
```diff
@@ -276,7 +277,7 @@ def ffmpeg_read(bpayload: bytes, sampling_rate: int) -> np.array:
     ffmpeg_command = [
         "ffmpeg",
         "-i",
-        "pipe:0",
+        bpayload,
```
lewtun, line 280: @Narsil this was the cause of the `Malformed soundfile` error 😅 

@osanseviero are you able to re-deploy this branch so we can use it with `superb`?

osanseviero: I don't have deployment powers, but we can ask people in the infra team to do the deployment if not

Narsil: This fix is not correct unfortunately.

We really don't want to read the filename, but it really means that this doesn't run in your case:
https://github.com/huggingface/huggingface_hub/pull/242/files#diff-db8bbac95c077540d79900384cfd524d451e629275cbb5de7a31fc1cd5d6c189R319

ffmpeg_read is really correct the way it is, and it's important to keep it that way for use in API mode, where the bytes are part of the requests (they are raw bytes and need to be sent as pipe:0)

Do you mind checking why ? (Maybe extension list does not contain all needed)

api-inference-community/api_inference_community/validation.py
```diff
@@ -308,7 +309,25 @@ def normalize_payload_image(bpayload: bytes) -> Tuple[Any, Dict]:
     return img, {}
 
 
+AUDIO_EXTENSIONS = {".mp3", ".wav", ".flac", ".mp4", ".webm", ".aac"}
```
lewtun, line 312: @Narsil this was the source of the problem. when splitting the filename and extension with

```
filename, ext = os.path.splitext(bpayload)
```

the `ext` variable includes the dot `"."` so we were never actually reading the audio file.

i've reverted my previous change so think we can now re-deploy this to test it out with `superb`!

api-inference-community/api_inference_community/validation.py
lewtun: thanks for the context @Narsil! i did not appreciate that we don't want to read the filenames. 

i've reverted the change and pushed a fix in `AUDIO_EXTENSIONS` [here](https://github.com/huggingface/huggingface_hub/pull/242#discussion_r681985597).

api-inference-community/api_inference_community/validation.py
Narsil: Thanks !

api-inference-community/api_inference_community/validation.py
Narsil: For info, turned out, for some reason the dataset was corrupted on drive and even though datasets was thought to be in cache, the actual audio files were not there.

api-inference-community/manage.py
Narsil: Ok, I'll keep that in mind.

Definitely getting borderline here. I made the current choice because adding types also has drawbacks (more complex to read and understand).

I think the biggest issue I have is with `_column` which for some pipelines won't be unique.

api-inference-community/api_inference_community/batch.py
Narsil: Done.

Narsil:
Merging this as it's already live to prevent further confusion.

Issue #113, opened 2021-06-16
Widget for text to image generation


osanseviero:
Input would be text
The output would be the image

mishig25:
More than happy to take this one. 
@osanseviero @julien-c please let me know if there's anything I need to improve in terms of the process I have taken for #100 and #112 
Otherwise, I'll follow similar approach and implement other widgets

osanseviero:
You're going full speed! ⚡ 

The approach has been great so far, I think you can go ahead with the widget.

mishig25:
Awesome! 🤗

mishig25:
Created draft PR #131

osanseviero:
Some example motivation for this :) 
https://discuss.huggingface.co/t/generate-gif-reply-to-english-text-with-vqgan-clip/7166

patil-suraj:
> Some example motivation for this :)
> https://discuss.huggingface.co/t/generate-gif-reply-to-english-text-with-vqgan-clip/7166

And this as well https://github.com/huggingface/transformers/pull/12281 😉

borisdayma:
Hi, I'm interested in this widget.
What's the progress?

osanseviero:
There is an existing PR - https://github.com/huggingface/huggingface_hub/pull/131 but at the moment this is not integrated to the Inference API (since there is no `transformers` pipeline for this nor in the other libraries). If you want to build a demo for this, consider using Spaces instead for the time being.

osanseviero:
Btw, if you have model repos with a text to image generation model, I would be interested if you share them + some example on how to do inference.

borisdayma:
Would it make sense to allow custom pipelines?
We're working on a mini DALL-E repo so we'll need something completely custom

osanseviero:
Yes, having a generic docker image to support custom inference is in our roadmap. We will require a very specific input/output requirement for the API to work, but other than that it will be up to the users to do their own implementations, including adding their custom dependencies.

FYI @Narsil @julien-c

julien-c:
They can already submit a PR against https://github.com/huggingface/huggingface_hub/tree/main/api-inference-community though no?

osanseviero:
@julien-c you mean in the form of a new docker image just for this use case?

julien-c:
Yes – the boundary between single model and "library" being rather slim as we've seen in other cases, wouldn't `mini-dall-e` be a good image name in `api-inference-community`? 
[...]

PR #246, opened 2021-08-02
Add validation for text to image in Inference API

osanseviero:
This is based on https://github.com/huggingface/huggingface_hub/issues/113 and https://github.com/huggingface/huggingface_hub/pull/131. The API contract allows defining how many images will be returned, which can go from 1 to 10. Let me know if you think we should not allow generating multiple images in a single go, but this resembles well the text to text Inference API.

FYI @mishig25

### DIFF
--- a/api-inference-community/api_inference_community/validation.py
+++ b/api-inference-community/api_inference_community/validation.py
@@ -103,10 +103,6 @@ class SummarizationParamsCheck(SharedGenerationParams):
     num_return_sequences: Optional[NumReturnSequences] = None
 
 
-class TextToImageParamsCheck(BaseModel):
-    num_return_images: Optional[NumReturnSequences] = None
-
-
 class ConversationalInputsCheck(BaseModel):
     text: str
     past_user_inputs: List[str]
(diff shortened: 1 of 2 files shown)


julien-c:
> this resembles well the text to text Inference API

the text generation models generate multiple "continuations" in parallel, in one inference (though they only generate one token per inference, they generate multiple sentences at the same time)

Is it the case for image generation models as well?

osanseviero:
My understanding is that usually images are generated one at a time (`model.generate` will only generate one image; the DALL.E demo makes multiple `generate` calls for example. Usual Keras models work the same way). 

In a second thought, if we want the inference API to just be a 1:1 representation of a model interface, then we can drop the parameter. WDYT @julien-c?

@borisdayma maybe you have additional insights on the inner workings of the model.

julien-c:
yep that's what i'm wondering as well, unless there's a strong reason to deviate from this! Let see if others have feedback

lewtun:
> In a second thought, if we want the inference API to just be a 1:1 representation of a model interface, then we can drop the parameter. WDYT @julien-c?

as an end-user i'm also very much in favour of having a 1:1 representation with the model interface! this way you don't have to ask lots of questions on the forum about how the inference api works :)

borisdayma:
That's right, generating 1 sample at a time makes probably the most sense for the widget.
In DALL·E mini specific case I'm not sure if we should activate the widget (right now we deactivated it) because users may not notice the difference between the widget and the demo (which takes top 8 of 128 so yields much better results).

osanseviero:
@lewtun I simplified the API inputs so it only requires an input string.

lewtun (approved):
LGTM 🚀 !

PR #247, opened 2021-08-02
Widget object detection

mishig25:
Implemented object detection widget.

Request to Inference API (same as input for Image Classification widget):
```js
const requestBody = { file }; // file is an image file
```

Response from Inference API:
```js
Array<{ 
    label: string; 
    score: number;
    box: {xmin: number; ymin: number; xmax: number; ymax: number;} 
}>
```
Sample output:
```js
[
  {"score": 0.9982, "label": "remote", "box": {"xmin": 40, "ymin": 70, "xmax": 175, "ymax": 117}},
  {"score": 0.9955, "label": "couch", "box": {"xmin": 0, "ymin": 1, "xmax": 639, "ymax": 473}},
  {"score": 0.9987, "label": "cat", "box": {"xmin": 345, "ymin": 23, "xmax": 640, "ymax": 368}}
]
```

Screenshots:
| initial state                                                                                                                                                         | success state                                                                                                                                                         | fail state                                                                                                                                                            |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot 2021-09-09 at 16 02 43 |Screenshot 2021-09-09 at 16 01 26  | Screenshot 2021-09-09 at 16 03 37 |

Check out the [interactive netlify demo](https://6143391258950cfa694afb8d--huggingface-widgets.netlify.app/facebook/detr-resnet-50)

cc: @Narsil @NielsRogge @osanseviero please feel free to make any UX suggestions


beurkinger:
The image can overflow  on mobile resolutions
Screen Shot 2021-09-15 at 11 29 48 AM

beurkinger:
On mobile, when I click on a selection box it both select the box AND open the file selection screen.

Some ideas for a work around:
- when an image have been uploaded to the dropzone, hide the "dropzone" component and display a "remove" button or icon to reset the state
- deactivate mouse clicks on the uploaded image through css (`pointer-events: none;` if memory serves me well), except for the selection boxes

beurkinger (changes requested):
see comments :)

mishig25:
@beurkinger 
1. fixed the mobile overflow issue
2. regarding `On mobile, when I click on a selection box it both select the box AND open the file selection screen.`
This only happens if you are switching to mobile chrome dev tools after the page is loaded (i.e mobile is being checked onMount). That is to say, this issue won't happen on an actual mobile phone. Try changing to chrom mobile dev tools and reload the page (the problem won't happen)

widgets/src/lib/InferenceWidget/widgets/ObjectDetectionWidget/SvgBoundingBoxes.svelte
```diff
@@ -0,0 +1,86 @@
...
+					stroke-width="2"
+					opacity={highlightIndex === -1 || highlightIndex === index
+						? "1"
+						: "0.0"}
+					fill-opacity={highlightIndex === -1 || highlightIndex === index
+						? "0.1"
+						: "0.0"}
+					class="transition duration-200 ease-in-out"
+					on:mouseover={() => mouseover(index)}
+					on:mouseout={mouseout}
```
beurkinger, line 81: Not a big deal and you don't have to change it, but I personally find it useful to have attributes/props ordered by name, and try to keep them that way in components

beurkinger:
Two things that bother me a little bit : 
- lack of consistency between the two "Browse for image" inputs on mobile (see screenshot below)
- the HTML rendered by the backend for the Object Detection input is different than the one applied by the FE, and you can see the transition between the two

Screen Shot 2021-09-15 at 4 21 07 PM

widgets/src/lib/InferenceWidget/widgets/ObjectDetectionWidget/SvgBoundingBoxes.svelte
mishig25: done

beurkinger:
Almost there ! The current CSS solution work great for desktop and mobile, BUT doesn't work on tablets.

That being said, I have the solution: I just remembered that I had added a tailwind screen breakpoint util (or whatever it is called) to tailwind.config.js in moon-landing:
```
screens: {
				'with-hover': { raw: '(hover: hover)' },
				'no-hover': { raw: '(hover: none)' },
			},
```

So if you have the right tailwind config, you should to do stuff like `with-hover:absolute` instead of using size breakpoints !

beurkinger (changes requested):
See comment, let's discuss it !

beurkinger (approved):
Nice !

PR #244, opened 2021-08-02
Add base generic inference and enable token classification

osanseviero:
FYI @lewtun @nateraw 

Repo template: https://huggingface.co/templates/token-classification
Repo example: https://huggingface.co/osanseviero/en_core_web_sm

PR #252, opened 2021-08-03
Add text to image pipeline to generic inference

osanseviero:
Similar to #244, this PR adds support for `text-to-image` pipeline.

Here is an [example repo](https://huggingface.co/osanseviero/BigGAN-deep-128) that uses BigGAN-deep-128 (the return of pytorch-pretrained-BigGAN). As can be seen, users specify their own `requirements.txt` and `pipeline.py`. I struggled a bit with the output, but it's a base64 string of the image. This PR should unblock #131. 

Here is a corresponding repo template https://huggingface.co/templates/text-to-image

### DIFF
--- a/api-inference-community/docker_images/common/app/pipelines/text_to_image.py
+++ b/api-inference-community/docker_images/common/app/pipelines/text_to_image.py
@@ -1,5 +1,9 @@
+from typing import TYPE_CHECKING
+
 from app.pipelines import Pipeline
 
+if TYPE_CHECKING:
+    from PIL import Image
 
 class TextToImagePipeline(Pipeline):
     def __init__(self, model_id: str):
@@ -11,13 +15,13 @@ class TextToImagePipeline(Pipeline):
             "Please implement TextToImagePipeline.__init__ function"
         )
 
-    def __call__(self, inputs: str) -> str:
+    def __call__(self, inputs: str) -> "Image.Image":
         """
         Args:
             inputs (:obj:`str`):
                 a string containing some text
         Return:
-            A :obj:`str`. A base64 string representing the image.
+            A :obj:`PIL.Image` with the raw image representation as PIL.
         """
         # IMPLEMENT_THIS
         raise NotImplementedError(
--- a/api-inference-community/docker_images/common/app/pipelines/text_to_image.py
+++ b/api-inference-community/docker_images/common/app/pipelines/text_to_image.py
@@ -1,0 +1,25 @@
+from app.pipelines import Pipeline
+
+
+class TextToImagePipeline(Pipeline):
+    def __init__(self, model_id: str):
+        # IMPLEMENT_THIS
+        # Preload all the elements you are going to need at inference.
+        # For instance your model, processors, tokenizer that might be needed.
+        # This function is only called once, so do all the heavy processing I/O here
+        raise NotImplementedError(
+            "Please implement TextToImagePipeline __init__ function"
+        )
+
+    def __call__(self, inputs: str) -> str:
+        """
+        Args:
+            inputs (:obj:`str`):
+                a string containing some text
+        Return:
+            A :obj:`str`. A base64 string representing the image.
+        """
+        # IMPLEMENT_THIS
+        raise NotImplementedError(
+            "Please implement TextToImagePipeline __call__ function"
+        )
--- a/api-inference-community/docker_images/common/tests/test_api_text_to_image.py
+++ b/api-inference-community/docker_images/common/tests/test_api_text_to_image.py
@@ -15,11 +15,11 @@ from tests.test_api import TESTABLE_MODELS
 )
 class TextToImageTestCase(TestCase):
     def setUp(self):
-        model_id = TESTABLE_MODELS["text-to-speech"]
+        model_id = TESTABLE_MODELS["text-to-image"]
         self.old_model_id = os.getenv("MODEL_ID")
         self.old_task = os.getenv("TASK")
         os.environ["MODEL_ID"] = model_id
-        os.environ["TASK"] = "text-to-speech"
+        os.environ["TASK"] = "text-to-image"
         from app.main import app
 
         self.app = app
--- a/api-inference-community/docker_images/common/tests/test_api_text_to_image.py
+++ b/api-inference-community/docker_images/common/tests/test_api_text_to_image.py
@@ -1,30 +1,26 @@
-import json
+import base64
 import os
+from io import BytesIO
 from unittest import TestCase, skipIf
 
+import PIL
 from app.main import ALLOWED_TASKS
-from parameterized import parameterized_class
 from starlette.testclient import TestClient
 from tests.test_api import TESTABLE_MODELS
 
 
 @skipIf(
-    "token-classification" not in ALLOWED_TASKS,
-    "token-classification not implemented",
+    "text-to-image" not in ALLOWED_TASKS,
+    "text-to-image not implemented",
 )
-@parameterized_class(
-    [{"model_id": model_id} for model_id in TESTABLE_MODELS["token-classification"]]
-)
-class TokenClassificationTestCase(TestCase):
+class TextToImageTestCase(TestCase):
     def setUp(self):
+        model_id = TESTABLE_MODELS["text-to-speech"]
         self.old_model_id = os.getenv("MODEL_ID")
         self.old_task = os.getenv("TASK")
-        os.environ["MODEL_ID"] = self.model_id
-        os.environ["TASK"] = "token-classification"
-
-        from app.main import app, get_pipeline
-
-        get_pipeline.cache_clear()
+        os.environ["MODEL_ID"] = model_id
+        os.environ["TASK"] = "text-to-speech"
+        from app.main import app
 
         self.app = app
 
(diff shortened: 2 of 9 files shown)


mishig25:
lgtm! 🚀

mishig25 (approved):

osanseviero:
FYI @nateraw

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,25 @@
+from app.pipelines import Pipeline
+
+
+class TextToImagePipeline(Pipeline):
+    def __init__(self, model_id: str):
+        # IMPLEMENT_THIS
+        # Preload all the elements you are going to need at inference.
```
lewtun, line 7: turbo nit:

```suggestion
        # Preload all the elements you are going to need for inference.
```

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,25 @@
+from app.pipelines import Pipeline
+
+
+class TextToImagePipeline(Pipeline):
+    def __init__(self, model_id: str):
```
lewtun, line 5: is the `model_id` a required argument? perhaps we can help the end-user by explaining what this is in the docstring?

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,25 @@
...
+        # IMPLEMENT_THIS
+        # Preload all the elements you are going to need at inference.
+        # For instance your model, processors, tokenizer that might be needed.
+        # This function is only called once, so do all the heavy processing I/O here
+        raise NotImplementedError(
+            "Please implement TextToImagePipeline __init__ function"
```
lewtun, line 11: turbo nit:

```suggestion
            "Please implement TextToImagePipeline.__init__ function"
```

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        Args:
+            inputs (:obj:`str`):
+                a string containing some text
+        Return:
+            A :obj:`str`. A base64 string representing the image.
+        """
+        # IMPLEMENT_THIS
+        raise NotImplementedError(
+            "Please implement TextToImagePipeline __call__ function"
```
lewtun, line 24: turbo nit:

```suggestion
            "Please implement TextToImagePipeline.__call__ function"
```

api-inference-community/docker_images/common/tests/test_api_text_to_image.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@skipIf(
+    "text-to-image" not in ALLOWED_TASKS,
+    "text-to-image not implemented",
+)
+class TextToImageTestCase(TestCase):
+    def setUp(self):
+        model_id = TESTABLE_MODELS["text-to-speech"]
+        self.old_model_id = os.getenv("MODEL_ID")
```
lewtun, line 19: not important, but i don't quite understand why there is a `model_id` and an `old_model_id`: what does the "old" refer to?

lewtun (approved):
looks great, thanks for adding this! could you share a screenshot of what the model output looks like?

i left a few tiny nits - feel free to ignore them if you want :)

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
osanseviero: Turbo nits are the best 👍

api-inference-community/docker_images/common/tests/test_api_text_to_image.py
osanseviero: This part is just copied from other tests. If I recall correctly, this is to ensure that when you run `pytest ...` you don't override environment variables in your computer. Does that make sense?

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
osanseviero: Yes, this is the docker image pipeline template (not the template shared to the user) that is copied when new libraries are added to the Hub.

I would prefer to keep this the same as the rest of the pipelines (see all other files under `pipelines/`) for the time being. We can send a different PR or commit to improve these. The end user of the `common` directory are library owners that are integrating to the Hub, so I don't think this is a big issue.

api-inference-community/docker_images/common/tests/test_api_text_to_image.py
lewtun: ah ok thanks for the context!

osanseviero:
`looks great, thanks for adding this! could you share a screenshot of what the model output looks like?`

The Inference pipeline returns a base 64 string with the image. I'm not fully satisfied with the output and if you look at the example repo it got quite complicated (which I'm sure it's due to my ignorance on some image stuff). 

Screen Shot 2021-08-03 at 5 22 12 PM

I won't merge this until @Narsil takes a look, since he probably has good insights on the output for this (not high priority so it can wait for a couple of extra days).

api-inference-community/docker_images/common/app/pipelines/text_to_image.py
lewtun: ok sounds good to me, thanks for the explanation!

lewtun:
> The Inference pipeline returns a base 64 string with the image. I'm not fully satisfied with the output and if you look at the example repo it got quite complicated (which I'm sure it's due to my ignorance on some image stuff).

seems good enough for [stack overflow](https://stackoverflow.com/a/57318959) 😄

api-inference-community/docker_images/common/tests/test_api_text_to_image.py
```diff
@@ -0,0 +1,67 @@
...
+from tests.test_api import TESTABLE_MODELS
+
+
+@skipIf(
+    "text-to-image" not in ALLOWED_TASKS,
+    "text-to-image not implemented",
+)
+class TextToImageTestCase(TestCase):
+    def setUp(self):
+        model_id = TESTABLE_MODELS["text-to-speech"]
```
Narsil, line 18: Isn't that supposed to be `text-to-image` ?

api-inference-community/docker_images/common/tests/test_api_text_to_image.py
```diff
@@ -0,0 +1,67 @@
...
+        self.old_model_id = os.getenv("MODEL_ID")
+        self.old_task = os.getenv("TASK")
+        os.environ["MODEL_ID"] = model_id
+        os.environ["TASK"] = "text-to-speech"
```
Narsil, line 22: `text-to-image` ?

api-inference-community/docker_images/generic/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,16 @@
+from app.pipelines import Pipeline
+
+
+class TextToImagePipeline(Pipeline):
```
Narsil, line 4: If you add only dummy stuff like this everywhere in `generic` why not just use `Pipeline` everywhere instead of having dummy subclass that does nothign ?

api-inference-community/docker_images/generic/tests/test_api_token_classification.py
```diff
@@ -11,14 +12,19 @@
     "token-classification" not in ALLOWED_TASKS,
     "token-classification not implemented",
 )
+@parameterized_class(
+    [{"model_id": model_id} for model_id in TESTABLE_MODELS["token-classification"]]
+)
```
Narsil, line 17: Not sure why you're parametrizing those tests as you're keeping only 1 model to test.
It's fine, but surprising nonetheless.

Narsil (approved):
Looks fine, but the multple dummy inheritance withing `generic` looks very bloated, why not have a single class handle all pipelines ?

Narsil:
Another thing, is that we can send raw bytes directly with proper mimecode if we're sending only 1 image per inference (which I guess is the case here).

api_inference_community/routes.py (call_pipe) should be the place to look at.

api-inference-community/docker_images/generic/app/pipelines/text_to_image.py
osanseviero: That was the initial plan, but unfortunately it can be the case that different pipelines require somewhat different behaviour. For example, the audio pipelines also required specifying sampling rate (this is done in #266). 

Maybe we could have a `DefaultPipeline`, `AudioPipeline`, etc. WDYT?

api-inference-community/docker_images/generic/tests/test_api_token_classification.py
osanseviero: Yes, right now we're only using one. Since I can imagine the test suite growing, specially for this use image which has many use cases, I imagined that the # of test models will probably grow over time.

osanseviero:
> Looks fine, but the multple dummy inheritance withing `generic` looks very bloated, why not have a single class handle all pipelines ?

Copied from the comment "That was the initial plan, but unfortunately it can be the case that different pipelines require somewhat different behaviour. For example, the audio pipelines also required specifying sampling rate (this is done in #266). Maybe we could have a `DefaultPipeline`, `AudioPipeline`, etc. WDYT?"

> Another thing, is that we can send raw bytes directly with proper mimecode if we're sending only 1 image per inference (which I guess is the case here).

Yes, the case here is to return one image at a time. Thanks for this pointer, I'll look into this.

osanseviero:
I sent #272 based on @Narsil feedback, which allows users to just return the `PIL.Image` and we take care of the encoding.

Before

> img = transforms.ToPILImage()(output[0])
> buffer = io.BytesIO()
> img.save(buffer, format="JPEG")
> img_str = base64.b64encode(buffer.getvalue()).decode('utf-8')
> return img_str

Now

> return transforms.ToPILImage()(output[0]) # or any other PIL.Image

osanseviero:
I'll merge this now to unblock @mishig25. The refactoring can follow in another PR.

PR #254, opened 2021-08-04
Deprecate audio source separation

osanseviero:
We now use `audio-to-audio`. All repos have been changed

[Internal change](https://github.com/huggingface/moon-landing/compare/deprecate_audio?expand=1)


julien-c:
lgtm

PR #265, opened 2021-08-06
Clarify/specify/document model card metadata, `model-index`, and pipeline/task types

julien-c:
⚠️ Warning: I got a little carried away with the https://github.com/huggingface/huggingface_hub/pull/265/commits/7d1d7c930cfd0755822e5b39242286480bd31a3d commit (😂), as it may very well break stuff in moon-landing, and it will require others JS-codebase PRs (like @mishig25's) to be rebased.

Let's see if maybe we want to split that change into a different PR

### DIFF
--- a/docs/hub/adding-a-library.md
+++ b/docs/hub/adding-a-library.md
@@ -246,7 +246,8 @@ The metadata held in model cards is the best way to supercharge your model. It i
 define tags for your library or framework, the type of model uploaded, the language, license, evaluation results,
 and more.
 
-The full model card specification can be seen below:
+The full model card specification is  [here](https://github.com/huggingface/huggingface_hub/blame/main/modelcard.md), it's also repeated below for convenience:
+
 ```yaml
 ---
 language:
@@ -281,7 +282,7 @@ model-index:
 ---
 ```
 
-None of the fields are required - but any added field will improve the discoverability of your model and open it to features such as the inference API. You can find more information on repos and model cards [here](/docs/hub/model-repos).
+None of the fields are required - but any added field will improve the discoverability of your model and open it to features such as the inference API. You can find more information on repos and model cards [here](/docs/hub/model-repos#model-card-metadata).
 
 ## Setting up the Inference API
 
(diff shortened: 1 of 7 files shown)


julien-c:
On further thought the migration path is simple enough that I think this is reviewable/mergeable. Will still tag @mishig25, let me know if i shouldn't merge this when it's reviewed!

julien-c:
CI is flakier than flakes but i've managed to spam it enough that it now passes 🔫 

✅

widgets/src/lib/interfaces/Types.ts
```diff
@@ -1,67 +1,60 @@
...
-	"zero-shot-classification" = "zero-shot-classification",
-	"translation" = "translation",
-	"summarization" = "summarization",
-	"conversational" = "conversational",
-	"feature-extraction" = "feature-extraction",
-	"text-generation" = "text-generation",
-	"text2text-generation" = "text2text-generation",
-	"fill-mask" = "fill-mask",
-	"sentence-similarity" = "sentence-similarity",
+	"text-classification"                                     = "Text Classification",
```
elishowk, line 14: I think this will break moon-landing code like `moon-landing/server/views/lib/modelInputSnippets.ts` or `server/views/components/InferenceModal/inferenceSnippets.ts` that uses values "as keys". But putting pretty names instead of `key: key` is a good refactoring though

julien-c: yes there will be an associated fix in moon-landing for sue

LysandreJik (approved):
Thank you for specifying the required/optional nature of elements in the modelcard.

docs/hub/adding-a-library.md
```diff
@@ -246,7 +246,8 @@ The metadata held in model cards is the best way to supercharge your model. It i
 define tags for your library or framework, the type of model uploaded, the language, license, evaluation results,
 and more.
 
-The full model card specification can be seen below:
+The full model card specification is  [here](https://github.com/huggingface/huggingface_hub/blame/main/modelcard.md), it's also repeated below for convenience:
```
osanseviero, line 249: Do we really want to maintain the spec in two places? Maybe we can delete the one below to avoid obsolete content

osanseviero (approved):
This looks great to me, thanks for this!

docs/hub/adding-a-library.md
julien-c: yes i think it makes sense

julien-c: i'll leave it for now but we can remove it later

PR #272, opened 2021-08-10
Return image bytes for text-to-image

osanseviero:


api-inference-community/api_inference_community/routes.py
```diff
@@ -108,6 +109,18 @@ def emit(self, record):
...
+            buf = io.BytesIO()
+            outputs.save(buf, format="JPEG")
+            buf.seek(0)
+            img_bytes = buf.read()
+            b64_img_bytes = base64.b64encode(img_bytes)
+            return Response(
+                b64_img_bytes,
+                headers=headers,
+                status_code=200,
+                media_type="image/jpeg",
```
osanseviero, line 122: Starlette uses `media_type` to automatically set the `content-type` header. Let me know if you would rather have `content-type` explicitly in `headers["content-type"]` for consistency.

Narsil: Why are you still base64encoding ? You don't need to anymore.

For the header, it's fine. I don't really think it's worth it (as it's passed as-is anyway) but it's fine.

osanseviero: Fixed

osanseviero:
Done, I'm now returning the raw bytes instead of the encoded bytes. This should also make usage simpler (I was using `response.text` instead of `response.content`, which was causing the confusion).

Narsil (approved):
LGTM

PR #273, opened 2021-08-10
Open source code related to the repo type (tag icon, display order, snippets)

osanseviero:
This PR open sources internal code of multiple UI components related to new model types:

* Inference API snippets
* Tag icon color
* Display order


mishig25 (commented):
Looks great! Totally agree with the idea of open-sourcing tag icon, display order, snippets.
Two comments:
1. Since [DefaultWidget.ts](https://github.com/huggingface/huggingface_hub/blob/types/widgets/src/lib/interfaces/DefaultWidget.ts) and [index.svelte ](https://github.com/huggingface/huggingface_hub/blob/types/widgets/src/routes/index.svelte) use some of the data that's in [inputSnippets.ts](https://github.com/huggingface/huggingface_hub/blob/1274c9fe6f387b1ebce00d8d5549d498a0043f3d/widgets/src/lib/inferenceSnippets/inputsSnippets.ts), could you think of any way to (import one file into another and) reuse some parts between those 3 files? I don't have a good answer atm; therefore, I'm asking
2. Nit: rename `snippets.ts` -> `index.ts` (so that we can just `import ... from widgets/src/lib/inferenceSnippets`), and rename inputsSnippets.ts` -> `inputs.ts`

osanseviero:
Thanks for your review!

1. I would rather not mix a code migration with a code refactoring. As of now, all code introduced in `huggingface_hub` is exactly the same as the existing one. We can refactor it later on. WDYT?
2. Thanks and done, that was a great suggestion!

mishig25:
For number 1, I agree. Therefore, everything is lgtm!

julien-c:
i feel like this should have been merged at the same time as the corresponding moon-landing PR

mishig25:
I see that's a better decision. Please let me know whether I should revert those changes and wait its moon-landing counterpart 👍

PR #276, opened 2021-08-13
Branch push to hub

LysandreJik:
This PR offers the ability to manage revisions (and therefore branches) with the `Repository` class. The revision management is introduced at two levels:
- When initializing a repository, with the `revision` keyword argument. If the revision exists (it is a reference to a commit, a branch, or a tag), then this revision will be checked out. If the revision does not exist, a new branch will be created with the branching point set to the current default branch's head.
- Using the `git_checkout` method to manually check out a `revision`. It has a `create_branch_ok` argument that is meant to behave as the `exist_ok` argument of `os.makedir` -> it checks out the revision if it exists, otherwise it creates the branch. I found this to be the most compatible with the way repositories would be used in scripts.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -117,22 +117,21 @@ def is_tracked_upstream(folder: Union[str, Path]) -> bool:
     Check if the current checked-out branch is tracked upstream.
     """
     try:
-        p = subprocess.run(
-            ["git", "status", "-sb"],
+        command = "git rev-parse --symbolic-full-name --abbrev-ref @{u}"
+        subprocess.run(
+            command.split(),
             stderr=subprocess.PIPE,
             stdout=subprocess.PIPE,
             encoding="utf-8",
+            check=True,
             cwd=folder,
         )
-        status = p.stdout
+        return True
     except subprocess.CalledProcessError as exc:
-        raise OSError(exc.stderr)
+        if "HEAD" in exc.stderr:
+            raise OSError("No branch checked out")
 
-    if "HEAD" in status:
-        raise OSError("No branch checked out.")
-
-    is_tracked = ".." in status
-    return is_tracked
+        return False
 
 
 @contextmanager
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -243,6 +243,7 @@ class Repository:
         use_auth_token: Union[bool, str] = True,
         git_user: Optional[str] = None,
         git_email: Optional[str] = None,
+        revision: Optional[str] = None,
     ):
         """
         Instantiate a local clone of a git repo.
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/repository.py
```diff
@@ -280,6 +307,29 @@ def __init__(
 
         self.lfs_enable_largefiles()
 
+        if revision is not None:
+            self.git_checkout(revision, create_branch_ok=True)
+
+    @property
+    def current_branch(self):
```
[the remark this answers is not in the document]
sgugger: I think it's goo to have this as a property.

sgugger (approved):
Very clean! As discussed offline, refactoring all the `subprocess` call in a util function would help a lot.

src/huggingface_hub/repository.py
```diff
@@ -112,6 +112,29 @@ def is_git_ignored(filename: Union[str, Path]) -> bool:
...
+            cwd=folder,
+        )
+        status = p.stdout
+    except subprocess.CalledProcessError as exc:
+        raise OSError(exc.stderr)
+
+    if "HEAD" in status:
+        raise OSError("No branch checked out.")
+
+    is_tracked = ".." in status
```
osanseviero, line 134: Could you expand a bit on this one? Is it because in the format of the status you get something like this?:

`## master...origin/master` 

I wonder if there's something a bit more error-proof, such as `git rev-parse --symbolic-full-name --abbrev-ref @{u}`, but this one is also harder to maintain/understand (this command gives a clear fatal error when there is no upstream configured).

src/huggingface_hub/repository.py
```diff
@@ -730,6 +785,45 @@ def git_push(self) -> str:
 
         return self.git_head_commit_url()
 
+    def git_checkout(self, revision, create_branch_ok=False):
+        """
+        git checkout a given revision
+
+        Specifying `create_branch_ok` to `True` will create the branch to the given revision if that revision doesn't exist.
```
osanseviero, line 792: I would put this in Args docstring for consistency with other methods (such as `push_to_hub`).

osanseviero (approved):
This is great and very concise/clean. Thanks for it!

src/huggingface_hub/repository.py
LysandreJik: All other methods in `Repository` are defined as this one is (except `push_to_hub` which has, unfortunately, a very uninformative description). I agree with you that `Args` is better, but I'd rather do it all in a single refactor. Does that sound good?

osanseviero: Yes, that sounds great!

lvwerra:
Hi @LysandreJik I gave this feature a try and ran into the issue that the upstream branch is not set:
```python
OSError: fatal: The current branch test-1 has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin test-1
```

Should a command like
```python
command = f"git branch --set-upstream-to origin/{revision}" #git>=1.8.0
```
be executed at the end of `git_checkout`? Or is this the wrong usage of this feature?

Steps to reproduce the error:

```
project_name = 'transformersbook/codeparrot-exp'
hf_repo = Repository(project_name, clone_from=project_name, revision='test-1')
model = GPT2LMHeadModel.from_pretrained(project_name)
tokenizer = AutoTokenizer.from_pretrained(project_name)

# SOME MODEL TRAINING

model.save_pretrained(project_name)
hf_repo.push_to_hub(commit_message=f'step {step}')
```

Issue #211, opened 2021-07-19
Only use git credentials, no more token


LysandreJik:
The current authentication system isn't ideal for git-based workflows. It isn't clear to users why they should first authenticate with `huggingface-cli`, then re-authenticate with `git push`.

It also isn't simple to `git push` in a colab notebook, a shell-less environment which can't prompt for username and password. All of these issues could be handled in a simpler way by only using git credentials, not an authentication token, in order to handle all git-related operations.

Currently looking at:
`git credentials fill`
`git credential-store store`
`git credential-store get`

julien-c:
> It isn't clear to users why they should first authenticate with huggingface-cli

Disclaimer; they shouldn't

PR #277, opened 2021-08-13
Git credentials

LysandreJik:
*Thoroughly tested in practice, will implement unit tests in a bit but the API will remain as it is so pinging you for review.*

Fixes https://github.com/huggingface/huggingface_hub/issues/211 and https://github.com/huggingface/moon-landing/issues/150

#### Problem

This PR offers a long-awaited overhaul of the login system through the `huggingface_hub` library. The way the login system is currently designed is complex and led to a lot of errors mid-training during the Jax/Flax community week - as well as a general misunderstanding of what was happening behind the scenes.

The core issue is that logging in through `huggingface_cli login` does *not* log the user through git. The user will have to reauthenticate when pushing. This is particularly an issue on Google Colab, where a simple `git push` is met by a fatal error:
```
fatal: could not read Username for 'https://huggingface.co': No such device or address
```
The workarounds are cumbersome (see [#12572](https://github.com/huggingface/transformers/issues/12572)).

#### Solving the issue

This PR offers to use the `git-credential` API to store the username and password in a way that git can leverage when authenticating the user on `huggingface.co`. It leverages the `git credential.helper store` helper, which stores the username/password under plain text in a file with the user's permission.

It does not set the global credential helper to `store` as this could disrupt the user's git configuration - it instead detects what is the current helper, and if it isn't `store`, mentions that users will need to reauthenticate. If users do not wish to reauthenticate, a command to be run (which will set the helper to `store`) is printed out.

Additionally, in order to improve the experience across platforms, this PR adjusts two configuration attributes which are linked to the above:
- Cloning a repository through `Repository(xxx, clone_from=yyy)` will set the helper of that repository to `store`. 
- Using the `Repository` without explicitly passing a `git_email` or a `git_user`, instead of leveraging the system's default, will leverage the token to be used by that `Repository`.

### DIFF
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -117,7 +117,7 @@ def tabulate(rows: List[List[Union[str, int]]], headers: List[str]) -> str:
     return "\n".join(lines)
 
 
-def currently_setup_credential_helper():
+def currently_setup_credential_helper(directory=None):
     try:
         output = subprocess.run(
             "git config --list".split(),
@@ -125,6 +125,7 @@ def currently_setup_credential_helper():
             stdout=subprocess.PIPE,
             encoding="utf-8",
             check=True,
+            cwd=directory,
         ).stdout.split("\n")
 
         current_credential_helper = ""
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -117,6 +117,26 @@ def tabulate(rows: List[List[Union[str, int]]], headers: List[str]) -> str:
     return "\n".join(lines)
 
 
+def currently_setup_credential_helper():
+    try:
+        output = subprocess.run(
+            "git config --list".split(),
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            encoding="utf-8",
+            check=True,
+        ).stdout.split("\n")
+
+        current_credential_helper = ""
+        for line in output:
+            if "credential.helper" in line:
+                current_credential_helper = line.split("=")[-1]
+    except subprocess.CalledProcessError as exc:
+        raise EnvironmentError(exc.stderr)
+
+    return current_credential_helper
+
+
 class BaseUserCommand:
     def __init__(self, args):
         self.args = args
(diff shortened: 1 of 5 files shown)


sgugger (approved):
Very clean! Thanks for investigating that bug and finding this solution!

src/huggingface_hub/hf_api.py
```diff
@@ -224,6 +225,80 @@ def __str__(self):
...
+
+        standard_input += "\n"
+        print(standard_input.encode("utf-8"))
+
+        process.stdin.write(standard_input.encode("utf-8"))
+        process.stdin.flush()
+        output = process.stdout.read()
+        print(output)
+        output = output.decode("utf-8")
+        print(output)
```
osanseviero, line 270: Should we remove these `print` statements?

src/huggingface_hub/hf_api.py
```diff
@@ -240,6 +315,9 @@ def login(self, username: str, password: str) -> str:
         r = requests.post(path, json={"username": username, "password": password})
         r.raise_for_status()
         d = r.json()
+
+        cased_username = self.whoami(d["token"])["name"]
```
osanseviero, line 319: No strong opinion.

is this the same as `username.lower()`? Maybe we could save an API call here, but maybe this is safer to be consistent with `erase_from_credential_store`

osanseviero (approved):
Thanks for this PR, this is a great improvement of quality of life for development!

src/huggingface_hub/commands/user.py
```diff
@@ -117,6 +117,27 @@ def tabulate(rows: List[List[Union[str, int]]], headers: List[str]) -> str:
...
+            stdout=subprocess.PIPE,
+            encoding="utf-8",
+            check=True,
+            cwd=directory,
+        ).stdout.split("\n")
+
+        current_credential_helper = ""
+        for line in output:
+            if "credential.helper" in line:
+                current_credential_helper = line.split("=")[-1]
```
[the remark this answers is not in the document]
LysandreJik: Indeed, I instead used an array of helpers. Thank you!

src/huggingface_hub/hf_api.py
LysandreJik: Absolutely! I went ahead the "correctly" cased way - but any casing works as long as it's identical across the helper methods. Thanks!

PR #282, opened 2021-08-16
Push to hub/commit with branches

LysandreJik:
Completes the `revision`-based API by adding branches management to `push_to_hub` and the `commit` context manager.

#### Using `revision` from `Repository.__init__`

```python
project_name = 'org/project'

# Checkout revision
hf_repo = Repository(project_name, clone_from=project_name, revision='test-1')

model = GPT2LMHeadModel.from_pretrained(project_name)
model.save_pretrained(project_name)

hf_repo.push_to_hub(commit_message=f'step {step}')
```

#### Using `git_checkout` from `Repository`

```python
project_name = 'org/project'

# Checkout revision
hf_repo = Repository(project_name, clone_from=project_name)
hf_repo.git_checkout("branch")

model = GPT2LMHeadModel.from_pretrained(project_name)
model.save_pretrained(project_name)

hf_repo.push_to_hub(commit_message=f'step {step}')
```

Note that the `push_to_hub` method interacts in a complex manner:
- The branch must first be checked out
- The files are saved
- Push to hub should be called

#### Using the `Repository.commit` context manager

Therefore, I would instead recommend using the `Repository.commit` context manager as I find it has a better API, as can be seen below. I think the two could eventually be merged into a single `with Repository.push_to_hub`. See details of the API with usage below:

```python
repo = Repository("local_directory", clone_from="Jikiwa/with-commit-1")

with repo.commit("Add files"):
    with open("file.txt", "w+") as f:
        f.write("Nice")

with repo.commit("Add files to brand-new-branch", branch="brand-new-branch"):
    with open("brand-new-file.txt", "w+") as f:
        f.write("Nice Nice")

repo.git_checkout("nice-branch", create_branch_ok=True)

# Will use lastly checked-out branch
with repo.commit("Add files in nice-branch"):
    with open("file.txt", "w+") as f:
        f.write("Nice")
```

A caveat with `git_checkout` is that it may result in a fatal error if some files are added/modified as the changes would be overridden by the checkout.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -735,7 +735,7 @@ class Repository:
         """
         if auto_lfs_track:
             tracked_files = self.auto_track_large_files(pattern)
-            if len(tracked_files) > 0:
+            if tracked_files:
                 logger.warning(
                     f"Adding files tracked by Git LFS: {tracked_files}. This may take a bit of time if the files are large."
                 )
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -112,6 +112,26 @@ def is_git_ignored(filename: Union[str, Path]) -> bool:
     return is_ignored
 
 
+def files_to_be_staged(pattern: str, folder: Union[str, Path]) -> List[str]:
+    try:
+        p = subprocess.run(
+            ["git", "ls-files", "-mo", pattern],
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            check=True,
+            encoding="utf-8",
+            cwd=folder,
+        )
+        if len(p.stdout.strip()):
+            files = p.stdout.strip().split("\n")
+        else:
+            files = []
+    except subprocess.CalledProcessError as exc:
+        raise EnvironmentError(exc.stderr)
+
+    return files
+
+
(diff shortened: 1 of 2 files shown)


lvwerra:
LGTM, thanks for improving the API!

lvwerra (approved):

src/huggingface_hub/repository.py
```diff
@@ -852,6 +882,21 @@ def commit(self, commit_message: str, track_large_files: bool = True):
 
         """
 
+        files_to_stage = files_to_be_staged(".", folder=self.local_dir)
+
+        if len(files_to_stage):
```
osanseviero, line 887: nit: I think we can directly do `if files_to_stage`

osanseviero (commented):
Overall seems good, but the tests are failing at the moment.

I was also wondering if we should add some documentation in https://github.com/huggingface/huggingface_hub/tree/main/src/huggingface_hub#managing-a-repository-with-repository, but this could be a follow up as well

NielsRogge:
LGTM, thank you for adding this!

src/huggingface_hub/repository.py
LysandreJik: Indeed!

julien-c (commented):
nice implementation, and I like the thoroughness of the tests

PR #293, opened 2021-08-24
Update Hub and huggingface_hub docs

osanseviero:
This PR does two things
* Updates the add a model guide based on feedback
* Does a general cleanup/update/simplification of `huggingface_hub` README.

### DIFF
--- a/docs/hub/adding-a-model.md
+++ b/docs/hub/adding-a-model.md
@@ -55,6 +55,10 @@ There is only one key difference if you have large files (over 10MB). These file
 
 2. Run `git lfs install` to initialize **git-lfs**:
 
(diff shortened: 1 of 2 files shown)


julien-c (approved):
good improvements

(nit) Not a fan of the 80-char line wrapping as it makes subsequent diffs hard to read

LysandreJik:
Was actually thinking of bumping up the `black` character limit to 119 like `transformers`, it greatly improves readability imo

julien-c:
hmm those are .md files

LysandreJik (approved):
Looks good to me, thanks for completing the docs with the docstrings!

LysandreJik:
> hmm those are .md files

Yes, I was talking about the Python files, not the files here. Bouncing back on your comment that 80 characters isn't necessarily more readable!

PR #301, opened 2021-09-01
🖼 Update timm image classification pipeline

nateraw:
Make `timm` image-classification pipeline work more generically. 

Works with:
- [this official timm model](https://huggingface.co/timm/eca_nfnet_l0)
- [this fine-tuned timm model](https://huggingface.co/nateraw/timm-resnet50-beans)


osanseviero:
Overall looks great! Thanks a lot for this really needed update.

* I think we should merge [this](https://github.com/huggingface/huggingface_hub/pull/304) PR as well first.
* Could you add a test model that is not imagenet for better coverage (add it to test_api)? You can see how to do tests with multiple models in [this example](https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/docker_images/sentence_transformers/tests/test_api.py).
  * Ideally there should be one model without labels that is for image net and another model with labels in its config. I don't think this is 100% needed, but would be great to have good coverage unless creating a dummy model would take too much time.

osanseviero (approved):
This looks great! Could you please run `RUN_DOCKER_TESTS=1 pytest -sv tests/test_dockers.py::DockerImageTests::test_timm` to make sure thing work well end-to-end?

Also adding @Narsil to have a second set of eyes in this PR

api-inference-community/docker_images/timm/app/pipelines/image_classification.py
```diff
@@ -4,16 +4,28 @@
...
+        state_dict = load_state_dict_from_hf(model_id)
+        self.model.load_state_dict(state_dict, strict=True)
+        self.model.eval()
+
+        self.transform = create_transform(**resolve_data_config({}, model=self.model))
+        self.top_k = min(self.model.num_classes, 5)
+
+        self.labels = self.hf_cfg.get("labels", None)
+        if self.labels is None:
+            if self.model.num_classes == 1000:
```
Narsil, line 25: That seems overly specific. Is there no way to fix the `labels` instead  (For the models that need it)?

Narsil (approved):
LGTM.
Not a big fan of the hardcoded == 1000 labels, but definitely not a blocker.

api-inference-community/docker_images/timm/tests/test_api_image_classification.py
```diff
@@ -11,12 +12,14 @@
...
+)
 class ImageClassificationTestCase(TestCase):
     def setUp(self):
-        model_id = TESTABLE_MODELS["image-classification"]
         self.old_model_id = os.getenv("MODEL_ID")
         self.old_task = os.getenv("TASK")
-        os.environ["MODEL_ID"] = model_id
+        os.environ["MODEL_ID"] = self.model_id
         os.environ["TASK"] = "image-classification"
         from app.main import app
```
osanseviero, line 25: Could you also add this below please?

```
@classmethod
def setUpClass(cls):
    from app.main import get_pipeline

    get_pipeline.cache_clear()
```

osanseviero: As in https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/docker_images/generic/tests/test_api_feature_extraction.py

nateraw: Tried this yesterday and it didn't work. however, I think I noticed the problem was that the app itself was a bit out of sync with the template. Will check back in on it

api-inference-community/docker_images/timm/app/pipelines/image_classification.py
nateraw: I suppose folks pushing timm models could be instructed to include the labels when they push. that way we wouldn't have to do any of this hard-coded business (which is horrifying to me as well).

In fact, let's do that. If I don't do it for this PR (running out of time on a friday), I'll issue another one early next week that does. 😄

osanseviero: Agree with @Narsil, the 1k hardcoding is not great. I think this responsibility should be for the users.

api-inference-community/docker_images/timm/tests/test_api_image_classification.py
osanseviero: You might want to make sure https://github.com/huggingface/huggingface_hub/pull/304 is in your branch as well :)

nateraw: done!

api-inference-community/docker_images/timm/app/pipelines/image_classification.py
```diff
@@ -4,16 +4,26 @@
...
-            self.model = timm.create_model("resnet50d", pretrained=True).eval()
-        elif "dpn92" in model_id:
-            self.model = timm.create_model("dpn92", pretrained=True).eval()
-        else:
-            raise EnvironmentError("Cannot detect the correct arch for this model")
+
+        self.hf_cfg, self.arch = load_model_config_from_hf(model_id)
+        self.model = timm.create_model(f"hf_hub:{model_id}", pretrained=True)
+        self.transform = create_transform(
+            **resolve_data_config(self.hf_cfg, model=self.model)
```
nateraw, line 18: Updated to use `self.hf_cfg` here instead of `{}` alone, which I had previously. This way, if the user defines any parameters in their `config.json` that that override the default config for transforms (size, std, mean), they'll be used instead of the defaults.

api-inference-community/docker_images/timm/app/pipelines/image_classification.py
```diff
@@ -4,16 +4,26 @@
...
+        )
+        self.model.eval()
+
+        self.top_k = min(self.model.num_classes, 5)
+
+        self.labels = self.hf_cfg.get("labels", None)
+        if self.labels is None:
+            self.labels = [f"LABEL_{i}" for i in range(self.model.num_classes)]
```
nateraw, line 26: Got rid of the `IMAGENET_LABELS` list at the end of this file. If a user doesn't provide labels, we'll just use `LABEL_<idx>`. We'll update the official examples to include labels so they look nice in the inference API/widget.

nateraw (commented):
I think we're good to go here! Docker tests passed. Just adding a couple comments for clarity/bookkeeping purposes 😄

PR #303, opened 2021-09-02
Fix links to filtered pages

beurkinger:
The /models and /datasets pages filtering system has been updated, but some of the links present in the documentation still relied on the old syntax


julien-c:
let's hold off on this before we decide if we actually want to change the URLs (I'd say we don't)

beurkinger:
@julien-c even if we decide to revert back the param change, 90% the code introduced by this PR will still be needed + it fixes the situation for current users

PR #307, opened 2021-09-03
Fix link to library tags image

lewtun:
This PR fixes a broken link to one of the images in the docs.

Before
Screen Shot 2021-09-03 at 10 19 54

After
Screen Shot 2021-09-03 at 10 19 47


osanseviero:
Thanks for this fix!

PR #308, opened 2021-09-03
Allow audio-classification in community API

osanseviero:
Please let me know if this looks good or if I'm missing something. 

This seems like a very complex PR!


Narsil:
Super hard. We can fly for `generic` with this. But we would need a real change in `common` to improve implementors life :)

Narsil (approved):
LGTM for what we want to achieve :)

PR #313, opened 2021-09-06
Fix Audio Classification pipeline

osanseviero:
Sort audio classification results by probability, not by label....

### DIFF
--- a/api-inference-community/docker_images/speechbrain/app/pipelines/audio_classification.py
+++ b/api-inference-community/docker_images/speechbrain/app/pipelines/audio_classification.py
@@ -32,6 +32,6 @@ class AudioClassificationPipeline(Pipeline):
         probs = torch.softmax(probs[0], dim=0)
         labels = self.model.hparams.label_encoder.decode_ndim(range(len(probs)))
         results = []
-        for label, prob in sorted(zip(labels, probs), reverse=True)[: self.top_k]:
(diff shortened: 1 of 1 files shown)


julien-c:
(just noting that I'm not sure this is consistent with `text-classification` where IIRC we keep the label list's order)

osanseviero:
@anton-l what do we do for `audio-classification` in `transformers`? Do we keep top_k and then sort by probability? Or do we keep top_k and then keep label list's order.

anton-l:
@osanseviero in the end the labels are ordered by probability (decreasing), regardless of top_k

julien-c:
(ah yeah if you do topk then it doesn't make a lot of sense to order by label list. For `text-classification` I think we keep label order because we don't do topk – I might be wrong)

osanseviero:
Yeah we do `top_k`, specially for speaker verification or language id there can be hundreds or thousands of labels

Narsil:
> (just noting that I'm not sure this is consistent with text-classification where IIRC we keep the label list's order)

@julien-c for transformers, by default the pipeline uses `~top_k=1` and returns only 1 result. For the API, there's an override with `return_all_scores=True` which will return all scores (no sorting). And it does not have a `top_k` argument.

It's definitely something that can be improved, especially to confirm with `image-classification` and `audio-classification`. Only issue is backward compatibility in transformers.

PR #315, opened 2021-09-06
Non blocking git push

LysandreJik:
This PR introduces a non-blocking git push as asked in https://github.com/huggingface/huggingface_hub/issues/169 and proposed by @cccntu in https://github.com/huggingface/huggingface_hub/pull/171.

It is currently painful to have to wait until a `git push` is over before continuing to something else. Since this uses nothing else than bandwidth, it is fine for it to happen behind the scenes.

This implementation:
- Implements a `blocking` parameter to `git_push` and the `commit` context manager; these variables default to `True`.
- In case of `blocking=False`, outputs a variable that can be monitored to follow the status of the `git push` command
- In case of `blocking=False`, outputs a variable that can be monitored to ensure the `git push` command has terminated
- These variables are added to the instance variable `command_queue` to be managed. More information about this can be found below.
- This implementation leverages the `atexit` official lib to ensure that all processes have completed before exiting the Python runtime.
- The progress bars are not managed when using a `blocking=False` parameter.
- Implements a warning when several commits are about to be pushed. The progress bars may be unreliable as a single progress bar may monitor the upload of several versions of a single file, therefore resulting in a jump between the two uploads on the progress bars.

It is recommended to monitor the progress of the current push before attempting a second one; however, several pushes may be launched concurrently. See below for the expected behaviors:

### What happens when two git pushes are done simultaneously?

If two identical git pushes are done, then no error in the workflow will happen. The remote correctly registers the updated version.

Two times more bandwidth is consumed as both pushes will upload all files.

### What happens when two git pushes are done simultaneously, with the second git push having one or several additional commits?

All commits are correctly registered on the remote. More bandwidth is consumed as both pushes will upload all files.

### What happens if one of the pushes errors out?

Given that the pushes are independent of each other as the remote isn't updated until a push is complete, the workflow should complete.

---

## The `command_queue`

This implementation implements a `command_queue` so that users may be aware of the current status of all commands launched.

It can be accessed as follows:

```py
# Access to command queue
command_queue = repo.command_queue

# Accessing a single command
last_command = repo.command_queue[-1]

# Vizualizing the command
print(last_command)
# [push command, status code: running, in progress. PID: 299167]

# Accessing the command's properties
last_command.status
last_command.is_done
last_command.title
last_command.stderr
last_command.stdout

# Accessing the commands that are currently in progress
commands_in_progress = repo.commands_in_progress

# Accessing the commands that failed
commands_failed = repo.commands_failed
```

### DIFF
--- a/src/huggingface_hub/README.md
+++ b/src/huggingface_hub/README.md
@@ -203,6 +203,10 @@ traditional Git methods:
 - `git_push()`
 - `git_checkout(branch)`
 
+The `git_push` method has a parameter `blocking` which is `True` by default. When set to `False`, the push will
+happen behind the scenes - which can be helpful if you would like your script to continue on while the push is 
+happening.
+
 LFS-tracking methods:
 
 - `lfs_track(pattern: Union[str, List[str]], filename: bool)`. Setting
@@ -221,6 +225,7 @@ On top of these unitary methods lie some useful additional methods:
   tracks large files (>10Mb) with `git-lfs`. The `track_large_files` argument can
   be set to `False` if you wish to ignore that behavior.
 
+These two methods also have support for the `blocking` parameter.
 
 Examples using the `commit` context manager:
 ```python
@@ -235,6 +240,46 @@ Examples using the `commit` context manager:
 >>> with Repository("torch-model", clone_from="<user>/torch-model", use_auth_token=True).commit("My cool model :)"):
 ...     torch.save(model.state_dict(), "model.pt")
   ```
+
+### Non-blocking behavior
+
+The pushing methods have access to a `blocking` boolean parameter to indicate whether the push should happen
+asynchronously.
+
+In order to see if the push has finished or its status code (to spot a failure), one should use the `command_queue`
+property on the `Repository` object.
+
+For example:
+
+```python
+from huggingface_hub import Repository
+
+repo = Repository("<local_folder>", clone_from="<user>/<model_name>")
+
+with repo.commit("Commit message", blocking=False):
+    # Save data
+
+last_command = repo.command_queue[-1]
+
+# Status of the push command
+last_command.status  
+# Will return the status code
+#     -> -1 will indicate the push is still ongoing
+#     -> 0 will indicate the push has completed successfully
+#     -> non-zero code indicates the error code if there was an error
+
+# if there was an error, the stderr may be inspected
+last_command.stderr
+
+# Whether the command finished or if it is still ongoing
+last_command.is_done
+
+# Whether the command errored-out.
+last_command.failed
+```
+
+When using `blocking=False`, the commands will be tracked and your script will exit only when all pushes are done, even
+if other errors happen in your script (a failed push counts as done).
 
 
 ### Need to upload very large (>5GB) files?
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -1,3 +1,4 @@
+import atexit
 import os
 import re
 import subprocess
@@ -6,7 +7,7 @@ import threading
 import time
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Iterator, List, Optional, Union
+from typing import Callable, Iterator, List, Optional, Tuple, Union
 
 from tqdm.auto import tqdm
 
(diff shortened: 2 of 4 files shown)


sgugger (approved):
Very nice! Thanks a lot for working on this!

src/huggingface_hub/repository.py
```diff
@@ -804,33 +872,87 @@ def git_commit(self, commit_message="commit files to HF hub"):
...
+
+        Args:
+            upstream (`str`, `optional`):
+                Upstream to which this should push. If not specified, will push
+                to the lastly defined upstream or to the default one (`origin main`).
+            blocking (`bool`, defaults to `True`):
+                Whether the function should return only when the push has finished.
+                Setting this to `False` will return an `CommandInProgress` object
+                which has an `is_done` property. This property will be set to
+                `True` when the push is finished.
```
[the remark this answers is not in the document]
LysandreJik: We use markdown!

src/huggingface_hub/README.md
```diff
@@ -236,6 +241,43 @@ Examples using the `commit` context manager:
...
+
+with repo.commit("Commit message", blocking=False):
+    # Save data
+
+last_command = repo.command_queue[-1]
+
+# Status of the push command
+last_command.status  
+# Will return the status code
+#     -> "running" will indicate the push is still ongoin
```
osanseviero, line 267: I'm not sure if mixing string with numbers is a good idea. Users will need to do non-intuitive handling if they want to rely in the status.

osanseviero:
```suggestion
#     -> "running" will indicate the push is still ongoing
```

src/huggingface_hub/README.md
```diff
@@ -236,6 +241,43 @@ Examples using the `commit` context manager:
...
+#     -> 0 will indicate the push has completed successfully
+#     -> non-zero code indicates the error code if there was an error
+
+# if there was an error, the stderr may be inspected
+last_command.stderr
```
osanseviero, line 272: no strong feeling, but since we have a `is_done` convenience method, maybe we could have a `failed` method that just checks if there's an error.

src/huggingface_hub/repository.py
```diff
@@ -344,6 +404,10 @@ def __init__(
         if revision is not None:
             self.git_checkout(revision, create_branch_ok=True)
 
+        # This ensures that all commands exit before exiting the Python runtime.
+        # This will ensure all pushes register on the hub, even if other errors happen in subsequent operations.
```
osanseviero, line 408: What happens if user tries to kill the process (for example ctrl + d) while things are getting pushed?

src/huggingface_hub/repository.py
```diff
@@ -397,7 +461,9 @@ def check_git_versions(self):
             )
         logger.info(git_version + "\n" + lfs_version)
 
-    def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = None):
+    def clone_from(
+        self, repo_url: str, use_auth_token: Union[bool, str, None] = None
+    ) -> bool:
```
osanseviero, line 466: Maybe I'm missing something, but where/why are we returning a bool here?

tests/test_repository.py
```diff
@@ -256,6 +256,110 @@ def test_add_commit_push(self):
...
+        with open(os.path.join(WORKING_REPO_DIR, "model.bin"), "w") as f:
+            f.write("hello")
+
+        repo.git_add()
+        repo.git_commit()
+        try:
+            url, result = repo.git_push(blocking=False)
+        except subprocess.CalledProcessError as exc:
+            print(exc.stderr)
+            raise exc
```
osanseviero, line 281: Why are we having a try/except here? This seems off for a test

osanseviero (approved):
This looks great! Thanks for the PR!

I mostly have some questions for my own learning/understanding :)

src/huggingface_hub/README.md
LysandreJik: Done!

src/huggingface_hub/repository.py
LysandreJik: If the user tries to kill the process, it will try to continue pushing:
```py
>>> repo.command_queue
[[push command, status code: running, in progress. PID: 95581]]
>>> (CTRL+C)
Waiting for the following commands to finish before shutting down: [[push command, status code: running, in progress. PID: 95581]].
```
If you try to exit again:
```
^CError in atexit._run_exitfuncs:
  File "/home/lysandre/Workspaces/Python/huggingface_hub/src/huggingface_hub/repository.py", line 1141, in wait_for_commands
    time.sleep(1)
KeyboardInterrupt
```
and then you're out!

src/huggingface_hub/repository.py
LysandreJik: Very nice catch, that's a mistake from my side

src/huggingface_hub/repository.py
osanseviero: That's perfect!

tests/test_repository.py
LysandreJik: Indeed, I suspect it was in an old test which got copy-pasted around. Fixed that in the last commit.

PR #317, opened 2021-09-07
Upgrade Espnet to latest community and use main repo

osanseviero:
This change fixes espnet inference since it was using an experimental fork before.


Narsil (approved):
LGTM. Models being slow is acceptable.

Maybe let's just add a comment so future testers won't be too surprised.

osanseviero:
The slowness issue is now fixed! Will merge this tomorrow. I did a copy of the GA workflow but it will be merged with the AllenNLP one in a follow-up

Narsil:
Nice, where is the slowness fix, can't find it here ?

osanseviero:
We now specify `beam_size=1` in ASR which makes inference considerably faster

PR #318, opened 2021-09-07
Add code snippet for speechbrain audio classification

osanseviero:


widgets/src/lib/interfaces/Libraries.ts
```diff
@@ -101,6 +101,13 @@ model = Speech2Text.from_pretrained(
 speech, rate = soundfile.read("speech.wav")
 text, *_ = model(speech)`;
 
+const espnetAudioClassification = (model: ModelData) =>
```
mishig25, line 104: this has to be `const speechbrainAudioClassification`, unless speechbrain is related to espnet?

osanseviero: Done, I messed working between two libraries at same time.

widgets/src/lib/interfaces/Libraries.ts
```diff
@@ -189,6 +189,13 @@ nlp = spacy.load("${nameWithoutNamespace(model.modelId)}")
 import ${nameWithoutNamespace(model.modelId)}
 nlp = ${nameWithoutNamespace(model.modelId)}.load()`;
 
+const speechbrainAudioClassification = (model: ModelData) =>
+`from speechbrain.pretrained import EncoderClassifier
+
+model = EncoderClassifier.from_hparams(
+  "${model.modelId}"
+)`;
+
```
mishig25, line 198: Nit: as in other speechbrain examples [here](https://github.com/huggingface/huggingface_hub/blob/b32ccaa4ab32d5b5ca40273d8a9f321332953de6/widgets/src/lib/interfaces/Libraries.ts#L199-L215),

does it need to follow the format:
```py
from speechbrain.pretrained import EncoderClassifier
audioclassification_model = EncoderClassifier.from_hparams("${model.modelId}") 
audioclassification_model.method("file.wav")
```

osanseviero: Done

mishig25 (approved):
lgtm!

PR #319, opened 2021-09-07
added option to create private repo

philschmid:
# What does this PR do? 

This PR adds the `repo_visibility` parameter to the `Repository` method to be able to create a private repository when pushing to the hub. 

usage:
```python
    repo = Repository(
        src_dir,
        clone_from=f"{user}/{repository_id}",
        use_auth_token=use_auth_token,
        repo_visibility="private"
    )
    repo.push_to_hub("Adding Infinity model to HF hub")
```

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -266,11 +266,11 @@ class Repository:
         local_dir: str,
         clone_from: Optional[str] = None,
         repo_type: Optional[str] = None,
-        repo_visibility: Optional[str] = None,
         use_auth_token: Union[bool, str] = True,
         git_user: Optional[str] = None,
         git_email: Optional[str] = None,
         revision: Optional[str] = None,
+        private: Optional[bool] = None,
     ):
         """
         Instantiate a local clone of a git repo.
@@ -290,8 +290,6 @@ class Repository:
                 repository url (e.g. ``'https://huggingface.co/philschmid/playground-tests'``).
             repo_type (``str``, `optional`):
                 To set when creating a repo: et to "dataset" or "space" if creating a dataset or space, default is model.
-            repo_visibility (``str``, `optional`):
-                visibility of the repository, needed for creating private repositories.
             use_auth_token (``str`` or ``bool``, `optional`, defaults ``None``):
                 huggingface_token can be extract from ``HfApi().login(username, password)`` and is used to authenticate against the hub
                 (useful from Google Colab for instance).
@@ -302,12 +300,14 @@ class Repository:
             revision (``str``, `optional`):
                 Revision to checkout after initializing the repository. If the revision doesn't exist, a
                 branch will be created with that revision name from the default branch's current HEAD.
+            private (``bool``, `optional`):
+                whether the repository is private or not.
         """
 
         os.makedirs(local_dir, exist_ok=True)
         self.local_dir = os.path.join(os.getcwd(), local_dir)
         self.repo_type = repo_type
-        self.repo_visibility = "public" if repo_visibility is None else repo_visibility
+        self.private = private
 
         self.check_git_versions()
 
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/repository.py
```diff
@@ -266,6 +266,7 @@ def __init__(
         local_dir: str,
         clone_from: Optional[str] = None,
         repo_type: Optional[str] = None,
+        repo_visibility: Optional[str] = None,
```
Narsil, line 269: Just happene d to have notifications for this repo.

Isn't `repo_visibility: str = "public"` much better than `Optional[str]` ? Strings are immutable so it's fine I think.

src/huggingface_hub/repository.py
```diff
@@ -431,12 +436,14 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
                 repo_url = repo_url.replace("https://", f"https://user:{token}@")
 
                 if namespace == user or namespace in valid_organisations:
+                    is_repo_private = True if repo_visibility == "private" else False
```
Narsil, line 439: Is that variable a string or a boolean, it seems pretty odd to switch from one to the other.

src/huggingface_hub/repository.py
julien-c: why not use an optional bool instead?

julien-c: (and name it the same as in `api.create_repo`)

philschmid: My intention was to have a similar UX and API to what you have when using the UI. There we have repository visibility with `"public"` & `"private"`. But not a strong opinion on this and happy to convert to the same as in `api.create_repo`.

julien-c: i'll defer to @LysandreJik and/or @osanseviero but personally i think it's better to have a bool

src/huggingface_hub/repository.py
```diff
@@ -404,6 +408,7 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
         If this folder is a git repository with linked history, will try to update the repository.
         """
         token = use_auth_token if use_auth_token is not None else self.huggingface_token
+        repo_visibility = "public" if token is None else self.repo_visibility
```
osanseviero, line 411: Maybe it's better to throw an error here so users don't accidentally publish a repo publicly.

src/huggingface_hub/repository.py
osanseviero: I think `private: Optional[bool] = None` would be better. That way we avoid having to change from string to bool and we're consistent with `.create_repo`.

src/huggingface_hub/repository.py
philschmid: Added the error to the top

src/huggingface_hub/repository.py
```diff
@@ -404,6 +408,10 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
         If this folder is a git repository with linked history, will try to update the repository.
         """
         token = use_auth_token if use_auth_token is not None else self.huggingface_token
+        if token is None and self.private is True:
```
LysandreJik, line 411:
```suggestion
        if token is None and self.private:
```

LysandreJik (approved):
LGTM

src/huggingface_hub/repository.py
```diff
@@ -404,6 +408,10 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
         If this folder is a git repository with linked history, will try to update the repository.
         """
         token = use_auth_token if use_auth_token is not None else self.huggingface_token
+        if token is None and self.private:
+            raise ValueError(
+                "You need to provide a Hugging Face Token in `use_auth_token` to work with private repositories"
```
osanseviero, line 413: nit: I think providing the token is more of an edge case for users working in colab or certain scenarios. Maybe we should mention that users should first login with `huggingface-cli login`  or manually provide a token.

osanseviero (approved):
Looks good! Thanks a lot for this!

src/huggingface_hub/repository.py
```diff
@@ -270,6 +270,7 @@ def __init__(
         git_user: Optional[str] = None,
         git_email: Optional[str] = None,
         revision: Optional[str] = None,
+        private: Optional[bool] = None,
```
Narsil, line 273: Sorry to insist, by `private: bool = False` seems strictly better.

The `None` is necessary when the default argument is mutable, other it's better to use a real default, no ?

osanseviero (approved):
LG! Thanks

PR #320, opened 2021-09-07
SDK version and Gradio latest

cbensimon:


julien-c (commented):
is our goal to also enable this for Gradio at some point? 

Or do we want to document how to install a specific version of `pip install gradio`from the Space itself?

(fyi @abidlabs)

julien-c:
Otherwise looks great to me but I'll let @osanseviero chime in!

osanseviero (approved):
This looks great!

As @julien-c suggested, we might want to mention how to install another version of Gradio, although if we plan to support `sdk_version` metadata for Gradio soon I guess we could skip that.

Such as: 
```
os.system('pip install gradio==2.3.0a0')
``` 
(and mention its cons).

cbensimon:
> is our goal to also enable this for Gradio at some point?

Yes we will at some point, so I think we don't have to document a temporary hacky solution.

Plus, at this stage, Gradio versions seem to only add features rather than bringing breaking changes so IMO it's unlikely that someone would want something else than latest. For now at least

PR #322, opened 2021-09-07
:bug: Fix timm code sample

nateraw:
In `timm`, you must add the prefix `hf_hub:` before any Hugging Face model identifier when using `timm.create_model`. 

The code sample currently being shown on the hub is incorrect, and will fail when you run it.


osanseviero (approved):
Looks good!

PR #323, opened 2021-09-08
Tags

AngledLuffa:
Add tags & a function to check if a repo is currently clean or not

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -1032,6 +1032,100 @@ class Repository:
                 except subprocess.CalledProcessError as exc:
                     raise EnvironmentError(exc.stderr)
 
+    def tag_exists(self, tag_name: str) -> bool:
+        """
+        Check if a tag exists or not
+        """
+        try:
+            git_tags = subprocess.run(
+                ["git", "tag"],
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                check=True,
+                encoding="utf-8",
+                cwd=self.local_dir,
+            ).stdout.strip()
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        git_tags = git_tags.split("\n")
+        return tag_name in git_tags
+
+    def delete_tag(self, tag_name: str, remote: str = "origin") -> bool:
+        """
+        Delete a tag, both local and remote, if it exists
+
+        Return True if deleted.  Returns False if the tag didn't exist
+        If remote is None, will just be updated locally
+        """
+        if not self.tag_exists(tag_name):
+            return False
+
+        try:
+            subprocess.run(
+                ["git", "tag", "-d", tag_name],
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                check=True,
+                encoding="utf-8",
+                cwd=self.local_dir,
+            ).stdout.strip()
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        if remote:
+            try:
+                subprocess.run(
+                    ["git", "push", remote, "--delete", tag_name],
+                    stderr=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    check=True,
+                    encoding="utf-8",
+                    cwd=self.local_dir,
+                ).stdout.strip()
+            except subprocess.CalledProcessError as exc:
+                raise EnvironmentError(exc.stderr)
+
+        return True
+
+    def add_tag(self, tag_name: str, message: str = None, remote: str = "origin"):
+        """
+        Add a tag at the current head and push it
+
+        If remote is None, will just be updated locally
+
+        If no message is provided, the tag will be lightweight.
+        if a message is provided, the tag will be annotated.
+        """
+        if message:
+            tag_args = ["git", "tag", "-a", tag_name, "-m", message]
+        else:
+            tag_args = ["git", "tag", tag_name]
+        try:
+            subprocess.run(
+                tag_args,
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                check=True,
+                encoding="utf-8",
+                cwd=self.local_dir,
+            ).stdout.strip()
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        if remote:
+            try:
+                subprocess.run(
+                    ["git", "push", "origin", tag_name],
+                    stderr=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    check=True,
+                    encoding="utf-8",
+                    cwd=self.local_dir,
+                ).stdout.strip()
+            except subprocess.CalledProcessError as exc:
+                raise EnvironmentError(exc.stderr)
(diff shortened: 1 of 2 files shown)


julien-c:
(Note for reviewers: i approved the workflows to get CI to run)

src/huggingface_hub/repository.py
```diff
@@ -869,14 +869,118 @@ def git_checkout(self, revision, create_branch_ok=False):
...
+                cwd=self.local_dir,
+            ).stdout.strip()
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        return True
+
+    def add_tag(self, tag_name: str, message: str):
+        """
+        Add a tag at the current head and push it
```
julien-c, line 928: git has two types of tags: `lightweight` and `annotated`. Do we also want to support lightweight tags (defaulting to annotated?) for completeness? Or we don't care?

AngledLuffa: I didn't care too much, but I could see renaming the method `add_annotated_tag` or just leaving message None to indicate lightweight

LysandreJik: I think having `message` as an optional string (lightweight when `None` and annotated when a message is specified would be super nice!

src/huggingface_hub/repository.py
```diff
@@ -869,14 +869,119 @@ def git_checkout(self, revision, create_branch_ok=False):
...
+                encoding="utf-8",
+                cwd=self.local_dir,
+            ).stdout.strip()
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
+        git_tags = git_tags.split("\n")
+        return tag_name in git_tags
+
+    def delete_tag(self, tag_name: str) -> bool:
```
LysandreJik, line 891: Having a `remote` argument here that would default to `"origin"` would be welcome!

src/huggingface_hub/repository.py
```diff
@@ -869,14 +869,119 @@ def git_checkout(self, revision, create_branch_ok=False):
...
+    ) -> str:
         """
         Helper to add, commit, and push files to remote repository on the HuggingFace Hub.
         Will automatically track large files (>10MB).
 
         Args:
             commit_message: commit message.
         """
+        if clean_ok and self.is_repo_clean():
+            logger.info("Repo currently clean.  Ignoring push_to_hub")
```
LysandreJik, line 983: Did you run into issues with a git repository that has nothing to commit?

LysandreJik (commented):
This looks great! Added a few comments, but it looks great.

It would be very nice to add a few tests - happy to handle that for you if you'd like! They live in `tests/test_repository.py`.

src/huggingface_hub/repository.py
AngledLuffa: Can do

src/huggingface_hub/repository.py
AngledLuffa: Yes, that would especially happen when programmatically adding ".zip" files to a repo that already had ".zip" as part of lfs, for example

src/huggingface_hub/repository.py
AngledLuffa: Sure

LysandreJik (approved):
This looks good to me, thank you @AngledLuffa! Will release a new version of `huggingface_hub` today so that you may leverage it.

LysandreJik:
There are some code quality issues, do you mind running `pip install -e .[quality]` followed by `make quality` at the root of your clone to patch them?

Regarding tests, please let me know if you'd like to handle them - and if not, if I can push on your branch directly to update them!

AngledLuffa:
> This looks good to me, thank you @AngledLuffa! Will release a new version of `huggingface_hub` today so that you may leverage it.

Thank you!  But there's no particular time horizon needed.  I pip installed my own branch in order to use my stuff

AngledLuffa:
> There are some code quality issues, do you mind running `pip install -e .[quality]` followed by `make quality` at the root of your clone to patch them?

Sure, sorry, that was sloppy of me.  I cut & paste the basis for the subprocess call

> Regarding tests, please let me know if you'd like to handle them - and if not, if I can push on your branch directly to update them!

It would certainly save me some time if you did that!  Both in terms of making the tests, and in terms of acquiring the familiarity for making tests (presumably there's some git repo mocking being done?)

AngledLuffa:
Also, what do you think of the TODO in push_to_hub?

```
        if clean_ok and self.is_repo_clean():
            logger.info("Repo currently clean.  Ignoring push_to_hub")
            return ""  # TODO: is there something else we should return here?
```

LysandreJik:
I have opened a PR here with tests: https://github.com/AngledLuffa/huggingface_hub/pull/1

LysandreJik:
> Also, what do you think of the TODO in push_to_hub?
> 
> ```
>         if clean_ok and self.is_repo_clean():
>             logger.info("Repo currently clean.  Ignoring push_to_hub")
>             return ""  # TODO: is there something else we should return here?
> ```

I think returning `None` here and switching the return type to `Optional[str]` might be better, what do you think? This should be reflected in the documentation too

AngledLuffa:
I think there's some problems with making the tags local only by default.
If tags aren't kept concurrent between local and remote, then what
mechanism is there for pushing them?  We would also need to add a separate
mechanism for adding or deleting tags remotely after those changes are
already made locally.

I'm thinking of the following sequence of events:

user adds a tag "foo" locally but not remotely
user makes and pushes a bunch of changes
user then tries to add a tag "foo" remotely

Does this update the local tag?  Does it push the local tag to the remote?
Or does it create a new tag on the remote with the same name but a
different hash?

LysandreJik:
I understand your concerns @AngledLuffa! The `Repository` is a manager for repositories both local and remote, so in my opinion, it doesn't really make sense to implement something remote-only.

In a standard git workflow, when you add a tag locally and push without the `--tags` flag, then the tag isn't pushed - and if you later commit and then try to add this tag again, you're met with a `tag xxx already exists`. I think we're mimicking the behavior of git quite nicely here.

If the user wants to push a tag to the remote, they'll then have to first delete their local tag, then add it again and push it. There's no way to push a previously defined tag - so for this we could add a `push_tag(tag: str)` method. What do you think?

AngledLuffa:
Sure, that logic works for me.  push_tag might be superfluous for now

LysandreJik:
I understand. Is this good to merge for you?

AngledLuffa:
Sure, that works for me!  I merged your change into my change, after all.
Thanks!

LysandreJik:
Fantastic! Will merge as soon as the tests are green.

AngledLuffa:
Seems the tests didn't pass.  Also, there was still this to address:

> I think returning None here and switching the return type to Optional[str] might
be better, what do you think? This should be reflected in the documentation
too

LysandreJik:
Ah yes the tests didn't pass but that's due to a network error; the network of the staging we're using is unfortunately flaky.

You're correct; I had forgotten about this TODO. Would you like to take care of it or should I go ahead and implement it?

AngledLuffa:
Sure, I went ahead and did it

LysandreJik:
Looks good to me @AngledLuffa!

PR #326, opened 2021-09-08
display git push warnings

elishowk:
- git push errors rejections should be displayed well, as discussed with @LysandreJik, we'll verify this once https://github.com/huggingface/moon-landing/pull/958 is in production.
- **But** git push will also output model-card warnings (but git push accepted) in stderr thatn we want huggingface_hub to display. cf. https://github.com/huggingface/moon-landing/pull/958


LysandreJik (approved):
Thank you for the changes! Looks good to me.

julien-c:
🎉!

PR #329, opened 2021-09-08
Add a widget to login in notebook

sgugger:
This PR adds a widget to be able to login to the HF website from a Jupyter Notebook. The command
```
! huggingface-cli login
```
does not work in a Jupyter notebook, as one cannot write in the output of the cell (it does work in a colab however).

With this PR, a user can do:
```
from huggingface_hub import notebook_login

notebook_login()
```
and they will get the following:
image

Filling the username and password will then login like the `huggingface-cli login` command.

### DIFF
--- a/src/huggingface_hub/__init__.py
+++ b/src/huggingface_hub/__init__.py
@@ -18,6 +18,7 @@
 
 __version__ = "0.0.16"
 
+from .commands.user import notebook_login
 from .constants import (
     CONFIG_NAME,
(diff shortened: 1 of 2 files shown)


osanseviero:
What kind of magic is this? I love it!!

LysandreJik (approved):
Wonderful! Tested o ncolab, works well. LGTM, thank you for working on it @sgugger

PR #330, opened 2021-09-08
Have our own List of license identifiers

julien-c:
Context/goal: have a stable link on our doc to link to, in the context of the upcoming metadata validation

Question: should we split that list into its own page? (i feel like its ok like this)

### DIFF
--- a/docs/hub/model-repos.md
+++ b/docs/hub/model-repos.md
@@ -47,7 +47,7 @@ You can find the detailed specification [here](https://github.com/huggingface/hu
 
 Some useful information on them:
 * All the tags can be used to filter the list of models on https://huggingface.co/models.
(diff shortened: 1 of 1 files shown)


osanseviero (approved):
Looks good! Thanks for the PR. I have a couple of comments/questions

1. Will we enforce the license tag to contain one of these? 

2. IMO it's not worth adding a new doc page for this. 

3. I feel that we should have some UI element to simplify adding these tags. For example, there could be some "add license" + dropdown somewhere in the UI. 

4. Maybe we should link this doc from https://raw.githubusercontent.com/huggingface/huggingface_hub/main/modelcard.md for the license tag

julien-c:
> 1. Will we enforce the license tag to contain one of these?

license can be unspecified but if it's there, yes it will be enforced that it's a valid identifier (we'll fix existing models)

> 3. UI selector on the Web UI

yes, on the roadmap!

elishowk (approved):

Pierrci (approved):

PR #339, opened 2021-09-09
Repo metadata load and save

### LINKED ISSUE #54: Feature Request: Have programatic way of adding metadata to a repo
If a user creates dozens of repos programmatically, and then realized they forgot adding some tag or any other info for the metadata, they will have to either go through each of them manually, re-generate the README.md, or change the README.md content manually which can be error-prone.

julien-c:
Tentatively close #54 

Example use case:
- goal: updating all licenses for Helsinki-NLP models with no license, to `apache-2.0` – see https://github.com/huggingface/transformers/issues/13328#issuecomment-916455471
- example script: https://gist.github.com/julien-c/b2dcde5df5d5e41ad7c4b594cb54aba3

### DIFF
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -13,7 +13,7 @@ TF2_WEIGHTS_NAME = "tf_model.h5"
 TF_WEIGHTS_NAME = "model.ckpt"
 FLAX_WEIGHTS_NAME = "flax_model.msgpack"
 CONFIG_NAME = "config.json"
-MODELCARD_NAME = "README.md"
+REPOCARD_NAME = "README.md"
 
 HUGGINGFACE_CO_URL_HOME = "https://huggingface.co/"
 
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -13,6 +13,7 @@ TF2_WEIGHTS_NAME = "tf_model.h5"
 TF_WEIGHTS_NAME = "model.ckpt"
 FLAX_WEIGHTS_NAME = "flax_model.msgpack"
 CONFIG_NAME = "config.json"
+MODELCARD_NAME = "README.md"
 
 HUGGINGFACE_CO_URL_HOME = "https://huggingface.co/"
 
--- a/src/huggingface_hub/repocard.py
+++ b/src/huggingface_hub/repocard.py
@@ -25,7 +25,7 @@ def metadata_load(local_path: Union[str, Path]) -> Optional[Dict]:
 
 def metadata_save(local_path: Union[str, Path], data: Dict) -> None:
     data_yaml = yaml.dump(data, sort_keys=False)
-    # keep dict order
+    # sort_keys: keep dict order
     content = Path(local_path).read_text() if Path(local_path).is_file() else ""
     match = REGEX_YAML_BLOCK.search(content)
     if match:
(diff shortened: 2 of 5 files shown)


sgugger (commented):
That's definitely useful, thanks a lot for adding this!

osanseviero:
Thanks so much for this! This is a very needed upgrade! ❤️‍🔥 

This will be very useful for all our 3rd party integrations if we ever need to modify all the repos. We should communicate about this PR once merged to all big libraries and orgs (e.g. Stanford NLP)

julien-c:
(current status: spamming the CI to try and get all transient errors to go green 😅)

src/huggingface_hub/repocard.py
```diff
@@ -0,0 +1,38 @@
+import re
+from pathlib import Path
+from typing import Dict, Optional, Union
+
+import yaml
+
+
+REGEX_YAML_BLOCK = re.compile(r"---[\n\r]+([\S\s]*?)[\n\r]+---[\n\r]")
+# exact same regex as in the hf hub server. Please keep in sync.
```
osanseviero, line 9: nit:

```suggestion
# exact same regex as in the Hub server. Please keep in sync.
REGEX_YAML_BLOCK = re.compile(r"---[\n\r]+([\S\s]*?)[\n\r]+---[\n\r]")
```

Also it would be good to add similar docstring in the Hub code

osanseviero (approved):
Good stuff!

src/huggingface_hub/constants.py
```diff
@@ -13,6 +13,7 @@
 TF_WEIGHTS_NAME = "model.ckpt"
 FLAX_WEIGHTS_NAME = "flax_model.msgpack"
 CONFIG_NAME = "config.json"
+MODELCARD_NAME = "README.md"
```
LysandreJik, line 16: You're setting `MODELCARD_NAME` here but using the `repocard` term elsewhere - what should it be? I would favor having all instances be `REPOCARD_NAME` if you're looking for a repo-type-agnostic name

src/huggingface_hub/repocard.py
```diff
@@ -0,0 +1,38 @@
...
+    match = REGEX_YAML_BLOCK.search(content)
+    if match:
+        yaml_block = match.group(1)
+        data = yaml.safe_load(yaml_block)
+        if isinstance(data, dict):
+            return data
+        else:
+            raise ValueError("repo card metadata block should be a dict")
+    else:
+        return None
```
LysandreJik, line 23: This should probably raise an error too, right? Running `metadata_load` on any random file will return nothing while I would expect it to fail if it doesn't find any metadata

LysandreJik: If that's the expected behavior then I would also put it in a test to ensure that it doesn't switch to raising an error in the future

LysandreJik (approved):
Thank you for working on that @julien-c, that's very useful!

As seen with @elishowk this morning it would be nice to eventually have the same validation here as what is enforced in the backend. `transformers` could then leverage it.

src/huggingface_hub/repocard.py
julien-c: as discussed offline we have a lot of repo cards (model cards, etc) that do not have a yaml block, and i think the expected behavior here is to return None. If need be, we might modify this in a subsequent PR

julien-c:
@LysandreJik attempted to solve merge conflict

This should be ready to merge, depending on final maintainer acceptance (😘 )

elishowk (approved):
LGTM

LysandreJik:
Will push an additional test and merge in a bit.

LysandreJik:
Thank you for your work @julien-c

elishowk:
> (current status: spamming the CI to try and get all transient errors to go green sweat_smile)

Note about that issue I witnessed working on PRs : huggingface_hub and transformers CI often fails because of moon-staging 504 timeouts. If needed, I can work on strengthening this staging server's capacities

PR #340, opened 2021-09-10
Add auth to snapshot download

lewtun:
This PR adds a `use_auth_token` argument to `snapshot_download` so private repos can be downloaded from the Hub. 

## Usage

```python

# Download with explicit token
filepath = snapshot_download("some-private-repo", use_auth_token="api_XXX")
# Download with token from cache
filepath = snapshot_download("some-private-repo", use_auth_token=True)
```

### DIFF
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -18,6 +18,7 @@ def snapshot_download(
     library_name: Optional[str] = None,
     library_version: Optional[str] = None,
     user_agent: Union[Dict, str, None] = None,
+    token: Union[str, None] = None,
 ) -> str:
     """
     Downloads a whole snapshot of a repo's files at the specified revision.
@@ -42,22 +43,16 @@ def snapshot_download(
         cache_dir = str(cache_dir)
 
     _api = HfApi()
-    model_info = _api.model_info(repo_id=repo_id, revision=revision)
+    model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
 
-    storage_folder = os.path.join(
-        cache_dir, repo_id.replace("/", REPO_ID_SEPARATOR) + "." + model_info.sha
-    )
+    storage_folder = os.path.join(cache_dir, repo_id.replace("/", REPO_ID_SEPARATOR) + "." + model_info.sha)
 
     for model_file in model_info.siblings:
-        url = hf_hub_url(
-            repo_id, filename=model_file.rfilename, revision=model_info.sha
-        )
+        url = hf_hub_url(repo_id, filename=model_file.rfilename, revision=model_info.sha)
         relative_filepath = os.path.join(*model_file.rfilename.split("/"))
 
         # Create potential nested dir
-        nested_dirname = os.path.dirname(
-            os.path.join(storage_folder, relative_filepath)
-        )
+        nested_dirname = os.path.dirname(os.path.join(storage_folder, relative_filepath))
         os.makedirs(nested_dirname, exist_ok=True)
 
         path = cached_download(
@@ -67,6 +62,7 @@ def snapshot_download(
             library_name=library_name,
             library_version=library_version,
             user_agent=user_agent,
+            use_auth_token=token,
         )
 
         if os.path.exists(path + ".lock"):
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -45,14 +45,20 @@ def snapshot_download(
     _api = HfApi()
     model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
 
-    storage_folder = os.path.join(cache_dir, repo_id.replace("/", REPO_ID_SEPARATOR) + "." + model_info.sha)
+    storage_folder = os.path.join(
+        cache_dir, repo_id.replace("/", REPO_ID_SEPARATOR) + "." + model_info.sha
+    )
 
     for model_file in model_info.siblings:
-        url = hf_hub_url(repo_id, filename=model_file.rfilename, revision=model_info.sha)
+        url = hf_hub_url(
+            repo_id, filename=model_file.rfilename, revision=model_info.sha
+        )
         relative_filepath = os.path.join(*model_file.rfilename.split("/"))
 
         # Create potential nested dir
-        nested_dirname = os.path.dirname(os.path.join(storage_folder, relative_filepath))
+        nested_dirname = os.path.dirname(
+            os.path.join(storage_folder, relative_filepath)
+        )
         os.makedirs(nested_dirname, exist_ok=True)
 
         path = cached_download(
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -4,7 +4,7 @@ from typing import Dict, Optional, Union
 
 from .constants import HUGGINGFACE_HUB_CACHE
 from .file_download import cached_download, hf_hub_url
-from .hf_api import HfApi
+from .hf_api import HfApi, HfFolder
 
 
 REPO_ID_SEPARATOR = "__"
@@ -18,7 +18,7 @@ def snapshot_download(
     library_name: Optional[str] = None,
     library_version: Optional[str] = None,
     user_agent: Union[Dict, str, None] = None,
-    token: Union[str, None] = None,
+    use_auth_token: Union[bool, str, None] = None,
 ) -> str:
     """
     Downloads a whole snapshot of a repo's files at the specified revision.
@@ -41,6 +41,13 @@ def snapshot_download(
         cache_dir = HUGGINGFACE_HUB_CACHE
     if isinstance(cache_dir, Path):
         cache_dir = str(cache_dir)
+
+    if use_auth_token:
+        token = HfFolder.get_token()
+        if token is None:
+            raise EnvironmentError(
+                "You specified use_auth_token=True, but a Hugging Face token was not found."
+            )
 
     _api = HfApi()
     model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
@@ -68,7 +75,7 @@ def snapshot_download(
             library_name=library_name,
             library_version=library_version,
             user_agent=user_agent,
-            use_auth_token=token,
+            use_auth_token=use_auth_token,
         )
 
         if os.path.exists(path + ".lock"):
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -42,7 +42,9 @@ def snapshot_download(
     if isinstance(cache_dir, Path):
         cache_dir = str(cache_dir)
 
-    if use_auth_token == True:
+    if isinstance(use_auth_token, str):
+        token = use_auth_token
+    elif use_auth_token:
         token = HfFolder.get_token()
         if token is None:
             raise EnvironmentError(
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/snapshot_download.py
```diff
@@ -18,6 +18,7 @@ def snapshot_download(
     library_name: Optional[str] = None,
     library_version: Optional[str] = None,
     user_agent: Union[Dict, str, None] = None,
+    token: Union[str, None] = None,
```
lewtun, line 21: I felt it was simpler to pass an explicit token instead of `use_auth_token` and then retrieving it, but happy to adjust if the latter is preferred

osanseviero: In my opinion it's better to retrieve the token so users don't have to get it themselves (`HfFolder.get_token()` should help you have some simple logic).

osanseviero: Also that way we're consistent with `cached_download`, `Repository`,  etc.

LysandreJik: I would also favor a `use_auth_token` based on how it's handled in other methods!

LysandreJik (commented):
Thanks for working on the implementation @lewtun, this was an oversight!

src/huggingface_hub/snapshot_download.py
lewtun: Thanks for the feedback @osanseviero @LysandreJik 🙌 !

I've refactored the PR to use `use_auth_token` instead :)

src/huggingface_hub/snapshot_download.py
```diff
@@ -41,8 +42,17 @@ def snapshot_download(
     if isinstance(cache_dir, Path):
         cache_dir = str(cache_dir)
 
+    if isinstance(use_auth_token, str):
```
lewtun, line 45: Here I aim to cover both cases where the user provides an explicit API token or we retrieve it from the cache

src/huggingface_hub/snapshot_download.py
```diff
@@ -41,8 +42,19 @@ def snapshot_download(
...
+        token = use_auth_token
+    elif use_auth_token:
+        token = HfFolder.get_token()
+        if token is None:
+            raise EnvironmentError(
+                "You specified use_auth_token=True, but a Hugging Face token was not found."
+            )
```
osanseviero, line 52: Not for this PR as it's unrelated, but maybe we should refactor this into a function (`get_token_or_raise`?) or add an optional param to error if no token. We have logic similar to this in multiple laces at the moment:

* https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/file_download.py#L331-L336
* https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/repository.py#L317
* + some other third party libraries for which a token is strictly required

lewtun: Yeah I agree this would be a nice improvement - happy to tackle it after this PR is merged (so I can make progress on SUPERB :))

lewtun:
I'm not sure why the unit tests are failing since they seem to be unrelated to the changes in this PR 🤔

src/huggingface_hub/snapshot_download.py
osanseviero: I agree with this!

osanseviero (commented):
Re: failing tests, some are unfortunately flaky at the moment :(  You can try to re-run the workflows.

Code looks good to me, it would be great if you add some unit tests. You can use  `private` in `Repository` to make a private repo. Then you could have a test without token that shows that the command fails, and another one with token that shows it works.

tests/test_snapshot_download.py
```diff
@@ -22,36 +24,52 @@ def setUpClass(cls):
...
+        with repo_public.commit("Add file to main branch"):
             with open("dummy_file.txt", "w+") as f:
                 f.write("v2")
             with open("dummy_file_2.txt", "w+") as f:
                 f.write("v3")
 
-        self.second_commit_hash = repo.git_head_hash()
+        self.second_commit_hash = repo_public.git_head_hash()
+
+        repo_private = Repository(
```
lewtun, line 49: There's a bit of code duplication here, but I couldn't see a way to convert an existing public repo to private directly with `huggingface_hub`

tests/test_snapshot_download.py
```diff
@@ -3,12 +3,14 @@
...
 from huggingface_hub.snapshot_download import snapshot_download
 from tests.testing_constants import ENDPOINT_STAGING, PASS, USER
 
 
-REPO_NAME = "dummy-hf-hub-{}".format(int(time.time() * 10e3))
+REPO_NAME_PUBLIC = "dummy-hf-hub-{}".format(int(time.time() * 10e3))
+REPO_NAME_PRIVATE = "dummy-hf-hub-{}".format(int(time.time() * 10e3))
 
 
 class SnapshotDownloadTests(unittest.TestCase):
```
lewtun, line 16: For some reason I am not able to run these tests locally due to a permissions error:

```
src/huggingface_hub/repository.py:322: in __init__
    self.clone_from(repo_url=clone_from)
src/huggingface_hub/repository.py:432: in clone_from
    whoami_info = api.whoami(token)
src/huggingface_hub/hf_api.py:330: in whoami
    r.raise_for_status()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Response [401]>

    def raise_for_status(self):
        """Raises :class:`HTTPError`, if one occurred."""
    
        http_error_msg = ''
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode('utf-8')
            except UnicodeDecodeError:
                reason = self.reason.decode('iso-8859-1')
        else:
            reason = self.reason
    
        if 400 <= self.status_code < 500:
            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        elif 500 <= self.status_code < 600:
            http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/api/whoami-v2
```

Is there some environment variables that need to be set when testing locally?

tests/test_snapshot_download.py
osanseviero: You can use `HfApi().update_repo_visibility` to make the repo private.

https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L638

tests/test_snapshot_download.py
```diff
@@ -88,3 +106,29 @@ def test_download_model(self):
...
+
+            with open(os.path.join(storage_folder, "dummy_file.txt"), "r") as f:
+                contents = f.read()
+                self.assertEqual(contents, "v1")
+
+            # folder name contains the revision's commit sha.
+            self.assertTrue(self.private_commit_hash in storage_folder)
+
+        # Test download fails without token
+        self._api.logout(self._token)
```
lewtun, line 132: This feels a bit hacky, but is needed because we define a `token` for all tests in `SnapshotDownloadTests`. The alternative would be to pass a dummy token like `api_XXX` to `snapshot_download`, but that doesn't test the default scenario where `use_auth_token=None`. Finally we could move these private repo tests to another test class, but that seems like overkill IMO

tests/test_snapshot_download.py
osanseviero: Yes, I think you need to set HUGGINGFACE_CO_STAGING = yes

osanseviero: Note you should change it back since it will impact `huggingface_hub` in general

tests/test_snapshot_download.py
lewtun: Thanks for the tip @osanseviero ! This is much simpler indeed :)

tests/test_snapshot_download.py
lewtun: Thanks for the tip!

lewtun:
> Code looks good to me, it would be great if you add some unit tests. You can use `private` in `Repository` to make a private repo. Then you could have a test without token that shows that the command fails, and another one with token that shows it works.
Thanks for the tips and suggestions - the new unit test should cover the three possible values of `use_auth_token` in `snapshot_download`

tests/test_snapshot_download.py
```diff
@@ -88,3 +90,60 @@ def test_download_model(self):
 
             # folder name contains the revision's commit sha.
             self.assertTrue(self.first_commit_hash in storage_folder)
+
+    def test_download_private_model(self):
```
osanseviero, line 94: I think this text can be greatly simplified. There's no need to download the repositories, you can use `HfApi.model_info` with and without the token and see if you get an error or if you can get info from the repo.

tests/test_snapshot_download.py
```diff
@@ -88,3 +90,60 @@ def test_download_model(self):
...
+        self._api.update_repo_visibility(self._token, REPO_NAME, private=True)
```
osanseviero, line 95: You might want to update it back to public no?

tests/test_snapshot_download.py
lewtun: OK sounds good, but then I think the test should be moved to `test_hf_api.py` since we'd no longer be testing a call of `snapshot_download`. I'll push a change so you can see what I mean

tests/test_snapshot_download.py
lewtun: Good catch!

tests/test_hf_api.py
```diff
@@ -450,6 +451,28 @@ def test_token_workflow(self):
         self.assertEqual(HfFolder.get_token(), None)
 
 
+class HfApiPrivateTest(HfApiCommonTestWithLogin):
```
lewtun, line 454: @osanseviero here is a new unit test for probing the model info with private repos. 

I'm not sure if this is exactly what you had in mind, but if it is OK I'll remove the existing test from `snapshot_download`

tests/test_hf_api.py
```diff
@@ -450,6 +451,28 @@ def test_token_workflow(self):
...
+    def setUp(self) -> None:
+        super().setUp()
+        self._api.create_repo(token=self._token, name=REPO_NAME)
+        self._api.update_repo_visibility(
+            token=self._token, name=REPO_NAME, private=True
+        )
+
+    def tearDown(self) -> None:
+        self._api.delete_repo(token=self._token, name=REPO_NAME)
```
lewtun, line 463: I opted to create / delete a new repo for this class of tests

tests/test_snapshot_download.py
osanseviero: Discussed offline since there was a misunderstanding from my side

tests/test_snapshot_download.py
lewtun: Done!

osanseviero (approved):
Looks good to me! Thanks a lot

lewtun:
@LysandreJik this PR is ready for a final pass / approval on your side :)

src/huggingface_hub/snapshot_download.py
LysandreJik: I agree as well!

LysandreJik (approved):
Looks great, thank you for iterating @lewtun!

LysandreJik:
Will do a release containing this commit later today.

PR #341, opened 2021-09-10
Use config for Speechbrain instead of the tags

osanseviero:
This PR changes the `speechbrain` Inference to use the class specified in the `config` instead of basing in the model card tag. This also upgrades the `speechbrain` version so it works with `EncoderASR`.
 
CC @TParcollet

### DIFF
--- a/.github/workflows/python-api-speechbrain.yaml
+++ b/.github/workflows/python-api-speechbrain.yaml
@@ -1,0 +1,29 @@
+# TODO: Merge this with allenNLP to have a single workflow for all docker images.
+name: speechbrain-docker
+
+on:
+  pull_request:
+    paths:
+      - "api-inference-community/docker_images/speechbrain/**"
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Install dependencies
+        working-directory: api-inference-community
+        run: |
+          pip install --upgrade pip
+          pip install pytest pillow httpx
+          pip install -e .
+      - run: RUN_DOCKER_TESTS=1 pytest -sv tests/test_dockers.py::DockerImageTests::test_speechbrain
+        working-directory: api-inference-community
--- a/api-inference-community/docker_images/speechbrain/app/pipelines/audio_classification.py
+++ b/api-inference-community/docker_images/speechbrain/app/pipelines/audio_classification.py
@@ -2,12 +2,17 @@ from typing import Dict, List
 
 import numpy as np
 import torch
+from app.common import ModelType, get_type
 from app.pipelines import Pipeline
 from speechbrain.pretrained import EncoderClassifier
 
 
 class AudioClassificationPipeline(Pipeline):
     def __init__(self, model_id: str):
+        model_type = get_type(model_id)
+        if model_type != ModelType.ENCODERCLASSIFIER:
+            raise ValueError(f"{model_type.value} is invalid for audio-classification")
+
         self.model = EncoderClassifier.from_hparams(source=model_id)
 
         self.top_k = 5
--- a/api-inference-community/docker_images/speechbrain/app/pipelines/audio_to_audio.py
+++ b/api-inference-community/docker_images/speechbrain/app/pipelines/audio_to_audio.py
@@ -1,50 +1,24 @@
-from enum import Enum
 from typing import List, Tuple
 
 import numpy as np
 import torch
+from app.common import ModelType, get_type
 from app.pipelines import Pipeline
-from huggingface_hub import HfApi
 from speechbrain.pretrained import SepformerSeparation, SpectralMaskEnhancement
-
-
-class ModelType(Enum):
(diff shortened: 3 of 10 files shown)


TParcollet:
I should update our tutorial to explain better how to upload a model to HF now ...

osanseviero:
> I should update our tutorial to explain better how to upload a model to HF now ...

Happy to help on that. And as discussed, later on a `push_to_hub` method would take care of it for the users.

api-inference-community/docker_images/speechbrain/app/pipelines/automatic_speech_recognition.py
```diff
@@ -1,19 +1,57 @@
...
+
+def interface_to_type(interface_str):
+    if interface_str == "EncoderASR":
+        return ModelType.ENCODER_ASR
+    elif interface_str == "EncoderDecoderASR":
+        return ModelType.ENCODER_DECODER_ASR
+    else:
+        raise ValueError(
+            f"Invalid interface: {interface_str} for Automatic Speech Recognition."
+        )
```
Narsil, line 24: seems like this could be reused and not duplicated (`from app.common import interface_to_type`)
With PIpeline validation (I mean supported by ASR vs AudioToAudio) done directly within the `__init__` function.

.github/workflows/python-api-speechbrain.yaml
```diff
@@ -0,0 +1,29 @@
+# TODO: Merge this with allenNLP to have a single workflow for all docker images.
+name: speechbrain-docker
+
+on:
+  pull_request:
+    paths:
+      - "api-inference-community/docker_images/speechbrain/**"
```
Narsil, line 7: Thank you for this !

api-inference-community/docker_images/speechbrain/app/pipelines/audio_to_audio.py
```diff
@@ -1,32 +1,48 @@
...
+    if info.config:
+        if "speechbrain" in info.config:
+            interface_str = info.config["speechbrain"].get(
+                "interface", "SepformerSeparation"
+            )
+        else:
+            interface_str = "SepformerSeparation"
+    else:
+        interface_str = "SepformerSeparation"
+    return interface_to_type(interface_str)
```
Narsil, line 36: Taste comments (feel free to ignore):

- I would use an enum covering all cases for all pipelines and have code `interface_to_type` be very generic.
- I would use same names as the strings in the configs:
  1. Would improve readability/reduce confusion
  2. Would enable `ModelType[interface_str]` code, which is probably easier to maintain than a series of if/else. (`ModelType[interface_str.upper()]` if we want to keep uppercase in the enum
- I would use STR values instead of integers in the enum. Less confusing when errors pop up, and shouldn't have any disadvantage over ints.

Other NIT, it's a bit longer but IMO shows intention a bit better.:

```python
def get_type(model_id):
    info = HfApi().model_info(repo_id=model_id)
    DEFAULT = "SepformerSeparation"
    if info.config:
        interface_str = info.config.get("speechbrain", {}).get("interface", DEFAULT)
    return interface_to_type(interface_str)
```

Narsil (commented):
Overall it looks good, 

I made a few suggestions for code cleanliness, I would be interested to know your opinion on them before approving this PR.

api-inference-community/docker_images/speechbrain/app/pipelines/automatic_speech_recognition.py
osanseviero: Resolving since this comment is duplicated with the other.

api-inference-community/docker_images/speechbrain/app/pipelines/audio_to_audio.py
osanseviero: Good feedback, thanks a lot

* I changed the enum to cover all cases
* I like that we can do `ModelType[interface_str]` although it's a bit less readable `ENCODERDECODERASR` now vs `ENCODER_DECODER_ASR` before
* I changed the value to be strings, which is also useful for error handling
* I decided to remove having a default interface entirely instead of your last suggestion. I feel it's better to not do any magic from our side. Having a default interface will likely lead to unintended consequences and unintuitive behavior for users. If there's no config or no interface specified in the config, let's just give an error.

api-inference-community/docker_images/speechbrain/app/pipelines/audio_classification.py
```diff
@@ -2,12 +2,17 @@
...
+from app.common import ModelType, get_type
 from app.pipelines import Pipeline
 from speechbrain.pretrained import EncoderClassifier
 
 
 class AudioClassificationPipeline(Pipeline):
     def __init__(self, model_id: str):
+        model_type = get_type(model_id)
+        if model_type != ModelType.ENCODERCLASSIFIER:
+            raise ValueError(f"{model_type.value} is invalid for audio-classification")
```
Narsil, line 14: Wait, weren't we supposed to suport ENCODERDECODER ?

Narsil: OK read further, wrong file.

Narsil (approved):
LGTM.

Btw, for ENCODERDECODER being hard to read we could go to ENCODER_DECODER with something like that:

https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case

Doesn't feel necessary right now, but definitely if this becomes a source of confusion. Would definitely add a test in that case too.

osanseviero:
SG! Let's keep things as they are for the time being and reconsider seeing how people interact with this to avoid unneeded complexity.

osanseviero:
Btw @Narsil thanks for useful, actionable reviews as always!

PR #345, opened 2021-09-13
Add model info test for private models

lewtun:
In PR #340, we observed that `HfApi.model_info` isn't explicitly tested for private repositories, where the absence of a token should throw a HTTP 404 error. This PR aims to cover this case. 

I took inspiration from `HfApiPublicTest` test class to create a private version.

### DIFF
--- a/tests/test_hf_api.py
+++ b/tests/test_hf_api.py
@@ -21,6 +21,7 @@ import time
 import unittest
 from io import BytesIO
 
+import requests
 from huggingface_hub.constants import REPO_TYPE_DATASET, REPO_TYPE_SPACE
 from huggingface_hub.file_download import cached_download
 from huggingface_hub.hf_api import (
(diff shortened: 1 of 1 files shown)


julien-c:
this is more of a test of the hub than of `huggingface_hub` no? (no objection on doing this, just curious)

lewtun:
> this is more of a test of the hub than of `huggingface_hub` no? (no objection on doing this, just curious)

You're right that this is mainly testing whether authentication works as expected for private repos, so it is similar in philosophy to the `CachedDownloadTests` ([link](https://github.com/huggingface/huggingface_hub/blob/d914261c9e67297a80dc07ac705d8087d4c4be63/tests/test_file_download.py#L56)), which also test that the expected HTTP errors are raised.

Happy to refactor / move things if preferred

julien-c:
no, that works for me!

tests/test_hf_api.py
```diff
@@ -434,6 +435,28 @@ def test_dataset_info(self):
...
+
+    def test_model_info(self):
+        # Test we cannot access model info without a token
+        with self.assertRaisesRegex(requests.exceptions.HTTPError, "404 Client Error"):
+            _ = self._api.model_info(repo_id=f"{USER}/{REPO_NAME}")
+        # Test we can access model info with a token
+        model_info = self._api.model_info(
+            repo_id=f"{USER}/{REPO_NAME}", token=self._token
+        )
+        self.assertIsInstance(model_info, ModelInfo)
```
julien-c, line 457: yes we are testing the actual behavior of huggingface_hub here so 👍

tests/test_hf_api.py
```diff
@@ -434,6 +435,28 @@ def test_dataset_info(self):
         self.assertEqual(dataset.sha, DUMMY_DATASET_ID_REVISION_ONE_SPECIFIC_COMMIT)
 
 
+class HfApiPrivateTest(HfApiCommonTestWithLogin):
+    def setUp(self) -> None:
+        super().setUp()
+        self._api.create_repo(token=self._token, name=REPO_NAME)
```
osanseviero, line 441: nit: alternatively you could use `create_repo` with `private=True` since the get go.

osanseviero (approved):
Thanks for splitting the PRs and sending this :)

tests/test_hf_api.py
lewtun: Thanks for the tip - done!

LysandreJik (approved):
Great idea and implementation @lewtun!

PR #347, opened 2021-09-14
force filename and remove `.lock`

philschmid:
# What does this PR do? 

This PR forces file for `hf_hub_download` download method to use the provided `filename`. That way the file will be saved with the file name and not with a hash. 

Question: the `snapshot_download` method also removes created `.lock` files, should we add an option to remove them as well?

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -529,7 +529,7 @@ def hf_hub_download(
         repo_id, filename, subfolder=subfolder, repo_type=repo_type, revision=revision
     )
 
-    path = cached_download(
+    return cached_download(
         url,
         library_name=library_name,
         library_version=library_version,
@@ -543,8 +543,3 @@ def hf_hub_download(
         use_auth_token=use_auth_token,
         local_files_only=local_files_only,
     )
-    # removes lock file if created
-    if os.path.exists(path + ".lock"):
-        os.remove(path + ".lock")
-        
-    return path
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -529,17 +529,22 @@ def hf_hub_download(
         repo_id, filename, subfolder=subfolder, repo_type=repo_type, revision=revision
     )
 
-    return cached_download(
+    path = cached_download(
         url,
         library_name=library_name,
         library_version=library_version,
         cache_dir=cache_dir,
         user_agent=user_agent,
         force_download=force_download,
-        force_filename=force_filename,
+        force_filename=filename,
         proxies=proxies,
         etag_timeout=etag_timeout,
         resume_download=resume_download,
         use_auth_token=use_auth_token,
         local_files_only=local_files_only,
     )
+    # removes lock file if created
+    if os.path.exists(path + ".lock"):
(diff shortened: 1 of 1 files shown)


osanseviero:
I think this PR would make `force_filename` in `hf_hub_download` useless. Should we delete the dead parameter if we decide to go with that direction?

___

Apart from that I wonder why not just use `force_filename` parameter, it seems to work well:

 ```
hf_hub_download("osanseviero/wine-quality", "config.yml", force_filename="config.yml")
>>> /root/.cache/huggingface/hub/config.yml
```

The issue of deleting the hash is that the caching will not work properly. If you modify `config.yml` in my example above, the file will not get reloaded. **That means that with this approach the downloaded files will never get updated.**

**Comment by Narsil** [comment]
Not to mention invalidating ALL users cache on every machine on the planet.

Not saying we shouldn't do this (obscure filenames are  obscure and it definitely would help when inspecting the cache).
But we need to but much more cautious about these changes, they have a tremendously big impact (forcing redownloading EVERY model for EVERY user + making a whole lot of disk space be useless now, not sure a lot of people clean their HF cache regularly)

**Comment by philschmid** [comment]
Not sure if I miss understood the method, but shouldn't be `hf_hub_download` an easy way to download files from a Repository on the hub to your local machine? For that, I as a user expect to receive a file with the same name as on the hub to a directory. 
If this assumption is wrong happy to close the PR again, but can you then let me know what is the preferred way of loading files from the hub (not the whole repository)?

**Comment by osanseviero** [comment]
This method downloads and caches a file from the Hub. That means that if you call `hf_hub_download` multiple times to a heavy model in the Hub, it's only downloaded the first time. If there's a change in the model in the Hub, it gets automatically re-downloaded. So `hf_hub_download` is a nice way to ensure you have the latest version of a file in the Hub.

Maybe we should consider adding another method or parameter for downloading a single file from the Hub to a given directory? (no caching)

**Comment by philschmid** [comment]
Okay, I then miss understood the purpose of `hf_hub_download`. 

> Maybe we should consider adding another method or parameter for downloading a single file from the Hub to a given directory? (no caching). 

I think we should over a method to load files easily from the hub to your preferred directory. That's so far the most common use case for me, whether it is in the `inference-toolkit` for sagemaker or using the hub as storage for any kind of file. For example if you want to bundle files into a container the `hf_hub_download` function is not well suited for it.

**Comment by osanseviero** [comment]
@LysandreJik and me had a recent discussion about a similar issue (although in that case it was about downloading a full repo, not a single file). WDYT @LysandreJik, is single-file downloading (no caching) something we want to support?

**Comment by LysandreJik** [comment]
I think `hf_hub_download` currently serves its purpose well but also offers to download files to a given repository using `cache_dir` and `force_filename`:

```py
>>> hf_hub_download("bert-base-cased", "config.json", cache_dir="/home/lysandre/test", force_filename="new_config.json")
```

Will correctly fetch the `new_config.json` file in the `/home/lysandre/test` directory. Maybe the documentation of that method could showcase this better.

This too will cache the file, which is IMO strictly better in every use-case, unless I'm missing something.

> Question: the snapshot_download method also removes created .lock files, should we add an option to remove them as well?

I think that's a good idea, yes, once the download is done.

PR #350, opened 2021-09-16
Use interface from config.json for speechbrain snippets

osanseviero:
Instead of heavily relying in the tags, `speechbrain` models now depend on the `config.json` file which specifies which interface should be used


widgets/src/lib/interfaces/Libraries.ts
```diff
@@ -189,46 +189,41 @@ nlp = spacy.load("${nameWithoutNamespace(model.modelId)}")
...
+		case "EncoderASR":
+			return "transcribe_file";
+		case "SpectralMaskEnhancement":
+			return "enhance_file";
+		case "SepformerSeparation":
+			return "separate_file";
+		default: {
+			return undefined;
+		}
+	}
```
mishig25, line 206: nit: could we remove the brackets?

mishig25 (approved):
lgtm!

PR #352, opened 2021-09-16
subprocess.run -> run_subprocess

LysandreJik:
Implement a `run_subprocess` method to remove most of the verbosity introduced by `subprocess.run`

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -245,7 +245,7 @@ def files_to_be_staged(pattern: str, folder: Union[str, Path]) -> List[str]:
         `List[str]`: List of files that are to be staged.
     """
     try:
-        p = run_subprocess(f"git ls-files -mo {pattern}".split(), folder)
+        p = run_subprocess("git ls-files -mo".split() + [pattern], folder)
         if len(p.stdout.strip()):
             files = p.stdout.strip().split("\n")
         else:
@@ -666,6 +666,7 @@ class Repository:
                     run_subprocess(
                         f"{'git clone' if self.skip_lfs_files else 'git lfs clone'} {repo_url} .".split(),
                         self.local_dir,
+                        env=env,
                     )
             else:
                 # Check if the folder is the root of a git repository
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -17,7 +17,7 @@ from huggingface_hub.repocard import metadata_load, metadata_save
 
 from .hf_api import HfApi, HfFolder, repo_type_and_id_from_hf_id
 from .lfs import LFS_MULTIPART_UPLOAD_COMMAND
-from .utils import logging
+from .utils import logging, run_subprocess
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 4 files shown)


src/huggingface_hub/repository.py
```diff
@@ -86,6 +86,17 @@ def __repr__(self):
         return f"[{self.title} command, status code: {status}, {'in progress.' if not self.is_done else 'finished.'} PID: {self._process.pid}]"
 
 
+def run_subprocess(command, folder, check=True) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        command.split(),
```
Narsil, line 91: This is wrong I think, you really want to pass list of arguments and not split arbitrarily.

For instance on `filename` containing spaces it would break:

`filename = "my filename.txt`,  command is now: `["xxx", "my", "filename.txt"], instead of expected ["xxx", "my filename.txt"].

julien-c: yes!

LysandreJik: Thanks for the review, will adapt the draft

src/huggingface_hub/utils/__init__.py
```diff
@@ -13,3 +13,25 @@
...
+
+import subprocess
+
+from .logging import get_logger
+
+
+logger = get_logger(__name__)
+
+
+def run_subprocess(command, folder, check=True) -> subprocess.CompletedProcess:
```
adrinjalali, line 25: docstring documentation here would be nice.

adrinjalali (commented):
This looks quite good. Pinging @LysandreJik as a reminder.

LysandreJik:
Thanks for the reminder! Would indeed like to update this and wrap this up.

LysandreJik:
Should be good to go!

src/huggingface_hub/utils/__init__.py
```diff
@@ -13,3 +13,47 @@
...
+import subprocess
+from typing import List
+
+from .logging import get_logger
+
+
+logger = get_logger(__name__)
+
+
+def run_subprocess(
```
adrinjalali, line 26: I've had bad experiences in including functions in `__init__` files since soon it'll cause some circular import issues. Could we move this to a `utils/_subprocess.py` file (I don't really feel strongly about the file name)?

LysandreJik: Yes, sounds good. Will move it around!

LysandreJik:
@merveenoyan are you aware of why this test might fail? 

```
______ HubKerasFunctionalTest.test_save_pretrained_task_name_deprecation _______

self = <tests.test_keras_integration.HubKerasFunctionalTest testMethod=test_save_pretrained_task_name_deprecation>

    def test_save_pretrained_task_name_deprecation(self):
        REPO_NAME = repo_name("save")
        model = self.model_init()
        model.build((None, 2))
    
        with pytest.warns(
            FutureWarning,
            match="`task_name` input argument is removed. Pass `tags` instead.",
        ):
            save_pretrained_keras(
                model,
                f"{WORKING_REPO_DIR}/{REPO_NAME}",
                tags=["test"],
                task_name="test",
>               save_traces=True,
            )
E           Failed: DID NOT WARN. No warnings of type (<class 'FutureWarning'>,) matching ('`task_name` input argument is removed. Pass `tags` instead.') were emitted. The list of emitted warnings is: [FutureWarning('`task_name` input argument is deprecated. Pass `tags` instead.')].
```

Failed before and after rebasing on current `main`.

adrinjalali (approved):

PR #353, opened 2021-09-16
Max line length 119

LysandreJik:
Sets the maximum line length to 119, which is the same as `transformers` and improves readability.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[tool.black]
+line-length = 119
+target-version = ['py39']
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,11 +40,11 @@ known_third_party =
     torch_xla
     tqdm
 
-line_length = 88
+line_length = 119
 lines_after_imports = 2
 multi_line_output = 3
 use_parentheses = True
 
 [flake8]
 ignore = E203, E501, E741, W503, W605
-max-line-length = 88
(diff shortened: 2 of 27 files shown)


adrinjalali (changes requested):
I would be much happier if we don't change it to 119.

119 would need a much wider screen for many contributors and reviewers to review code and browse on github. So I'd be happier if we don't increase the default. Here's an example of how it looks after the change. I can't see the end of the line w/o having to scroll.

88 is also more accepted and is the norm in the community. 119 is more of a common default for corporate codebases, partly because in those code bases you don't have a strong focus on reviewing and what looks good to review code. The focus is more on how it looks when developing code.

Screenshot from 2022-03-07 12-48-16

julien-c:
i think consistency with other huggingface open source codebases is important, no?

adrinjalali:
To answer that question, I think we should think about whether we want the same governance/development style for all the repos.

Our open source repositories each play a different role in the community. Some are developed more in house than others, and they also have different users. Some have data scientists as users (like transformers), some are more integrated with the closed source products than others,  and some have other developers of other libraries as users (like huggingface_hub). That to me means they follow different development/community building/governance styles, and that can also reflect on how the projects look and what the standards are. For that, I would suggest that consistency with other huggingface repos is not the main factor.

LysandreJik:
Closing

PR #355, opened 2021-09-20
Improve save and load repocard metadata

elishowk:
# Problem
## Loading and Saving repocard shouldn't modify implicitely the YAML.

- loading and saving override the current newlines format.
- also pyyaml changes "no" to False and encode UTF-8 so that round-trip is impossible.

# Summary

## Newlines

- Python3 by default read files and normalize new lines. This PR takes care of that.

## Replacing pyyaml by ruamel.yaml

- Added a test case `RepocardTest.test_metadata_roundtrip` that works fine with ruamel.yaml and not with pyyaml. 
#### Bug 1 : 'no' string parsed to False

```
________________________________________ RepocardTest.test_metadata_roundtrip ________________________________________

self = <tests.test_repocard.RepocardTest testMethod=test_metadata_roundtrip>

    def test_metadata_roundtrip(self):
        filename = "dummy_target.md"
        filepath = Path(REPOCARD_DIR) / filename
        filepath.write_text(ROUND_TRIP_MODELCARD_CASE)
        metadata = metadata_load(filepath)
>       self.assertDictEqual({
            "language": "no",
            "datasets": "CLUECorpusSmall",
            "widget": [{ "text": "北京是[MASK]国的首都。" }],
        }, metadata)
E       AssertionError: {'language': 'no', 'datasets': 'CLUECorpusSmall', 'wid[30 chars]。'}]} != {'language': False, 'datasets': 'CLUECorpusSmall', 'wi[31 chars]。'}]}
E         {'datasets': 'CLUECorpusSmall',
E       -  'language': 'no',
E       ?              ^^^^
E
E       +  'language': False,
E       ?              ^^^^^
E
E          'widget': [{'text': '北京是[MASK]国的首都。'}]}

tests/test_repocard.py:132: AssertionError
```

#### Bug 2 : UTF-8 string broken.

```
________________________________________ RepocardTest.test_metadata_roundtrip ________________________________________

self = <tests.test_repocard.RepocardTest testMethod=test_metadata_roundtrip>

    def test_metadata_roundtrip(self):
        filename = "dummy_target.md"
        filepath = Path(REPOCARD_DIR) / filename
        filepath.write_text(ROUND_TRIP_MODELCARD_CASE)
        metadata = metadata_load(filepath)
        metadata_save(filepath, metadata)
        content = filepath.read_text()
>       self.assertEqual(content, ROUND_TRIP_MODELCARD_CASE)
E       AssertionError: '\n---\nlanguage: false\ndatasets: CLUECorpusSmall\nwidget:[88 chars]le\n' != '\n---\nlanguage: no\ndatasets: CLUECorpusSmall\nwidget:\n-[35 chars]le\n'
E
E         ---
E       - language: false
E       ?           ^^^^^
E       + language: no
E       ?           ^^
E         datasets: CLUECorpusSmall
E         widget:
E       - - text: "\u5317\u4EAC\u662F[MASK]\u56FD\u7684\u9996\u90FD\u3002"
E       + - text: 北京是[MASK]国的首都。
E         ---
E
E         # Title

tests/test_repocard.py:139: AssertionError
```


src/huggingface_hub/repocard.py
```diff
@@ -24,15 +25,26 @@ def metadata_load(local_path: Union[str, Path]) -> Optional[Dict]:
...
-            content[: match.start()] + f"---\n{data_yaml}---\n" + content[match.end() :]
-        )
-    else:
-        output = f"---\n{data_yaml}---\n{content}"
-
-    Path(local_path).write_text(output)
+    # try to preserve newlines
+    linebrk = "\n"
+    # this is known not to work with ^M linebreaks, so ^M are replaced by \n
+    with open(local_path, "r") as readme:
```
elishowk, line 31: this first open is meant to ensure we detect the real newline character in the file.
and to read its contents

src/huggingface_hub/repocard.py
```diff
@@ -24,15 +25,26 @@ def metadata_load(local_path: Union[str, Path]) -> Optional[Dict]:
...
+        if type(readme.newlines) is tuple:
+            linebrk = readme.newlines[0]
+        if type(readme.newlines) is str:
+            linebrk = readme.newlines
+        content = readme.read()
+
+    if content:
+        with open(local_path, "w", newline="") as readme:
```
elishowk, line 39: `newline=""` is the only way not the modify line endlings

src/huggingface_hub/repocard.py
```diff
@@ -24,15 +24,28 @@ def metadata_load(local_path: Union[str, Path]) -> Optional[Dict]:
...
+            data_yaml = yaml.dump(data, sort_keys=False, line_break=linebrk)
```
julien-c, line 39: keep the comment please

julien-c (commented):
This changes the behavior when local_path does not exist, where we still want to create a file and output the metadata block

(@LysandreJik was right that we needed to add a test for this 😆 )

Other than that, looks good!

src/huggingface_hub/repocard.py
elishowk:
```suggestion
            data_yaml = yaml.dump(data, sort_keys=False, line_break=linebrk)
            # sort_keys: keep dict order
```

src/huggingface_hub/repocard.py
elishowk:
```suggestion
    with open(local_path, "r", newline="") as readme:
```

src/huggingface_hub/repocard.py
```diff
@@ -24,15 +24,29 @@ def metadata_load(local_path: Union[str, Path]) -> Optional[Dict]:
...
-    match = REGEX_YAML_BLOCK.search(content)
-    if match:
-        output = (
-            content[: match.start()] + f"---\n{data_yaml}---\n" + content[match.end() :]
-        )
-    else:
-        output = f"---\n{data_yaml}---\n{content}"
-
-    Path(local_path).write_text(output)
+    # try to preserve newlines
```
elishowk, line 27:
```suggestion
    # try to preserve newlines
    # docs about open()  newline="" parameter:
    # https://docs.python.org/3/library/functions.html?highlight=open#open
```

elishowk:
I'm proposing to replace pyyaml by [ruamel.yaml](https://yaml.readthedocs.io/en/latest/overview.html), which is better maintained and optimized for safe loading with preservation of the existing formatting. WDYT @LysandreJik ?

julien-c:
it's also what we use in our other codebases

elishowk:
The reason I choose `ruamel.yaml` instead of `pyyaml` is because the latter have really bad loading/dumping side effects unacceptable when we need a safe & conservative "round trip" loading/dumping. PyYAML's official indeed [but full of not-fixed bugs](https://github.com/yaml/pyyaml/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc). 
Also [PyYAML does not support YAML 1.2 yet](https://github.com/yaml/pyyaml/issues/116) whereas [ruamel supports it by default](https://yaml.readthedocs.io/en/latest/pyyaml.html#defaulting-to-yaml-1-2-support) (fixing the yes/no to boolean bug) 
## Real bugs found in the hub.

- Example for UTF-8 bug : 
```
-- text: "Fui a la librería a comprar un <mask>."
+- text: "Fui a la librer\xEDa a comprar un <mask>."
```


- Example for language norwegian "no", [explanations](https://docs.saltproject.io/en/latest/topics/troubleshooting/yaml_idiosyncrasies.html#true-false-yes-no-on-off):

```
-language: no
-license: CC-BY 4.0
+language: false
+license: cc-by-4.0
```

It's your call, but I prefer ruamel.yaml

LysandreJik:
Thank you for the explanation @elishowk - this explanation, along with a PR description would be very useful to understand the core issue and the solution you provide, as it's impossible to review without having the full context.

Now that you have explained, I understand why you chose `ruamel.yaml`. Could you provide a sample use-case that you use in your setup that works with `ruamel.yaml` but doesn't with `pyyaml`, so that I may take a closer look? Thank you.

elishowk:
Hi @LysandreJik,

Sorry for the lack of context. I started a PR and discovered bugs while working on real-world metadata, and didn't always thought about explaining all the context.

So, now I pushed back `ruamel.yaml` for your consideration, and I added 1 test case in test_repocard.py that shows and ensures the problem I encountered with `pyyaml`.

I completed the top-description of this PR https://github.com/huggingface/huggingface_hub/pull/355#issue-737552707


I hope this give you all the elements to review 😅

Cheers

LysandreJik (approved):
Thank you for the explanation. Looks good, only left a nit.

LysandreJik:
Thank you @elishowk! Feel free to merge when green.

PR #356, opened 2021-09-20
Add pyopenjtalk to espnet dependencies

osanseviero:
Add `pyopenjtalk` to dependencies for `espnet`, which is required for models as discovered in https://github.com/espnet/espnet/pull/3326#discussion_r663513293


Narsil:
Update coming up

PR #364, opened 2021-09-22
Add utility to get repo name

sgugger:
This PR adds a small utility function to get the full repo name from a model ID which:
- extracts the username if no organization is passed and returns username/model_id
- just returns organization/model_id otherwise

This function is used in all examples of Transformers.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -318,13 +318,17 @@ class HfApi:
         write_to_credential_store(username, password)
         return d["token"]
 
-    def whoami(self, token: str) -> Dict:
+    def whoami(self, token: Optional[str] = None) -> Dict:
         """
         Call HF API to know "whoami".
 
         Args:
-            token (``str``): Hugging Face token.
+            token (``str``, `optional`):
+                Hugging Face token. Will default to the locally saved token if not provided.
         """
+        if token is None:
+            token = HfFolder.get_token()
+
         path = "{}/api/whoami-v2".format(self.endpoint)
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -318,13 +318,17 @@ def login(self, username: str, password: str) -> str:
...
         Call HF API to know "whoami".
 
         Args:
-            token (``str``): Hugging Face token.
+            token (``str``, `optional`):
+                Hugging Face token. Will default to the locally saved token if not provided.
         """
+        if token is None:
+            token = HfFolder.get_token()
+
```
LysandreJik, line 331: WDYT @osanseviero? I think this is a great improvement in terms of user experience and should also be used to reopen conversations about:
- having the `HfApi.login` saving the token
- having the token default to the locally saved token when not passed through the `HfApi`

src/huggingface_hub/hf_api.py
```diff
@@ -797,6 +801,35 @@ def upload_file(
         d = r.json()
         return d["url"]
 
+    def get_full_repo_name(
+        self,
+        model_id: str,
+        organization: Optional[str] = None,
+        token: Optional[str] = None,
```
LysandreJik, line 808: In a similar fashion to what is done here

LysandreJik (approved):
LGTM!

src/huggingface_hub/hf_api.py
osanseviero: Agree it's a great improvement! Thanks @sgugger! (related issue: #154)

PR #365, opened 2021-09-22
append uuid to test dummy repo names

elishowk:


elishowk:
At this point, 504 timeouts still happens in `create_repo` endpoint.

For example, there's been one failure in each test run (python3.6 and python3.9), for the same RepositoryTest unittest class, not the same function :

https://github.com/huggingface/huggingface_hub/pull/365/checks?check_run_id=3677284317
https://github.com/huggingface/huggingface_hub/pull/365/checks?check_run_id=3676979803

elishowk:
Another random 504 case with HfApi `list_repos_objs` endpoint
https://github.com/huggingface/huggingface_hub/runs/3697440120

elishowk:
Hi maintainers !

I propose this tests code improvements that takes care of generating unique repo names for each test separately. 504 errors still happens on two hub server endpoint that are gonna be worked on independently.

This PR doesn't terminate this task of fixin CI's flakyness, but it's good first step I think.

WDYT ?

tests/test_hf_api.py
```diff
@@ -19,6 +19,7 @@
 import tempfile
 import time
 import unittest
+import uuid
```
elishowk, line 22: Any better soluttion for generating a randon unique id is fine by me :-)

PR #367, opened 2021-09-23
Add adapter-transformers to `api-inference-community`

calpt:
This PR adds a new image for the [adapter-transformers](https://github.com/Adapter-Hub/adapter-transformers) library, currently including pipelines for text classification, QA & token classification.

(cc @osanseviero)


osanseviero:
Thanks for this!!! Super exciting to have this in production 🚀

The existing failing test is unrelated.

In the meantime, could you also add a test in test_dockers please? (https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/tests/test_dockers.py)

And if you create a copy of this file and modify for adapter transformers it would be great https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-api-allennlp.yaml

api-inference-community/docker_images/adapter_transformers/requirements.txt
```diff
@@ -0,0 +1,3 @@
+starlette==0.14.2
+api-inference-community==0.0.19
+git+https://github.com/Adapter-Hub/adapter-transformers#egg=adapter-transformers
```
osanseviero, line 3: This could alternatively be `adapter-transformers` no? Or is there any particular reason to install directly from head? (it's usually better to have `adapter-transformers==version` so we can reliably test/debug things)

api-inference-community/docker_images/adapter_transformers/requirements.txt
```diff
@@ -0,0 +1,3 @@
+starlette==0.14.2
```
osanseviero, line 1: I think we also need torch here (see failing docker test)

api-inference-community/docker_images/adapter_transformers/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY ./prestart.sh /app/
+
+
+# Most DL models are quite large in terms of memory, using workers is a HUGE
+# slowdown because of the fork and GIL with python.
+# Using multiple pods seems like a better default strategy.
+# Feel free to override if it does not make sense for your library.
+ARG max_workers=1
+ENV MAX_WORKERS=$max_workers
+ENV HUGGINGFACE_HUB_CACHE=/data
```
osanseviero, line 18: Do you use `TORCH_HOME` env variable in `adapter-transformers` as well?

api-inference-community/docker_images/adapter_transformers/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN pip install --no-cache-dir -r requirements.txt
+COPY ./prestart.sh /app/
+
+
+# Most DL models are quite large in terms of memory, using workers is a HUGE
+# slowdown because of the fork and GIL with python.
+# Using multiple pods seems like a better default strategy.
+# Feel free to override if it does not make sense for your library.
+ARG max_workers=1
+ENV MAX_WORKERS=$max_workers
```
osanseviero, line 17: I think you also use `HF_HOME` in `adapter-transformers`

api-inference-community/docker_images/adapter_transformers/app/pipelines/base.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, model_id: str):
+        raise NotImplementedError("Pipelines should implement an __init__ method")
+
+    @abstractmethod
+    def __call__(self, inputs: Any) -> Any:
+        raise NotImplementedError("Pipelines should implement a __call__ method")
+
+    @staticmethod
+    def _load_pipeline_instance(pipeline_class, adapter_id):
+        adapter_info = get_adapter_info(adapter_id, source="hf")
```
osanseviero, line 18: Nice refactoring here!

osanseviero (commented):
Thanks for adding the docker tests. The PR looks great overall! I left a couple of minor comments to ensure everything works as intended

osanseviero:
I still see some errors in the docker test for `text-classification`. Feel free to take a look [here](https://github.com/huggingface/huggingface_hub/pull/367/checks?check_run_id=3720669809).

In particular I see 

> Token indices sequence length is longer than the specified maximum sequence length for this model (1026 > 512). Running this sequence through the model will result in indexing errors.

I'm not sure if this is the cause of the errors, though. Testing with `pytest -sv docker_images/adapter_transformers/app/pipelines/text_classification.py` should help (and if you get 400 instead of 200 you can always print the message to debug further - pro level debugging 😎 ). Let me know if you would like some help and I can debug a bit.

api-inference-community/docker_images/adapter_transformers/requirements.txt
calpt: I implemented this using the current dev version of adapter-transformers, including some syntax that isn't released in PyPI yet. This should happen in the coming days, will change to `adapter-transformers==2.2.0` then.

api-inference-community/docker_images/adapter_transformers/requirements.txt
calpt: Thanks :)

api-inference-community/docker_images/adapter_transformers/Dockerfile
calpt: Yes, added that, thanks

calpt:
@osanseviero Thanks for the very quick reviews! Regarding the currently failing tests, I'm currently unable to reproduce e.g. the one you linked on my machine (doesn't return HTTP 500 but HTTP 400 for me, which passes the assertion). Not sure why this is the case yet (ran using `pytest -sv tests/test_dockers.py::DockerImageTests::test_adapter_transformers`).

api-inference-community/docker_images/adapter_transformers/requirements.txt
osanseviero: SG, thanks!

osanseviero:
You're right that `pytest -sv docker_images/adapter_transformers/app/pipelines/text_classification.py` passes.

`RUN_DOCKER_TESTS=1 pytest -sv tests/test_dockers.py::DockerImageTests::test_adapter_transformers` gives me a 500 at the following part

```
with open(
  os.path.join(os.path.dirname(__file__), "samples", "malformed.flac"),
  "rb",
) as f:
  data = f.read()
response = httpx.post(url, data=data, timeout=timeout)
self.assertIn(response.status_code, {200, 400})
```

The exact error is 

> Token indices sequence length is longer than the specified maximum sequence length for this model (1026 > 512). Running this sequence through the model will result in indexing errors

Not sure why this happens in the docker test but not in the unit test. I can take another look tomorrow unless you have any ideas

osanseviero:
Hey @calpt. We were able to take a look at it and the issue was that the model was breaking with very long sequences, which then raised a 500 which is not good. So we are now re-raising the error as a 400. And the long sequence was actually `malformed.flac` which is a long string of 0, so it was utf-8 decodable.

Anyways, the error is fixed now by the re-raising. It's not a super elegant solution, but should unblock this PR and we're really looking forward to having adapter transformers in the inference API.

Narsil:
LGTM

Narsil (approved):

osanseviero (approved):

calpt:
Hey @osanseviero, thanks for looking into this again and providing a workaround. Happy to have adapter-transformers integrated in the inference API :)

PR #371, opened 2021-09-23
Generic is all you need

osanseviero:
This PR removes the mapping of type to exactly the same pipeline for all types. This means that if a new repo type is added to the API, then `generic` already supports it out of the box.

I'm not 100% sure if we should keep the unit tests under docker_images. I would prefer to leave them since they help in debugging quite easily for any other repo.

cc @nateraw

### DIFF
--- a/.github/workflows/python-api-generic.yaml
+++ b/.github/workflows/python-api-generic.yaml
@@ -1,0 +1,29 @@
+# TODO: Merge this with allenNLP to have a single workflow for all docker images.
+name: generic-docker
+
+on:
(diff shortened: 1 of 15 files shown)


Narsil:
Docker tests, let's keep (they are testing specific code)

The unittest on the other hand cannot really do a lot since all the logic is remote (well we can keep 1 to test we're downloading correctly but that's about it IMO).

osanseviero:
Discussed offline. The unit tests are specially useful for debugging purposes so we'll keep them for the time being.

PR #376, opened 2021-09-30
Fix model-repos.md title level

elishowk:
## Why

Currently the title fixed does not appear in the sidebar of the docs : https://huggingface.co/docs/hub/model-repos#how-can-i-fork-or-rebase-a-repository-with-lfs-pointers

This PR fixes that


osanseviero (approved):
Thanks for the fix!

PR #379, opened 2021-09-30
Make token optional in all HfApi methods.

sgugger:
This PR makes the `token` argument of all HfApi methods optional and default to the saved token. As you can see in the updated docs, the result is quite nice in the sense a user never has to retrieve their token (and thus get some chance of leaking it by mistake).

Since this changes the order of the argument and is thus a breaking change, some internal magic is implemented to detect if the argument at the old place of the token is the token, when the argument at the place of the token is not a valid token, and then do some switch.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -589,6 +589,9 @@ class HfApi:
         return DatasetInfo(**d)
 
     def _is_valid_token(self, token: str):
+        """
+        Determines whether `token` is a valid token or not.
+        """
         try:
             self.whoami(token=token)
             return True
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -588,10 +588,17 @@ class HfApi:
         d = r.json()
         return DatasetInfo(**d)
 
+    def _is_valid_token(self, token: str):
+        try:
+            self.whoami(token=token)
+            return True
+        except HTTPError:
+            return False
+
     def create_repo(
         self,
-        token: str,
         name: str,
+        token: Optional[str] = None,
         organization: Optional[str] = None,
         private: Optional[bool] = None,
         repo_type: Optional[str] = None,
@@ -616,6 +623,23 @@ class HfApi:
             URL to the newly created repo.
         """
         path = "{}/api/repos/create".format(self.endpoint)
+        if token is None:
+            token = HfFolder.get_token()
+            if token is None:
+                raise EnvironmentError(
+                    "You need to provide a `token` or be logged in to Hugging Face with "
+                    "`huggingface-cli login`."
+                )
+        elif not self._is_valid_token(token):
+            if self._is_valid_token(name):
+                warnings.warn(
+                    "`create_repo` now takes `token` as an optional positional argument. "
+                    "Be sure to adapt your code!",
+                    FutureWarning,
+                )
+                token, name = name, token
+            else:
+                raise ValueError("Invalid token passed!")
 
         if repo_type not in REPO_TYPES:
             raise ValueError("Invalid repo type")
@@ -650,8 +674,8 @@ class HfApi:
 
     def delete_repo(
         self,
-        token: str,
         name: str,
+        token: Optional[str] = None,
         organization: Optional[str] = None,
         repo_type: Optional[str] = None,
     ):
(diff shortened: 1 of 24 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -328,6 +328,10 @@ def whoami(self, token: Optional[str] = None) -> Dict:
         """
         if token is None:
             token = HfFolder.get_token()
+        if token is None:
+            raise ValueError(
+                "You need to pass a valid `token` or login by using `huggingface-cli login`"
+            )
```
LysandreJik, line 334: Yes, very nice improvement

src/huggingface_hub/hf_api.py
```diff
@@ -563,10 +588,17 @@ def dataset_info(
         d = r.json()
         return DatasetInfo(**d)
 
+    def _is_valid_token(self, token: str):
+        try:
+            self.whoami(token=token)
+            return True
+        except HTTPError:
+            return False
```
LysandreJik, line 596: Nice helper, would it be possible to add a bit of documentation?

src/huggingface_hub/hf_api.py
```diff
@@ -591,6 +623,23 @@ def create_repo(
...
+                    "`huggingface-cli login`."
+                )
+        elif not self._is_valid_token(token):
+            if self._is_valid_token(name):
+                warnings.warn(
+                    "`create_repo` now takes `token` as an optional positional argument. "
+                    "Be sure to adapt your code!",
+                    FutureWarning,
+                )
+                token, name = name, token
```
LysandreJik, line 640: Perfect!

LysandreJik (commented):
This is a very welcome change which should make many things clearer. I think having the `token` as first parameter in most methods isn't ideal, as it is generally not the key parameter of the method.

Ont nit, however: would it be possible to split the PR in two, with one PR changing the API and the second changing the tests? Modifying both at the same time is a bit dangerous as we try to prevent breaking changes.

@osanseviero, this PR introduces a change in the methods', but Sylvain also introduced mechanisms to be backwards compatible as it would otherwise be very breaking. This will, however, change in a future version with future warnings removed.

Do you know if libraries have an unpinned version of `huggingface_hub`, and, if so, if they use any of the changed methods here? Namely, `create_repo`, `delete_repo`, `update_repo_visibility`, and `upload_file`? Thank you.

LysandreJik:
Also cc @stevhliu, could you link which documentation pages should be updated to reflect this change? Thanks a lot

stevhliu:
I think these are the only two pages that will need to be updated:

- [Integrate a library with the Hub
](https://huggingface.co/docs/hub/adding-a-library)
- [How to integrate upstream utilities in your library](https://huggingface.co/docs/hub/how-to-upstream)

Let me know if I can help update these :)

sgugger:
I've already updated those to the new API, normally.

osanseviero:
Sorry for the delay with my response. Thanks for this PR @sgugger!!

I think there are 4 actionable repos (1 officially supported, 3 non official)

**Officially supported libraries**

* I think we definitively need to change this (unpinned, not using keyword args)
https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L437-L458

* AllenNLP is fine but we can simplify the code with this PR in a later stage
https://github.com/allenai/allennlp/blob/main/allennlp/common/push_to_hf.py#L136-L150

* spaCy is also fine, and again we can simplify it with this change
https://github.com/explosion/spacy-huggingface-hub/blob/main/spacy_huggingface_hub/push.py#L81-L89


**Non official**

* Glasses https://github.com/FrancescoSaverioZuppichini/glasses/blob/master/glasses/utils/weights/storage/hubs/HFModelHub.py#L202-L210
* HuggingTweets https://github.com/borisdayma/huggingtweets/blob/8a4cdefd8b7ef9fef56c76190b5afb97b444108b/huggingtweets-demo.ipynb
* HuggingArtitsts https://github.com/AlekseyKorshuk/huggingartists/blob/master/huggingartists-demo.ipynb

LysandreJik:
Thank you for doing the research @osanseviero!

@Narsil this touches to some of the code in api-inference-community, could you take a quick look at the changes?

LysandreJik (approved):
LGTM! As seen with @sgugger and @osanseviero, the only library that uses an unpinned version of `huggingface_hub` and that would potentially break is `sentence-transformers` cc @nreimers.

The workarounds in `create_repo` should prevent any breaking change, however.

PR #385, opened 2021-10-01
Link to new guides

stevhliu:
This PR adds links to the new guides in the Hub docs sidebar.

### DIFF
--- a/docs/hub/_sections.yml
+++ b/docs/hub/_sections.yml
@@ -19,8 +19,17 @@
 - local: spaces
   title: Spaces documentation (beta)
 
-- local: adding-a-library
-  title: Integrating a library to the Hub
+- local: tutorial-add-library
+  title: Integrate a library with the Hub
+
+- local: how-to-downstream
+  title: How to download files from the Hub
+
+- local: how-to-upstream
+  title: How to create repositories and upload files to the Hub
+
+- local: how-to-inference
+  title: How to set up Inference API with your library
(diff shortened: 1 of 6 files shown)


docs/hub/_sections.yml
```diff
@@ -19,8 +19,17 @@
 - local: spaces
   title: Spaces documentation (beta)
 
-- local: adding-a-library
-  title: Integrating a library to the Hub
+- local: tutorial-add-library
```
osanseviero, line 22: I think we should replace `adding-a-library` entirely but keeping the same name.

osanseviero: No strong opinion so if you prefer to have the tutorial prefix it's fine

osanseviero:
Thanks a lot!!

docs/hub/_sections.yml
```diff
@@ -20,7 +20,16 @@
...
+  title: Integrate a library with the Hub
+
+- local: how-to-downstream
+  title: How to download files from the Hub
+
+- local: how-to-upstream
+  title: How to create repositories and upload files to the Hub
+
+- local: how-to-inference
+  title: How to set up Inference API with your library
```
julien-c, line 32: to me this sounds a bit confusing vs. how to actually spawn an inference API if you're a library author

I'd suggest changing to something like `How to call the Inference API from Python code`

WDYT?

PR #387, opened 2021-10-05
don't check if 'json' in filename when loading config file

nateraw:
Checking if json is in the filename can be problematic when loading from the hub, as the config file name might be a hash.


osanseviero (approved):
LGTM

LysandreJik (approved):
LGTM!

PR #388, opened 2021-10-05
Root methods instead of `HfApi`

LysandreJik:
This PR offers methods as root methods instead of leveraging the `HfApi`. Since the `HfApi` is essentially stateless in most use-cases, there is essentially no need to add an additional layer between the user and what they want to do.

This offers repository management methods at the root of `huggingface_hub`.

I've updated the docs alongside it, @stevhliu I should not have forgotten anything unless I'm mistaken. 

This is fully backward compatible.


julien-c:
i think this makes sense

sgugger (approved):
The resulting API is so much nicer :heart_eyes:

PR #391, opened 2021-10-06
Add PipelineCategory and PIPELINE_TAG_CATEGORIES to Types.ts

beurkinger:


osanseviero (approved):
This looks good to me! Thanks for the PR

widgets/src/lib/interfaces/Types.ts
```diff
@@ -57,8 +58,33 @@ export enum FinerGrainedTaskType {
 	"speech-enhancement"                                      = "Speech Enhancement",
 }
 
+export type PipelineCategory = "audio" | "computer-vision" | "nlp" | "other";
 
-export const ALL_PIPELINE_TYPES = Object.keys(PipelineType) as (keyof typeof PipelineType)[];
+export const PIPELINE_TAG_CATEGORIES: Record<keyof typeof PipelineType, PipelineCategory> = {
```
julien-c, line 63:
```suggestion
export const PIPELINE_TAG_MODALITIES: Record<keyof typeof PipelineType, PipelineModality> = {
```

suggestion

widgets/src/lib/interfaces/Types.ts
```diff
@@ -57,8 +58,33 @@ export enum FinerGrainedTaskType {
 	"speech-enhancement"                                      = "Speech Enhancement",
 }
 
+export type PipelineModality = "audio" | "computer-vision" | "nlp" | "other";
```
julien-c, line 61: (nit) if we write "nlp", why do we write "computer-vision" and not "cv"?

beurkinger: works for me !

PR #405, opened 2021-10-12
TextGeneration widgets use text area

mishig25:
Modified Text Generation widgets to use text areas, rather than single line text inputs

Screenshot 2021-10-12 at 11 02 19

[Netlify demo here](https://61654f41bb82bbd611dffda1--huggingface-widgets.netlify.app/)

cc: @julien-c @VictorSanh @NielsRogge


julien-c:
maybe it should be auto-resizing (up to a certain max height) to auto-fit its content?

Also outside of this scope but I'm wondering if we should do a widget more like OpenAI API's (or https://transformer.huggingface.co/) where the text is generated inside the input itself (maybe rendered differently)?? @osanseviero @merveenoyan @LysandreJik

PR #408, opened 2021-10-12
Add fastai as supported library

omarespejel:
Required for UI items in the Hub and as a step towards the integration of fastai.

- I followed the integration made for sklearn in this document
- fastai models are usually stored and imported from *.pkl formats so this would make sense for practitioners
- The 'espejelomar/fastai-pet-breeds-classification' model can be used as an example of how this would work:

```
from fastai.learner import load_learner

model = load_learner(
    hf_hub_download('espejelomar/fastai-pet-breeds-classification', filename="model.pkl")
    )
```

### DIFF
--- a/widgets/src/lib/interfaces/Libraries.ts
+++ b/widgets/src/lib/interfaces/Libraries.ts
@@ -17,7 +17,7 @@ export enum ModelLibrary {
 	'speechbrain'            = 'speechbrain',
 	'tensorflowtts'          = 'TensorFlowTTS',
 	'timm'                   = 'Timm',
-	'fastai'           	 = 'Fastai',
+	'fastai'           	 = 'fastai',
 	'transformers'           = 'Transformers',
 };
 


osanseviero:
I think one big challenge here is that this does not work with custom code, so loading a model with `load_learner` will not work most of the time. 

* One alternative is to merge this as is, similar to how we've done for sklearn. It's not perfect, but is a first step towards a solution.
* Another alternative is to properly do the integration in the fastai codebase OR have a mixin as we've done for keras/torch that allows users to push and load models from the Hub.

WDYT @nateraw @LysandreJik @julien-c ? (and @muellerzr if you have other ideas)

osanseviero:
FYI some time ago I did a demo of FastAI in the Hub with the Generic Inference API which has a similar idea

* model repo: https://huggingface.co/osanseviero/fastai_cat_vs_dog
* pipeline file: https://huggingface.co/osanseviero/fastai_cat_vs_dog/blob/main/pipeline.py (which imports other files, so it's a bit different)

Just sharing in case it's useful

PR #416, opened 2021-10-17
Add fastai upstream and downstream capacities

omarespejel:
## Premise.
As we have a [huggingface_hub/keras_mixin.py](https://github.com/huggingface/huggingface_hub/blob/10e7445506fe547c916a21ca7535a88dc11c915c/src/huggingface_hub/keras_mixin.py) implementation to allow for easy upstream and downstream tasks for Keras library, I considered creating an equivalent for the fastai library. A similar approach was suggested [here](https://github.com/huggingface/huggingface_hub/pull/408#issuecomment-941018069) so to allow easy loading and pushing from the hub.

Goal:
1. Add upstream support: push a fastai learner to the HF Hub. See **`save_fastai_learner`** and **`push_to_hub_fastai`**.
2. Add downstream support: download a fastai learner from the hub. See **`from_pretrained_fastai`**.

* Note: The code verifies one of the most recent versions of fastai is being used.

## Problem trying to be solved.
Few fastai models are present in the Hub (only three until today). This would provide the fastai community an easy way to push and load their models to the Hub.

## Examples of how it would look.
- Load a learner with: 

```
from huggingface_hub import from_pretrained_fastai
learner = from_pretrained_fastai("espejelomar/fastai-pet-breeds-classification")
```

- Push a learner with: 
```
from huggingface_hub import push_to_hub_fastai
push_to_hub_fastai(learner, 
                   repo_path_or_name='data/',
                   repo_url='https://huggingface.co/espejelomar/pruebita',
                   use_auth_token='*********************',
                   git_user='espejelomar',  
                   git_email='espejelomar@gmail.com',
                   config=config)
```

## Limitations and next steps.
- Going from storing/downloading a `fastai.learner` to saving the weights directly into de hub. See [PR #408's comments](https://github.com/huggingface/huggingface_hub/pull/408) for more information.
- The current approach assumes the model is stored in .pkl format. The name of the .pkl does not matter since the code looks for a .pkl file with a `fastai.learner` stored.

## PRs/Tasks involved.
- [Add fastai as supported library #408](https://github.com/huggingface/huggingface_hub/pull/408) - this PR, looking to add a downstream support description into fastai model cards, would be obsolete since we would have a new approach for downstream tasks.
- Add in [huggingface_hub/file_download.py](https://github.com/huggingface/huggingface_hub/blob/10e7445506fe547c916a21ca7535a88dc11c915c/src/huggingface_hub/file_download.py) a function to verify fastai is properly installed. This task is currently being done directly in the present code.
- Add to [huggingface_hub.constants](https://github.com/huggingface/huggingface_hub/blob/10e7445506fe547c916a21ca7535a88dc11c915c/src/huggingface_hub/constants.py) the constant FASTAI_LEARNER_NAME to have the same name in all the .pkl models pushed to the hub.

### DIFF
--- a/src/huggingface_hub/fastai_mixin.py
+++ b/src/huggingface_hub/fastai_mixin.py
@@ -1,0 +1,227 @@
+import json
+import logging
+import os
+from pathlib import Path
+from typing import Any, Dict, Optional, Union
+
+logger = logging.getLogger(__name__)
+
+try:
+  from fastai.learner import load_learner
+except ImportError as error:
+        logger.error(error.__class__.__name__ + ": fastai version above or equal to 2.5.1 required. Run, for example, `pip install fastai==2.5.1`.")
+
+
+from huggingface_hub import ModelHubMixin
+from huggingface_hub.constants import CONFIG_NAME
+# TODO - add to huggingface_hub.constants the constant FASTAI_LEARNER_NAME: same name to all the .pkl models pushed to the hub.
+from huggingface_hub.hf_api import HfFolder, HfApi
+from huggingface_hub.repository import Repository
+from huggingface_hub.snapshot_download import snapshot_download
+
+
+logger = logging.getLogger(__name__)
+
+
+def save_fastai_learner(
+    learner, save_directory: str, config: Optional[Dict[str, Any]] = None
+):
+    """Saves a fastai learner to save_directory in pickle format. Use this if you're using Learners.
+
+    learner:
+        The `fastai.learner` you'd like to save.
+    save_directory (:obj:`str`):
+        Specify directory in which you want to save the fastai learner.
+    config (:obj:`dict`, `optional`):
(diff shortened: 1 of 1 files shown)


muellerzr:
Hey Omar, just now taking a look at this :)

Excellent work. I think one area we could improve upon is rather than limiting/pinning the fastai version to be used, I think a more stable route could be to store the versions of fastai and its dependencies being used. As it stands, it should be something like so:

1. Check if `fastai` <= 1.0.61, if so, store that version used
2. If not, check if `fastai` < 3.0.0, if so, store that version being used
3. If we're using 2.x, we should then also note the `fastcore` version being used.

This is needed because fastai v2 is two libraries, fastai and fastcore, and a common deployment issue is not getting the fastcore versions correct. 

Likely this can live in the `config.json` and we do a check when *loading* the model, not necessarily the mixin module

omarespejel:
Hi Zach! Thank you for your kind comment 🤗.

Great! Let me proceed with these suggested changes. I agree they are required.

PR #449, opened 2021-11-01
Refresh Spaces docs

stevhliu:
As discussed in #443, this PR reorganizes the Spaces docs from a question-answer format to something with more structure and narrative. Some things to note:

- The link to the RAG app is broken. Do we want to include a link to the Facebook (or Meta? 🥸) [RAG blog post](https://ai.facebook.com/blog/retrieval-augmented-generation-streamlining-the-creation-of-intelligent-natural-language-processing-models/) instead?
- Still need to find a home for the Changelog. Maybe we can include them as [release](https://github.com/huggingface/datasets/releases) notes as Datasets does?


julien-c:
can you please tag @cbensimon the lead engineer for Spaces?

Personally I'm not a huge fan of this rewrite (and i think we still need the changelog – this is not a library so GitHub release notes are not really an option)

stevhliu:
Thanks for the feedback! :) Is there some aspect of the rewrite you aren't a fan of?

Maybe we can create a new Changelog section on the side for the Hub as a whole? 🤔  This way we can also document changes to other aspects of the Hub.

julien-c:
> Is there some aspect of the rewrite you aren't a fan of?

a bit swamped right now but I'll give it a go soon

I would just keep the changelog where it is for now

julien-c:
rebased on master + tweaks

julien-c (approved):

stevhliu:
LGTM, merging now :)

PR #450, opened 2021-11-01
Git prune

LysandreJik:
Adds the option to prune local files to prevent the local folder from exploding in size.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -840,6 +840,27 @@ class Repository:
 
         return files_to_be_tracked_with_lfs
 
+    def lfs_prune(self, recent=False):
+        """
+        git lfs prune
+        """
+        args = "git lfs prune".split()
+        if recent:
+            args.append("--recent")
+        try:
+            with lfs_log_progress():
+                result = subprocess.run(
+                    args,
+                    stderr=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    check=True,
+                    encoding="utf-8",
+                    cwd=self.local_dir,
+                )
+                logger.info(result.stdout)
+        except subprocess.CalledProcessError as exc:
+            raise EnvironmentError(exc.stderr)
+
     def git_pull(self, rebase: Optional[bool] = False):
         """
         git pull
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -960,7 +960,7 @@ class Repository:
                 Setting this to `False` will return an `CommandInProgress` object
                 which has an `is_done` property. This property will be set to
                 `True` when the push is finished.
-            auto_lfs_prune (`bool`, defaults to `True`):
+            auto_lfs_prune (`bool`, defaults to `False`):
                 Whether to automatically prune files once they have been pushed to the remote.
         """
         command = "git push"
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/repository.py
```diff
@@ -29,20 +29,27 @@ def __init__(
...
     @property
     def is_done(self) -> bool:
         """
         Whether the process is done.
         """
-        return self._is_done()
+        result = self._is_done()
+
+        if result and self._post_method is not None:
+            self._post_method()
```
sgugger, line 50: Very smart!

sgugger (approved):
Very nice new API and cool new tests, thanks a lot!

src/huggingface_hub/repository.py
patrickvonplaten: I don't know the code too well, but it's a bit weird to me that a property "is this command done" now calls a `"post"` method. Wouldn't it be cleaner to separate the `self._post_method()` from `is_done`?

tests/test_repository.py
```diff
@@ -735,6 +735,168 @@ def test_delete_tag(self):
         repo.delete_tag("v4.6.0", remote="origin")
         self.assertFalse(repo.tag_exists("v4.6.0", remote="origin"))
 
+    def test_lfs_prune(self):
```
patrickvonplaten, line 738: Great tests!

src/huggingface_hub/repository.py
sgugger: In that case the user will need to manually check when an async git commit is finished to call the post method, which is what we are trying to avoid.

src/huggingface_hub/repository.py
```diff
@@ -986,12 +1017,16 @@ def status_method():
                 is_done_method=lambda: process.poll() is not None,
                 status_method=status_method,
                 process=process,
+                post_method=self.lfs_prune if auto_lfs_prune else None,
```
sgugger, line 1020: Maybe only do this when `blocking=False`? When `blocking=True`, we can run the instruction at the end.

LysandreJik: This is already the case, this whole block is under `if not blocking`!

https://github.com/huggingface/huggingface_hub/blob/4152f15e7c28fb2afda3f3ece4b5abfd0b8a564e/src/huggingface_hub/repository.py#L1006-L1025

src/huggingface_hub/repository.py
LysandreJik: I understand what you mean Patrick, let me give you some context: we're leveraging the `subprocess.Popen` approach to launch the process in a non-blocking manner. This approach means we're spawning a subprocess and keeping track of it in the `CommandInProgress` utility.

The process does not have the ability to let the Python runtime know when it's done: it's up to the Python runtime to check when it is done by checking its `status`.

Therefore, here, we have three options:
1. Check the status regularly and when it is done run the `git lfs prune`
2. Let the user check the status, and when it is done run the `git lfs prune`
3. Don't do any of that and let the user handle the `git lfs prune` when they feel like it

We've chosen to do 2., as that's the least intensive way of doing things; but approach 1. might be better as the `git lfs prune` might take a bit of time for larger files, which is currently blocking when the user checks for `is_done`. For now I'm going to add a log message when it is pruning, but I'm open to switching from 2. to 1. if you feel strongly.

3. is not a great user API.

src/huggingface_hub/repository.py
sgugger: Ah sorry, I didn't catch that :-)

src/huggingface_hub/repository.py
patrickvonplaten: Good for me!

patrickvonplaten (approved):

PR #451, opened 2021-11-03
Add video classification to api_inference_community

nateraw:


api-inference-community/api_inference_community/validation.py
```diff
@@ -355,3 +359,11 @@ def normalize_payload_nlp(bpayload: bytes, task: str) -> Tuple[Any, Dict]:
...
     check_inputs(inputs, task)
     return inputs, parameters
+
+
+def normalize_payload_video(bpayload: bytes) -> Tuple[Any, Dict]:
+
+    from pytorchvideo.data.encoded_video_pyav import EncodedVideoPyAV
+
+    clip = EncodedVideoPyAV(BytesIO(bpayload))
+    return clip, {}
```
nateraw, line 369: We return `EncodedVideo` here because it allows user to access video/audio slices of the input video dynamically. Same idea as providing `PIL.Image` for image-based tasks.

api-inference-community/requirements.txt
```diff
@@ -5,3 +5,6 @@ parameterized==0.8.1
 pillow==8.2.0
 huggingface_hub>=0.0.12
 datasets==1.10
+pytorchvideo==0.1.3
+torch==1.9.0
```
nateraw, line 9: This `torch` dependency is the exact reason why I wrote [this package](https://github.com/nateraw/encoded-video). With my version, we can forego the need for `torch`. Can include that logic here.

What do you think @Narsil?

nateraw (commented):
The idea here is to use `pytorchvideo` directly for now while we figure things out. Will issue separate PR for `pytorchvideo` pipelines 😄

api-inference-community/api_inference_community/validation.py
Narsil: We shouldn't do that IMO, pytorch is not a standard so we shouldn't try to use it here.

The common grounds for a video, is IMHO a numpy array for a clip, and ideally a streaming series of clips, because AFAIK, most videos will explode RAM/GPU RAM usage if we try to feed everything at once, is that assumption correct ?

Do we need to have streaming capabilities for this to work ?

numpy albeit working and be relatively agnostic in the python ecosystem would actually be lacking more metadata compared to real video, (FPS, color scheme, audio channels, sampling rate.). Not sure how much standardized we want to go here.

For `audio`, the problem is reversed, the pipeline needs to declare it's intended `sampling_rate`, maybe we could go the same route, and have video pipelines declare there expected metadata, then feeding them that data in numpy arrays would be OK IMO.

WDYT ?

Narsil: Also, `av` could help do the actual interpretation of the bytes ( `ffmpeg` is ok too).

api-inference-community/requirements.txt
nateraw: We are in agreement here! We clearly need an agnostic solution.

Issue #452, opened 2021-11-03
ImportError: cannot import name 'TypeAlias' from 'typing_extensions'


lhoestq:
In `hugginggface_hub` 0.1.0 on python 3.7.2 I'm getting this error when importing `huggingface_hub`:

```python
~/.virtualenvs/hf-hub/lib/python3.7/site-packages/huggingface_hub/__init__.py in <module>
     53     whoami,
     54 )
---> 55 from .hub_mixin import ModelHubMixin, PyTorchModelHubMixin
     56 from .inference_api import InferenceApi
     57 from .keras_mixin import (

~/.virtualenvs/hf-hub/lib/python3.7/site-packages/huggingface_hub/hub_mixin.py in <module>
      9 from .file_download import hf_hub_download, is_torch_available
     10 from .hf_api import HfApi, HfFolder
---> 11 from .repository import Repository
     12 from .utils import logging
     13 

~/.virtualenvs/hf-hub/lib/python3.7/site-packages/huggingface_hub/repository.py in <module>
     13 
     14 from huggingface_hub.constants import REPO_TYPES_URL_PREFIXES, REPOCARD_NAME
---> 15 from huggingface_hub.repocard import metadata_load, metadata_save
     16 
     17 from .hf_api import ENDPOINT, HfApi, HfFolder, repo_type_and_id_from_hf_id

~/.virtualenvs/hf-hub/lib/python3.7/site-packages/huggingface_hub/repocard.py in <module>
      6 
      7 import yaml
----> 8 from huggingface_hub.repocard_types import (
      9     ModelIndex,
     10     SingleMetric,

~/.virtualenvs/hf-hub/lib/python3.7/site-packages/huggingface_hub/repocard_types.py in <module>
      4 from typing import Any, List, Optional
      5 
----> 6 from typing_extensions import TypeAlias
      7 
      8 

ImportError: cannot import name 'TypeAlias' from 'typing_extensions' (/Users/quentinlhoest/.virtualenvs/hf-hub/lib/python3.7/site-packages/typing_extensions.py)
```

I think this is because my `typing_extensions` is outdated. Maybe let's add a minimum version to `typing_extensions` in `setup.py` ?

LysandreJik:
Indeed, good catch!

otvj776:
pip uninstall typing_extensions
pip install typing_extensions==4.0.0

PR #453, opened 2021-11-03
Fix typing-extensions minimum version

lhoestq:
To be able to do
```python
from typing_extensions import TypeAlias
```

we need to have `typing-extensions>=3.7.4.3`

Fix https://github.com/huggingface/huggingface_hub/issues/452


LysandreJik (approved):
Great, thank you!

PR #459, opened 2021-11-05
Fix argument order in `create_repo` for `Repository.clone_from`

sgugger:
This fixes the deprecation warning one gets when creating a `Repository` object.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -540,8 +540,8 @@ class Repository:
 
                 if namespace == user or namespace in valid_organisations:
                     api.create_repo(
-                        token,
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/repository.py
```diff
@@ -540,8 +540,8 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
 
                 if namespace == user or namespace in valid_organisations:
                     api.create_repo(
-                        token,
                         repo_id,
+                        topken=token,
```
LysandreJik, line 544:
```suggestion
                        token=token,
```
:)

PR #461, opened 2021-11-08
Remove beta Spaces

albertvillanova:
Remove "beta" from Spaces in the docs side navigation menu.


julien-c:
we decided to keep it for now as it's still in Public Beta (expected to change in the coming months)

Thanks! will close for now :)

PR #462, opened 2021-11-08
clean_ok should be True by default

LysandreJik:
Should default to `True` so that two consecutive `push_to_hub` do not end up failing. Will greatly reduce friction in examples/downstream usage.


sgugger (approved):
Yes, thanks a lot for this fix!

PR #470, opened 2021-11-15
Fix adapter transformers

osanseviero:
Update test suite to have same as common and update inference output to be a nested list as expected by the widget.


Narsil (approved):
LGTM.

PR #472, opened 2021-11-15
Add ability to skip large files

### LINKED ISSUE #424: [Feature request] Add ability to skip large files when creating Repository
When instantiating a `Repository` object, it would be great to skip large files, similar to how you can prepend a git clone with the following env var: GIT_LFS_SKIP_SMUDGE=1. This is handy when updating existing checkpoints from the hub (otherwise it first takes a long time to instantiate the `Repository` object).

NielsRogge:
Fixes #424 

Started a draft to add the ability to skip large files when git cloning a repo from the hub (as I need it ASAP).

To do:

- [ ] add corresponding test

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -365,7 +365,7 @@ class Repository:
         git_email: Optional[str] = None,
         revision: Optional[str] = None,
         private: bool = False,
-        skip_lfs: bool = False,
+        skip_lfs_files: bool = False,
     ):
         """
         Instantiate a local clone of a git repo.
@@ -397,7 +397,7 @@ class Repository:
                 branch will be created with that revision name from the default branch's current HEAD.
             private (``bool``, `optional`, defaults to ``False``):
                 whether the repository is private or not.
-            skip_lfs (``bool``, `optional`, defaults to ``False``):
+            skip_lfs_files (``bool``, `optional`, defaults to ``False``):
                 whether to skip git-LFS files or not.
         """
 
@@ -406,7 +406,7 @@ class Repository:
         self.repo_type = repo_type
         self.command_queue = []
         self.private = private
-        self.skip_lfs = skip_lfs
+        self.skip_lfs_files = skip_lfs_files
 
         self.check_git_versions()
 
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/repository.py
```diff
@@ -365,6 +365,7 @@ def __init__(
         git_email: Optional[str] = None,
         revision: Optional[str] = None,
         private: bool = False,
+        skip_lfs: bool = False,
```
LysandreJik, line 368: The `skip_lfs` argument isn't super clear, what about `download_lfs_fles` set to `True` by default?

src/huggingface_hub/repository.py
```diff
@@ -384,7 +385,7 @@ def __init__(
                 repository url (e.g. ``'https://huggingface.co/philschmid/playground-tests'``).
             repo_type (``str``, `optional`):
                 To set when creating a repo: et to "dataset" or "space" if creating a dataset or space, default is model.
-            use_auth_token (``str`` or ``bool``, `optional`, defaults ``None``):
+            use_auth_token (``str`` or ``bool``, `optional`, defaults to ``True``):
```
LysandreJik, line 388: Good, thanks

src/huggingface_hub/repository.py
```diff
@@ -566,9 +570,12 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
             # checks if repository is initialized in a empty repository or in one with files
             if len(os.listdir(self.local_dir)) == 0:
                 logger.warning(f"Cloning {clean_repo_url} into local empty directory.")
+                prefix = ""
+                if self.skip_lfs:
+                    prefix = "GIT_LFS_SKIP_SMUDGE=1 "
```
LysandreJik, line 575: The `subprocess.run` method takes as argument the `env` parameter, which contains the values for the environment variables. I'd rather we leverage this as it's the most cross-platform way to handle things. 

See the docs here https://docs.python.org/3/library/subprocess.html

Example usage here https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment

LysandreJik (commented):
Great, thanks for your PR! Left a few comments.

Please also write tests leveraging the `staging` endpoint, like it's already done for other `Repository` tests. In order to test this, please use tiny files that are sill LFS tracked (thanks to their extension, for example). You can check for the size of the file to check if the pointer was downloaded or if the full file was downloaded.

src/huggingface_hub/repository.py
NielsRogge: I've renamed it to `skip_lfs_files`, if that's ok for you

julien-c: Your call, but I also personally think that `skip_lfs` (or `skip_lfs_files`) defaulting to False is clearer (and consistent w/ the env var)

LysandreJik: Yes, sounds good

LysandreJik:
Needed this for another PR, will add the tests in the corresponding PR.

PR #473, opened 2021-11-15
Typo fix: huggingface-cli for login

snoop2head:
Minor fix for the command line interface when using huggingface-cli login function.


LysandreJik (approved):
Thank you @snoop2head!

PR #474, opened 2021-11-16
Update some widgets to use `textarea` rather than `input`

mishig25:
1. Updated widgets (FillMaskWidget (which is TextClsWidget), TokenClsWidget, TextToSpeechWidget) to use `textarea` as their input element, rather than `input`. 
2. Made `textarea` input height auto-growable until 500px

[Checkout netlfy demo here](https://6193b4a5ec227207e6ff4874--huggingface-widgets.netlify.app/)

cc: @Pierrci @lewtun 

Closes https://github.com/huggingface/moon-landing/issues/1523

<video src="https://user-images.githubusercontent.com/11827707/141996190-8c8f551f-f4d0-48b6-b451-071ea81737f7.mov" controls autoplay loop/>

PR #483, opened 2021-11-19
Hotfix fillmask widget

mishig25:
While adding a hack to bypass firefox (noted [here](https://github.com/huggingface/huggingface_hub/pull/474#pullrequestreview-808286966)), I broke the fillmask widget. This pr is a hotfix for that.


mishig25:
cc: @gary149 @Pierrci

PR #486, opened 2021-11-19
Fix filename for external dependencies in a Space

lewtun:
This PR fixes a mismatch between the filename `package.txt` needed to install external dependencies in a Space and what is actually used in the Docker logs:

```
Step 5/25 : COPY packages.txt /root/packages.txt
```

This problem was reported by a participant in the 🤗 course event 😉


julien-c (approved):
good catch, cc'ing @cbensimon

PR #487, opened 2021-11-19
Remove "beta" label from Spaces documentation

lewtun:
Since Spaces is now out of beta, this PR updates the section title to remove this label. You can see the current state in the screenshot below:

Screen Shot 2021-11-19 at 18 56 14

### DIFF
--- a/docs/hub/_sections.yml
+++ b/docs/hub/_sections.yml
@@ -17,7 +17,7 @@
   title: Hub API Endpoints
 
 - local: spaces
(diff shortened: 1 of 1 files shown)


Pierrci:
I think we keep the `beta` label on purpose (see https://github.com/huggingface/huggingface_hub/pull/461) :)

lewtun:
> I think we keep the `beta` label on purpose (see #461) :)

Oh sorry for not seeing this in the close PRs! Closing this for the same reason :)

PR #488, opened 2021-11-22
stanza pos-tagging added

merveenoyan:
Added pos-tagging in stanza widgets.

cc: @osanseviero

### DIFF
--- a/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
+++ b/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
@@ -36,9 +36,9 @@ class TokenClassificationPipeline(Pipeline):
         """
         doc = self.model(inputs)
 
+        entities = []
         if "ner_model_path" in self.model.config.keys():
 
-            entities = []
             for entity in doc.entities:
                 entity_dict = {
                     "entity_group": entity.type,
@@ -48,20 +48,20 @@ class TokenClassificationPipeline(Pipeline):
                     "score": 1.0,
                 }
                 entities.append(entity_dict)
-            return entities
-
+            
         else:
-            pos_tags = []
 
             for sent in doc.sentences:
-                for word in sent.words:
-                    pos_dict = {
-                        "pos_group": word.upos,
-                        "word": word.text,
-                        "start": word.start_char,
-                        "end": word.end_char,
+                for entity in sent.words:
+                    
+                    entity_dict = {
+                        "entity_group": entity.upos,
+                        "word": entity.text,
+                        "start": entity.start_char,
+                        "end": entity.end_char,
                         "score": 1.0,
                     }
-            pos_tags.append(pos_dict)
 
-            return pos_tags
+                    entities.append(entity_dict)
+
+        return entities
--- a/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
+++ b/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
@@ -48,12 +48,12 @@ class TokenClassificationPipeline(Pipeline):
                     "score": 1.0,
                 }
                 entities.append(entity_dict)
-            
+
         else:
 
             for sent in doc.sentences:
                 for entity in sent.words:
-                    
+
                     entity_dict = {
                         "entity_group": entity.upos,
                         "word": entity.text,
--- a/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
+++ b/api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
@@ -36,14 +36,32 @@ class TokenClassificationPipeline(Pipeline):
         """
         doc = self.model(inputs)
 
-        entities = []
-        for entity in doc.entities:
-            entity_dict = {
-                "entity_group": entity.type,
-                "word": entity.text,
-                "start": entity.start_char,
-                "end": entity.end_char,
-                "score": 1.0,
-            }
-            entities.append(entity_dict)
-        return entities
+        if "ner_model_path" in self.model.config.keys():
+
+            entities = []
+            for entity in doc.entities:
+                entity_dict = {
+                    "entity_group": entity.type,
+                    "word": entity.text,
+                    "start": entity.start_char,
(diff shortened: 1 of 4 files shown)


api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
```diff
@@ -36,14 +36,32 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
...
+                entities.append(entity_dict)
+            return entities
+
+        else:
+            pos_tags = []
+
+            for sent in doc.sentences:
+                for word in sent.words:
+                    pos_dict = {
+                        "pos_group": word.upos,
```
Narsil, line 59: I don' t `pos_group`  will be understood by the widget.

Is it possible to stick to `entity_group` for the name of the group ?

That will keep the pipeline simpler too probably.

```python
use_type = "ner_model_path" in self.model.config.keys()
...

   "entity_group" : entity.type if use_type else entity.upos

```

Seems slightly cleaner to me too (removing the outer `if/else`).

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
```diff
@@ -36,14 +36,32 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
...
+                        "word": word.text,
+                        "start": word.start_char,
+                        "end": word.end_char,
+                        "score": 1.0,
+                    }
+            pos_tags.append(pos_dict)
```
osanseviero, line 65: is this append right? Should it be indented so it's within the loop at same level of pos_dict definition?

Narsil (changes requested):
I think we should keep `entity_group` key regardless for the widget to work.

It should make the code simpler everywhere too.

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
merveenoyan: ok it's my stupidity, fixing it

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
merveenoyan: ok I'll do `entity_group`
Thanks for the feedback!

api-inference-community/docker_images/stanza/tests/test_api_token_classification.py
```diff
@@ -59,10 +59,16 @@ def test_simple(self):
...
+        if "entity_group" in (el.keys() for el in content):
+            self.assertEqual(
+                set(k for el in content for k in el.keys()),
+                {"entity_group", "word", "start", "end", "score"},
+            )
+        elif "pos_group" in (el.keys() for el in content):
+            self.assertEqual(
+                set(k for el in content for k in el.keys()),
+                {"pos_group", "word", "start", "end", "score"},
+            )
```
osanseviero, line 71: Quick suggestions
* Try avoiding duplicating too much content if possible
* We still have to check that one of `pos_group` and `entity_group` is present in the keys. 


Something like this, but maybe there are better ideas
```suggestion
        keys = set(el.keys() for el in content)
        self.assertTrue(any((True for x in keys if x in ["pos_group", "entity_group"])))
        key = "entity_group" if "entity_group" in keys else "pos_group"
        self.assertEqual(
              set(k for el in content for k in el.keys()),
              {key, "word", "start", "end", "score"},
          )
```

osanseviero (commented):
Thanks for this PR. I have a couple of suggestions :)

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
merveenoyan: What you're iterating over is different depending on the key so it can't simplify the code but I'll return the same name & variables

merveenoyan: in NER I iterate over `doc.entities` (models that have NER have POS tagging too) and in POS tagging I iterate over `.words` instead so I don't know a neat way to make it into one loop

api-inference-community/docker_images/stanza/tests/test_api_token_classification.py
merveenoyan: I got rid of it altogether, I'll see if it works

Narsil (approved):
LGTM

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
```diff
@@ -37,13 +37,31 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
...
-            entity_dict = {
-                "entity_group": entity.type,
-                "word": entity.text,
-                "start": entity.start_char,
-                "end": entity.end_char,
-                "score": 1.0,
-            }
-            entities.append(entity_dict)
+        if "ner_model_path" in self.model.config.keys():
+
```
osanseviero, line 41: huge nit: remove this newline

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
```diff
@@ -37,13 +37,31 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
...
+                    "entity_group": entity.type,
+                    "word": entity.text,
+                    "start": entity.start_char,
+                    "end": entity.end_char,
+                    "score": 1.0,
+                }
+                entities.append(entity_dict)
+
+        else:
+
```
osanseviero, line 53: same here

api-inference-community/docker_images/stanza/app/pipelines/token_classification.py
```diff
@@ -37,13 +37,31 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
...
+            for sent in doc.sentences:
+                for entity in sent.words:
+
```
osanseviero, line 56: and here :)

osanseviero (approved):
Thanks! Looks good to me as well

Issue #366, opened 2021-09-22
Tracking integration for text-nearest


osanseviero:
## Tracking integration of task - text-nearest

Naming is tentative 

Note that you're not expected to do all of the following steps. This PR helps track all the steps required to get a new task fully supported in the Hub 🔥

- [ ] Integration with Inference API. Select at least one of the following:
  - [ ] Added a `transformers` `pipeline`
  - [ ] Added to Community Inference API for 3rd party library
  - [ ] Added to Community Inference API for generic
- [ ] Added basic UI elements (icon, order specification, etc)
- [ ] Added a widget

Integration guide: https://hf.co/docs/hub/adding-a-task

We can do this with gensim or fasttext models for example for obtaining closest words with nearest neighbors. Example repo: https://huggingface.co/Hellisotherpeople/debate2vec

CC @mishig25 in case you would like to help with the widget, this is pretty much the same as the `text-classification` widget

julien-c:
isn't it exactly the `text-classification` coarse-grained task though?

osanseviero:
@Narsil and me were discussing about this yesterday and thought a new task would likely be better. Although the input/output specifications could be the same, this is really not a classification problem so we don't think we should model the problem as a coarse-grained subtask of classification.

osanseviero:
But if you feel strongly about this and want to avoid having too many tasks we can also just model it as such for the time being

julien-c:
no strong opinion, but if it's exactly the same API & widget, i would tend to model this as a finer-grained sub task of text-classification

Not 100% sure though!

Narsil:
The widget sort of are exactly the same, the full goal seems a bit more like `text-to-text` (as in `audio-to-audio` and not `text2text`) which would encompass both IMO.

That being said, I do agree that saying it's a `text-classification` is a bit misleading, and we came up with this idea for a name.
Having slept on it, Maybe we can start by keeping `text-classification` (much less code has to be written) and go from there later. That way we can keep mostly everything as is, and it's just slightly misleading on the model page. The README could help convey that information better.

julien-c:
if it's just on the model page, we can definitly add `text-nearest` to https://github.com/huggingface/huggingface_hub/blob/8e2b37c830d43ad3d81b866f2dc3c7778a6da358/widgets/src/lib/interfaces/Types.ts#L51-L58 and find a way to use that to override the widget's name when model has both tags (or something like that)

mishig25:
[...]

PR #490, opened 2021-11-22
Add fasttext to Inference API

osanseviero:
This adds `text-classification` and `feature-extraction`


api-inference-community/docker_images/fasttext/app/pipelines/text_classification.py
```diff
@@ -0,0 +1,29 @@
...
+        Args:
+            inputs (:obj:`str`):
+                a string containing some text
+        Return:
+            A :obj:`list`:. The object returned should be a list of one list like [[{"label": 0.9939950108528137}]] containing:
+                - "label": A string representing what the label/class is. There can be multiple labels.
+                - "score": A score between 0 and 1 describing how confident the model is for this label/class.
+        """
+        if len(inputs.split()) > 1:
+            raise ValueError("Expected input is a single word")
```
Narsil, line 24: That's odd ;) but ok I guess.

api-inference-community/tests/test_dockers.py
```diff
@@ -79,6 +79,19 @@ def test_espnet(self):
             "espnet/kamo-naoyuki_mini_an4_asr_train_raw_bpe_valid.acc.best",
         )
 
+    def test_fasttext(self):
+        self.framework_docker_test(
+            "fasttext",
+            "text-classification",
+            "osanseviero/fasttext_nearest",
```
Narsil, line 86: This is OK for tests, but because repos can only have 1 pipeline_tag on the hub, maybe create a duplicate here so we can also test directly on the hub later ?

Narsil (approved):
LGTM.

Some comments but this can go as is !

api-inference-community/docker_images/fasttext/app/pipelines/text_classification.py
osanseviero: The decision at https://github.com/huggingface/huggingface_hub/issues/366 was to model this task as a text-classification and just modify the input. Maybe in the modified widget we could just enforce a one word input (cc @mishig25)

Narsil: Don't think it's worth it, just probably a little odd on the user end (I am not sure how to fix it properly, this is good enough, and would provide a good message so that's ok to me.

api-inference-community/tests/test_dockers.py
osanseviero: Good point. Now we have https://huggingface.co/osanseviero/fasttext_embedding

PR #498, opened 2021-11-25
Fix skip_lfs_file

LysandreJik:
Fixes the `skip_lfs_file` and adds a test.

The `GIT_LFS_SKIP_SMUDGE` parameter only works with `git clone` and not with `git lfs clone`.

The difference between those two is that `git lfs clone` does the equivalent of the following commands:
```
GIT_LFS_SKIP_SMUDGE=1 git clone git@something.something.git
cd something
git lfs pull
```

We use `git lfs clone` as it provides a simpler environment for progress bars and therefore better feedback for the user. When specifying that the LFS files should not be fetched, the method will be changed to `git clone` using the `GIT_LFS_SKIP_SMUDGE` parameter.

PS: Next PR will try to refactor the `test_repository.py` file so that it takes less time than it currently does.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -571,11 +571,16 @@ class Repository:
             if len(os.listdir(self.local_dir)) == 0:
                 logger.warning(f"Cloning {clean_repo_url} into local empty directory.")
                 env = os.environ.copy()
+
                 if self.skip_lfs_files:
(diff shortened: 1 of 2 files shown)


NielsRogge (approved):
LGTM! Thanks for fixing.

osanseviero (approved):

PR #499, opened 2021-11-26
Adding `image-segmentation` 

Narsil:
Making tasks validation extremely explicit (so we don't miss it).


osanseviero (approved):
Thanks for taking care of this!

PR #505, opened 2021-11-29
[Snapshot download] Allow to load local repo id with snapshot download

patrickvonplaten:
This PR adds more function params to the snapshot download method. @LysandreJik @osanseviero - are there any reasons why we wouldn't add those?

### DIFF
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -1,15 +1,19 @@
-import glob
 import os
+from glob import glob
 from pathlib import Path
 from typing import Dict, Optional, Union
 
 from .constants import DEFAULT_REVISION, HUGGINGFACE_HUB_CACHE
 from .file_download import cached_download, hf_hub_url
 from .hf_api import HfApi, HfFolder
+from .utils import logging
 
 
 REPO_ID_SEPARATOR = "__"
 # ^ make sure this substring is not allowed in repo_ids on hf.co
+
+
+logger = logging.get_logger(__name__)
 
 
 def snapshot_download(
@@ -66,33 +70,73 @@ def snapshot_download(
     # if we have no internet connection we will look for the
     # last modified folder in the cache
     if local_files_only:
-        # list all possible folders that correspond to the repo_id
-        repo_id_folders = glob.glob(
-            os.path.join(cache_dir, repo_id_flattened + "." + revision + ".*")
+        # possible repos have <path/to/cache_dir>/<flatten_repo_id> prefix
+        repo_folders_prefix = os.path.join(cache_dir, repo_id_flattened)
+
+        # list all possible folders that can correspond to the repo_id
+        # and are of the format <flattened-repo-id>.<revision>.<commit-sha>
+        # now let's list all cached repos that have to be included in the revision
+        # 1) cached repos of format <repo_id>.{revision}.<any-hash>
+        repo_folders_branch = glob(repo_folders_prefix + "." + revision + ".*")
+
+        # 2) cached repos of format <repo_id>.{revision}
+        # -> in this case {revision} has to be a commit sha
+        repo_folders_commit_only = glob(repo_folders_prefix + "." + revision)
+
+        # 3) cached repos of format <repo_id>.<any-branch>.{revision}
+        # -> in this case {revision} also has to be a commit sha
+        repo_folders_branch_commit = glob(repo_folders_prefix + ".*." + revision)
+
+        # combine all possible fetched cached repos
+        repo_folders = (
+            repo_folders_branch + repo_folders_commit_only + repo_folders_branch_commit
         )
 
-        if len(repo_id_folders) == 0:
+        if len(repo_folders) == 0:
             raise ValueError(
                 "Cannot find the requested files in the cached path and outgoing traffic has been"
                 " disabled. To enable model look-ups and downloads online, set 'local_files_only'"
                 " to False."
             )
 
-        # find last modified folder
-        storage_folder = max(
-            repo_id_folders,
-            key=os.path.getmtime,
+        # check if repo id was previously cached from a commit sha revision
+        # and passed {revision} is not a commit sha
+        # in this case snapshotting repos locally might lead to unexpected
+        # behavior the user should be warned about
+        repo_folders_commit_any = set(glob(repo_folders_prefix + ".*")) - set(
+            glob(repo_folders_prefix + ".*.*")
         )
+        if (
+            len(repo_folders_commit_any) > 0
+            and len(repo_folders_commit_only + repo_folders_branch_commit) == 0
+        ):
+            logger.warn(
+                f"The repo {repo_id} was previously downloaded from a commit hash revision "
+                f"and has created the following cached directories {repo_folders_commit_any}."
+                f" In this case, trying to load a repo from the branch {revision} in offline "
+                "mode might lead to unexpected behavior by not taking into account the latest "
+                "commits."
+            )
 
+        # find last modified folder
+        storage_folder = max(repo_folders, key=os.path.getmtime)
+
+        # get commit sha
         repo_id_sha = storage_folder.split(".")[-1]
         model_files = os.listdir(storage_folder)
     else:
         # if we have internet connection we retrieve the correct folder name from the huggingface api
         _api = HfApi()
         model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
-        storage_folder = os.path.join(
-            cache_dir, repo_id_flattened + "." + revision + "." + model_info.sha
-        )
+
+        storage_folder = os.path.join(cache_dir, repo_id_flattened + "." + revision)
+
+        # if passed revision is not identical to the commit sha
+        # then revision has to be a branch name, e.g. "main"
+        # in this case make sure that the branch name is included
+        # cached storage folder name
+        if revision != model_info.sha:
+            storage_folder += f".{model_info.sha}"
 
         repo_id_sha = model_info.sha
         model_files = [f.rfilename for f in model_info.siblings]
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -18,7 +18,11 @@ def snapshot_download(
     library_name: Optional[str] = None,
     library_version: Optional[str] = None,
     user_agent: Union[Dict, str, None] = None,
+    proxies=None,
+    etag_timeout=10,
+    resume_download=False,
     use_auth_token: Union[bool, str, None] = None,
+    local_files_only=False,
 ) -> str:
     """
     Downloads a whole snapshot of a repo's files at the specified revision.
@@ -79,7 +83,11 @@ def snapshot_download(
             library_name=library_name,
             library_version=library_version,
             user_agent=user_agent,
+            proxies=proxies,
+            etag_timeout=etag_timeout,
+            resume_download=resume_download,
             use_auth_token=use_auth_token,
+            local_files_only=local_files_only,
         )
 
         if os.path.exists(path + ".lock"):
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -75,8 +75,11 @@ def snapshot_download(
 
         # list all possible folders that can correspond to the repo_id
         # and are of the format <flattened-repo-id>.<revision>.<commit-sha>
-        # now let's list all cached repos that have to be included in the revision
+        # now let's list all cached repos that have to be included in the revision.
+        # There are 3 cases that we have to consider.
+
         # 1) cached repos of format <repo_id>.{revision}.<any-hash>
+        # -> in this case {revision} has to be a branch
         repo_folders_branch = glob(repo_folders_prefix + "." + revision + ".*")
 
         # 2) cached repos of format <repo_id>.{revision}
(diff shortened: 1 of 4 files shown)


patrickvonplaten:
This PR is needed for the integration with https://github.com/kensho-technologies/pyctcdecode/pull/32

patrickvonplaten:
Ah ok - we need internet connection for `_api.model_info(...)` - that's why `local_files_only` doesn't make sense here

LysandreJik:
Nice! We'll need some tests :)

patrickvonplaten:
@LysandreJik @osanseviero - I kind of need to be able to use the snapshot download method for local files only. I've added a working solution to this PR that works as follows:

- Take repo_id and transform into a local folder name. E.g. `kensho/lm_model` into `kensho__lm_model` 
- Then list all cached local folder names, e.g. `[".cache/hub/kensho__lm_model.12341234", ".cache/hub/kensho__lm_model.123411234"]`
- Take the folder hash that was last modified
- Load all files from there

So the somewhat "magic" part here is to take the last modified folder name. Would such a logic be ok for you @LysandreJik ?

LysandreJik:
Thanks for your contribution, @patrickvonplaten! I understand the need for a `snapshot_download` with a `local_files_only` parameter. 

The only issue I can see here is that it does not play well at all with the `revision` parameter:

```py
snapshot_download("xxx")
snapshot_download("xxx", revision="yyy")

snapshot_download("xxx", local_files_only=True)  # Will return the one with revision "yyy"
```

I can't think of a way to go around this limitation, which seems quite troublesome to me. It does seem like a very rare case to me so maybe putting a warning is enough, wdyt?

The following use-case doesn't work either:

```py
snapshot_download("xxx", revision="zzz")
snapshot_download("xxx", revision="yyy")

snapshot_download("xxx", revision="yyy", local_files_only=True)  # Will return the one with revision "yyy"
```

julien-c:
Out of curiosity, why doesn't the canonical method of loading all files by name (like in `transformers` for instance) not work in your use case?

patrickvonplaten:
> Out of curiosity, why does the canonical method of loading all files by name (like in `transformers` for instance) not work in your use case?

The problem is that for https://github.com/kensho-technologies/pyctcdecode the language model is essentially loaded as follows: https://github.com/kensho-technologies/pyctcdecode/blob/b50b5ae39ebb047c42765be0dce2c27137bd9474/pyctcdecode/language_model.py#L352

meaning that it is not loaded by a hard-coded name, but rather the "last file" that is found in the folder `language_model`. We could give the file a name, but then we also have the problem that the file can be both in `.arpa` format or `.bin` format. So we would require an ugly flag `"format="bin"` anymays....

patrickvonplaten:
> Thanks for your contribution, @patrickvonplaten! I understand the need for a `snapshot_download` with a `local_files_only` parameter.
> 
> The only issue I can see here is that it does not play well at all with the `revision` parameter:
> 
> ```python
> snapshot_download("xxx")
> snapshot_download("xxx", revision="yyy")
> 
> snapshot_download("xxx", local_files_only=True)  # Will return the one with revision "yyy" even if the "xxx" should have been the one chosen as it was the latest commit
> ```
> 
> I can't think of a way to go around this limitation, which seems quite troublesome to me. It does seem like a very rare case to me so maybe putting a warning is enough, wdyt?
> 
> The following use-case doesn't work either:
> 
> ```python
> snapshot_download("xxx", revision="zzz")
> snapshot_download("xxx", revision="yyy")
> 
> snapshot_download("xxx", revision="zzz", local_files_only=True)  # Will return the one with revision "yyy"
> ```

Ah yeah this is indeed a big problem. The revision is however nowhere locally saved (it's only encoded in the saved sha no?) -> so there is no way that we can retrieve the correct revision when we don't have internet connection no?

patrickvonplaten:
> > Thanks for your contribution, @patrickvonplaten! I understand the need for a `snapshot_download` with a `local_files_only` parameter.
> > The only issue I can see here is that it does not play well at all with the `revision` parameter:
> > ```python
> > snapshot_download("xxx")
> > snapshot_download("xxx", revision="yyy")
> > 
> > snapshot_download("xxx", local_files_only=True)  # Will return the one with revision "yyy" even if the "xxx" should have been the one chosen as it was the latest commit
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I can't think of a way to go around this limitation, which seems quite troublesome to me. It does seem like a very rare case to me so maybe putting a warning is enough, wdyt?
> > The following use-case doesn't work either:
> > ```python
> > snapshot_download("xxx", revision="zzz")
> > snapshot_download("xxx", revision="yyy")
> > 
> > snapshot_download("xxx", revision="zzz", local_files_only=True)  # Will return the one with revision "yyy"
> > ```
> 
> Ah yeah this is indeed a big problem. The revision is however nowhere locally saved (it's only encoded in the saved sha no?) -> so there is no way that we can retrieve the correct revision when we don't have internet connection no?

We could force the user to have to provide the repo structure when doing `local_files_only` or else throw a warning that the last updated folder is used? What do you think @LysandreJik ?

patrickvonplaten:
@LysandreJik , @julien-c - I now went for the solution where we update the cache folder names to 
`<repo_id>.<revision>.<hash_of_folder>`

I think that being able to load repos from a specific revision locally is quite important and that we should enable this functionality. At the same time, we cannot assume that the filenames in a repo are always hardcoded since some repositories 
like https://github.com/kensho-technologies/pyctcdecode/blob/b50b5ae39ebb047c42765be0dce2c27137bd9474/pyctcdecode/language_model.py#L353 load non-hardcoded filenames. 

I don't see another way as saving the revision into the name of the cached folder which can then be correctly looked up when `local_files_only=True`. 
This change will require a new cached download for people using `snapshot_download` which I think is ok given that `snapshot_download` is not yet used that widely (IMO).

What do you think?

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +60,46 @@ def snapshot_download(
...
-        url = hf_hub_url(
-            repo_id, filename=model_file.rfilename, revision=model_info.sha
+        repo_id_sha = storage_folder.split(".")[-1]
+        model_files = os.listdir(storage_folder)
+    else:
+        # if we have internet connection we retrieve the correct folder name from the huggingface api
+        _api = HfApi()
+        model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
+        storage_folder = os.path.join(
+            cache_dir, repo_id_flattened + "." + revision + "." + model_info.sha
```
LysandreJik, line 94: If the `revision` and `model_info.sha` are identical (which they are if you explicitly specify the `revision` parameter to a commit hash), then this will put it twice in the name. Is this expected? i.e., doing:

```py
snapshot_download("xxx", revision="yyy")
# ~/.cache/xxx.yyy.yyy
```

LysandreJik: (not necessarily a problem)

tests/test_snapshot_download.py
```diff
@@ -155,3 +155,62 @@ def test_download_private_model(self):
         self._api.update_repo_visibility(
             token=self._token, name=REPO_NAME, private=False
         )
+
+    def test_download_model_local_only(self):
```
LysandreJik, line 159: Can you also add a test with no explicit `revision` passed?

LysandreJik (commented):
Thanks for iterating on the implementation. I think there's either a small issue with the implementation or that we should align on the behavior of this method.

My understanding is that `snapshot_download` should always fetch the latest update on the default branch of the repository (`main` right now, but should be updated later to reflect the default git branch in the future) when no `revision` is passed. If a `revision` is passed and it is a branch, will fetch the latest update on that branch. If it is a tag or a commit hash, then will fetch the exact version referenced.

With `local_files_only`, this becomes a bit tricky: if we handle `main` (or any other branch name), then the aforementioned behavior cannot be strictly respected. Let's see why:

```py
# Fetch repository on branch `main`
directory = snapshot_download("xxx")

# <Update of the repository>

directory = snapshot_download("xxx", local_files_only=True)
```
This will work correctly and return the correct `directory`.

```py
# Fetch repository on branch `main`
directory = snapshot_download("xxx")

# <Update of the repository with commit #6fr4de on branch main>

directory = snapshot_download("xxx", revision="6fr4de")

directory = snapshot_download("xxx", local_files_only=True)
```
This will now return the directory with the `main` identifier, even though the second call to `snapshot_download` called a more recent update of that same branch (`main`).

TBH I think this is mostly a nit than an actual issue. In most cases, users that indicate the revision parameter in a `snapshot_download` will also indicate it in the same `snapshot_download` when calling it with `local_files_only=True`.

I guess to be explicit here, I'd vote to log a warning when not specifying a specific revision:
```
WARNING: You called `snapshot_download` without specifying an explicit revision and with no networking. Will return the latest local download of the `main` branch.
```

---

Another, smallish issue I have is that the following doesn't work:

```py
snapshot_download("xxx", revision="main")
# Returns ~/.cache/xxx.main.<COMMIT_SHA>

snapshot_download("xxx", revision="main", local_files_only=True)
# Returns ~/.cache/xxx.main.<COMMIT_SHA>

snapshot_download("xxx", revision="<COMMIT_SHA>", local_files_only=True)
# ValueError: Cannot find the requested files in the cached path and outgoing traffic has been disabled. To enable model look-ups and downloads online, set 'local_files_only' to False
```

---

It seems that most OS (except Windows) have a maximum filename length of 255, and a maximum path of 4096 characters. With ~40 characters per hash and a maximum of two hashes per directory name (revision + file hash), it seems we should be good to go. Windows should work to since https://github.com/huggingface/huggingface_hub/pull/40

julien-c:
I might be missing something, but why don't we enforce that if `local_files_only=True` then `revision` must be a full commit hash? 

This way there's no potential indirection. If it's used mostly for offline tests, it will just work out of the box.

What am I missing here?

LysandreJik:
Hmmm if we do that then we lose quite a big portion of the API I think, no? We're mostly dealing with edge cases here, but I think the following is more important than the rest:

```py
snapshot_download("xxx")

# Should work
snapshot_download("xxx", local_files_only=True)
```

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +64,86 @@ def snapshot_download(
...
+        # 2) cached repos of format <repo_id>.{revision}
+        # -> in this case {revision} has to be a commit sha
+        repo_folders_commit_only = glob(repo_folders_prefix + "." + revision)
 
-    storage_folder = os.path.join(
-        cache_dir, repo_id.replace("/", REPO_ID_SEPARATOR) + "." + model_info.sha
-    )
+        # 3) cached repos of format <repo_id>.<any-branch>.{revision}
+        # -> in this case {revision} also has to be a commit sha
+        repo_folders_branch_commit = glob(repo_folders_prefix + ".*." + revision)
```
LysandreJik, line 88: And this isn't necessarily a `branch` + a `commit`, I'd say it's an `explicit_revision` + a `folder_sha` (even if the latter is ultimately a commit hash)

How about `repo_folders_explicit_revision_and_sha` ?

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +64,86 @@ def snapshot_download(
...
+                "mode might lead to unexpected behavior by not taking into account the latest "
+                "commits."
+            )
+
+        # find last modified folder
+        storage_folder = max(repo_folders, key=os.path.getmtime)
+
+        # get commit sha
+        repo_id_sha = storage_folder.split(".")[-1]
+        model_files = os.listdir(storage_folder)
```
LysandreJik, line 126: OK!

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +64,86 @@ def snapshot_download(
...
+        model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
+
+        storage_folder = os.path.join(cache_dir, repo_id_flattened + "." + revision)
+
+        # if passed revision is not identical to the commit sha
+        # then revision has to be a branch name, e.g. "main"
+        # in this case make sure that the branch name is included
+        # cached storage folder name
+        if revision != model_info.sha:
+            storage_folder += f".{model_info.sha}"
```
LysandreJik, line 139: I wonder if this is necessary or superfluous - does it hurt to keep things simple at the cost of having a duplicate of sha in the saved folder name?

tests/test_snapshot_download.py
```diff
@@ -155,3 +163,148 @@ def test_download_private_model(self):
...
+            )
+
+            snapshot_download(
+                f"{USER}/{REPO_NAME}",
+                revision=self.first_commit_hash,
+                cache_dir=tmpdirname,
+            )
+
+            # now load from cache and make sure warning to be raised
+            with self.assertWarns(Warning):
```
LysandreJik, line 270: Good

LysandreJik (approved):
This looks good to me, only left a few comments. Thanks for working on it and iterating, @patrickvonplaten!

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +64,86 @@ def snapshot_download(
...
+    # last modified folder in the cache
+    if local_files_only:
+        # possible repos have <path/to/cache_dir>/<flatten_repo_id> prefix
+        repo_folders_prefix = os.path.join(cache_dir, repo_id_flattened)
+
+        # list all possible folders that can correspond to the repo_id
+        # and are of the format <flattened-repo-id>.<revision>.<commit-sha>
+        # now let's list all cached repos that have to be included in the revision
+        # 1) cached repos of format <repo_id>.{revision}.<any-hash>
+        repo_folders_branch = glob(repo_folders_prefix + "." + revision + ".*")
```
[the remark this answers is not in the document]
patrickvonplaten: I think this has to be a branch actually because it's not possible that either:

`<repo_id>.{revision}.` (not the second `'.'` which disables things like `<repo_id>.<sha>` or `<repo_id>.<branch>.<sha>`

src/huggingface_hub/snapshot_download.py
```diff
@@ -53,18 +64,86 @@ def snapshot_download(
...
+                "Cannot find the requested files in the cached path and outgoing traffic has been"
+                " disabled. To enable model look-ups and downloads online, set 'local_files_only'"
+                " to False."
+            )
 
-    for model_file in model_info.siblings:
-        url = hf_hub_url(
-            repo_id, filename=model_file.rfilename, revision=model_info.sha
+        # check if repo id was previously cached from a commit sha revision
+        # and passed {revision} is not a commit sha
```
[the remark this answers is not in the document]
patrickvonplaten: This `repo_folders_commit_any = set(glob(repo_folders_prefix + ".*")) - set(glob(repo_folders_prefix + ".*.*")` is only `>` 0 if there has not been a sha revision snapshot and `len(repo_folders_commit_only + repo_folders_branch_commit) == 0` makes sure that the revision has to be a branch

src/huggingface_hub/snapshot_download.py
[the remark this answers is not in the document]
LysandreJik: Indeed, I understand

src/huggingface_hub/snapshot_download.py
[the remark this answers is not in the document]
LysandreJik: Thanks for explaining, I understand

src/huggingface_hub/snapshot_download.py
LysandreJik: After f2f explanation I understand why this is necessary :)

PR #507, opened 2021-11-29
Access tokens

LysandreJik:
Offers the access tokens as the main login handler.

When doing `huggingface-cli login`, a `Token: ` prompt will appear, preceded by the following:
```
        To login, `huggingface_hub` now requires a token generated from https://huggingface.co/settings/token.
        To login with username and password instead, interrupt with Ctrl+C.
```

If the user does Ctrl+C, it redirects automatically to a Username/Password prompt:

file

### DIFF
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -209,7 +209,7 @@ class LogoutCommand(BaseUserCommand):
         except HTTPError as e:
             if e.response.status_code == 400:
                 # Logging out with an access token will return a client error.
-                pass
+                HfApi.unset_access_token()
             else:
                 raise e
         print("Successfully logged out.")
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -164,11 +164,19 @@ class LoginCommand(BaseUserCommand):
         _|    _|  _|    _|  _|    _|  _|    _|    _|    _|    _|_|  _|    _|      _|        _|    _|  _|        _|
         _|    _|    _|_|      _|_|_|    _|_|_|  _|_|_|  _|      _|    _|_|_|      _|        _|    _|    _|_|_|  _|_|_|_|
 
+        To login, `huggingface_hub` now requires a token generated from https://huggingface.co/settings/token.
+        To login with username and password instead, interrupt with Ctrl+C.
         """
         )
-        username = input("Username: ")
-        password = getpass()
-        _login(self._api, username, password)
+
+        try:
+            token = getpass("Token: ")
+            _login(self._api, token=token)
+
+        except KeyboardInterrupt:
+            username = input("\rUsername: ")
+            password = getpass()
+            _login(self._api, username, password)
 
 
 class WhoamiCommand(BaseUserCommand):
@@ -196,7 +204,14 @@ class LogoutCommand(BaseUserCommand):
             print("Not logged in")
             exit()
         HfFolder.delete_token()
-        self._api.logout(token)
+        try:
+            self._api.logout(token)
+        except HTTPError as e:
+            if e.response.status_code == 400:
+                # Logging out with an access token will return a client error.
+                pass
+            else:
+                raise e
         print("Successfully logged out.")
 
 
@@ -398,6 +413,7 @@ def _login(hf_api, username=None, password=None, token=None):
     elif not hf_api._is_valid_token(token):
         raise ValueError("Invalid token passed.")
 
+    hf_api.set_access_token(token)
     HfFolder.save_token(token)
     print("Login successful")
     print("Your token has been saved to", HfFolder.path_token)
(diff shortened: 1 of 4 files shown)


tests/testing_constants.py
```diff
@@ -1,6 +1,7 @@
 USER = "__DUMMY_TRANSFORMERS_USER__"
 FULL_NAME = "Dummy User"
 PASS = "__DUMMY_TRANSFORMERS_PASS__"
+TOKEN = "hf_94wBhPGp6KrrTH3KDchhKpRxZwd6dmHWLL"
```
sgugger, line 4: Is this not sensitive at all?

sgugger (approved):
Thanks a lot for adding this!

tests/testing_constants.py
julien-c: not super critical, but probably cleaner to put it into a GitHub Action secret (wdyt @SBrandeis )

src/huggingface_hub/commands/user.py
```diff
@@ -164,11 +164,19 @@ def run(self):
...
-        _login(self._api, username, password)
+
+        try:
+            token = getpass("Token: ")
+            _login(self._api, token=token)
+
+        except KeyboardInterrupt:
+            username = input("\rUsername: ")
+            password = getpass()
+            _login(self._api, username, password)
```
julien-c, line 179: do we really need to still support this? what's the timeline to remove it?

src/huggingface_hub/hf_api.py
```diff
@@ -405,6 +405,14 @@ def logout(self, token: Optional[str] = None) -> None:
         r = requests.post(path, headers={"authorization": "Bearer {}".format(token)})
         r.raise_for_status()
 
+    @staticmethod
+    def set_access_token(access_token: str):
+        write_to_credential_store("username", access_token)
```
julien-c, line 410:
```suggestion
        write_to_credential_store("user", access_token)
```
I'd favor using user (i think we do this somewhere else)

julien-c: and maybe define a constant `USERNAME_PLACEHOLDER` to make intent clearer (and share it w/ the tests)

src/huggingface_hub/commands/user.py
julien-c: in my opinion we could already remove it

src/huggingface_hub/hf_api.py
SBrandeis, line 410: I feel like `hf_access_token` or the like would be more explicit than `username` for someone looking at their git credential manager from outside `huggingface_hub`, WDYT ?

tests/testing_constants.py
SBrandeis: Either that or use an ostensibly fake token? (like `hf_fakefakefake` or smth)

src/huggingface_hub/commands/user.py
```diff
@@ -196,7 +204,14 @@ def run(self):
             print("Not logged in")
             exit()
         HfFolder.delete_token()
-        self._api.logout(token)
+        try:
+            self._api.logout(token)
+        except HTTPError as e:
+            if e.response.status_code == 400:
+                # Logging out with an access token will return a client error.
+                HfApi.unset_access_token()
```
SBrandeis, line 212: I think we want to call this even if there is no exception since it's always set by `_login`

src/huggingface_hub/commands/user.py
SBrandeis, line 179: Nice 👍

SBrandeis (commented):
Thanks for implementing this @LysandreJik!

tests/testing_constants.py
LysandreJik: It's still used to ensure it works. But this doesn't seem sensitive at all to me, it's exactly like the user/pass before

julien-c: it's not sensitive but it _looks_ sensitive to an outside eye. Is there any drawback to passing it through GitHub secrets (or locally through an env variable)?

Otherwise, let's add a comment below it that says "this is not sensitive because it's only on a sandboxed CI instance" or something?

src/huggingface_hub/commands/user.py
sgugger: I think we need to offer the same API for a bit, but we could indicate clearly it's deprecated and will be removed in n months.

tests/testing_constants.py
LysandreJik: Passing it through GitHub secrets means that it will be a bit tough to run the CI locally; I'm happy to do that, but I'd prefer the comment to indicate it's not sensitive!

src/huggingface_hub/commands/user.py
LysandreJik: I'll leave it like this for this release (0.2.0) with a deprecation warning, but will remove as soon as it's released, if that works for you.

src/huggingface_hub/hf_api.py
LysandreJik: How about `hf_user` as a middle ground between @julien-c's proposal and yours?

julien-c: OK

I just don't want people to think that they need to replace it (it's just a dummy as it's not actually used)

src/huggingface_hub/commands/user.py
julien-c: yes ok it works for me

tests/testing_constants.py
julien-c: SG

PR #510, opened 2021-11-30
Deprecate `list_repos_objs` as probably no-one uses it

julien-c:
it was initially implemented just for backward compat to the old S3 system

I doubt anyone's using it, and it feels "unnatural" as an API now that every file is repo-centric


LysandreJik (approved):
Yes!!!

SBrandeis (approved):
👍

PR #511, opened 2021-11-30
Follow up to #277: tone down git-credential warning slightly

julien-c:
cc @SBrandeis

### DIFF
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -407,7 +407,7 @@ def _login(hf_api, username=None, password=None, token=None):
         print(
             ANSI.red(
                 "Authenticated through git-credential store but this isn't the helper defined on your machine.\nYou "
-                "will have to re-authenticate when pushing to the Hugging Face Hub. Run the following command in your "
-                "terminal to set it as the default\n\ngit config --global credential.helper store"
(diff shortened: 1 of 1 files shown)


LysandreJik:
Are you sure of the will > might?

julien-c:
Personally I'd even favor removing this warning completely given that when interacting with a repo through `Repository` (the library's main point of interaction with a repo) from what I understand it's going to set the user's local git config to use it anyways

LysandreJik:
Yeah but if using `git clone` manually you'll be asked to authenticate again

LysandreJik:
I'd be very in favor of keeping it, even if I understand the tone-down

julien-c:
and i know very few people (no-one outside of Google Colab?) who use `store` as their system wide git credential helper so this error message would be displayed for everyone IMO

julien-c:
Ok then let's keep the message but apply this PR? WDYT? or feel free to close, no big deal =)

LysandreJik (approved):
LGTM

LysandreJik:
I understand where you're coming from and the potential issues this can bring. I'll take a second look at having a cross-credential-helper solution and if I still fail to identify a good solution, then let's revisit the warning problem.

PR #517, opened 2021-12-02
Fix the progress bars when cloning a repository

LysandreJik:
The environment changes that had to happen for the `GIT_LFS_SKIP_SMUDGE` ended up modifying the progress bars for that specific command.

A test should be setup to ensure this doesn't happen.


osanseviero (approved):
Thanks for the fix!

As you mentioned a test to ensure this does not happen would be nice

LysandreJik:
Will merge this to have the progress bars on `main` and work on a test as soon as I have a bit of time.

PR #522, opened 2021-12-06
Token in env variables

FrancescoSaverioZuppichini:
## Why
Sometimes (tests, docker, ... etc) it's easier to pass the token required to use the hub in a env variable, something like:

```
export HUGGING_FACE_HUB_TOKEN=<YOUR_TOKEN>
```

This avoids the need of getting the token, create a folder, and store it inside

## How

I've edited the logic inside `hf_api.HFolder.get_token`. Pasted below for convenience

```python
@classmethod
    def get_token(cls) -> Optional[str]:
        """
        Get token or None if not existent. A token can be also provided using env variables

        >>> export HUGGING_FACE_HUB_TOKEN=<YOUR_TOKEN>
        """

        path_token: Path = Path(cls.path_token)
        token: Optional[str] = os.environ.get("HUGGING_FACE_HUB_TOKEN")

        if token is None:
            # fall back to disk
            if path_token.exists():
                with path_token.open("r") as f:
                    token = f.read()
        return token
```

## Test it
I've added a few new line in the tests but even if I've mocked the `os.environ` dict the token is still `None`, I can use a little help :pray: 

Thanks

Francesco

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -22,7 +22,7 @@ import warnings
 from io import BufferedIOBase, RawIOBase
 from os.path import expanduser
 from typing import IO, Dict, Iterable, List, Optional, Tuple, Union
-
+from pathlib import Path
 import requests
 from requests.exceptions import HTTPError
 
@@ -1127,15 +1127,22 @@ class HfFolder:
             f.write(token)
 
     @classmethod
-    def get_token(cls):
+    def get_token(cls) -> Optional[str]:
         """
-        Get token or None if not existent.
+        Get token or None if not existent. A token can be also provided using env variables
+
+        >>> export HUGGING_FACE_HUB_TOKEN=<YOUR_TOKEN>
         """
-        try:
-            with open(cls.path_token, "r") as f:
-                return f.read()
-        except FileNotFoundError:
-            pass
+
+        path_token: Path = Path(cls.path_token)
+        token: Optional[str] = os.environ.get("HUGGING_FACE_HUB_TOKEN")
+
+        if token is None:
+            # fall back to disk
+            if path_token.exists():
+                with path_token.open("r") as f:
+                    token = f.read()
+        return token
 
     @classmethod
     def delete_token(cls):
--- a/tests/test_hf_api.py
+++ b/tests/test_hf_api.py
@@ -21,7 +21,7 @@ import time
 import unittest
 import uuid
 from io import BytesIO
-
+from unittest import mock
 import pytest
 
 import requests
@@ -596,6 +596,10 @@ class HfFolderTest(unittest.TestCase):
         # ^^ not an error, we test that the
         # second call does not fail.
         self.assertEqual(HfFolder.get_token(), None)
+        with mock.patch.dict(os.environ, {"HUGGING_FACE_HUB_TOKEN": token}):
+            self.assertEqual(HfFolder.get_token(), token)
+        with mock.patch.dict(os.environ, {"HUGGING_FACE_HUB_TOKEN": None}):
+            self.assertEqual(HfFolder.get_token(), None)
 
 
 @require_git_lfs


muellerzr:
I wouldn't be surprised if we don't want users to do this, or at least have a way to revert it, for the same reasoning mentioned here:

https://github.com/huggingface/huggingface_hub/pull/547#issuecomment-996984015

What do you think @osanseviero

LysandreJik:
Hey @FrancescoSaverioZuppichini, sorry for getting back so late on this - the issue with this approach is that it won't configure git properly as it won't put the token in the git-credential store, so it's only a partial fix.

Could you let me know of an issue you encountered with docker, so that we may take a look together and see the best resolution?

FrancescoSaverioZuppichini:
Hi @LysandreJik , thank you for your reply. Apologies for the noob question, how is the token placed in the git-credential store?  

To use the hub in docker, you have to create the folder and paste the token inside it. Something like:

```
mkdir <HF_TOKEN_FOLDER>
echo $HF_TOKEN > <HF_TOKEN_FOLDER>/token
```

This has to be done at build time you also (probably) need to define the token variable with the `ARG` keyword.

In my opinion, it would be cool to let the code directly set everything up if it sees the token in the correct env variable. What do you think?

tests/test_hf_api.py
```diff
@@ -616,6 +616,10 @@ def test_token_workflow(self):
         # ^^ not an error, we test that the
         # second call does not fail.
         self.assertEqual(HfFolder.get_token(), None)
+        with mock.patch.dict(os.environ, {"HUGGING_FACE_HUB_TOKEN": token}):
+            self.assertEqual(HfFolder.get_token(), token)
+        with mock.patch.dict(os.environ, {"HUGGING_FACE_HUB_TOKEN": None}):
+            self.assertEqual(HfFolder.get_token(), None)
```
LysandreJik, line 622: Cool :)

src/huggingface_hub/hf_api.py
```diff
@@ -1139,15 +1139,22 @@ def save_token(cls, token):
...
-            pass
+
+        path_token: Path = Path(cls.path_token)
+        token: Optional[str] = os.environ.get("HUGGING_FACE_HUB_TOKEN")
+
+        if token is None:
+            # fall back to disk
+            if path_token.exists():
+                with path_token.open("r") as f:
+                    token = f.read()
```
LysandreJik, line 1156: Actually, the try/except block was more pythonic here. I don't think it's a huge deal, but the try/except is more reliable and more up-to-date than the check with the `if path_token.exists()` which can return information that will be different by the time it is consumed.

The approach you implemented is based off of the [look-before-you-leap](https://docs.python.org/2.7/glossary.html#term-lbyl) paradigm, but in Python we favor the [easier-to-ask-for-forgiveness-than-permission](https://docs.python.org/2.7/glossary.html#term-eafp) paradigm.

Would you mind reverting this change? Other than that, feel free to merge :)

LysandreJik (approved):
Sorry for taking so long to get back to this. I agree with you regarding the need, thanks for offering a PR! Only left a comment, ~but once that is taken care of, you can feel free to merge.~ I'm looking at how to enable that from git's perspective and will get back to you.

Thanks, @FrancescoSaverioZuppichini!

LysandreJik:
Actually, before merging, let me take a quick look at how we can handle that from git's side.

LysandreJik:
Your proposition of toggling it in the credential store could be done by calling the following method: [`write_to_credential_store`](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L343-L356).

I think the environment variable will not change during the runtime, so setting it at the beginning would be the best. The tooling right now is designed to read from the folder, so the simplest way to handle it right now would be to right the token to the folder, equivalent to a login; but that would kind of defeat the purpose of having it as an environment variable.

LysandreJik:
Thinking more about it, you're correct @adrinjalali. Having it as an environment variable without any impact on the git-credentials logging system sounds good to me.

Apart from the try/except statement, should be good to go. Thanks!

src/huggingface_hub/hf_api.py
FrancescoSaverioZuppichini: Reverted :)

FrancescoSaverioZuppichini:
Tried to rebase it, something didn't work. Gonna open a new PR

PR #530, opened 2021-12-10
[Hotfix][API] card_data => cardData on /api/datasets

julien-c:
api change: /api/datasets[...] now returns cardData not card_data (for consistency with models)

We didn't find any evidence for people using this API attribute yet so are hotfixing it on master


Pierrci (approved):

julien-c:
Note that the CI failure was not intentional and I will fix it as soon as I have some time!

Issue #513, opened 2021-11-30
List models by author from HfApi


FrancescoSaverioZuppichini:
**Is your feature request related to a problem? Please describe.**
I would like to search for a specific author using `HfApi`. I've tried to use the `filter` option as following but the list is empty

```python
from huggingface_hub import HfApi
from pprint import pprint

api = HfApi()
pprint(api.list_models(filter="glasses")) 
pprint(api.list_models(filter="organization:glasses"))
pprint(api.list_models(filter="author:glasses"))
```
[glasses](https://huggingface.co/glasses) is an organization with saved models on hugging face. 

**Describe the solution you'd like**
It appears is indeed possible to search by authors, 

![image](https://user-images.githubusercontent.com/15908060/144102573-30e54633-dc31-4acc-8c81-fcfcae06c6a3.png)

<del>I've tried to look online for the endpoint used by `Hfapi` (so` https://huggingface.co/api/models`) but I was not able to find them.</del>

[Edit], [here](https://huggingface.co/docs/hub/endpoints) they are

Something in the right direction can be archived by listing all the files inside an organization repo

```python
from huggingface_hub import HfApi, HfFolder
from pprint import pprint
import requests


api = HfApi()
headers = {"authorization": f"Bearer {HfFolder.get_token()}"}
params = {"organization": "glasses"}
res = requests.get(f"{api.endpoint}/api/repos/ls", headers=headers, params=params)

pprint(res.json())
```

[Edit] Also,  `.list_repos_objs` from `HfApi` can be used to list the files.

But probably there is a better way

Thanks!

Cheers,

Francesco

julien-c:
This is not document nor implemented in the `huggingface_hub.HfApi` wrapper yet (cc @LysandreJik @muellerzr @osanseviero), but you can do https://huggingface.co/api/models?author=glasses to get models by author 

you can also do https://huggingface.co/api/models?search=glasses to get model names that contain the string "glasses"

There exist equivalent endpoints for datasets and spaces.

`list_models(filter=...)` applies on tags, i.e. the repo labels you see in the header of the model page on hf.co

julien-c:
If you'd like to implement this in `huggingface_hub` feel free to give it a shot!

FrancescoSaverioZuppichini:
Hi @julien-c , thank you for the highlights! I'll work on that ;)

PR #531, opened 2021-12-11
Search by authors and string

FrancescoSaverioZuppichini:
## Why
Searching by `authors` and `search` (so just a string) is currently missing, linked to #513 

## How

I've edited the logic inside `hf_api.HfApi.list_models`. Pasted below for convenience

```python

  def list_models(....):
     ....
     if filter is not None:
          params.update({"filter": filter})
          params.update({"full": True})
      if author is not None: # <- here
          params.update({"author": author})
      if search is not None: # <- here
          params.update({"search": search})
      .....
```

[EDIT] After discussing it we also included the same logic in `list_datasets`
## Test it
I've added a few new lines in the tests under `HfApiPublicTest`, maybe a test with a complex query + `authors`, `search` is not a bad idea.

**Since I've edited my `main` branch on the `fork` this branch has the commits from the `main` branch that contains the code for 
#522. Unfortunately, I am not a git wizard so I've removed that code. I am not sure what's the best way to clean it up while having my other PR open (#522)**

Thanks

Francesco

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -522,6 +522,8 @@ class HfApi:
     def list_datasets(
         self,
         filter: Union[str, Iterable[str], None] = None,
+        author: Optional[str] = None,
+        search: Optional[str] = None,
         sort: Union[Literal["lastModified"], str, None] = None,
         direction: Optional[Literal[-1]] = None,
         limit: Optional[int] = None,
@@ -546,6 +548,30 @@ class HfApi:
 
                     >>> # List only the datasets in russian for language modeling
                     >>> api.list_datasets(filter=("languages:ru", "task_ids:language-modeling"))
+            author (:obj:`str`, `optional`):
+                A string which identify the author of the returned models
+                Example usage:
+
+                    >>> from huggingface_hub import HfApi
+                    >>> api = HfApi()
+
+                    >>> # List all datasets from google
+                    >>> api.list_datasets(author="google")
+
+                    >>> # List only the text classification datasets from google
+                    >>> api.list_datasets(filter="text-classification", author="google")
+            search (:obj:`str`, `optional`):
+                A string that will be contained in the returned models
+                Example usage:
+
+                    >>> from huggingface_hub import HfApi
+                    >>> api = HfApi()
+
+                    >>> # List all datasets with "text" in their name
+                    >>> api.list_datasets(search="text")
+
+                    >>> #List all datasets with "text" in their name made by google
+                    >>> api.list_datasets(search="text", author="google")
             sort (:obj:`Literal["lastModified"]` or :obj:`str`, `optional`):
                 The key with which to sort the resulting datasets. Possible values are the properties of the `DatasetInfo`
                 class.
@@ -562,6 +588,10 @@ class HfApi:
         params = {}
         if filter is not None:
             params.update({"filter": filter})
+        if author is not None:
+            params.update({"author": author})
+        if search is not None:
+            params.update({"search": search})
         if sort is not None:
             params.update({"sort": sort})
         if direction is not None:
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -420,6 +420,8 @@ class HfApi:
     def list_models(
         self,
         filter: Union[str, Iterable[str], None] = None,
+        author: Optional[str] = None,
+        search: Optional[str] = None,
         sort: Union[Literal["lastModified"], str, None] = None,
         direction: Optional[Literal[-1]] = None,
         limit: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -451,6 +453,30 @@ def list_models(
 
                     >>> # List only the models from the AllenNLP library
                     >>> api.list_models(filter="allennlp")
+            author (:obj:`str`, `optional`):
+                A string which identify the author of the returned models
```
julien-c, line 457:
```suggestion
                A string which identify the author (user or organization) of the returned models
```

julien-c:
The `list_models`-related changeset of this PR looks good to me, and you could also apply it to `list_datasets` which has the same API

src/huggingface_hub/hf_api.py
```diff
@@ -21,7 +21,7 @@
 from io import BufferedIOBase, RawIOBase
 from os.path import expanduser
 from typing import IO, Dict, Iterable, List, Optional, Tuple, Union
-
+from pathlib import Path
```
muellerzr, line 24:
```suggestion
```

src/huggingface_hub/hf_api.py
```diff
@@ -1159,6 +1189,16 @@ def delete_token(cls):
...
+    @classmethod
+    def delete_token(cls):
+        """
+        Delete token. Do not fail if token does not exist.
+        """
+        try:
+            os.remove(cls.path_token)
+        except FileNotFoundError:
+            pass
+
```
muellerzr, line 1201:
```suggestion
```

tests/test_hf_api.py
```diff
@@ -21,7 +21,7 @@
 import unittest
 import uuid
 from io import BytesIO
-
+from unittest import mock
```
muellerzr, line 24:
```suggestion
```

tests/test_hf_api.py
```diff
@@ -615,7 +629,6 @@ def test_token_workflow(self):
         HfFolder.delete_token()
         # ^^ not an error, we test that the
         # second call does not fail.
```
muellerzr, line 631:
```suggestion
        # second call does not fail.
        self.assertEqual(HfFolder.get_token(), None)
```

src/huggingface_hub/hf_api.py
```diff
@@ -420,6 +420,8 @@ def unset_access_token():
     def list_models(
         self,
         filter: Union[str, Iterable[str], None] = None,
+        author: Optional[str] = None,
```
muellerzr, line 423: I think this might be better named as `author_or_organization` instead of just author.

muellerzr (changes requested):
Thank you for this! 

I saw your note about some git issues, try accepting my suggestions in the review and that should handle it just fine. 

I also have a few suggestions in there

src/huggingface_hub/hf_api.py
julien-c: just chiming in to say that on the hf.co hub side (and in APIs and the doc probably) we use `author` to mean `user_or_organization`

My 2 cents would be to keep `author` as a umbrella for both

muellerzr: Sounds good with me then.

FrancescoSaverioZuppichini:
> The `list_models`-related changeset of this PR looks good to me, and you could also apply it to `list_datasets` which has the same API

Changed the signature and the logic of `list_datasets` to match `list_models` :) Tests are green on my local machine

FrancescoSaverioZuppichini:
> Thank you for this!
> 
> I saw your note about some git issues, try accepting my suggestions in the review and that should handle it just fine.
> 
> I also have a few suggestions in there

Hi @muellerzr , I've included your commits, thank you for helping me out

FrancescoSaverioZuppichini (commented):
I've included the changes

julien-c:
LGTM but i'll let @muellerzr and @LysandreJik chime in!

muellerzr:
@FrancescoSaverioZuppichini looks like you're not passing the code quality tests, can you do the following please? :)

```bash
pip install -U black isort flake8
```
And then from the root of the repository:
```bash
make style; make quality
```
That should make the quality tests pass

tests/test_hf_api.py
```diff
@@ -472,6 +471,20 @@ def test_list_models(self):
         self.assertGreater(len(models), 100)
         self.assertIsInstance(models[0], ModelInfo)
 
+    @with_production_testing
+    def test_list_models_author(self):
+        _api = HfApi()
+        models = _api.list_models(author="google")
```
muellerzr, line 477: These tests aren't actually ensuring that the query is returning items that have `author="google"`, just that something was returned. I'd recommend changing each of these to look something like the following:

```python
# After your previous two
self.assertTrue("google" in models[0].author)
```
(And then same for where you would expect bert to be)

muellerzr (changes requested):

tests/test_hf_api.py
FrancescoSaverioZuppichini: done!

LysandreJik (approved):
Looks good to me, thanks @FrancescoSaverioZuppichini!

muellerzr (approved):

PR #484, opened 2021-11-19
Render newlines in code outputs

mishig25:
When there is a newline, render output of TextGenerationWidget as code output (pre with new line rendering)

[Try here](https://6197ca7027904a0a2fd3ff80--huggingface-widgets.netlify.app/) the `huggingface-course/codeparrot-ds` checkpoint

<video src="https://user-images.githubusercontent.com/11827707/142653609-b7edca42-70ee-4128-9d97-41d6ae6486fe.mov" controls autoplay loop/>

cc: @lvwerra

PR #539, opened 2021-12-15
Revert "Render newlines in code outputs"

mishig25:
Reverts huggingface/huggingface_hub#484

#484 was doing:
if output contains `\n` newline character, then render like code output
otherwise, render as a normal text output


Pierrci:
Is it handled somewhere else now, or are we removing this special case completely?

mishig25:
I think @gary149 wants to remove it for now.

Basically, the problem is that:
currently, if the output string contains new line character (i.e. `\n`), it is being rendered as a code output with <pre> htmlelement. However, Victor found this approach produces a lot of false positives (i.e. rendering as code outputs that should be rendered as normal widget output text).

So I think we are removing this special case completely for now

and come up with something later

gary149 (approved):

PR #545, opened 2021-12-16
[snapshot_download] upgrade to canonical separator

julien-c:
ping @severo

### DIFF
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -9,8 +9,8 @@ from .hf_api import HfApi, HfFolder
 from .utils import logging
 
 
-REPO_ID_SEPARATOR = "__"
-# ^ make sure this substring is not allowed in repo_ids on hf.co
+REPO_ID_SEPARATOR = "--"
+# ^ this substring is not allowed in repo_ids on hf.co
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/snapshot_download.py
```diff
@@ -9,8 +9,9 @@
 from .utils import logging
 
 
-REPO_ID_SEPARATOR = "__"
-# ^ make sure this substring is not allowed in repo_ids on hf.co
+REPO_ID_SEPARATOR = "--"
+# ^ this substring is not allowed in repo_ids on hf.co
+# and is the canonical one we use for serialization of repo ids elsewhere.
```
julien-c, line 14: (means that you could get the repo id from its snapshot path on disk, if ever needs be)

patrickvonplaten (approved):
Good for me! Think it will however force all repos to be downloaded & cached again. But should not be a big deal

LysandreJik (approved):
Ok LGTM!

julien-c:
> Think it will however force all repos to be downloaded & cached again

yes but we already did that recently no? :)

but yeah i feel like now we should be at the final format so shouldn't change after that

julien-c:
pinging @severo for a final review then let's merge!

severo (approved):
OK!

I put the moonlanding PR for reference: https://github.com/huggingface/moon-landing/pull/1657

PR #547, opened 2021-12-17
Add test directions

muellerzr:
Adds a small markdown file that contains directions for running the test suite, since it took me a moment to get everything setup on my side.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,14 @@
+# Running Tests
+
+To run the test suite, please perform the following from the root directory of this repository:
+
+1. `pip install -e .[testing]`
+
+      This will install all the testing requirements.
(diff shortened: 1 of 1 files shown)


osanseviero:
If you do `export HUGGINGFACE_CO_STAGING=1` that will impact you later unless you revert it. 

One solution is to do
`HUGGINGFACE_CO_STAGING pytest -sv ./tests/` 
and then it should work.

osanseviero:
I'm ok with going with this proposal in `/tests/README/md`, but we should consider creating a `CONTRIBUTING.md` for this.

muellerzr:
Should be good to go now @osanseviero :)

osanseviero (approved):
Thanks for this!

Issue #535, opened 2021-12-14
New Filtering API for `HfApi`


muellerzr:
I'll be working on the implementation of this. This feature will introduce a `Filter` class which allows for a nice user-centric searching API for the Hub + a `search_by_filter` method of some degree.

A rough example of what this could look like is the following:
```python
api = HfApi()
newFilter = Filter(author_or_organization='facebook')
api.search_models_by_filter(filter=newFilter)
```

Aimed search parameters:
- Author or Organization (`author_or_organization`)
- Framework, such as TF or Torch (`framework`)
- Trained dataset (`trained_dataset`)
- Language (`language`)
- Model name, both partial and exact (`model_name`)
- Task (`task`)
- CO2 (`co2`)

PR #553, opened 2021-12-20
Implement a Model Filter class

muellerzr:
This PR introduces a new way to search the hub: the `ModelFilter` class, discussed in https://github.com/huggingface/huggingface_hub/issues/535

It is a simple Enum at first to the user, allowing them to specify what they want to search for, such as:
```python
f = ModelFilter(author="microsoft", model_name="wavlm-base-sd", framework="pytorch")
```
From there, they can pass in this filter to the new `list_models_by_filter` function in `HfApi` to search through it:
```python
models = api.list_modes(filter=f)
```
Of the proposed tasks, here is what can be currently completed in the current state of the backend API:
- [x] Author or Organization (`author_or_organization`)
- [x] Framework, such as TF or Torch (`framework`)
- [x] Trained dataset (`trained_dataset`)
- [x] Language (`language`)
- [x] Model name, both partial and exact (`model_name`)
- [x] Task (`task`)

<s>After this, the same needs to be done for the Dataset searching as well.</s>


muellerzr:
I also renamed `utils.tags` to `utils.api_endpoint_helpers` as that's more inline with what that file contains

src/huggingface_hub/hf_api.py
```diff
@@ -506,6 +506,85 @@ def list_models(
         d = r.json()
         return [ModelInfo(**x) for x in d]
 
+    def list_models_from_filter(self, model_filter: ModelFilter):
```
LysandreJik, line 509: As you correctly guessed, I'm not a huge fan of offering another API to do the same thing - I'd rather this be merged in `list_models` with the support for the `ModelFilter` class.

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -14,6 +14,82 @@
...
+    the REST API. For all parameters capitalization does not matter.
+
+    Args:
+        author_or_organization (:obj:`str` or :class:`List`, `optional`):
+            A string or list of strings that can be used to identify models on the hub
+            by the original uploader, such as `facebook` or `huggingface`
+            Example usage:
+
+                >>> from huggingface_hub import Filter
+                >>> new_filter = Filter(author_or_organization="facebook")
```
LysandreJik, line 33: These examples should be updated to `ModelFilter`

src/huggingface_hub/hf_api.py
```diff
@@ -506,6 +506,85 @@ def list_models(
...
+            model_filter (:obj: `ModelFilter`):
+                An Enum class which handles storing of query information
+                Example usage:
+
+                    >>> from huggingface_hub import HfApi, ModelFilter
+                    >>> api = HfApi()
+
+                    >>> # List only the text classification models
+                    >>> f = ModelFilter(task="text-classification")
+                    >>> api.list_models_from_filter(filter=f)
```
LysandreJik, line 522: I think the `ModelFilter` class has good potential to be helpful, but right now it doesn't feel very different from the `list_models` method.

Doing the following doesn't seem necessarily simpler
```
>>> f = ModelFilter(task="text-classification")
>>> api.list_models_from_filter(filter=f)
```

than doing this
```
>>> api.list_models(filter="text-classification")
```

LysandreJik: I think it has very good potential to be super useful by leveraging your PR here https://github.com/huggingface/huggingface_hub/pull/537, i.e., offering auto-completion, or am I missing something?

LysandreJik: What if you'd like to filter with several tasks, how would you handle that? I think offering multiple selectors (language, task) per `ModelFilter` makes sense, but then I'd also expect the ability to offer multiple selectors from the same category

LysandreJik (commented):
Thanks for working on it! I've left a few comments so that we may work on the API together :)

src/huggingface_hub/hf_api.py
muellerzr: Threw that into https://github.com/huggingface/huggingface_hub/pull/556

IMO: This PR stays a bit dormant until all of the other necessary bits are needed to *only* add in a new filter API function, and this would be for complex queries ideally.

muellerzr:
@LysandreJik here are my initial thoughts:

The API here should be designed for *complex* queries, when it would just be too tiresome as a user to throw it into just `list_models`. For example:

```python
args = ModelSearchArguments()
f = ModelFilter(framework=[args.library.pytorch, args.library.TensorFlow], model_name="bert", tasks=[args.pipeline_tag.Summarization, args.pipeline_tag.TokenClassification])

api.list_models_from_filter(f)
```

With the role of the `_from_filter` function being turn the filter into something we can breakdown and pass into `list_models` itself, giving us a layered API of sorts.

osanseviero:
@muellerzr, I think it's confusing since `list_models` already has a filter parameter. We already support both iterables and strings...maybe we could also support `ModelFilter`s as input for this parameter. I think having a second method that does the same thing just for convenience to handle a new type won't scale/maintain nicely

muellerzr:
That was another thought I had I didn't like. Will think on this more over and tomorrow come up with a proposal for what that'd look like.

muellerzr:
@LysandreJik @osanseviero I think I've come to a solution we can all agree upon:

`list_models` now takes in a `ModelFilter`, which then unpacks it. I changed the logic for `if filter is not None` to be the following:
```python
        if filter is not None:
            if isinstance(filter, ModelFilter):
                params = self._unpack_filter(filter)
            else:
                params.update({"filter": filter})
            params.update({"full": True})
```
This keeps the old methods consistent, and still keeps `full:True` (which wasn't something I included in `ModelFilter`.

I'll go ahead and update this for the Datasets search as well, as this does seem to be the better way :)

muellerzr:
Okay! We're good for review!

Now I will say: adding all the related tests took us from 9.28s -> 17.78s to complete all of them for `HfApiPublicTest` (where they live)

So, my question for @LysandreJik (and probably @julien-c):

How do you think we should prioritize testing in this case? Since we're trying to make sure every edge case of the filter gets called, do we accept the time it takes? (It'd be more stable that way)

julien-c:
i don't think +10 sec is a big deal in tests (it's more on the Repository tests that we want to shave off some whole minutes IIRC)

LysandreJik:
Cool, this works well! Some feedback:
- For some reason the autocompletion doesn't work for me anymore with the updated code, whereas it worked with the previous version (`ModelTags`, I believe)
- Tests are cool. Could you also add some tests that test the `list_models` or `list_datasets` alongside a custom `ModelSearchArguments`?
- The documentation doesn't show how to use the two hand in hand either, but I think that's where most of the value is

LysandreJik (approved):
This looks good to me! @osanseviero could you give it a second look? @FrancescoSaverioZuppichini also, since you were involved in the `author` PR and you might have some feedback :)

FrancescoSaverioZuppichini:
> This looks good to me! @osanseviero could you give it a second look? @FrancescoSaverioZuppichini also, since you were involved in the `author` PR and you might have some feedback :)

All good from my side :)

src/huggingface_hub/hf_api.py
```diff
@@ -529,17 +533,20 @@ def list_models(
...
-                    >>> # List only the russian models compatible with pytorch
-                    >>> api.list_models(filter=("ru", "pytorch"))
-
-                    >>> # List only the models trained on the "common_voice" dataset
-                    >>> api.list_models(filter="dataset:common_voice")
+                    >>> # Using ModelFilter and SearchArguments to find text classification in both PyTorch and TensorFlow
+                    >>> filt = ModelFilter(task=args.pipeline_tags.TextClassification, framework=[args.library.PyTorch, args.library.TensorFlow])
 
-                    >>> # List only the models from the AllenNLP library
-                    >>> api.list_models(filter="allennlp")
```
osanseviero, line 542: Could we keep one example with a filter for a 3rd party library to have as a pointer?

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+    Args:
+        author (:obj:`str`, `optional`):
+            A string that can be used to identify models on the hub
+            by the original uploader (author or organization), such as `facebook` or `huggingface`
+            Example usage:
+
+                >>> from huggingface_hub import Filter
+                >>> new_filter = ModelFilter(author_or_organization="facebook")
+
+         framework (:obj:`str` or :class:`List`, `optional`):
```
osanseviero, line 122: should this be library instead? That way we are consistent with `ModelSearchArguments` and the filters at left in https://huggingface.co/models

WDYT?

src/huggingface_hub/hf_api.py
```diff
@@ -606,9 +618,66 @@ def list_models(
...
+                if "dataset:" not in dataset:
+                    dataset = f"dataset:{dataset}"
+                filter_tuple.append(dataset)
+
+        # Handling framework
+        if model_filter.framework:
+            if not isinstance(model_filter.framework, (list, tuple)):
+                model_filter.framework = [model_filter.framework]
+            for framework in model_filter.framework:
+                filter_tuple.append(framework)
```
osanseviero, line 659: Could this be `filter_tuple.extend(framework)` instead?

src/huggingface_hub/hf_api.py
```diff
@@ -606,9 +618,66 @@ def list_models(
...
+
+        # Handling tags
+        if model_filter.tags:
+            if not isinstance(model_filter.tags, (list, tuple)):
+                tags.append(model_filter.tags)
+            else:
+                tags += model_filter.tags
```
osanseviero, line 666: I think all of this can be simplified to `tags.extend` as well and then there's no need to do the type checking

osanseviero: And also https://github.com/huggingface/huggingface_hub/pull/553/files#diff-71a3e23532e1355e79edd51b351bd514f0aa55b946e4c530b0c95f647773ac5dR640-R643

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+@dataclass
+class DatasetFilter:
+    """A class that converts human-readable dataset search parameters into ones compatible with
+    the REST API. For all parameters capitalization does not matter.
+
+    Args:
+        author (:obj:`str`, `optional`):
+            A string or list of strings that can be used to identify datasets on the hub
```
osanseviero, line 28: nit here and below

```suggestion
            A string or list of strings that can be used to identify datasets on the Hub
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+        language_creators (:obj:`str` or :class:`List`, `optional`):
+            A string or list of strings that can be used to identify datasets on the hub
+            with how the data was curated, such as `crowdsourced` or `machine_generated`
+            Example usage:
+
+                >>> from huggingface_hub import DatasetFilter
+                >>> new_filter = DatasetFilter(language_creator="crowdsourced")
+
+        languages (:obj:`str` or :class:`List`, `optional`):
+            A string representing a two-character language to filter datasets by on the hub
```
osanseviero, line 59: Should this say "or list of strings" for consistency with other docstrings?

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+    languages: Union[str, List[str]] = None
+    multilinguality: Union[str, List[str]] = None
+    size_categories: Union[str, List[str]] = None
+    task_categories: Union[str, List[str]] = None
+    task_ids: Union[str, List[str]] = None
+
+
+@dataclass
+class ModelFilter:
+    """A class that converts human-readable model search parameters into ones compatible with
```
osanseviero, line 110: Should this link to hf.co/models and mention the filters in the left?

osanseviero (approved):
Thanks so much for this PR :rocket: 

I left a couple of nits, readability improvements and questions. Thanks once more! This is great

src/huggingface_hub/utils/endpoint_helpers.py
muellerzr: Makes sense with me, will change. Nice catch

src/huggingface_hub/hf_api.py
muellerzr: Some of it can be, but in some cases we check to make sure the string tag is there (such as `dataset:`), so in some cases it should stay

muellerzr: Since a string is an array of char, it's not quite as simple as that. I did refactor it to just be a type check and extend either the value or convert to a list, but simply using `.extend` alone doesn't work.

src/huggingface_hub/hf_api.py
muellerzr: See note here https://github.com/huggingface/huggingface_hub/pull/553#discussion_r779558602

src/huggingface_hub/utils/endpoint_helpers.py
muellerzr: It should not in this case, as the filters available here are actually different than those that are provided in hf.co/models and hf.co/datasets. Considerably more with datasets in particular.

LysandreJik (approved):
Looks good to me, thanks for iterating @muellerzr!

PR #556, opened 2021-12-21
Namespace Objects for Search Parameters

muellerzr:
# PR that adds the following:
## Namespace Objects for Search Parameters
With a goal of adding more tab-completion to the library, this PR introduces two objects:
- `DatasetSearchArguments`
- `ModelSearchArguments`
These two `AttributeDictionary` objects contain all the valid information we can extract from a model as tab-complete parameters. We also include the `author_or_organization` and `dataset` (or `model`) `_name` as well through careful string splitting. 

## Tests Added:
Considering there is nothing "new" done in `DatasetSearchArguments` that isn't already tested in `DatasetTags` or `AttributeDictionary.search` I chose not to add any tests here (especially because each call to the `SearchArguments()` classes calls the api twice)

If tests are wanted, I can include them.


muellerzr:
Realized I never quite said how they're related. It's a quality of life need for our users. For example, say I want to use the `ModelSearchArguments`, and I'd like to know if there are models with `resnet` I can use. I do:

```python
args = ModelSearchArguments()
args.model_name.search("resnet")
```
To return all of the models that contain the name resnet

LysandreJik (approved):
Cool, I like this a lot. Thanks for iterating with me on this, I think this should provide some very useful tooling for the `model_list`/`dataset_list`, and should provide a good platform on which we can build some useful methods in `transformers` and others.

LysandreJik:
As said offline; the `search` argument was a bit overkill for a first version, so thanks for removing it for now - I think these namespaces are quite simple and can be further iterated once we identify community usage/requests.

PR #560, opened 2021-12-22
[Fairseq] Add fairseq pipeline

patrickvonplaten:
This adds a pipeline for fairseq


api-inference-community/docker_images/fairseq/requirements.txt
```diff
@@ -0,0 +1,3 @@
+api-inference-community==0.0.19
+g2p_en
+git+git://github.com/pytorch/fairseq.git
```
osanseviero, line 3: This might be a bit problematic. Could we add a specific commit here so we know what is deployed? If not, it gets hard to know what was in the deployed version which makes debugging harder

cc @Narsil

osanseviero (commented):
Overall looks great! Thanks for the PR. Two small requests

* Could you also add a test under `test_dockers` please? (https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/tests/test_dockers.py). It's just a couple of lines but will allow us to run tests in an end-to-end way.
* Could you create a copy of this file but for fairseq? https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-api-adapter-transformers.yaml This will run a GitHub Actions workflow

api-inference-community/docker_images/fairseq/requirements.txt
patrickvonplaten: This should only stay there for 1,2 weeks until fairseq has a new version.

patrickvonplaten:
Closing in favor of https://github.com/huggingface/huggingface_hub/pull/561

PR #561, opened 2021-12-23
[Fairseq] Add fairseq TTS model to hub

Narsil:


api-inference-community/docker_images/fairseq/app/pipelines/text_to_speech.py
```diff
@@ -11,7 +12,9 @@ class TextToSpeechPipeline(Pipeline):
     def __init__(self, model_id: str):
         # hardcoded stuff can later be moved to a config
         model, cfg, task = load_model_ensemble_and_task_from_hf_hub(
-            model_id, arg_overrides={"vocoder": "griffin_lim", "fp16": False}
+            model_id,
+            arg_overrides={"vocoder": "griffin_lim", "fp16": False},
+            cache_dir=os.getenv("HUGGINGFACE_HUB_CACHE"),
```
Narsil, line 17: Use `HUGGINGFACE_HUB` to make docker it the cache (See the dockerfile this is defined as `/data` and this is mounted as an external disk on the kube cluster to make the model files persistent.

api-inference-community/docker_images/fairseq/app/pipelines/text_to_speech.py
```diff
@@ -39,6 +42,8 @@ def __call__(self, inputs: str) -> Tuple[np.array, int]:
         """
         tokenized_inputs = self._tokenize(inputs)
 
+        if not tokenized_inputs:
+            return np.zeros((0,)), self.sampling_rate
```
Narsil, line 46: If the inputs is `''` then there's a crash later. We can simply return nil in that case

patrickvonplaten (commented):
Thanks a lot!

api-inference-community/docker_images/fairseq/requirements.txt
```diff
@@ -1,3 +1,4 @@
-api-inference-community==0.0.19
-g2p_en
-git+git://github.com/pytorch/fairseq.git
+api-inference-community==0.0.20
+g2p_en==2.1.0
+librosa==0.8.1
```
Narsil, line 3: Pinning all version + adding librosa which wasn't there.
+ use latest `api-inference-community`.

api-inference-community/tests/test_dockers.py
```diff
@@ -79,6 +79,12 @@ def test_espnet(self):
             "espnet/kamo-naoyuki_mini_an4_asr_train_raw_bpe_valid.acc.best",
         )
 
+    def test_fairseq(self):
```
Narsil, line 82: Adding a docker test (Checking that launching the docker works and most use cases show correct errors + cache is working.

api-inference-community/docker_images/fairseq/app/pipelines/text_to_speech.py
patrickvonplaten: Thanks for the explanation

PR #563, opened 2021-12-23
Attempt to improve the docs for users wanting to add a new framework.

Narsil:
Reviewers, do you think those changes would help you implement what's missing ?

I try to make all the docker stuff a bit separate from the minimal work (which worked great IMO)
It's just extra steps to go a step beyond in getting things approved faster.


patrickvonplaten (approved):
That's very nice!

PR #564, opened 2021-12-23
Add flag to use basic auth

SBrandeis:
This PR adds a ~`use_basic` flag to use Basic authentication instead of a Bearer token~ a `override_auth_header` parameter to pass custom authorization header when talking to the hub

cc @lhoestq (wdyt of implementing this in datasets as well ?)

### DIFF
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -55,6 +55,11 @@ class UserCommands(BaseHuggingfaceCLICommand):
         ls_parser.add_argument(
             "--organization", type=str, help="Optional: organization namespace."
         )
+        ls_parser.add_argument(
+            "--use-basic",
+            action="store_true",
+            help="Optional: use Basic authentication",
+        )
         ls_parser.set_defaults(func=lambda args: ListReposObjsCommand(args))
         repo_create_parser = repo_subparsers.add_parser(
             "create", help="Create a new repo on huggingface.co"
@@ -83,6 +88,11 @@ class UserCommands(BaseHuggingfaceCLICommand):
             "--yes",
             action="store_true",
             help="Optional: answer Yes to the prompt",
+        )
+        repo_create_parser.add_argument(
+            "--use-basic",
+            action="store_true",
+            help="Optional: use Basic authentication",
         )
         repo_create_parser.set_defaults(func=lambda args: RepoCreateCommand(args))
 
@@ -221,7 +231,11 @@ class ListReposObjsCommand(BaseUserCommand):
             print("Not logged in")
             exit(1)
         try:
-            objs = self._api.list_repos_objs(token, organization=self.args.organization)
+            objs = self._api.list_repos_objs(
+                token,
+                organization=self.args.organization,
+                use_basic=self.args.use_basic,
+            )
         except HTTPError as e:
             print(e)
             print(ANSI.red(e.response.text))
@@ -288,6 +302,7 @@ class RepoCreateCommand(BaseUserCommand):
                 organization=self.args.organization,
                 repo_type=self.args.type,
                 space_sdk=self.args.space_sdk,
+                use_basic=self.args.use_basic,
             )
         except HTTPError as e:
             print(e)
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -55,11 +55,6 @@ class UserCommands(BaseHuggingfaceCLICommand):
         ls_parser.add_argument(
             "--organization", type=str, help="Optional: organization namespace."
         )
-        ls_parser.add_argument(
-            "--use-basic",
-            action="store_true",
-            help="Optional: use Basic authentication",
-        )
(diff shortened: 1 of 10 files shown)


src/huggingface_hub/commands/user.py
```diff
@@ -55,6 +55,11 @@ def register_subcommand(parser: ArgumentParser):
         ls_parser.add_argument(
             "--organization", type=str, help="Optional: organization namespace."
         )
+        ls_parser.add_argument(
+            "--use-basic",
+            action="store_true",
+            help="Optional: use Basic authentication",
+        )
```
julien-c, line 62: as discussed offline i don't think we need to expose this through the CLI (at least for now)

src/huggingface_hub/commands/user.py
```diff
@@ -84,6 +89,11 @@ def register_subcommand(parser: ArgumentParser):
             action="store_true",
             help="Optional: answer Yes to the prompt",
         )
+        repo_create_parser.add_argument(
+            "--use-basic",
+            action="store_true",
+            help="Optional: use Basic authentication",
+        )
```
julien-c, line 96: same

src/huggingface_hub/file_download.py
```diff
@@ -328,14 +330,18 @@ def cached_download(
         )
     }
     if isinstance(use_auth_token, str):
-        headers["authorization"] = f"Bearer {use_auth_token}"
+        auth = cast(
+            AuthHeaders, auth_headers(token=use_auth_token, use_basic=use_basic)
+        )
+        headers["authorization"] = auth["authorization"]
```
julien-c, line 336: do we really need the cast here? (you've been spoiled by Typescript type inference no? 🤣 )

src/huggingface_hub/utils/auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+
+class AuthHeaders(TypedDict):
+    authorization: str
```
julien-c, line 13: neat :)

julien-c (commented):
as discussed offline i'd suggest it'd be simpler/less confusing if we just have a `auth_header` optional override parameter (and not expose it through CLI or anything)

src/huggingface_hub/file_download.py
SBrandeis: Yes we need the cast, as `auth_header` return type is `Optional` and here it can not be None

src/huggingface_hub/file_download.py
```diff
@@ -328,14 +330,21 @@ def cached_download(
         )
     }
     if isinstance(use_auth_token, str):
-        headers["authorization"] = f"Bearer {use_auth_token}"
+        auth = cast(
+            AuthHeaders,
+            auth_header(token=use_auth_token, override=override_auth_header),
+        )
+        headers["authorization"] = auth["authorization"]
```
lhoestq, line 337: Here `auth` can end up being `None` if `auth_header` returns None right ?

In this case `auth["authorization"]` can fail.

It would also be more pythonic to have something like this:
```suggestion
        auth = auth_header(token=use_auth_token, override=override_auth_header)
        if auth:
            headers["authorization"] = auth["authorization"]
```

src/huggingface_hub/utils/auth.py
```diff
@@ -0,0 +1,31 @@
...
```
lhoestq, line 13: Nice ! Why not use a dataclass here ?

In your code you assume that the `authorization` field always exist in an instance of `AuthHeaders`, but it's ont the case for an empty `AuthHeaders()` for example. So if you want to fix that, I guess you can either use `auth.get("authorization")` instead of `auth["authorization"]`, or switch to a dataclass (as you want)

src/huggingface_hub/utils/auth.py
```diff
@@ -0,0 +1,31 @@
...
+
+def auth_header(
+    token: Optional[str] = None, override: Optional[str] = None
+) -> Optional[AuthHeaders]:
+    if override is not None:
+        return {"authorization": override}
+    return {"authorization": f"Bearer {token}"} if token is not None else None
+
+
+def basic_auth_header(username: str, password: str) -> str:
```
lhoestq, line 24: Cool !

lhoestq (commented):
Cool ! In `datasets` it would be nice to not have to add yet another parameter in `load_dataset`. Maybe we can just extent `use_auth_token` ?

Currently `use_auth_token` is `Optional[str]`, but if a dict is passed it could be interpreted as a custom auth header. Therefore if `basic_auth_header()` could return a dict instead of a string, I think that would be helpful :)

In `huggingface_hub` we can have the same behavior for consistency

julien-c (commented):
now looks good to me 👌

SBrandeis:
Closing in favor of https://github.com/huggingface/moon-landing/issues/1714

PR #566, opened 2021-12-30
[Snapshot download] allow some filenames to be ignored

patrickvonplaten:
The snapshot is incredible useful if one doesn't know all the exact filenames, but it has the downside of **always** downloading all contents in a repo. While we can't prevent the model to re-download unchanged files if the sha of the whole model repo has changed, we could allow to only download certain files that pass a passed regex. 

*E.g.* PyCTCDecode makes use of `snapshot_download` in https://huggingface.co/patrickvonplaten/wav2vec2-large-xlsr-53-spanish-with-lm/tree/main but in only needs the files `language_model` and `alphabet.json`. It would be nice to be able to specify this with a `allowed_file_regex` parameter. This would prevent to download all three model files in https://huggingface.co/patrickvonplaten/wav2vec2-large-xlsr-53-spanish-with-lm/tree/main unnecessarily.


tests/test_snapshot_download.py
```diff
@@ -308,3 +308,27 @@ def test_download_model_local_only_multiple(self):
...
+                cache_dir=tmpdirname,
+                allowed_file_regex="dummy*",
+            )
+
+            # folder contains the two files contributed and the .gitattributes
+            folder_contents = os.listdir(storage_folder)
+            self.assertEqual(len(folder_contents), 2)
+            self.assertTrue("dummy_file.txt" in folder_contents)
+            self.assertTrue("dummy_file_2.txt" in folder_contents)
+            self.assertTrue(".gitattributes" not in folder_contents)
```
patrickvonplaten, line 327: make sure that `.gitattributes` is **not** in folder_contents here

LysandreJik (commented):
Thanks for your contribution, this is very useful! Would you mind adding:
1. Some documentation about it cc @stevhliu @osanseviero 
2. A test to ensure that lists of regexes are also correctly handled

Thank you!

patrickvonplaten:
Thanks for the pointer @osanseviero . I've slightly changed the PR to accept both `allowed` and `ignored` regex and to base the regex on Unix style rather than Python `re`-like style. 

The reasons are:
- Unix style is more intuitive in our case as we always deal with file names. E.g. it's a very common case to only download certain file extensions which is intuitively `*.json` with Unix style but not at all intuitive with Python's `re` regex (see https://stackoverflow.com/questions/16087450/python-regular-expression-match-with-file-extension)
- Another common use case is to exclude certain files / file extensions. "Not"-matching regex is not really intuitive so I think it's nicer to just add another argument for this. 

I've also added some docs. Would be great if you could take a second look @osanseviero @LysandreJik @stevhliu

src/huggingface_hub/snapshot_download.py
```diff
@@ -152,7 +155,22 @@ def snapshot_download(
...
+    ignore_regex = [ignore_regex] if isinstance(ignore_regex, str) else ignore_regex
+
     for model_file in model_files:
+        # verify that regex is allowed, if not any `allow_regex` is true then no download
+        if allow_regex is not None and not any(
+            fnmatch(model_file, r) for r in allow_regex
+        ):
+            continue
+
+        # verify that regex is not forbidden, if any `ignore_regex` is true then no download
```
osanseviero, line 168:
```suggestion
        # if there's a denylist, skip download if file does matches any regex
```

tests/test_snapshot_download.py
```diff
@@ -308,3 +308,43 @@ def test_download_model_local_only_multiple(self):
...
+
+            with open(os.path.join(storage_folder, "dummy_file.txt"), "r") as f:
+                contents = f.read()
+                self.assertEqual(contents, "v2")
+
+            # folder name contains the revision's commit sha.
+            self.assertTrue(self.second_commit_hash in storage_folder)
```
osanseviero, line 338: I feel this is a bit orthogonal to what we're testing

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
 >>> snapshot_download(repo_id="lysandre/arxiv-nlp", revision="main")
 ```
 
-In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
\ No newline at end of file
+In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
+
+To only download some files of a repository with `snapshot_download`, you can make 
+make use of the `allow_regex` and `ignore_regex` arguments to specific
```
osanseviero, line 88:
```suggestion
use of the `allow_regex` and `ignore_regex` arguments to specify 
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
 ```
 
-In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
\ No newline at end of file
+In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
+
+To only download some files of a repository with `snapshot_download`, you can make 
+make use of the `allow_regex` and `ignore_regex` arguments to specific 
+which files shall be downloaded.
+Both a single regex or a list of regex are accepted for `allow_regex` and `ignore_regex`.
```
osanseviero, line 90:
```suggestion
Both a single regex or a list of regexes are accepted for `allow_regex` and `ignore_regex`. 
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
+The regex matching is based on [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) which means that it provides support for Unix shell-style wildcards.
+
+To download, *e.g.*, only the JSON configuration files, you could make use of `allow_regex`:
+
+```python
+>>> from huggingface_hub import snapshot_download
+>>> snapshot_download(repo_id="lysandre/arxiv-nlp", allow_regex="*.json")
+```
+
+If, on the other hand, you might want to always exclude any files of `.msgpack`
```
osanseviero, line 100:
```suggestion
If, on the other hand, you want to always exclude any files of `.msgpack` 
```

osanseviero (commented):
Thanks a lot for this PR Patrick! I left a couple of comments but overall it looks great!

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
 >>> snapshot_download(repo_id="lysandre/arxiv-nlp", revision="main")
 ```
 
-In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
\ No newline at end of file
+In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
```
stevhliu, line 85:
```suggestion
In general, it is usually better to manually download files with `hf_hub_download` (if you already know the file name) to avoid re-downloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
+
+To only download some files of a repository with `snapshot_download`, you can make
```
stevhliu, line 87: Help set some context for why this is useful :)

```suggestion
However, you don't want to always download the contents of an entire repository with `snapshot_download`. Even if you don't know the file name and only know the file type, you can download specific files with `allow_regex` and `ignore_regex`.
```

docs/hub/how-to-downstream.md
stevhliu, line 90:
```suggestion
`allow_regex` and `ignore_regex` accept either a single regex or a list of regexes. 
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
 
-In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
\ No newline at end of file
+In general, it is usually better to manually download files with `hf_hub_download` (if you already know which files you need) to avoid redownloading an entire repository. `snapshot_download` is helpful when your library's downloading utility is a helper, and unaware of which files need to be downloaded.
+
+To only download some files of a repository with `snapshot_download`, you can make 
+make use of the `allow_regex` and `ignore_regex` arguments to specific 
+which files shall be downloaded.
+Both a single regex or a list of regex are accepted for `allow_regex` and `ignore_regex`. 
+The regex matching is based on [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) which means that it provides support for Unix shell-style wildcards.
```
stevhliu, line 91:
```suggestion
The regex matching is based on [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) which means it provides support for Unix shell-style wildcards.
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
+
+To download, *e.g.*, only the JSON configuration files, you could make use of `allow_regex`:
```
stevhliu, line 93:
```suggestion
For example, you can use `allow_regex` to only download JSON configuration files:
```

docs/hub/how-to-downstream.md
stevhliu, line 100:
```suggestion
On the other hand, `ignore_regex` can be used to exclude certain files from being downloaded. The following example ignores the `.msgpack` and `.h5` file extensions:
```

docs/hub/how-to-downstream.md
```diff
@@ -82,4 +82,32 @@ Download a whole repository as shown in the following:
...
+If, on the other hand, you might want to always exclude any files of `.msgpack` 
+or `.h5` extensions, you could make use of `ignore_regex`:
+
+```python
+>>> from huggingface_hub import snapshot_download
+>>> snapshot_download(repo_id="lysandre/arxiv-nlp", ignore_regex=["*.msgpack", "*.h5"])
+```
+
+Passing a regex can be especially useful when repositories contain files that 
+are are never expected to be downloaded by `snapshot_download`.
```
stevhliu, line 109:
```suggestion
are never expected to be downloaded by `snapshot_download`.
```

stevhliu (commented):
Super cool feature! 🤩

One question I have (sorry if it is a bit naive or if this type of scenario isn't likely!) is will `allow_regex` also look inside a folder for a specific file? For example, in the repo [here](https://huggingface.co/patrickvonplaten/wav2vec2-large-xlsr-53-spanish-with-lm/tree/main), if you only want `json` files, will `allow_regex` also look inside the `language_model` folder and download the `attrs.json` file in there?

patrickvonplaten:
> Super cool feature! star_struck
> 
> One question I have (sorry if it is a bit naive or if this type of scenario isn't likely!) is will `allow_regex` also look inside a folder for a specific file? For example, in the repo [here](https://huggingface.co/patrickvonplaten/wav2vec2-large-xlsr-53-spanish-with-lm/tree/main), if you only want `json` files, will `allow_regex` also look inside the `language_model` folder and download the `attrs.json` file in there?

Yes, this works as expected :-) In this case the filenames would be:

- config.json
- language_model/attrs.json

both those filenames would match the `*.json` regex

LysandreJik (approved):
LGTM, thanks @patrickvonplaten!

patrickvonplaten:
Thanks for the reviews! Think I got an **F** in spelling and grammar haha

patrickvonplaten:
CI failures seem unrelated -> merging

PR #569, opened 2022-01-04
Update fairseq TTS pipeline

kahne:


osanseviero:
This seems to have a couple of issues as seen in the [Docker tests](https://github.com/huggingface/huggingface_hub/runs/4712488285?check_suite_focus=true) during installation of requirement (`pkuseg`). I was not able to reproduce unfortunately, maybe @Narsil has some ideas otherwise I can dive a bit more into this.

osanseviero (approved):
Looks good from my side, thanks @Narsil and @kahne!

Narsil:
It seems the issue lies in the `pkuseg` that requires `numpy` to already be installed as it uses it in its installation script.

I took the liberty of modifying 
`self.generator = self.task.build_generator(self.model, cfg)` into `self.generator = self.task.build_generator(model, cfg)` (first is the raw model, second is a model list)

Also adding another layer of protection against bogus data ("\x00\x00" as an input gets transformed into nil tensor which triggers and error).

Issue #47, opened 2021-05-20
Discussion: download and upload models have a somewhat inconsistent API 


osanseviero:
Hi all. Just a small-ish thing to discuss.

Methods that get information about/from repos take the full model id (e.g. `hf_api.model_info("osanseviero/my-model")` or `snapshot_download("osanseviero/my-model)`).

On the other hand, `create_repo` takes as two separate arguments the model name and the organization (if any). I wonder if this could lead to confusions in the future, and, if so, we should check if the name param has a `/` and determine if there's an org in there.

nreimers:
I think this makes a lot of sense. Actually implemented it in the `save_to_hub` from sentence-transformers

Users are used to pull images like `{org}/{name}`, so it makes sense to have a `push_to_hub("{org}/{name}")`

osanseviero:
@LysandreJik @sgugger @julien-c in case you have any suggestions/ideas :) 

We now have better error messages when this error happens, so it might not be such a big issue.

LysandreJik:
I wouldn't mind having `org/name` be interpreted as `name, organization=org` if specified, I think it would be coherent with the downloading methods.

julien-c:
Yes I'd be supportive of this as well

severo:
Related: while working on the models' data in https://observablehq.com/@huggingface/kaggle-dataset-huggingface-modelhub, I have been surprised that some of the "modelId"s are not "user/model", but just "model" (I understand all of them are models provided by Hugging Face as the organization). eg: https://huggingface.co/bert-base-uncased

LysandreJik:
Yes, this is for historic purposes. The first models had no organization and were therefore at the root.

Unfortunately, for backwards compatibility purposes, it is now complicated to change the behavior.

PR #574, opened 2022-01-04
Bring consistency to download and upload APIs

muellerzr:
Closes https://github.com/huggingface/huggingface_hub/issues/47

This PR aims to bring some consistency in how the api can be used, by letting a user both pass in a specified `organization` as well as using `organization/model_name` for functions revolving around uploading and downloading


muellerzr:
@osanseviero mostly interested on if you think I missed any. There weren't many instances of this happening from a user perspective I could *see*, but one may have trickled past me

src/huggingface_hub/hf_api.py
```diff
@@ -811,6 +811,15 @@ def create_repo(
...
             raise ValueError("Invalid repo type")
 
+        if "/" in name:
+            if organization is not None:
+                if organization != name.split("/")[1]:
+                    raise ValueError(
+                        f"Both a `name` with an organization and an `organization` were passed in that do not align ({name}, {organization}). Please ensure these align to the same organization"
+                    )
+                else:
+                    name, organization = name.split("/")
```
osanseviero, line 821: I would actually do the split first and then check if length is 2 (avoid osan/repo/anotherthing) and raise a more informative value error

osanseviero (commented):
Overall looks good. I think you cover all the cases

LysandreJik (approved):
Agree with @osanseviero's comment, otherwise this looks good!

osanseviero (approved):

muellerzr:
Based on advice from @LysandreJik I forgo that logic for a complete refactor, making use of the `repo_type_and_id_from_hf_id` functionality, since we already have a function that *can*  check if it's a valid namespace.

src/huggingface_hub/hf_api.py
```diff
@@ -808,6 +808,17 @@ def create_repo(
...
+        checked_name = repo_type_and_id_from_hf_id(name)
+        name = checked_name[2]
+
+        if organization is None:
+            organization = checked_name[1]
+        elif organization != checked_name[1]:
+            raise ValueError(
+                f"""Passed `organization` and `name` organization are not the same ({organization}, {checked_name[1]}).
+            Please either include the organization in only `name` or the `organization` parameter, such as `api.create_repo({checked_name[0]}, organization={organization})` or `api.create_repo({checked_name[1]}/{checked_name[2]})`"""
+            )
```
LysandreJik, line 820: Nice

LysandreJik (approved):
Looks good to me!

PR #576, opened 2022-01-07
Add a HowTo on searching the Hub efficiently

muellerzr:
Here is the start to my doc on searching the hub efficiently with the new `ModelFilter` and `ModelSearchArguments` classes. Feedback is welcome, as well as anything else you would like or expect to see in the documentation.

cc @thomwolf :)


docs/hub/searching-the-hub.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+
+Below is a graphic repeating the process for finding both our `Text Classification` requirement, as well as our `glue` requirement:
+
+![](../assets/hub/search_text_classification.gif)
+
+![](../assets/hub/search_glue.gif)
+
+Now that we have the pieces, we just need to combine them all together for something the api can use, through the `ModelFilter` and `DatasetFilter` classes. These take those outputs we just had, and transform them into something the API can use conviently:
```
osanseviero, line 126:
```suggestion
Now that all the pieces are there, the last step is to combine them all for something the API can use through the `ModelFilter` and `DatasetFilter` classes. The classes transform the outputs of the previous step into something the API can use conveniently:
```

docs/hub/searching-the-hub.md
```diff
@@ -0,0 +1,189 @@
...
+    	config: None
+    	private: False
+    	downloads: 6
+    	library_name: transformers
+    	likes: 0
+    }
+
+
+
+As we can see, it found the models we wanted that fit all of our criteria. We can even take it further by passing in an array for each of the parameters we had before. For example, now we'll look for the same configuration, but also include `TensorFlow` in the filter:
```
osanseviero, line 159:
```suggestion
As you can see, it found the models that fit all the criteria. You can even take it further by passing in an array for each of the parameters from before. For example, let's take a look for the same configuration, but also include `TensorFlow` in the filter:
```

docs/hub/searching-the-hub.md
```diff
@@ -0,0 +1,189 @@
...
+
+![](../assets/hub/search_text_classification.gif)
+
+![](../assets/hub/search_glue.gif)
+
+Now that we have the pieces, we just need to combine them all together for something the api can use, through the `ModelFilter` and `DatasetFilter` classes. These take those outputs we just had, and transform them into something the API can use conviently:
+
+
+```python
+>>> from huggingface_hub import ModelFilter, DatasetFilter
```
osanseviero, line 130:
```suggestion
>>> from huggingface_hub import ModelFilter
```

Removed as it was not used

osanseviero (approved):
This looks great! This was a nice read!

I left a bunch of nit comments, but please feel free to just integrate what you like and skip what you don't. Two "bigger" things
* I tried to avoid first person (https://developers.google.com/style/person), but this is not a strict rule so feel free to keep as is if you prefer.
* Should we have a quick example at the end for datasets?

Thank you!

(cc @stevhliu)

docs/hub/searching-the-hub.md
muellerzr: Opting to keep it in for a Dataset example at the end

docs/hub/searching-the-hub.md
```diff
@@ -0,0 +1,189 @@
+# Searching the Hub Efficiently with Python
+
+In this tutorial we will explore how to interact and search the Hub with the `HfApi` class to help find available models and datasets quickly.
+
+## The Basics
+
+The `huggingface_hub` comes packaged with an interface that can interact with the Hub in the `HfApi` class:
```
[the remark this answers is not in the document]
muellerzr: Went with:
`huggingface_hub` is a Python library that allows anyone to freely extract useful information from the Hub, as well as downloading and publishing models. You can install it with:
```bash
pip install huggingface_hub
```

It comes packaged with an interface that can interact with the Hub in the `HfApi` class:

osanseviero (approved):
Looks good! Thank you!

PR #579, opened 2022-01-07
Shared components directory

beurkinger:
The problem:  right now it's impossible to use certain components in widgets without duplicating them in the widgets folder, which means we end up with several versions of the same components to manage and keep up-to-date. This is for instance the case for the Dropdown component. There are already some duplicate between some files present in this repo and `moon-landing` .

As discussed with @mishig25, @Pierrci and @pngwn : 
- rename `/widgets` to `/js` and make it the directory for shared components/utils/types between this repo and `moon-landing`
- move `Dropdown` and its related components to the new shared components folder
- move `ViewUtils.ts` to `/js/src/lib/utils`
- dedupe `moon-landing` and this repo's `ViewUtils` files
- two functions in `ViewUtils.ts` were very specific to some widgets, so I moved them to the `InferenceWidget/shared/helpers.ts` file

Despite moving some files, this should not change anything in terms of behavior or functionality.

Note: the PR has a lot of file changes but these are essentially files moving / import updates ;)


pngwn:
After some conversations with @beurkinger, I think the best approach here would be to set up a JS monorepo with pnpm. It would resolve some of the issues with this PR (importing node_modules from a bare folder will fail, for example) but still allow separation of shared concerns.

Real monorepo tooling will take care of imports between 'packages' (for example import shared widgets or types into other packages) without using deep relative paths or aliases, and since this tooling generates hard links for module imports it is robust and works across all tooling, unlike aliases which you need to teach all tools about manually.

This also paves a simpler path to publishing some of these components to npm, which seems like a step we may want to take with this being an open source repo containing components, as pnpm will help us with versions, allowing easy independent versioning of various components and widgets.

I also think the same approach should be taken in some other repos but I'll raise that separately.

Pierrci:
As discussed offline, moving the `shared` dir inside `widgets/src/lib` should do the trick (and ok to - finally! - rename `widgets` to something more generic like `js`).

The monorepo architecture is something that might be worth considering if indeed we decide to open-source more things / publish packages, but I think that for now, the current setup is working well for our needs.

beurkinger:
After a chat with @Pierrci and @mishig25, we decided to simply continue using the `widgets` folder as our "shared js" directory, but rename it to `js` to makes it clearer its content is not only centered around widgets. It's the easiest, most reasonable way to go right now :)

beurkinger:
While we're at it, and since `/widgets` is going to become our official "shared js" directory (at least for now), may I suggest that we organize the inside of the `widgets/src/lib` directory a bit better ? Right now we're dumping everything together.

Maybe we could have something like 
- `lib/components`
- `lib/utils`
- `lib/interfaces` (or `lib/types`)
- `lib/constants`
?

Also right now `lib/interfaces` is a bit of a mess - it contains types, constants, and even a .md files. Maybe we could take this as an opportunity to clean that up ?

@Pierrci @mishig25 @pngwn

pngwn:
This looks sensible to me. Agree that the interface folder could be clean up a little although I don't have an issue with folders having markdown files to document the internals.

I have no preference on naming.

beurkinger:
Woops, I had not seen that the .md file was used for documenting the doc content, my bad, I definitely should have checked before opening my mouth.

beurkinger:
@Pierrci I recently noticed we have a copy of the server's `ViewUtils.ts` file in `hf_hub/widgets/lib/InferenceWidgets/shared/`.

For the sake of cleaning things up and avoiding duplication, shall we put a single version of this file in the new `hf_hub/js/lib/utils` folder and make it the "official" ViewUtils.ts file used by the server ?

beurkinger:
Follow-up to my previous message: 

@Pierrci and I had a chat, and we decided that the most reasonable way to go was to have a `hf_hub/js/lib/utils/ViewUtils.ts` file containing the functions that are required by both `hf_hub` and `moon-landing`, and to keep in the `server/lib/ViewUtils.ts` the functions that are only used by the moon-landing server.

That way we won't have to update `hf_hub` every time we have to make a change in a `ViewUtils.ts` function that is only used by the server.

Pierrci (approved):

PR #584, opened 2022-01-11
Switch /api/{login|logout} to /{login|logout}

coyotte508:
Both endpoints currently exist, but `/api/login` and `/api/logout` are deprecated


julien-c:
i think we should remove them, auth is token-based now

@SBrandeis will review, this is his baby :)

Pierrci:
> i think we should remove them, auth is token-based now

+1, I don't think it makes sense to keep those methods here

SBrandeis:
> i think we should remove them

Yes, let's push the users towards token-based authentication. I'd advocate to remove them from the docs, but leave the `huggingface_hub` package untouched for now (cc @LysandreJik )

SBrandeis (approved):

LysandreJik (approved):

julien-c (approved):

PR #589, opened 2022-01-12
Misc nits for task pages

merveenoyan:
Creating this draft PR in case if we have any nits reported from pre-prod.
@osanseviero @beurkinger @gary149


merveenoyan:
Put dots back @Pierrci 
I will merge this to Tasks EOD @osanseviero (waiting in case if we have any other feedbacks)

PR #590, opened 2022-01-12
Tasks

Pierrci:
There didn't seem to be any PR open for the `tasks` branch - easier to understand all the changes in conjunction w/ https://github.com/huggingface/moon-landing/pull/1651 @osanseviero @beurkinger @merveenoyan


LysandreJik:
Will this be merged in `huggingface_hub`? For `transformers` we recently switched to using a dataset on the hub to host images, as otherwise it significantly weights down repositories.

If this PR is to be merged, it would weigh down the repo by at least a couple MBs. Would it be possible to host them on a hub dataset, like we do [here](https://huggingface.co/datasets/huggingface/documentation-images) for example?

beurkinger:
@LysandreJik @Pierrci 

Regarding hosting hugginface_hub task assets on a Hugging Face dataset:
- I don't think "weighting down" the repositories is an issue right now - we're talking about some pictures and audio files. People can probably clone the whole thing on their smartphone connection  in a couple of seconds. And for subsequent push/pull, only relevant files are updated anyway.
- If possible I personally prefer having the assets where they're supposed to be used, rather than in another place. Simply makes things easier to manage.
- One of the goal of putting tasks in hugginface_hub is that people can easily edit the tasks, but if the assets become hosted in another place they don’t have access to, doesn't it beat the whole purpose ?
- Hosting assets on datasets feels like we're abusing our own dataset system (which is supposed to be used to host ML stuff). Why not use a service/place that is actually dedicated to this ?

julien-c:
Just on this specific point:

> Hosting assets on datasets feels like we're abusing our own dataset system (which is supposed to be used to host ML stuff).

not really, this is still content related to ML, i don't think it is a complete abuse of our system. And the open source team has been doing it for a while

LysandreJik:
Hey @beurkinger, thanks for your comment. I agree with you that what I offer is not necessarily ideal regarding this exact PR: what I want to optimize for is external contributions. Cloning a repository which weighs ~17MB is fine (current size of hfh), but if PRs like this one pile up it wouldn't be surprising to see its size 10x over the coming months. 

Unfortunately, as I'm sure you're aware, once an asset has been added to the repository, and unless we rewrite the history (which we should never do for public-facing repositories), it will be part of its history forever. I think having some of the burden on us in terms of maintenance is not an issue if it makes external contributions simpler.

> One of the goal of putting tasks in hugginface_hub is that people can easily edit the tasks, but if the assets become hosted in another place they don’t have access to, doesn't it beat the whole purpose ?

I agree with you so feel free to keep it here if you're expecting external contributors to edit these tasks. My point still stands for other/future assets.

beurkinger:
> Unfortunately, as I'm sure you're aware, once an asset has been added to the repository, and unless we rewrite the history (which we should never do for public-facing repositories), it will be part of its history forever

I was going to suggest rewriting the history, but if this is not an option then you're making a good point on the long term.

Apart from not having the assets available on the repo, the main thing I'm wary about is using datasets for that purpose, because it's definitely not the usage this system was built for, and I feel this idea is kind of a "if all you have is a hammer, everything looks like a nail" situation ;) 

I'm definitely not an expert on the subject, but could we not use Git LFS to make the assets part of the repo without increasing its size, if things get out of control ?

> if you're expecting external contributors to edit these tasks

This is an interesting point - are we sure we do @osanseviero (not a loaded question, just making sure) ?

I'd be curious to have your opinion on the matter @Pierrci @pngwn

LysandreJik (approved):
Indeed, we could eventually use `git-lfs` for that purpose but it does make contributing a bit harder (but I suppose every HF-user/contributor should be aware of git-lfs by now!)

No opposition to merging this PR like this and revisiting the issue at a later point.

beurkinger:
@LysandreJik yes, I was thinking that since these assets have already been put on the repo, it's too late for them anyway :D

julien-c:
not until the branch is merged, no

merveenoyan:
@julien-c @beurkinger I've put assets [here](https://huggingface.co/datasets/huggingface/task-page-images) earlier to make your life easier, if you want to do it

beurkinger:
@julien-c @merveenoyan changing the assets location means rewriting some code, updating both repos, and review again. We would like to publish the new pages in the next hour, but that would set us back for no major gain IMHO. I'll let you decide if it's worth it.

merveenoyan:
@julien-c @beurkinger can we just release like this and change the path to assets later?

beurkinger:
We can. Downside is that the files will stay in the fit history. I think that removing them from the history in time would not be such a big deal, even if the repo is public, as it's only used by moon-landing AFAIK, and these are not crucial files. Then again I'm not incredibly familiar with the do and don'ts the open-source community.

julien-c:
@beurkinger this repo is *not* only used by `moon-landing` multiple other projects depend on it

PR #592, opened 2022-01-14
Fixing AllenNLP.

Narsil:


api-inference-community/docker_images/allennlp/app/pipelines/question_answering.py
```diff
@@ -1,6 +1,8 @@
+# Even though it is not imported, it is actually required, it downlaods some stuff.
```
osanseviero, line 1: This should be before `allennlp_models`

osanseviero (approved):
Looks good, thanks!

api-inference-community/docker_images/allennlp/app/pipelines/question_answering.py
Narsil: Done !

PR #594, opened 2022-01-14
Realtime ASR widget

mishig25:
# Add `realtime` capability to ASR widget

Add realtime asr capability to a model's ASR widget. To do so, we need:
1. User needs HF token? (@Narsil can provide more details)
2. In the modelCard, user needs to turn on flag [widget_realtime_asr](https://github.com/huggingface/huggingface_hub/blob/e0e5f74dddf68bcb214c18b21cf6c9c8f9da3353/widgets/src/lib/interfaces/Types.ts#L214) (open for suggestion on this flag's naming cc: @julien-c @anton-l)

# How to test locally:
1. hardcode your token [here](https://github.com/huggingface/huggingface_hub/blob/e0e5f74dddf68bcb214c18b21cf6c9c8f9da3353/widgets/src/lib/InferenceWidget/shared/WidgetRealtimeRecorder/Recorder.ts#L16)
2. `npm run dev`
3. visit http://localhost:3000/facebook/wav2vec2-base-960h

## Screenshots
| initial state | success state |
|---|---|
| Screenshot 2022-01-16 at 17 09 48 | Screenshot 2022-01-16 at 17 09 36 |
| token error | mic access error |
| Screenshot 2022-01-16 at 17 10 04 | Screenshot 2022-01-16 at 17 10 34 |
| warning: mic input is empty | warning: realtime asr output is empty |
| Screenshot 2022-01-16 at 17 11 08 | Screenshot 2022-01-16 at 17 11 32 |

cc: @anton-l @patrickvonplaten @osanseviero


widgets/src/lib/InferenceWidget/shared/WidgetRealtimeRecorder/Recorder.ts
```diff
@@ -0,0 +1,61 @@
...
+		const node = new AudioWorkletNode(this.audioContext, "vumeter");
+		node.port.onmessage = (event) => {
+			const base64String = btoa(String.fromCharCode(...new Uint8Array(event.data.buffer.buffer)));
+			const message = {
+				raw: base64String,
+				sampling_rate: event.data.sampling_rate,
+			};
+			this.socket.send(JSON.stringify(message));
+		};
+		this.microphone.connect(node).connect(this.audioContext.destination);
```
mishig25, line 49: On chrome, this line and 
https://github.com/huggingface/huggingface_hub/blob/e0a232a0b19b4e49008373ec9d0fc39ea5b593ee/widgets/static/capture.js#L19-L26
doesn't run
cc: @Narsil

Narsil:
> User needs HF token? (@Narsil can provide more details)

You hardcoded the use of GPU (which does provide the nicest experience) but GPUs is only allowed for customers, so token won't work for most users.

Currently, the websocket is also only allowed for customers, but this is changeable.

CPU will work too, it will just be slower/higher latency.

widgets/src/lib/Icons/IconMagicWand.svelte
```diff
@@ -0,0 +1,22 @@
+<script>
```
beurkinger, line 1:
```suggestion
<script lang="ts">
```

let's take the habit of adding `lang="ts"` to our script tags, since it is becoming mandatory :)

Pierrci:
+1 on what @Narsil said, we should always use CPU for widgets - is the experience much much worse compared to GPU here?

Narsil:
```python
import datetime
import platform
import subprocess
import os
import base64
import asyncio
import json
import websockets
import curses

# nlp = pipeline("automatic-speech-recognition", device=0)


start = datetime.datetime.now()

sampling_rate = 48_000
chunk_length_s = 0.1

MODEL_ID = "facebook/wav2vec2-base-960h"
COMPUTE_TYPE = "cpu"  # or "gpu"
SENT = 0
AUTH = True
# URI = f"ws://localhost:8000/asr/live/{COMPUTE_TYPE}/{MODEL_ID}"
# URI = f"ws://localhost:8001/asr/live"
URI = f"wss://api-inference.huggingface.co/asr/live/{COMPUTE_TYPE}/{MODEL_ID}"


async def send(websocket):
    global SENT, AUTH
    # You need to login with a first message as headers are not forwarded
    # for websockets
    if AUTH:
        await websocket.send(f"Bearer {os.getenv('HF_API_TOKEN')}")
    format_for_conversion = "f32le"
    microphone = ffmpeg_microphone(
        sampling_rate=sampling_rate, chunk_length_s=chunk_length_s, format_for_conversion=format_for_conversion
    )
    print("> Start talking")
    for i, bytes_audio in enumerate(microphone):
        inputs = {
            "raw": base64.b64encode(bytes_audio).decode("utf-8"),
            "sampling_rate": sampling_rate,
            "format": format_for_conversion,
        }
        # payload = {"id": i, "inputs": inputs}
        await websocket.send(json.dumps(inputs))
        SENT += 1
        await asyncio.sleep(0.01)


async def recv(websocket, stdscr):
    global SENT
    full_text = ""
    while True:
        data = await websocket.recv()
        payload = json.loads(data)
        stdscr.addstr(0, 0, f"Sent {SENT} -> {payload.get('partial', True)}")
        text = payload.get("text", "")
        stdscr.addstr(1, 0, repr(payload))
        stdscr.addstr(2, 0, full_text + text)
        if not payload.get("partial", True):
            full_text += text

        stdscr.clrtoeol()
        stdscr.refresh()


def ffmpeg_microphone(
    sampling_rate: int,
    chunk_length_s: float,
    format_for_conversion: str = "f32le",
):
    ar = f"{sampling_rate}"
    ac = "1"
    if format_for_conversion == "s16le":
        size_of_sample = 2
    elif format_for_conversion == "f32le":
        size_of_sample = 4
    else:
        raise ValueError("Unhandled format `{format_for_conversion}`. Please use `s16le` or `f32le`")

    system = platform.system()
    if system == "Linux":
        format_ = "alsa"
        input_ = "default"
    elif system == "Darwin":
        format_ = "avfoundation"
        input_ = ":0"
    elif system == "Windows":
        format_ = "dshow"
        input_ = "default"

    ffmpeg_command = [
        "ffmpeg",
        "-f",
        format_,
        "-i",
        input_,
        "-ac",
        ac,
        "-ar",
        ar,
        "-f",
        format_for_conversion,
        "-fflags",
        "nobuffer",
        "-hide_banner",
        "-loglevel",
        "quiet",
        "pipe:1",
    ]
    chunk_len = int(round(sampling_rate * chunk_length_s)) * size_of_sample
    iterator = _ffmpeg_stream(ffmpeg_command, chunk_len)
    for item in iterator:
        yield item


def _ffmpeg_stream(ffmpeg_command, buflen: int):
    """
    `bufout` is used to enable streaming partial
    """
    bufsize = 2 ** 24  # 16Mo
    try:
        with subprocess.Popen(ffmpeg_command, stdout=subprocess.PIPE, bufsize=bufsize) as ffmpeg_process:
            while True:
                raw = ffmpeg_process.stdout.read(buflen)
                if raw == b"":
                    break
                yield raw
    except FileNotFoundError as error:
        raise ValueError("ffmpeg was not found but is required to stream audio files from filename") from error


async def main(stdscr):
    curses.noecho()
    curses.cbreak()
    async with websockets.connect(URI) as websocket:
        # inputs and parameters are classic, "id" is a way to track that query
        future = send(websocket)
        future_r = recv(websocket, stdscr)
        await asyncio.gather(future, future_r)


def mmain(stdscr):
    loop = asyncio.get_event_loop()
    if loop.is_running():
        # When running in notebooks
        import nest_asyncio

        nest_asyncio.apply()
    results = loop.run_until_complete(main(stdscr))


curses.wrapper(mmain)
```
Here's a script so everyone can test locally if you want to test the difference.
The latency is particularly bad when the sample length reaches 10s or so (before we consider it "full"). It still viewable, but definitely not as good as GPU.

Pierrci:
Thanks @Narsil!

I see 2 paths here:
1. Stay on CPU, and maybe add a message in the widget stating "computation on CPU for test purpose only, favor short texts" or something similar
2. Enable a GPU fair-use quota for all HF-logged users (for asr models only?)

I would favor 1. for now as it's probably the simplest/fastest approach, but we can revisit 2. later if more need arise for making GPU accessible to all users in some (limited) way? cc @julien-c

widgets/static/audioProcessor.js
```diff
@@ -0,0 +1,38 @@
+class AudioDataExtractor extends AudioWorkletProcessor {
+    _updateIntervalInMS;
+    _sampleInFrames;
+    _index;
+    _buffer;
+
+    constructor() {
+        super();
+        this._updateIntervalInMS = 50;
```
patrickvonplaten, line 9: This is a bit too small IMO. It means that audio chunks are only 0.05 seconds long no? Think we should increase this to 0.2 seconds. In my experiments, the quality was significantly better when increasing the chunk length and a delay of 0.2 is still very good!

mishig25:
Superceded by https://github.com/huggingface/huggingface_hub/pull/738

PR #607, opened 2022-01-19
Add how to search the hub to the sidebar

muellerzr:
Looks like it was missed when the article was added to the documentation.


docs/hub/_sections.yml
```diff
@@ -21,7 +24,7 @@
 
 - local: adding-a-library
   title: Integrate a library with the Hub
-
+
```
osanseviero, line 27: nit: please remove

osanseviero (approved):

julien-c:
can we check why the CI is red? (even if it's not linked to this particular change)

LysandreJik:
Can reproduce, will patch. Thanks for raising the issue @julien-c

PR #608, opened 2022-01-19
Remove rogue API token

LysandreJik:
The `apiToken` is only available on the `valid_org` organization in the staging environment.

Thanks for raising the issue, @julien-c.


LysandreJik:
This is the wrong fix - investigating.

LysandreJik:
Merging as seen offline with @SBrandeis

PR #609, opened 2022-01-19
Set a limit to what valid Model Tags and Dataset Tags can be, and filter results with them for namespaces

muellerzr:
Currently there is a "bug" where a user can upload a model or dataset, and have it be an invalid pipeline tag, such as `sequence_classification` rather than `sequence-classification`.

This becomes an issue when we have our helper classes such as `Model` and `DatasetSearchArguments`, as there's two references to `sequence-classification` it needs to point to, and one overrides the other.

In the ideal situation, we have a way to catch a user from uploading a model or dataset with invalid tags, but this would be the starting point I think to improved standardization.


osanseviero:
Standardization actually exists a bit in the Hub backend side. Some quick context:

We have metadata validation in the hub when pushing for things such as the license (https://github.com/huggingface/moon-landing/issues/396). The idea is to expand this in a progressive way (e.g. language would be a natural follow-up). When users push wrong tag, they get nice warning for it.

One concern I have with this is that https://github.com/merveenoyan/huggingface_hub/blob/main/widgets/src/lib/interfaces/Types.ts and this might just diverge, that's what I usually like cross-language tools to have this info. If @LysandreJik feels this PR is the right direction, I would just suggest to also add a clear comment both here and in `Types.ts` asking people to update both files.

osanseviero:
Also cc @julien-c  on this.

@muellerzr, do you have the # of model/dataset repos with wrong tags? That sounds as something to be fixed :)

muellerzr:
The number is not something I have. But I found my problematic one very quickly when I was looking at:

```python
import requests

api = HfApi()
path = f"{api.endpoint}/api/datasets-tags-by-type"
r = requests.get(path)
o = r.json()
for item in o["task_categories"]:
    if "text-classification" in item['id'] or "text_classification" in item["id"]:
        print(item)
```
We find two "text-classifications" (only the first is the valid one. The second only has one item):
```
{'id': 'task_categories:text-classification', 'label': 'text-classification', 'type': 'task_categories'}
{'id': 'task_categories:text_classification', 'label': 'text_classification', 'type': 'task_categories'}
```

muellerzr:
@osanseviero reached out to the authors, 3/5 of their datasets had issues, so once those are fixed I'll close this

Issue #598, opened 2022-01-16
Use all parameters of the `keras.models.save_model`


ariG23498:
With the current implementation of saving a keras model we are missing out on a number of different parameters as mentioned in the [official docs of `tf.keras.models.save_model`](https://www.tensorflow.org/api_docs/python/tf/keras/models/save_model)
```python
tf.keras.models.save_model(
    model, filepath, overwrite=True, include_optimizer=True, save_format=None,
    signatures=None, options=None, save_traces=True
)
```

From the `push_to_hub_keras` code:
https://github.com/huggingface/huggingface_hub/blob/aa4994cf98db77cd709ea7a107a83b39c0d8f91b/src/huggingface_hub/keras_mixin.py#L53

I would love to see more control on the `push_to_hub_keras` API.

merveenoyan:
@gante @osanseviero @ariG23498 we could introduce `include_optimizer` as an optional parameter that is set to default as False. For other parameters you indicated here we could maybe include signatures which would allow users willing to use their model on TF lite. I don't see a need to change the rest. WDYT?

merveenoyan:
Answered it in the previous issue but my 2 cents:
```
@gante @ariG23498 @osanseviero
save_traces registers every custom layer to SavedModel by default and we don't need to register custom objects, it's my bad. I thought the error @ariG23498 got was related to that because the error message indicated this.
Now, we only need to change include_optimizer and maybe we could change signatures for TF Lite users, related to #598.
```

merveenoyan:
Hello 👋  I'm currently on it. Will inform you when I open the PR 🤗

PR #616, opened 2022-01-21
include_optimizer parameter introduced to push_to_hub_keras()

merveenoyan:
This PR addresses the discussion we had in [#598](https://github.com/huggingface/huggingface_hub/issues/598).
I did a quick and dirty implementation and haven't tested yet so it's a WIP, I want to hear the review first.

For context: this parameter prevents models with custom optimizers (like the ones built with tfa) to be serialized and is by default set to True so I set it to False and made it an optional parameter.

cc: @osanseviero

### DIFF
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 
 def save_pretrained_keras(
-    model, save_directory: str, config: Optional[Dict[str, Any]] = None
+    model, save_directory: str, include_optimizer: Optional[Bool], config: Optional[Dict[str, Any]] = None
 ):
     """Saves a Keras model to save_directory in SavedModel format. Use this if you're using the Functional or Sequential APIs.
 
@@ -27,6 +27,8 @@ def save_pretrained_keras(
         Specify directory in which you want to save the Keras model.
     config (:obj:`dict`, `optional`):
         Configuration object to be saved alongside the model weights.
+    include_optimizer(:obj:`bool`, `optional`):
+        Whether or not to include optimizer in serialization.
     """
     if is_tf_available():
         import tensorflow as tf
@@ -50,7 +52,7 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
-    tf.keras.models.save_model(model, save_directory)
+    tf.keras.models.save_model(model, save_directory, include_optimizer)
 
 
 def from_pretrained_keras(*args, **kwargs):
@@ -69,6 +71,7 @@ def push_to_hub_keras(
     git_user: Optional[str] = None,
     git_email: Optional[str] = None,
     config: Optional[dict] = None,
+    include_optimizer: Optional[bool] = False,
 ):
     """
     Upload model checkpoint or tokenizer files to the 🤗 Model Hub while synchronizing a local clone of the repo in
@@ -104,6 +107,8 @@ def push_to_hub_keras(
             will override the ``git config user.email`` for committing and pushing files to the hub.
         config (:obj:`dict`, `optional`):
             Configuration object to be saved alongside the model weights.
+        include_optimizer (:obj:`bool`, `optional`):
+            Whether or not to include optimizer during serialization.
 
     Returns:
         The url of the commit of your model in the given repository.
@@ -150,7 +155,7 @@ def push_to_hub_keras(
     )
     repo.git_pull(rebase=True)
 
-    save_pretrained_keras(model, repo_path_or_name, config=config)
+    save_pretrained_keras(model, repo_path_or_name, config=config, include_optimizer=include_optimizer)
 
     # Commit and push!
     repo.git_add(auto_lfs_track=True)
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -167,7 +167,11 @@ def push_to_hub_keras(
     repo.git_pull(rebase=True)
 
     save_pretrained_keras(
-        model, repo_path_or_name, config=config, include_optimizer=include_optimizer, **model_save_kwargs
+        model,
+        repo_path_or_name,
+        config=config,
+        include_optimizer=include_optimizer,
+        **model_save_kwargs,
     )
 
     # Commit and push!
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -20,7 +20,7 @@ def save_pretrained_keras(
     model,
     save_directory: str,
     config: Optional[Dict[str, Any]] = None,
-    include_optimizer: Optional[Bool] = False,
+    include_optimizer: Optional[bool] = False,
     **model_save_kwargs,
 ):
     """Saves a Keras model to save_directory in SavedModel format. Use this if you're using the Functional or Sequential APIs.
@@ -167,7 +167,7 @@ def push_to_hub_keras(
     repo.git_pull(rebase=True)
 
     save_pretrained_keras(
-        model, repo_path_or_name, config=config, include_optimizer=include_optimizer
+        model, repo_path_or_name, config=config, include_optimizer=include_optimizer, **model_save_kwargs
     )
 
     # Commit and push!
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -17,8 +17,8 @@ logger = logging.getLogger(__name__)
 
 
 def save_pretrained_keras(
-    model, save_directory: str, include_optimizer: Optional[Bool], config: Optional[Dict[str, Any]] = None
-):
+    model, save_directory: str, config: Optional[Dict[str, Any]] = None, include_optimizer: Optional[Bool] = False, 
+    **model_save_kwargs: Optional[Dict] = None):
     """Saves a Keras model to save_directory in SavedModel format. Use this if you're using the Functional or Sequential APIs.
 
     model:
@@ -29,6 +29,8 @@ def save_pretrained_keras(
         Configuration object to be saved alongside the model weights.
     include_optimizer(:obj:`bool`, `optional`):
         Whether or not to include optimizer in serialization.
+    **model_save_kwargs(:obj:`dict`, `optional`):
+        Arguments other than default arguments that can be passed to tf.keras.models.save_model().
     """
     if is_tf_available():
         import tensorflow as tf
@@ -52,7 +54,7 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
-    tf.keras.models.save_model(model, save_directory, include_optimizer)
+    tf.keras.models.save_model(model, save_directory, include_optimizer=include_optimizer, **model_save_kwargs)
 
 
 def from_pretrained_keras(*args, **kwargs):
@@ -72,6 +74,7 @@ def push_to_hub_keras(
     git_email: Optional[str] = None,
     config: Optional[dict] = None,
     include_optimizer: Optional[bool] = False,
+    **model_save_kwargs: Optional[dict] = None
 ):
     """
     Upload model checkpoint or tokenizer files to the 🤗 Model Hub while synchronizing a local clone of the repo in
@@ -109,6 +112,8 @@ def push_to_hub_keras(
             Configuration object to be saved alongside the model weights.
         include_optimizer (:obj:`bool`, `optional`):
             Whether or not to include optimizer during serialization.
+        **model_save_kwargs(:obj:`dict`, `optional`):
+            Arguments other than default arguments that can be passed to tf.keras.models.save_model().
 
     Returns:
         The url of the commit of your model in the given repository.
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -34,7 +34,7 @@ def save_pretrained_keras(
     include_optimizer(:obj:`bool`, `optional`):
         Whether or not to include optimizer in serialization.
     model_save_kwargs(:obj:`dict`, `optional`):
-        Arguments other than default arguments that can be passed to tf.keras.models.save_model().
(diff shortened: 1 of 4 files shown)


gante (commented):
Should solve the issue 👍

osanseviero:
Thanks for this PR!

Suggestion for future: before sending PRs for review, it's always good idea to test it at least locally since you might realize it does not work for X reason (wrong argument, weird TF bug, etc).

Anyways, overall looks good :+1:  feel free to open the PR for official review once you've tested it and added some automated test as well

ariG23498:
Hey folks,

My two cents on this would be to use a `**model_save_kwargs` instead of `include_optimizer`. This would help us in using all the parameters of the `tf.keras.models.save_model` API with minimal changes.

WDYT?

merveenoyan:
@osanseviero @ariG23498's suggestion is neat. I don't know which choice to pick because we'll set it to False by default to handle the custom optimizers (like the ones that tfa offers) user has to set it to True if they want to push a model with one of the default optimizers. Maybe take `include_optimizer` only and take the rest with something like `**model_save_kwargs`?

osanseviero:
Yes, @merveenoyan. I think that sounds like a right approach and consistent with what we do in transformers and other places (setting `include_optimizer` to false and then other save args through `**model_save_kwargs**`)

src/huggingface_hub/keras_mixin.py
```diff
@@ -150,7 +166,9 @@ def push_to_hub_keras(
     )
     repo.git_pull(rebase=True)
 
-    save_pretrained_keras(model, repo_path_or_name, config=config)
+    save_pretrained_keras(
+        model, repo_path_or_name, config=config, include_optimizer=include_optimizer
+    )
```
ariG23498, line 171: @merveenoyan I think we need to pass the `**kwargs` argument here as well.

merveenoyan:
I tested this in multiple ways, first one was directly pushing it by `include_optimizer = True` which was successful.

Second one was adding `save_traces = False` (this is one of the `**model_save_kwargs`) which didn't push the custom layers (and given that I didn't implement `get_config` it failed to load, which is expected already). This is normally set as `True` by TensorFlow. Pushing is just fine but `from_pretrained_keras() `doesn't work (again, expected, up to user to set it to something different than default).
Ekran Resmi 2022-01-27 13 42 35
Ekran Resmi 2022-01-27 13 44 05

I also didn't include optimizer and saved the traces, then when loading I was asked to compile manually.
Ekran Resmi 2022-01-27 13 53 03

src/huggingface_hub/keras_mixin.py
```diff
@@ -27,6 +31,10 @@ def save_pretrained_keras(
         Specify directory in which you want to save the Keras model.
     config (:obj:`dict`, `optional`):
         Configuration object to be saved alongside the model weights.
+    include_optimizer(:obj:`bool`, `optional`):
+        Whether or not to include optimizer in serialization.
+    **model_save_kwargs(:obj:`dict`, `optional`):
```
osanseviero, line 36: In the docstring you can just put the name, no need for `**`. Same in the other docstring
```suggestion
    model_save_kwargs(:obj:`dict`, `optional`):
```

ariG23498:
@merveenoyan I think this works! Thanks for the great work 😄

gante (approved):
LGTM

merveenoyan:
I'm writing automated tests at the moment, I will not merge yet. 🤓

merveenoyan:
@osanseviero last week I've removed the pytest assertions from the two new test cases (they weren't going to be raised so). Should I merge?

src/huggingface_hub/keras_mixin.py
```diff
@@ -27,6 +31,10 @@ def save_pretrained_keras(
         Specify directory in which you want to save the Keras model.
     config (:obj:`dict`, `optional`):
         Configuration object to be saved alongside the model weights.
+    include_optimizer(:obj:`bool`, `optional`):
+        Whether or not to include optimizer in serialization.
+    model_save_kwargs(:obj:`dict`, `optional`):
+        Arguments other than default arguments that can be passed to tf.keras.models.save_model().
```
osanseviero, line 37: For consistency with other mixins and methods, maybe let's use 

`model_save_kwargs will be passed to `tf.keras.models.save_model()``

(e.g. https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hub_mixin.py#L49)

src/huggingface_hub/keras_mixin.py
```diff
@@ -104,6 +116,10 @@ def push_to_hub_keras(
             will override the ``git config user.email`` for committing and pushing files to the hub.
         config (:obj:`dict`, `optional`):
             Configuration object to be saved alongside the model weights.
+        include_optimizer (:obj:`bool`, `optional`):
+            Whether or not to include optimizer during serialization.
+        model_save_kwargs(:obj:`dict`, `optional`):
+            Arguments other than default arguments that can be passed to tf.keras.models.save_model().
```
osanseviero, line 122: Same here :)

tests/test_keras_integration.py
```diff
@@ -178,7 +178,43 @@ def test_save_pretrained(self):
...
+        model.build((None, 2))
+
+        save_pretrained_keras(
+            model, f"{WORKING_REPO_DIR}/{REPO_NAME}", include_optimizer=True
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+        self.assertEqual(len(files), 4)
```
osanseviero, line 204: I think this test should focus in seeing if the optimizer is saved. For example, you could load the model and then check that the optimizer test is there

tests/test_keras_integration.py
```diff
@@ -178,7 +178,43 @@ def test_save_pretrained(self):
...
+
```
osanseviero, line 205: A second test that  shows that the optimizer is not there by default would be a good idea

tests/test_keras_integration.py
```diff
@@ -178,7 +178,43 @@ def test_save_pretrained(self):
...
+            model,
+            f"{WORKING_REPO_DIR}/{REPO_NAME}",
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+        self.assertEqual(len(files), 4)
+
+    def test_save_pretrained_optimizer_weights(self):
```
osanseviero, line 191: nit: Should it be `optimizer_state` instead?

tests/test_keras_integration.py
```diff
@@ -250,6 +286,30 @@ def test_push_to_hub(self):
...
+        push_to_hub_keras(
+            model,
+            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+            config={"num": 7, "act": "gelu_fast"},
+            include_optimizer=True,
+            save_traces=False,
```
osanseviero, line 312: The code after this `push_to_hub_keras` tests for the same thing as `test_push_to_hub`. If the two tests have same code, we are not testing for different behavior. For example, you could test that loading fails or that traces are not there

osanseviero (commented):
Thanks a lot for this PR! Overall looks good, but I think the tests could improve a bit :smile:

tests/test_keras_integration.py
merveenoyan: there's no such thing. I load the model and check if `model.optimizer` is `None` or not

merveenoyan: will fix it

osanseviero:
Feel free to pin the `black` version :smile:

merveenoyan:
AFTER FIGHTING WITH BLACK FOR HOURS @osanseviero

tests/test_keras_integration.py
merveenoyan: I check if the custom layer exception is raised.

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+            model,
+            f"{WORKING_REPO_DIR}/{REPO_NAME}",
+        )
         files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
 
         self.assertIn("saved_model.pb", files)
         self.assertIn("keras_metadata.pb", files)
         self.assertEqual(len(files), 4)
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+        self.assertIsNone(model.optimizer)
```
osanseviero, line 192: I would suggest to avoid testing this as well in the same test. Maybe let's have a test which is
`test_save_pretrained_does_not_save_optimizer_state_by_default` which has this

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+
+        save_pretrained_keras(
+            model, f"{WORKING_REPO_DIR}/{REPO_NAME}", include_optimizer=True
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+
+        self.assertEqual(len(files), 4)
```
osanseviero, line 208: I think we can remove this since it's already tested in `test_save_pretrained`

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+        self.assertIsNotNone(model.optimizer)
```
osanseviero, line 210: Nice!!

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+        save_pretrained_keras(
+            model, f"{WORKING_REPO_DIR}/{REPO_NAME}", include_optimizer=True
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+
+        self.assertEqual(len(files), 4)
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
```
osanseviero, line 209: Maybe let's use another variable name when loading since `model` is the original model. Maybe `loaded_model`?

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+        save_pretrained_keras(
+            model,
+            f"{WORKING_REPO_DIR}/{REPO_NAME}",
+        )
         files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
 
         self.assertIn("saved_model.pb", files)
         self.assertIn("keras_metadata.pb", files)
         self.assertEqual(len(files), 4)
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
```
osanseviero, line 191: Nit: for all and all other uses of `from_pretrained_keras`, avoid assigning to a variable named `model` since you already have that above.

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+
+        self.assertEqual(len(files), 4)
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+        self.assertIsNotNone(model.optimizer)
+
+    def test_save_pretrained_kwargs(self):
```
osanseviero, line 212: Maybe a better name
```suggestion
    def test_save_pretrained_kwargs_load_fails_without_traces(self):
```

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+            save_traces=False,
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+        self.assertEqual(len(files), 4)
+
+        model = from_pretrained_keras(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+        self.assertRaises(ValueError, msg="Exception encountered when calling layer*")
```
osanseviero, line 231: Nice test!

tests/test_keras_integration.py
```diff
@@ -178,12 +179,56 @@ def test_save_pretrained(self):
...
+            model,
+            f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            include_optimizer=False,
+            save_traces=False,
+        )
+        files = os.listdir(f"{WORKING_REPO_DIR}/{REPO_NAME}")
+
+        self.assertIn("saved_model.pb", files)
+        self.assertIn("keras_metadata.pb", files)
+        self.assertEqual(len(files), 4)
```
osanseviero, line 228: Feel free to remove things if they are well tested in other tests

osanseviero (approved):
Thanks a lot for this PR! I left a couple of improvement ideas but looks almost ready to merge! :rocket: :rocket:

tests/test_keras_integration.py
merveenoyan: this is already tested in the test below but not the DummyModel's weights so I'll remove this part only

tests/test_keras_integration.py
merveenoyan: I check the Exception of traces and that's the only assertion, shouldn't this be resolved?

merveenoyan:
@osanseviero on my local (checked black version) the quality test passes but here it fails. Any guesses?

PR #618, opened 2022-01-21
Improvements non git mixin

muellerzr:
This is the start of getting the `improvements-non-git-mixin` up to speed and worked towards its completion.

Basic changes found here:
- `upload_file` now takes in a `commit_message`. This was part of the intended implementation, so trickled it up. This also allows for the back-ends to write custom messages for if something is a weight, tokenizer, etc, mimicking how Repository can do so already. 
- Added a test in `test_hubmixin` that makes sure the timings for uploading a new file to a blank repo, and uploading a new file to a repo with fairly large files is about the same. Currently the fake repo uploads 5 .5mb files (locally this took 12 seconds to accomplish. If we want to lower that to one .5mb file that's okay)


LysandreJik (approved):
This is quite a nice improvement to the existing PR. Feel free to merge that PR in mine, and then let's discuss what is needed on it before we can merge.

What needs to be taken care of before we can merge in `main`, in my opinion:
- Taking care of the 5GB files and chunking them before uploading them. Ideally the size would be given manually to the method, similarly to how it's done in `datasets`' `push_to_hub` method which already respects that paradigm (see https://github.com/huggingface/datasets/pull/3098)
- I may be mistaken but I don't think progress bars are implemented yet for `upload_file`
- Very explicit documentation mentioning that for complex workflows, `Repository` should be used as it respects the git workflow

Anything else? Exciting to see progress on this!

muellerzr:
Nope! That should be about it for this PR. Will make the adjustment + merge, then will work towards the list

PR #619, opened 2022-01-21
[WIP] Propose a non-git mixin

LysandreJik:


LysandreJik:
Will close this PR, this is the one that should be followed: https://github.com/huggingface/huggingface_hub/pull/847

Will add it to the v0.9 milestone so that it makes the next release.

Issue #557, opened 2021-12-22
Creating a repo does not allow using model as repo_type


osanseviero:
Using `create_repo` does not accept `repo_type="model"` right now. This is counter intuitive, specially as the params docstring says 

> repo_type: Set to "dataset" or "space" if creating a dataset or space, default is model

But in reality only `dataset` and `space` are [allowed](https://github.com/huggingface/huggingface_hub/blob/7b7b6e18ad6b770ca7cce3187512e91dc22a4dee/src/huggingface_hub/constants.py#L36).

Here is an example painful usage in which I need to have conditionals to handle spaces/datasets vs models https://huggingface.co/spaces/osanseviero/fork_a_repo/blob/main/app.py#L14-L20

cc @LysandreJik @muellerzr

LysandreJik:
Indeed, there's a bit of a discrepancy as initially there were only model repositories. I think it's fine if the backend remains the same (using `None`), but the frontend should probably allow `"model"` as a valid `repo_type`.

julien-c:
agree with @LysandreJik solution

PR #620, opened 2022-01-21
Include "model" in repo_type to keep consistency 

muellerzr:
Closes https://github.com/huggingface/huggingface_hub/issues/557

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1050,7 +1050,7 @@ class HfApi:
                 Authentication token, obtained with :function:`HfApi.login` method. Will default to the stored token.
 
             repo_type (``str``, Optional):
-                Set to :obj:`"dataset"` or :obj:`"space"` if uploading to a dataset or space, :obj:`None` or :obj:`"model"` if uploading to a model. Default is :obj:`None`.
+                Set to :obj:`"dataset"` or :obj:`"space"` if the file is in a dataset or space, :obj:`None` or :obj:`"model"` if in a model. Default is :obj:`None`.
 
             revision (``str``, Optional):
                 The git revision to commit from. Defaults to the :obj:`"main"` branch.
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -611,7 +611,7 @@ class HfApi:
         """
         Get the list of files in a given repo.
         """
-        if repo_type is None:
+        if repo_type is None or repo_type == "model":
             info = self.model_info(
                 repo_id, revision=revision, token=token, timeout=timeout
             )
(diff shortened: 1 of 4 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -743,7 +743,8 @@ def create_repo(
 
         json = {"name": name, "organization": organization, "private": private}
         if repo_type is not None:
-            json["type"] = repo_type
+            if repo_type != "model":
+                json["type"] = repo_type
```
LysandreJik, line 747: For clarity I'd keep the same logic you implemented everywhere else, i.e., keeping the `if repo_type ... None ... "model"`:

```diff
-        if repo_type is not None:
+        if repo_type is not None and repo_type != "model":
-            if repo_type != "model":
-                json["type"] = repo_type
+            json["type"] = repo_type
```

LysandreJik (approved):
LGTM, only left a nit :)

src/huggingface_hub/hf_api.py
muellerzr: Adjusted

LysandreJik (approved):
LGTM!

src/huggingface_hub/hf_api.py
```diff
@@ -742,19 +742,19 @@ def create_repo(
...
-                    raise ValueError(
-                        "No space_sdk provided. `create_repo` expects space_sdk to be one of "
-                        f"{SPACES_SDK_TYPES} when repo_type is 'space'`"
-                    )
-                if space_sdk not in SPACES_SDK_TYPES:
-                    raise ValueError(
-                        f"Invalid space_sdk. Please choose one of {SPACES_SDK_TYPES}."
-                    )
-                json["sdk"] = space_sdk
+        elif repo_type == "space":
```
LysandreJik, line 747: Actually shouldn't this be an `if` rather than `elif` given that if `repo_type == "space"` would have entered the previous branch of the if statement? 

Would be nice to add a test for that specific behavior to ensure it's not broken.

Thanks @julien-c for raising!

src/huggingface_hub/hf_api.py
```diff
@@ -742,19 +742,19 @@ def create_repo(
             raise ValueError("Invalid repo type")
 
         json = {"name": name, "organization": organization, "private": private}
-        if repo_type is not None:
+        if repo_type is not None and repo_type != "model":
```
julien-c, line 745: on the server side you can probably pass type="model" to create a model repo

(so you can do `json["type"] = repo_type` also in that case)

src/huggingface_hub/hf_api.py
```diff
@@ -822,7 +822,7 @@ def delete_repo(
             raise ValueError("Invalid repo type")
 
         json = {"name": name, "organization": organization}
-        if repo_type is not None:
+        if repo_type is not None and repo_type != "model":
```
julien-c, line 825: same comment here, you can remove that `and` clause

src/huggingface_hub/hf_api.py
muellerzr: So we're saying just get rid of the `None` entirely and just force the user to do explicitly `model`? (And set that as a default)

julien-c: no (i mean, not necessarily), just that you can remove the `and repo_type != "model" ` clause if you want

But if it's simpler yeah you can also get rid of `None` (i'd try to staying backward compatible if possible though)

LysandreJik: Yes, staying backward compatible is super important, indeed

muellerzr: Got it, okay. Just making sure. Didn't think we wanted to go down that path

PR #621, opened 2022-01-24
[docs] Add `other` to licenses

julien-c:
cc @Pierrci


osanseviero:
What's the use case of this? Doesn't it go against the goal to standardize the licensing? (My worry is people exploiting this one)

osanseviero (approved):
I see this is in the license picker so approving :)

julien-c:
yes @osanseviero the discussion with @Pierrci was inside [internal PR](https://github.com/huggingface/moon-landing/pull/1642) but maybe buried a little deep so hard to follow

TL;DR: with the new GUI users who would want to assign a license that's not in the list could use "other" and add a `LICENSE` file to their repo). I wouldn't expect many users to use this as the license list is rather comprehensive already

PR #622, opened 2022-01-24
Add audio dependency to tests

LysandreJik:
Adds the audio dependency for tests as it is now required to download audio datasets.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -37,6 +37,9 @@ jobs:
           pip install --upgrade pip
           pip install .[testing]
 
+          apt install -y libsndfile1-dev
+          pip install datasets[audio]
+
       - run: pytest -sv ./tests/
 
   build_pytorch:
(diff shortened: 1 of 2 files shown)


muellerzr:
Might need sudo? 

> The Linux and macOS virtual machines both run using passwordless sudo. When you need to execute commands or install tools that require more privileges than the current user, you can use sudo without needing to provide a password.

.github/workflows/python-tests.yml
```diff
@@ -37,6 +37,9 @@ jobs:
           pip install --upgrade pip
           pip install .[testing]
 
+          apt install -y libsndfile1-dev
```
LysandreJik, line 40: Indeed, got too used to the self-hosted machines

```suggestion
          sudo apt install -y libsndfile1-dev
```

LysandreJik:
Still a failure related to the audio test of the inference API cc @osanseviero

Will take a look at it this afternoon

osanseviero:
From the error you should add `soundfile` to the testing dependencies our add `soundfile` to `datasets` `audio` installation https://github.com/huggingface/datasets/blob/master/setup.py#L104.

osanseviero:
Thanks for this PR!

PR #625, opened 2022-01-25
Fix pasting copied text inside TextAreaWidget

mishig25:
This PR fixes a bug where copied text was being pasted with a style inside the span text area (which is the default span behaviour)

Test it [here on netlify](https://61f16ed04efc61d1044b4c73--huggingface-widgets.netlify.app/facebook/bart-large-cnn)

### DIFF
--- a/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
+++ b/js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
@@ -1,39 +1,40 @@
 <script>
+	import { onMount, tick } from "svelte";
+
 	import WidgetLabel from "../WidgetLabel/WidgetLabel.svelte";
 
 	export let label: string = "";
 	export let placeholder: string = "Your sentence here...";
 	export let value: string;
 
-	function handlePaste(e: ClipboardEvent) {
-		e.preventDefault();
+	let textAreaEl: HTMLTextAreaElement;
 
-		const text = e.clipboardData.getData("text/plain");
-		const sel = window.getSelection();
-		const range = sel.getRangeAt(0);
-		range.insertNode(document.createTextNode(text));
-		range.collapse(false);
+	const HEIGHT_LIMIT = 500 as const;
+
+	async function resize() {
+		if (!!textAreaEl) {
+			textAreaEl.style.height = "";
+			await tick();
+			textAreaEl.style.height =
+				Math.min(textAreaEl.scrollHeight, HEIGHT_LIMIT) + "px";
+		}
+	}
+
+	$: {
+		value;
+		resize();
 	}
 </script>
 
 <WidgetLabel {label}>
 	<svelte:fragment slot="after">
-		<span
-			bind:textContent={value}
-			on:paste={handlePaste}
+		<textarea
+			bind:this={textAreaEl}
+			bind:value
 			class="{label
 				? 'mt-1.5'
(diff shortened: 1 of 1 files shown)


js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
```diff
@@ -1,29 +1,34 @@
...
+
+	function insertTextAtCaret(text: string) {
+		const sel = window.getSelection();
+		const range = sel.getRangeAt(0);
+		range.insertNode(document.createTextNode(text));
+		range.collapse(false);
 	}
+
+	onMount(() => {
+		spanEl.addEventListener("paste", (e) => {
```
beurkinger, line 20: You should remove the eventListener on unMount :)

beurkinger: And to achieve this, also probably make the arrow function a named handling function in the body of the component

beurkinger (changes requested):

js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
mishig25: Great suggestion!

js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
```diff
@@ -5,26 +5,22 @@
...
 	}
 </script>
 
 <WidgetLabel {label}>
 	<svelte:fragment slot="after">
 		<span
 			bind:textContent={value}
-			bind:innerHTML={innterHTML}
-			bind:this={spanEl}
+			on:paste={handlePaste}
```
beurkinger, line 23: oh, nice  !

js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
```diff
@@ -5,26 +5,22 @@
...
-	let spanEl: HTMLSpanElement;
-	const REGEX_SPAN = /<span .+>(.*)<\/span>/gms;
-	$: {
-		if (spanEl && innterHTML && REGEX_SPAN.test(innterHTML)) {
-			innterHTML = innterHTML.replace(REGEX_SPAN, (_, txt) => {
-				return txt;
-			});
-			spanEl.blur();
-		}
+	function handlePaste(e: ClipboardEvent) {
```
beurkinger, line 8: It's still buggy: if you copy some text into the area, click outside of it, and select an example, the content is not changed.

Also if you paste with the text content selected, the text is appended at the beginning of the content instead of replacing it.

I feel  this way of doing things is basically breaking the way Svelte handle input / state.

If you're using a span only because textarea don't adjust its height with content (why do we want this anyway ?), I strongly suggest you use a textarea and add some simple code to handle resizing according to its content, it's very easy.

beurkinger (changes requested):

js/src/lib/components/InferenceWidget/shared/WidgetTextarea/WidgetTextarea.svelte
beurkinger: (we tried implementing contentEditable element in a React project I work on, and we entered a world of pain doing so - like is seemingly the case here :D )

mishig25: https://github.com/huggingface/huggingface_hub/pull/625/commits/9b34fece7caee5897cf0074f1af326d44c227bb0 wdyt

beurkinger (approved):
@mishig25 looks like it's working much better ! 

thanks for taking the time to address this

PR #626, opened 2022-01-25
Introduce upload slicing

muellerzr:
The aim of this PR is to allow us to slice larger files into smaller ones, and upload those to the repo. 

To be implemented:

Downloading the chunks and rebuilding them directly inside `download_file`.

What size should we determine the file would need to be "chunked"? Right now it is set to 1mb. 

*Many* tests.

cc @LysandreJik


muellerzr:
Per slack discussion:

- Use a naming convention with an easy-to-read suffix, such as `filename.ext.{000, 001, 002}-of-010` so we know how many there are and what number we are on
- To reconstruct: check the list of files inside the repo to see if `filename.ext` is missing. If so but we have `filename.ext.XYZ`, reconstruct the file. 
- With 20gb upload now the limit, the need for slicing/sharding is for extremely large files, so won't happen too often

julien-c:
related: https://github.com/huggingface/transformers/issues/13548

julien-c:
This can now be closed no?

LysandreJik:
Yes!

PR #633, opened 2022-01-31
Update SB3 Libraries.ts

simoninithomas:
Hey 👋 ,

Since we now need to specify the extension of the file during the downstream task, I added ".zip".


osanseviero (approved):
Looks good!

PR #638, opened 2022-02-01
Support list_models and list_datasets with token arg

### LINKED ISSUE #634: List all datasets or models using token
As of now, `list_models` and `list_datasets` don't use the token, so they only list public models. If we could pass the token here, it would be very useful to do things such as retrieving all datasets (including private) of an organization.

cc @muellerzr @LysandreJik 

https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L511

muellerzr:
Closes #634

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -605,6 +605,11 @@ class HfApi:
                 token = use_auth_token
             else:
                 token = HfFolder.get_token()
+                if token is None:
+                    raise EnvironmentError(
+                        "You need to provide a `token` to `use_auth_token` or be logged in to Hugging Face with "
+                        "`huggingface-cli login`."
+                    )
         headers = {"authorization": f"Bearer {token}"} if use_auth_token else None
         params = {}
         if filter is not None:
@@ -788,6 +793,11 @@ class HfApi:
                 token = use_auth_token
             else:
                 token = HfFolder.get_token()
+                if token is None:
+                    raise EnvironmentError(
+                        "You need to provide a `token` to `use_auth_token` or be logged in to Hugging Face with "
+                        "`huggingface-cli login`."
+                    )
         headers = {"authorization": f"Bearer {token}"} if use_auth_token else None
         params = {}
         if filter is not None:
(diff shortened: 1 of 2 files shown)


LysandreJik (approved):
Yes, LGTM! Thanks @muellerzr.

src/huggingface_hub/hf_api.py
```diff
@@ -592,9 +593,19 @@ def list_models(
                 This is set to `True` by default when using a filter.
             fetch_config (:obj:`bool`, `optional`):
                 Whether to fetch the model configs as well. This is not included in `full` due to its size.
-
```
osanseviero, line 595: nit: revert

muellerzr:
cc @albertvillanova so it's on your radar :)

src/huggingface_hub/hf_api.py
```diff
@@ -764,9 +776,19 @@ def list_datasets(
...
+                a valid `auth_token` can be passed in as a string.
         """
         path = f"{self.endpoint}/api/datasets"
+        if use_auth_token:
+            if isinstance(use_auth_token, str):
+                if not self._is_valid_token(use_auth_token):
+                    raise ValueError("Invalid token passed!")
+                token = use_auth_token
+            else:
+                token = HfFolder.get_token()
```
osanseviero, line 790: Consider something like this for consistency

```suggestion
                token = HfFolder.get_token()
                if token is None:
                  raise EnvironmentError(
                    "You need to provide a `token` or be logged in to Hugging Face with "
                    "`huggingface-cli login`."
                  )
```

osanseviero (approved):
Looks good! Thanks a lot!

src/huggingface_hub/hf_api.py
muellerzr: Opted for this error as it's a bit more specific towards what's happening:
```python
                        "You need to provide a `token` to `use_auth_token` or be logged in to Hugging Face with "
                        "`huggingface-cli login`."
```

osanseviero: You're right

src/huggingface_hub/hf_api.py
osanseviero: ignore!

PR #639, opened 2022-02-01
Include cardData in list_models and list_datasets

muellerzr:
To support listing the carbon emissions as well as any other metadata that can be returned with `cardData:True`, this PR now includes that as a parameter to both `list_models` and `list_datasets`.

Note: By default this is `True` because timing wise I did not see a difference between with vs without `cardData`, and returning the metadata for each model and dataset only *increases* the understandability of each one IMO.

cc @julien-c @LysandreJik


tests/test_hf_api.py
```diff
@@ -780,6 +788,14 @@ def test_filter_models_with_complex_query(self):
             ["pytorch" in model.tags and "tf" in model.tags for model in models]
         )
 
+    @with_production_testing
+    def test_filter_models_with_cardData(self):
+        _api = HfApi()
+        models = _api.list_models("co2_eq_emissions", cardData=True)
+        self.assertTrue([hasattr(model, "cardData") for model in models])
+        models = _api.list_models("co2_eq_emissions", cardData=False)
+        self.assertTrue([not hasattr(model, "cardData") for model in models])
```
LysandreJik, line 797: What is this test's aim? From my local tests, I see that `list_models("co2_eq_emissions", cardData=False)` does return models with `cardData`.

The assertion checks that the list (which is made up of all `False` values) is truthy, which it is because a list is truthy as long as it has length.

I would expect a test such as the following:
```suggestion
        models = _api.list_models("co2_eq_emissions", cardData=False)
        self.assertTrue(all([not hasattr(model, "cardData") for model in models]))
```

But that would fail right now

tests/test_hf_api.py
```diff
@@ -858,7 +874,7 @@ def test_end_to_end_thresh_6M(self):
         REMOTE_URL = self._api.create_repo(
             name=self.REPO_NAME_LARGE_FILE,
             token=self._token,
-            lfsmultipartthresh=6 * 10 ** 6,
+            lfsmultipartthresh=6 * 10**6,
```
LysandreJik, line 877: Seems like you upgraded to black==22.1 already :)

LysandreJik (approved):
Thanks for your PR!

Would it be possible to add a bit of docs regarding how this attribute can be leveraged, for example with the `co2_eq_emissions`?

https://huggingface.co/docs/hub/searching-the-hub

I observed the same, putting `cardData` to `True` or `False` does not change the time required to list all models, so it's fine for me to put it enabled by default.

julien-c:
> I observed the same, putting `cardData` to `True` or `False` does not change the time required to list all models, so it's fine for me to put it enabled by default.

can you compare the underlying API calls? I think you're not seeing too much bc we have good bandwidth, but the underlying calls outputs from /api/models would have very different content sizes imo.

tests/test_hf_api.py
muellerzr: That actually seems to be an issue with the backend API, no? Because the assumption should be that calling with `cardData:False` should not return the `cardData`, as @julien-c noted here: https://github.com/huggingface/moon-landing/issues/1584#issuecomment-991344206

But this is actually inconsistent, telling me that something might be wrong with our backend. Why? When I do:
```python
o = api.list_models(cardData=False, full=False)
i = 0
models = []
for model in o:
    if hasattr(model, 'cardData'): 
        i+= 1
        models.append(model)
keys = set()
non_dict = []
for model in models: 
    if isinstance(model.cardData, dict):
        for key in list(model.cardData.keys()):
            keys.add(key)
    else:
        non_dict += [model]
```
What I discover is that there are models that return a modelInfo with a `co2_emissions` key, even though that should only happen when we set `cardData` to `True`. Along with this, there is a broken model that has the wrong card data, it's a list rather than a dictionary. Is that expected?

(This is also shown when I do this via a raw `requests`)

muellerzr: @julien-c this may also tie into your thinking on how they should have very different content sizes

julien-c: not sure, on my side i was just mentioning that `wget https://huggingface.co/api/models` and `wget https://huggingface.co/api/models?cardData=true` have 2x size diff

src/huggingface_hub/hf_api.py
```diff
@@ -786,6 +795,8 @@ def list_datasets(
         if full is not None:
             if full:
                 params.update({"full": True})
+        if cardData is not None:
+            params.update({"cardData": True})
```
LysandreJik, line 799: I'd apply the same as for the `full`:
```suggestion
        if cardData is not None:
            if cardData:
                params.update({"cardData": True})
```

This way, if it's passed as `False` then it's still handled correctly.

julien-c:
lgtm!

tests/test_hf_api.py
LysandreJik, line 797:
```suggestion
        models = _api.list_models("co2_eq_emissions")
        self.assertTrue(all([not hasattr(model, "cardData") for model in models]))
```

tests/test_hf_api.py
```diff
@@ -672,6 +672,14 @@ def test_list_datasets_search(self):
         self.assertGreater(len(datasets), 10)
         self.assertIsInstance(datasets[0], DatasetInfo)
 
+    @with_production_testing
+    def test_filter_datasets_with_cardData(self):
+        _api = HfApi()
+        datasets = _api.list_datasets(cardData=True)
+        self.assertTrue([hasattr(dataset, "cardData") for dataset in datasets])
+        datasets = _api.list_datasets(cardData=False)
+        self.assertTrue([not hasattr(dataset, "cardData") for dataset in datasets])
```
LysandreJik, line 681:
```suggestion
        datasets = _api.list_datasets()
        self.assertTrue(all([not hasattr(dataset, "cardData") for dataset in datasets]))
```

LysandreJik (approved):
Other than the tests, LGTM

muellerzr:
Merging PR. For transparency:

```python
        self.assertTrue(all([not hasattr(dataset, "cardData") for dataset in datasets]))
```
This will *always* fail because it is hard coded into the `DatasetResult` to set `cardData` to `None`, even if it's not present.

@LysandreJik let me know if you'd like me to clean that up a bit with some `setattr`'s instead of hard-coding inputs.

PR #648, opened 2022-02-03
Enhance FA (Persian) examples

adrinjalali:
This PR enhances the existing examples and adds examples to the empty sections.

It also replaces space with ZWNJ (zero width non joiner) where applicable, but leaves the ZWNJ out sometimes, since people sometimes leave it out.

This PR also changes some examples to use words and place names more common in the Persian language.

Ping @m3hrdadfi since I think you worked on the existing examples.


julien-c (commented):
so awesome to have someone on the team who knows Farsi now! 🙏

osanseviero (approved):
Thank you very much! This is cool :fire:

PR #651, opened 2022-02-04
TensorBoard for Keras models

merveenoyan:
TensorBoard FTW

### DIFF
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -100,8 +100,8 @@ def push_to_hub_keras(
             repository will be created in your namespace (unless you specify an :obj:`organization`) with
             :obj:`repo_name`.
         tensorboard_dir (:obj:`str`, `optional`):
-            TensorBoard logging directory. Specify this in case you have used TensorBoard callback and would like
-            to host TensorBoard on your model repository.
+            TensorBoard logging directory to be pushed. The Hub automatically hosts 
+            and displays a TensorBoard instance if log files are included in the repository.
         commit_message (:obj:`str`, `optional`):
             Message to commit while pushing. Will default to :obj:`"add model"`.
         organization (:obj:`str`, `optional`):
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -72,7 +72,7 @@ def push_to_hub_keras(
     model,
     repo_path_or_name: Optional[str] = None,
     repo_url: Optional[str] = None,
-    tensorboard_dir: Optional[str] = None,
+    log_dir: Optional[str] = None,
     commit_message: Optional[str] = "Add model",
     organization: Optional[str] = None,
     private: Optional[bool] = None,
@@ -99,8 +99,8 @@ def push_to_hub_keras(
             Specify this in case you want to push to an existing repository in the hub. If unspecified, a new
             repository will be created in your namespace (unless you specify an :obj:`organization`) with
             :obj:`repo_name`.
-        tensorboard_dir (:obj:`str`, `optional`):
-            TensorBoard logging directory to be pushed. The Hub automatically hosts 
+        log_dir (:obj:`str`, `optional`):
+            TensorBoard logging directory to be pushed. The Hub automatically hosts
             and displays a TensorBoard instance if log files are included in the repository.
         commit_message (:obj:`str`, `optional`):
             Message to commit while pushing. Will default to :obj:`"add model"`.
@@ -178,8 +178,8 @@ def push_to_hub_keras(
         include_optimizer=include_optimizer,
         **model_save_kwargs,
     )
-
-    copytree(tensorboard_dir, f"{repo_path_or_name}/logs")
+    if log_dir is not None:
+        copytree(log_dir, f"{repo_path_or_name}/logs")
     # Commit and push!
     repo.git_add(auto_lfs_track=True)
     repo.git_commit(commit_message)
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -179,7 +179,7 @@ def push_to_hub_keras(
         **model_save_kwargs,
     )
 
-    copytree(tensorboard_dir, repo_path_or_name/logs)
+    copytree(tensorboard_dir, f"{repo_path_or_name}/logs")
     # Commit and push!
     repo.git_add(auto_lfs_track=True)
     repo.git_commit(commit_message)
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/keras_mixin.py
```diff
@@ -97,6 +99,9 @@ def push_to_hub_keras(
             Specify this in case you want to push to an existing repository in the hub. If unspecified, a new
             repository will be created in your namespace (unless you specify an :obj:`organization`) with
             :obj:`repo_name`.
+        tensorboard_dir (:obj:`str`, `optional`):
```
osanseviero, line 102: I feel `log_dir` might be a nicer name so we're consistent both with `tensorflow` and `transformers`. WDYT?

src/huggingface_hub/keras_mixin.py
```diff
@@ -97,6 +99,9 @@ def push_to_hub_keras(
...
+            TensorBoard logging directory. Specify this in case you have used TensorBoard callback and would like
+            to host TensorBoard on your model repository.
```
osanseviero, line 104: WDYT of something along these lines?
```suggestion
            TensorBoard logging directory to be pushed. The Hub automatically hosts 
            and displays a TensorBoard instance if log files are included in the repository.
```

tests/test_keras_integration.py
```diff
@@ -292,6 +292,36 @@ def test_push_to_hub(self):
...
+        model.build((None, 2))
+        push_to_hub_keras(
+            model,
+            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            tensorboard_dir=f"{WORKING_REPO_DIR}/tb_log_dir",
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+            config={"num": 7, "act": "gelu_fast"},
```
osanseviero, line 310: We can remove the config I think :smile:

tests/test_keras_integration.py
```diff
@@ -292,6 +292,36 @@ def test_push_to_hub(self):
...
+            include_optimizer=False,
```
osanseviero, line 311: No need to set this flag

tests/test_keras_integration.py
```diff
@@ -292,6 +292,36 @@ def test_push_to_hub(self):
...
+        )
+        self.assertEqual(
+            os.path.exists(f"{WORKING_REPO_DIR}/{REPO_NAME}/logs/tensorboard.txt"), True
+        )
+
+        model_info = HfApi(endpoint=ENDPOINT_STAGING).model_info(
+            f"{USER}/{REPO_NAME}",
+        )
+
+        self.assertEqual(model_info.modelId, f"{USER}/{REPO_NAME}")
```
osanseviero, line 321: The test should be just about if the tensorboard directory is in the repo, no need to check for the `modelId`, this was already tested

tests/test_keras_integration.py
```diff
@@ -292,6 +292,36 @@ def test_push_to_hub(self):
...
+            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            tensorboard_dir=f"{WORKING_REPO_DIR}/tb_log_dir",
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+            config={"num": 7, "act": "gelu_fast"},
+            include_optimizer=False,
+        )
+        self.assertEqual(
```
osanseviero, line 313: You can use `self.assertTrue` instead

src/huggingface_hub/keras_mixin.py
```diff
@@ -174,6 +179,7 @@ def push_to_hub_keras(
         **model_save_kwargs,
     )
 
+    copytree(tensorboard_dir, f"{repo_path_or_name}/logs")
```
osanseviero, line 182: A lot of tests are failing because `tensorboard_dir` is None by default. You should only do this step when it's not `None`

See https://github.com/huggingface/huggingface_hub/runs/5067079206?check_suite_focus=true#step:5:209

tests/test_keras_integration.py
```diff
@@ -292,6 +292,36 @@ def test_push_to_hub(self):
 
         self._api.delete_repo(name=f"{REPO_NAME}", token=self._token)
 
+    def test_push_to_hub_tensorboard(self):
+        os.mkdir(f"{WORKING_REPO_DIR}/tb_log_dir")
```
osanseviero, line 296: I think this fails because `{WORKING_REPO_DIR}` does not exist yet. You can use `os.makedirs` instead. https://github.com/huggingface/huggingface_hub/runs/5067079206?check_suite_focus=true#step:5:262

osanseviero (commented):
Thanks for this PR. The tests are failing, so I left a couple of notes :smile:

merveenoyan:
@osanseviero I did resolved everything

tests/test_keras_integration.py
```diff
@@ -292,6 +292,28 @@ def test_push_to_hub(self):
...
+        push_to_hub_keras(
+            model,
+            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            log_dir=f"{WORKING_REPO_DIR}/tb_log_dir",
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+        )
+        self.assertTrue(
```
osanseviero, line 311: This tests if the txt file is in the local repo (great!). You will also need to add an assert to see if the file is in the remote repo. For this, you can get `model_info.siblings` which shows all the files in the repo

osanseviero (approved):
Thanks a lot! I left one last comment to improve the test but it looks great!!

tests/test_keras_integration.py
```diff
@@ -292,6 +292,30 @@ def test_push_to_hub(self):
...
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+        )
+        model_info = HfApi(endpoint=ENDPOINT_STAGING).model_info(
+            f"{USER}/{REPO_NAME}",
+        )
+        model_files = [f.rfilename for f in model_info.siblings]
+        self.assertTrue("logs/tensorboard.txt" in model_files)
```
osanseviero, line 315: Alternatively
```suggestion
        self.assertTrue("logs/tensorboard.txt" in model_info.siblings)
```

osanseviero (approved):
Looks good!

tests/test_keras_integration.py
osanseviero: Bad suggestion, ignore D:

PR #653, opened 2022-02-04
Remove paying subscription mention from docstring

cakiki:
Private repositories do not need a paying subscription; this PR removes that from the docstring of the `push_to_hub` function.

cc @osanseviero


osanseviero (commented):
Thanks! Could you also remove this from https://github.com/huggingface/huggingface_hub/blob/3ab328a8cc0b57c731cd270a10149fe9bfe3a3a4/src/huggingface_hub/keras_mixin.py#L106 please?

osanseviero:
I think that last commit might have modified an unrelated file. Would you mind removing it?

cakiki:
I have no idea how that got in there, sorry 😅

osanseviero (approved):
Thank you very much for the PR! :rocket:

PR #654, opened 2022-02-04
Add security status

McPatate:
Is this the correct way to add a parameter ?

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -519,6 +519,7 @@ class HfApi:
         full: Optional[bool] = None,
         fetch_config: Optional[bool] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
+        security_status: Optional[bool] = None,
     ) -> List[ModelInfo]:
         """
         Get the public list of all the models on huggingface.co
@@ -635,6 +636,8 @@ class HfApi:
                 del params["full"]
         if fetch_config is not None:
             params.update({"config": fetch_config})
+        if security_status is not None:
+            params.update({"securityStatus": security_status})
         r = requests.get(path, headers=headers, params=params)
         r.raise_for_status()
         d = r.json()
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -897,9 +897,9 @@ class HfApi:
             else f"{self.endpoint}/api/models/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        status_query_param = "?securityStatus=True" if securityStatus else ""
+        status_query_param = {"securityStatus":True} if securityStatus else None
         path += status_query_param
-        r = requests.get(path, headers=headers, timeout=timeout)
+        r = requests.get(path, headers=headers, timeout=timeout, params=status_query_param)
         r.raise_for_status()
         d = r.json()
         return ModelInfo(**d)
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -897,9 +897,11 @@ class HfApi:
             else f"{self.endpoint}/api/models/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        status_query_param = {"securityStatus":True} if securityStatus else None
+        status_query_param = {"securityStatus": True} if securityStatus else None
         path += status_query_param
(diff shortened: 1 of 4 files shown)


McPatate:
I'm not sure I understand what you're asking.

To give you some context, I'm adding this so that users can see the security status of a given repository. The object added to `ModelInfo` will look like this :
```json
{
    "securityStatus" : {
         "containsInfected": bool,
         "infectionTypes": List<str>
    }
}
```

This was requested by @juliensimon for private model hub users :)

McPatate:
Also for reference@LysandreJik : https://github.com/huggingface/moon-landing/pull/2017

LysandreJik:
Ah, I think I'm misunderstanding something here. Right now you're adding a parameter to the `list_models` method, which aims to be a filtering method for the model repositories.

Adding a parameter there for security status would mean being able to filter models according to their security status (which is sensible!). I'm asking how do you envision the request to behave behind it, as with this added parameter, we'll now do requests with the `securityStatus=True` HTTP parameter. 

As I understand it, adding it here means that you'd like to be able to toggle the visibility of the `securityStatus` of the returned model repositories on or off in the response.

Would it make sense to add the security status to the `ModelInfo` itself?  https://github.com/huggingface/huggingface_hub/blob/df18b714c09a43313f9ff116a6534c12efc961cd/src/huggingface_hub/hf_api.py#L150-L190

LysandreJik:
Jumped on a quick call with @McPatate and I get it; thanks for taking the time.

The implementation looks good to me, but ideally there should also be a test that ensures the correct value is returned (or not returned!). For this to work, the PR moon-landing should be merged first, of course.

julien-c:
i don't think this is on the right endpoint, @McPatate – it should be on the `model_info` endpoint as discussed before

This is where the lookup has been implemented in the moon-landing repo (i.e. on a model specific basis)

julien-c:
(underlying endpoint has been merged on the hub side)

muellerzr (changes requested):
Nice addition!

One little nit, mostly with keeping the styles/formatting similar

src/huggingface_hub/hf_api.py
```diff
@@ -896,6 +897,8 @@ def model_info(
             else f"{self.endpoint}/api/models/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
+        status_query_param = "?securityStatus=True" if securityStatus else ""
```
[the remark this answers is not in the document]
McPatate: Well, I'd like this to be falsy if the user inputs `False`.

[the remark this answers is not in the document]
muellerzr: `params = {"securityStatus":True} if securityStatus else None`?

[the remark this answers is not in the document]
McPatate: Ah neat I just learned smthing :D

src/huggingface_hub/hf_api.py
```diff
@@ -897,9 +897,9 @@ def model_info(
             else f"{self.endpoint}/api/models/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        status_query_param = "?securityStatus=True" if securityStatus else ""
+        status_query_param = {"securityStatus":True} if securityStatus else None
         path += status_query_param
-        r = requests.get(path, headers=headers, timeout=timeout)
+        r = requests.get(path, headers=headers, timeout=timeout, params=status_query_param)
```
McPatate, line 902: Tell me if this works for you @muellerzr

src/huggingface_hub/hf_api.py
```diff
@@ -896,7 +897,11 @@ def model_info(
             else f"{self.endpoint}/api/models/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        r = requests.get(path, headers=headers, timeout=timeout)
+        status_query_param = {"securityStatus": True} if securityStatus else None
+        path += status_query_param
```
muellerzr, line 901:
```suggestion
```
Since we already pass in `status_query_param` to `params`, this isn't needed.

src/huggingface_hub/hf_api.py
muellerzr: @McPatate close! See my suggestion below. Otherwise good to go!

src/huggingface_hub/hf_api.py
McPatate: Sorry about this !

PR #657, opened 2022-02-05
Add precision about binary LFS files

coyotte508:
The hub now prevents binary files over 1MB from being committed (unless LFS)


julien-c (approved):
thanks for thinking of updating the doc! 

also cc'ing @LysandreJik and @osanseviero as this might affect some usages

Issue #635, opened 2022-02-01
Audio-to-regions widget and community API for pyannote.audio


hbredin:
Opening an issue as per @osanseviero's [suggestion on Twitter](https://twitter.com/osanseviero/status/1488512516162572290). 
Issue imported from https://github.com/pyannote/pyannote-audio/issues/835

---

[pyannote.audio 2.0](https://github.com/pyannote/pyannote-audio/tree/develop) will bring a unified pipeline API:

```python
from pyannote.audio import Pipeline
pipeline = Pipeline.from_pretrained("pyannote/speaker-diarization")
output = pipeline("audio.wav")   # or pipeline({"waveform": np.ndarray, "sample_rate": int})
```

where `output` is a [`pyannote.core.Annotation`](http://pyannote.github.io/pyannote-core/structure.html#annotation) instance.

I just [created a space](https://huggingface.co/spaces/pyannote/pretrained-pipelines) that allows to test a bunch of pipelines shared on Hugginface Hub but it would be nice if those were testable directly in their own model card.

My understanding is that two things needs to happen 
* pyannote pipeline support must be added to [Hugginface Inference API](https://github.com/huggingface/huggingface_hub/tree/main/api-inference-community)
* a new widget (based on [wavesurfer.js](https://wavesurfer-js.org/) like the aforementioned [space](https://huggingface.co/spaces/pyannote/pretrained-pipelines)?) must be added to the list of [Huggingface widgets](https://huggingface.co/docs/hub/main#what-are-all-the-possible-taskwidget-types)
![image](https://user-images.githubusercontent.com/1080837/151987868-d525347f-0b2c-479a-a140-7a77d8fe4dd8.png)

osanseviero:
This is a cool proposal! In Twitter I did mention we can model this task with `audio-to-audio` and it would already work by outputting multiple audios. But having a nice custom widget more specific for the task would be very cool!

cc @mishig25 @julien-c WDYT?

Narsil:
This is very cool ! 

Definitely a good target for audio-to-audio as a starter (no widget needed).
`audio-segmentation` seems like a good fit for what you're trying to do (does not exist yet, but should cover multiple use cases)

julien-c:
`audio-token-classification`? 😱
`audio-to-structured`?

not sure of the best new task type to keep some generality

But yeah could be cool to have it

Narsil:
> `audio-token-classification`? scream

[...]

PR #660, opened 2022-02-07
Pipeline for Pyannote Audio

J-Petiot:
As discuss here #635 , we decide with @hbredin to use automatic_speech_recognition while waiting for a more specific widget.


hbredin:
Hey @J-Petiot, for quality check to pass, I think you need to run [`black`](https://github.com/psf/black) code formatter on your files and commit the updated files.

api-inference-community/docker_images/pyannote/requirements.txt
```diff
@@ -0,0 +1,4 @@
+starlette==0.14.2
+api-inference-community==0.0.19
+git+https://github.com/pyannote/pyannote-audio.git@develop#egg=pyannote-audio
+speechbrain == 0.5.10
```
osanseviero, line 4: Is speechBrain needed? I think it can be removed

osanseviero (commented):
Thanks for the PR! 🚀 🔥 

A couple of comments
* From the discussion in the related issue, I don't think ASR is the right direction, since that would be a strong hack, and then your models would appear under ASR in the Hub. I think `audio-to-audio` is a good initial thing you could try: you output multiple labeled audios, so you can output one audio per segment with it's corresponding label.
* Could you please also add a docker test such as https://github.com/huggingface/huggingface_hub/blob/main/api-inference-community/tests/test_dockers.py#L221 ?
* Could you please duplicate one of these GA workflows so we run end-to-end tests with Docker? https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-api-espnet.yaml

Thanks again for the PR!

hbredin:
> * From the discussion in the related issue, I don't think ASR is the right direction, since that would be a strong hack, and then your models would appear under ASR in the Hub. I think `audio-to-audio` is a good initial thing you could try: you output multiple labeled audios, so you can output one audio per segment with it's corresponding label. ([example](https://huggingface.co/speechbrain/sepformer-wsj02mix))some time 

Until a proper `audio-segmentation` widget is available, I purposedly ask @J-Petiot to output text (`start_time end_time label`) rather than audio, as I believe it would be much more useful to the end-user.

Having the model output `num_speakers` audio files (containing the concatenation of audio segments of each speaker?) does not seem  very useful to me... Also, depending on the input audio file and the pipeline, `num_speakers` might be huuuuge.

api-inference-community/docker_images/pyannote/requirements.txt
hbredin: It is necessary for `pyannote/speaker-diarization` to work.
However, I don't plan to make it a hard `pyannote-audio` dependency.

osanseviero:
Take into account that this would also impact the discoverability/tags of your models. They would show under ASR which is not great. 

WDYT @Narsil?

hbredin:
> Take into account that this would also impact the discoverability/tags of your models. 
> They would show under ASR which is not great.

`audio-to-audio` does not make much sense for speaker diarization either. 

I suggest we put this PR on hold for now and start working on it again once the actual task (and widget) is available.
A good starting point fort the widget could be the part of hf.co/spaces/pyannote/pretrained-pipelines that interfaces with wavesurfer.js

hbredin:
Follow-up question: are `audio-segmentation` task and widget on HF short-term roadmap?

api-inference-community/docker_images/pyannote/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY ./prestart.sh /app/
+
+
+# Most DL models are quite large in terms of memory, using workers is a HUGE
+# slowdown because of the fork and GIL with python.
+# Using multiple pods seems like a better default strategy.
+# Feel free to override if it does not make sense for your library.
+ARG max_workers=1
+ENV MAX_WORKERS=$max_workers
+ENV HUGGINGFACE_HUB_CACHE=/data
```
Narsil, line 18: Can you confirm that the cache env key is using this ? Or does PYANNOTE defines another variable to locate the cache directory ? 

(This is important for the API to load the pods faster, by sharing a disk for the model files)

api-inference-community/docker_images/pyannote/app/pipelines/automatic_speech_recognition.py
```diff
@@ -0,0 +1,38 @@
...
+
+class AutomaticSpeechRecognitionPipeline(Pipeline):
+    def __init__(self, model_id: str):
+        # IMPLEMENT_THIS
+        # Preload all the elements you are going to need at inference.
+        # For instance your model, processors, tokenizer that might be needed.
+        # This function is only called once, so do all the heavy processing I/O here
+        # IMPLEMENT_THIS : Please define a `self.sampling_rate` for this pipeline
+        # to automatically read the input correctly
+        self.sampling_rate = 16000
```
Narsil, line 18: I see the Pypeline is able to resample on its own.

Is there a way to get the real value of the expected sampling rate from `Pypeline` ?
It will just avoid an unecessary resampling, since the inputs will be sent to you directly correctly sampled.

api-inference-community/docker_images/pyannote/app/pipelines/automatic_speech_recognition.py
```diff
@@ -0,0 +1,38 @@
...
+            the detected langage from the input audio
+        """
+        wav = torch.from_numpy(inputs).unsqueeze(0)
+        output = self.model({"waveform": wav, "sample_rate": self.sampling_rate})
+        regions = "".join(
+            [
+                f"|{str(datetime.timedelta(seconds=segment.start))[:-3]} - {str(datetime.timedelta(seconds=segment.end))[:-3]} : {label} |"
+                for segment, _, label in output.itertracks(yield_label=True)
+            ]
+        )
```
Narsil, line 37: This works, but isn't it very close to `str(output)` ?

api-inference-community/docker_images/pyannote/requirements.txt
Narsil: Perfect to add it here then.

hbredin:
> I had in mind the output to look something like:
> 
> ```json
> [{"label": "Speaker 1", "start": 0.1, "stop": 3.2}, ...]
> ```
> 
> Do you think that format would support all 4:
> 
>     * `speaker-diarization`
>     * `overlapped-speech-detection`
>     * `speaker-segmentation`
>     * `voice-activity-detection`
> 
Yes. It is perfectly fine for this type of pipeline.

api-inference-community/docker_images/pyannote/Dockerfile
hbredin: `pyannote` uses `PYANNOTE_CACHE`.

api-inference-community/docker_images/pyannote/app/pipelines/automatic_speech_recognition.py
hbredin: There is no easy way to query this...
FWIW, all existing models (for now) have been trained with `sample_rate = 16000`.

api-inference-community/docker_images/pyannote/app/pipelines/automatic_speech_recognition.py
hbredin: Very close indeed.

api-inference-community/docker_images/pyannote/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM tiangolo/uvicorn-gunicorn:python3.8
+LABEL maintainer="me <me@example.com>"
+
+# Add any system dependency here
+# RUN apt-get update -y && apt-get install libXXX -y
```
Narsil, line 5: I think you need to add `ffmpeg` here.

To check you can test the docker test locally.

```bash
RUN_DOCKER_TESTS=1 pytest -sv tests/test_dockers::DockerImageTests::test_pyannote
```

api-inference-community/docker_images/pyannote/Dockerfile
Narsil: Ok, do you mind adding

```
ENV PYANNOTE_CACHE=/data
```

api-inference-community/docker_images/pyannote/app/pipelines/automatic_speech_recognition.py
Narsil: Let's stick with this then.

Narsil:
Btw @J-Petiot , `make style` should fix the quality thing.

api-inference-community/docker_images/pyannote/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM tiangolo/uvicorn-gunicorn:python3.8
+LABEL maintainer="me <me@example.com>"
```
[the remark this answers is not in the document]
J-Petiot: Ok I'll put this mail.

api-inference-community/docker_images/pyannote/Dockerfile
J-Petiot: > Btw @J-Petiot , `make style` should fix the quality thing.

Ok thanks it should be better.

mishig25:
> `[{"label": "Speaker 1", "start": 0.1, "stop": 3.2}, ...]` output format

looks good to me.

in terms of the widget, 
As discussed in https://github.com/huggingface/huggingface_hub/issues/635, we can make a widget that uses https://wavesurfer-js.org/

Please let me know if I should start working on it

cc: @gary149 for design

Narsil:
You can try !

hbredin:
What should I put in my model card to activate inference?

hbredin:
image
Awesome!

julien-c:
really cool!!

julien-c:
(how long does the model takes to load for you guys? seems a bit slow on my side)

Narsil:
It's actually quite fast (10s-20s). But I did manage to trigger a bug, where it would keep hanging instead of showing the result. 
Looking into that.

Narsil:
> Please let me know if I should start working on it

If no one has any objections, I think we can get the ball rolling ! 
I'll add support in `api-inference-community` too.

Is everyone ok with the name `audio-segmentation` (with maybe other aliases later which are even more specific ?)

PR #668, opened 2022-02-07
Enable filtering by carbon emission

muellerzr:
This enables filtering via a carbon emission threshold when searching models at the base level.

If we like this, I'll also go include it to the `ModelFilter` before merging.

cc @sashavor

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -38,6 +38,7 @@ from .utils.endpoint_helpers import (
     DatasetTags,
     ModelFilter,
     ModelTags,
+    _filter_emissions,
 )
 
 
@@ -513,6 +514,7 @@ class HfApi:
         filter: Union[ModelFilter, str, Iterable[str], None] = None,
         author: Optional[str] = None,
         search: Optional[str] = None,
+        emissions_threshold: Optional[Tuple[float, float]] = None,
         sort: Union[Literal["lastModified"], str, None] = None,
         direction: Optional[Literal[-1]] = None,
         limit: Optional[int] = None,
@@ -581,6 +583,9 @@ class HfApi:
                     >>> #List all models with "bert" in their name made by google
                     >>> api.list_models(search="bert", author="google")
 
+            emissions_threshold (:obj:`Tuple`, `optional`):
+                A tuple of two ints or floats representing a minimum and maximum carbon footprint
+                to filter the resulting models with
             sort (:obj:`Literal["lastModified"]` or :obj:`str`, `optional`):
                 The key with which to sort the resulting models. Possible values are the properties of the `ModelInfo`
                 class.
@@ -644,7 +649,10 @@ class HfApi:
         r = requests.get(path, params=params, headers=headers)
         r.raise_for_status()
         d = r.json()
-        return [ModelInfo(**x) for x in d]
+        res = [ModelInfo(**x) for x in d]
+        if emissions_threshold is not None:
+            return _filter_emissions(res, *emissions_threshold)
+        return res
 
     def _unpack_model_filter(self, model_filter: ModelFilter):
         """
--- a/src/huggingface_hub/utils/endpoint_helpers.py
+++ b/src/huggingface_hub/utils/endpoint_helpers.py
@@ -14,8 +14,53 @@ Helpful utility functions and classes in relation to exploring API endpoints
 with the aim for a user-friendly interface
 """
 
+import math
+import re
 from dataclasses import dataclass
 from typing import List, Union
+
+
+def _filter_emissions(
+    models,
+    minimum_threshold: Union[int, float] = None,
+    maximum_threshold: Union[int, float] = None,
+):
+    """Filters a list of models for those that include an emission tag
+    and limit them to between two thresholds
+
+    Args:
+        models (:obj:`ModelInfo` or :class:`List`):
+            A list of `ModelInfo`'s to filter by
+        minimum_threshold (:obj:`int` or :obj:`float`):
+            A minimum carbon threshold to filter by, such as 1.
+        maximum_threshold (:obj:`int` or :obj:`float):
+            A maximum carbon threshold to filter by, such as 10.
+    """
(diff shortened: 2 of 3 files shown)


sashavor:
This is cool! Next, we need to add more models with emission data! :seedling:

nateraw:
For `transformers` support out of the box, should we add `co2_eq_emissions` to `TrainingSummary`? Can be done automatically via `TrainingSummary.from_trainer` if `codecarbon` is installed. WDYT?

julien-c:
@nateraw yes i think it would be great (and we had brainstormed this some time ago i think)

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -14,10 +14,55 @@
...
+        minimum_threshold (:obj:`int` or :obj:`float`):
+            A minimum carbon threshold to filter by, such as 1.
+        maximum_threshold (:obj:`int` or :obj:`float):
+            A maximum carbon threshold to filter by, such as 10.
+    """
+    if minimum_threshold is None and maximum_threshold is None:
+        raise ValueError(
+            "Both `minimum_threshold` and `maximum_threshold` cannot both be `None`"
+        )
+    if not minimum_threshold:
```
LysandreJik, line 43: (nit) this will be `False` if `minimum_treshold` is set to `0`, should it be  `if minimum_treshold is not None` ?

LysandreJik (approved):
Cool addition!

src/huggingface_hub/utils/endpoint_helpers.py
muellerzr: Makes sense. Went with
```
if minimum_threshold is None:
```
(As this sets `minimum_threshold` to be a default value)

src/huggingface_hub/hf_api.py
```diff
@@ -650,8 +658,14 @@ def list_models(
...
         d = r.json()
         res = [ModelInfo(**x) for x in d]
-        if emissions_threshold is not None:
-            return _filter_emissions(res, *emissions_threshold)
+        if emissions_thresholds is not None:
+            if cardData is None:
+                logging.error(
+                    "Warning! `emissions_threshold` passed without setting `cardData=True`. No filtering can be performed"
+                )
+                return res
```
LysandreJik, line 665: Cool error! I wonder if it wouldn't be even better to raise in that case?

muellerzr: Agreed, makes sense. Will toss it in and then merge :smile:

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -14,10 +14,55 @@
...
+    models,
+    minimum_threshold: Union[int, float] = None,
+    maximum_threshold: Union[int, float] = None,
+):
+    """Filters a list of models for those that include an emission tag
+    and limit them to between two thresholds
+
+    Args:
+        models (:obj:`ModelInfo` or :class:`List`):
+            A list of `ModelInfo`'s to filter by
```
osanseviero, line 33:
```suggestion
            A list of `ModelInfo`'s to filter by.
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -14,10 +14,55 @@
...
+            "Both `minimum_threshold` and `maximum_threshold` cannot both be `None`"
+        )
+    if minimum_threshold is None:
+        minimum_threshold = -1
+    if maximum_threshold is None:
+        maximum_threshold = math.inf
+    emissions = []
+    for i, model in enumerate(models):
+        if hasattr(model, "cardData"):
+            if isinstance(model.cardData, dict):
```
osanseviero, line 50: Is this type check really necessary?

src/huggingface_hub/hf_api.py
```diff
@@ -644,7 +656,15 @@ def list_models(
...
         d = r.json()
-        return [ModelInfo(**x) for x in d]
+        res = [ModelInfo(**x) for x in d]
+        if emissions_thresholds is not None:
+            if cardData is None:
+                raise ValueError(
+                    "`emissions_thresholds` were passed without setting `cardData=True`."
+                )
+            else:
+                return _filter_emissions(res, *emissions_thresholds)
```
osanseviero, line 666: Should this return the model indices of the models? Or should it return the list results? (I feel that we want the list results, since this is essentially a filter).

osanseviero (approved):

osanseviero:
I left a couple of comments and approved since @LysandreJik is ok with this. I think we should try to keep consistency in the results of `list_models`.

src/huggingface_hub/utils/endpoint_helpers.py
muellerzr: For right now there's some bad `cardData` stored, so yeah. (And we don't have guards in place for this yet)

src/huggingface_hub/hf_api.py
muellerzr: My vote is the list results for similar reasons

muellerzr:
As we think about standardizing this more, it would be good to add checks on what we would consider "valid carbon emissions data".

E.g., since for right now users can more or less upload what we want, in some cases we have a number, in some cases we have a number and the letters "Kg", and in worst cases we have instances of whole dictionaries:
```python
{'emissions': '28.53 in grams of CO2', 'source': 'calculated using this webstie https://mlco2.github.io/impact/#compute', 'training_type': 'pretraining', 'geographical_location': 'NA', 'hardware_used': 'TPUv3-8 for about 180 hours or 7.5 days'}
```

It would be good if we could stick to one format, as for right now there's not a clean way to test the occurrence of nested dictionaries cleanly. (and why the `_filter_emissions` func is an onslaught of nested if's). But that's out of the scope of this little pr :) 

@julien-c @nateraw @sashavor

julien-c:
yep i think @sashavor's "spec" was that the unit is grams, no? https://huggingface.co/docs/hub/model-repos#carbon-footprint-metadata

we can probably enforce some validation (that this property is a number) hub-side at some point

PR #671, opened 2022-02-08
Upgrade black to version ~=22.0

LysandreJik:
Upgrades the `huggingface_hub` library quality tool `black` to use the first stable release of `black`, version 22.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ extras["testing"] = [
 ]
 
 extras["quality"] = [
-    "black==21.4b0",
+    "black~=22.0",
     "isort>=5.5.4",
     "flake8>=3.8.3",
 ]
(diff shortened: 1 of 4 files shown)


muellerzr:
We should also update the [setup.py](https://github.com/huggingface/huggingface_hub/blob/main/setup.py#L41) so it trickles into the CI/CD

setup.py
```diff
@@ -38,7 +38,7 @@ def get_version() -> str:
 ]
 
 extras["quality"] = [
-    "black==21.4b0",
+    "black~=22.0",
```
LysandreJik, line 41: It is updated no?

muellerzr: Yes it is, sorry I glanced over that! 

Peeking at it more, *eventually* this CI/CD may fail due to us not pinning the versions of black like we do with the setup.py https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-api-quality.yaml#L19

muellerzr (approved):

osanseviero (approved):
Thanks for the PR! 🔥 

cc people working in the repo @merveenoyan @nateraw @Narsil so you update your black version as well

Issue #672, opened 2022-02-08
Repository does not work with Spaces


osanseviero:
If I create a space and then do

```python
repo = Repository(local_dir=local_dir, clone_from="https://huggingface.co/space/osanseviero/this-is-a-test12345", repo_type="space")
```

I get following error

ValueError: No space_sdk provided. `create_repo` expects space_sdk to be one of ['gradio', 'streamlit', 'static'] when repo_type is 'space'`

cc @nateraw @LysandreJik @muellerzr

nateraw:
Did you mean `repo = Repository(local_dir=local_dir, clone_from="https://huggingface.co/spaces/osanseviero/this-is-a-test12345", repo_type="space")` with "spaces" in the URL? (I haven't tried, just a thought)

osanseviero:
Actually with spaces it also fails.

I tried to debug a bit more. It fails when I'm the owner of the space and works when it's another unrelated space

osanseviero:
This is the src https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/repository.py#L545-L552

muellerzr:
Nice catch. this is because `Repository` won't actually take in a `space_sdk` during it's init, so nothing get's passed when performing https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L1081

Pretty sure the solution would be to give `Repository` a `space_sdk` param that can be passed down? WDYT @LysandreJik

osanseviero:
I don't love the idea because it's annoying when combined with `create_repo` or when cloning an existing space created through the UI. We're expecting the user to specify the sdk both in `create_repo` and again in `Repository`.

One solution could be to:
* Check before internal `create_repo` call if the repo exists. If it exists, let's not call it.
* Add a `space_sdk` param as well, but now it won't be obligatory for all spaces

julien-c:
isn't it weird API-wise that `Repository`'s init creates a repo for you on the fly if needed?

Shouldn't it be a different method to better separate intents, like a static `Repository.create(...) -> Repository`?

julien-c:
what's `huggingface_hub` policy on breaking changes? we can maybe keep BC for a few months then drop it?

LysandreJik:
[...]

PR #674, opened 2022-02-08
Improve Repository handling of spaces

muellerzr:
Closes https://github.com/huggingface/huggingface_hub/issues/672

<s>Only thing I might need some help with is staging the tests, as this was a silent failure. Right now it err's out with `InvalidToken`. Is there something different I should be doing? TIA</s>

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -8,12 +8,12 @@ import time
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
-from urllib.error import HTTPError
 
 from tqdm.auto import tqdm
 
 from huggingface_hub.constants import REPO_TYPES_URL_PREFIXES, REPOCARD_NAME
 from huggingface_hub.repocard import metadata_load, metadata_save
+from requests.exceptions import HTTPError
 
 from .hf_api import ENDPOINT, HfApi, HfFolder, repo_type_and_id_from_hf_id
 from .lfs import LFS_MULTIPART_UPLOAD_COMMAND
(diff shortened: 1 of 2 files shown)


muellerzr:
Sorry for the noise, un-drafting. Auto-import was wrong and made all the tests fail. 

Let me know if we *shouldn't* warn the user that we're creating a new repo

src/huggingface_hub/repository.py
```diff
@@ -543,14 +556,21 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
...
+                        )
+                        api.create_repo(
+                            repo_id,
+                            token=token,
+                            repo_type=self.repo_type,
+                            organization=namespace,
+                            exist_ok=True,
+                            private=self.private,
+                            space_sdk=space_sdk,
+                        )
```
LysandreJik, line 573: Not super enthusiastic about that part with the double API call but I understand it's needed for spaces we want to clone when no `space_sdk` is passed.

src/huggingface_hub/repository.py
```diff
@@ -543,14 +556,21 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
...
+                    except HTTPError:
+                        logger.warning(
+                            f"Repo {namespace}/{repo_id} does not exist, creating a new repo"
+                        )
+                        api.create_repo(
+                            repo_id,
+                            token=token,
+                            repo_type=self.repo_type,
+                            organization=namespace,
+                            exist_ok=True,
```
LysandreJik, line 570: I guess this can be removed now

LysandreJik (approved):
Thanks for working on it!

LysandreJik:
Closing for same reason as https://github.com/huggingface/huggingface_hub/pull/852

PR #676, opened 2022-02-09
Ensure `post_method` is only executed once

sgugger:
Currently, the `post_method` of a `CommandInProgress` is executed each time the flag `is_done` is accessed. This can yield to errors as seen in the Transformers CI (with some LFS pruning attempted in folders that don't exist anymore).

This PR destroys the `post_method` once executed, which ensures it is only executed once.


LysandreJik (approved):
Great catch :) Thanks for fixing!

PR #678, opened 2022-02-10
Add fastai mixin

omarespejel:
Sorry, I created a new PR (previous one: [PR #416](https://github.com/huggingface/huggingface_hub/pull/416)) I had a couple of mistakes with git, and seemed easier to create a new one. Thanks @muellerzr for your comments!

## What
- Add fastai upstream and downstream capabilities **for versions `fastai>=2.4`** ([link](https://github.com/fastai/fastai/releases/tag/2.4)) and `fastcore>=1.3.27` ([link](https://github.com/fastai/fastcore/releases/tag/1.3.27)).  
- Inform users that **lower versions of fastai are not supported *yet*** (TODO -- Examine whether it is worth implementing previous versions).

## Why
- **Supporting version [fastai1](https://github.com/fastai/fastai1) might not be worth it.** Loading and pushing `Learners` involves several changes and the fastai1 library has not been updated for over a year.
- **Supporting versions `2.0.6>= fastai <2.4`  involve complex changes**.  `Fastai` was updated due to what appear to be modifications to [ pytorch/serialization.py](https://github.com/pytorch/pytorch/blob/master/torch/serialization.py#L989). For the sake of agility in our first release, I suggest releasing this version without supporting these previous versions.
- **`fastai>=2.4` versions work with `fastcore>=1.3.27` version** which is installed automatically when `fastai>=2.4` is installed.

## How?
- Following huggingface_hub practices, `file_download.py` checks for fastcore and fastai availability and versions.
- The fastai and fastcore versions used to train the Learner are **automatically stored in a `config.json` when pushed to hub.**
- A `README.md` is automatically generated, if none, with the **`fastai` tag.**

## Testing?
- Tested that all `fastai>=2.4` versions work with the present code and with the `fastcore==1.3.27` version. 

## Next?
- **Examine whether it is worth implementing `fastai <2.4` versions.** I will ask on the fastai forums to assess how many users would require this support.
- **Update description on how to load fastai models** in [Libraries.ts](https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Libraries.ts).

### DIFF
--- a/src/huggingface_hub/fastai_mixin.py
+++ b/src/huggingface_hub/fastai_mixin.py
@@ -1,0 +1,227 @@
+import json
+import logging
+import os
+from pathlib import Path
+from typing import Any, Dict, Optional, Union
+
+logger = logging.getLogger(__name__)
+
+try:
+  from fastai.learner import load_learner
+except ImportError as error:
+        logger.error(error.__class__.__name__ + ": fastai version above or equal to 2.5.1 required. Run, for example, `pip install fastai==2.5.1`.")
+
+
+from huggingface_hub import ModelHubMixin
+from huggingface_hub.constants import CONFIG_NAME
+# TODO - add to huggingface_hub.constants the constant FASTAI_LEARNER_NAME: same name to all the .pkl models pushed to the hub.
+from huggingface_hub.hf_api import HfFolder, HfApi
+from huggingface_hub.repository import Repository
+from huggingface_hub.snapshot_download import snapshot_download
+
+
+logger = logging.getLogger(__name__)
+
+
+def save_fastai_learner(
+    learner, save_directory: str, config: Optional[Dict[str, Any]] = None
+):
+    """Saves a fastai learner to save_directory in pickle format. Use this if you're using Learners.
+
+    learner:
+        The `fastai.learner` you'd like to save.
+    save_directory (:obj:`str`):
+        Specify directory in which you want to save the fastai learner.
+    config (:obj:`dict`, `optional`):
+        Configuration object with the labels of the model. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
+    
+    TODO - Save weights and model structure instead of the built learner.
+    """
+
+    # creating path
+    os.makedirs(save_directory, exist_ok=True)
+
+    # saving config
+    if config:
+        if not isinstance(config, dict):
+            raise RuntimeError(
+                f"Provided config should be a dict. Got: '{type(config)}'"
+            )
+        path = os.path.join(save_directory, CONFIG_NAME)
+        with open(path, "w") as f:
+            json.dump(config, f)
+    
+    # saving learner
+    learner.export(os.path.join(save_directory, 'model.pkl'))
+
+
+def from_pretrained_fastai(*args, **kwargs):
+    return FastaiModelHubMixin.from_pretrained(*args, **kwargs)
+
+
+def push_to_hub_fastai(
+    learner,
+    repo_path_or_name: Optional[str] = None,
+    repo_url: Optional[str] = None,
+    commit_message: Optional[str] = "Add model",
+    organization: Optional[str] = None,
+    private: Optional[bool] = None,
+    api_endpoint: Optional[str] = None,
+    use_auth_token: Optional[Union[bool, str]] = True,
+    git_user: Optional[str] = None,
+    git_email: Optional[str] = None,
+    config: Optional[dict] = None,
+):
+    """
+    Upload learner checkpoint files to the 🤗 Model Hub while synchronizing a local clone of the repo in
+    :obj:`repo_path_or_name`.
+
+    Parameters:
+        model:
+            The `fastai.learner' you'd like to push to the hub. 
+        repo_path_or_name (:obj:`str`, `optional`):
+            Can either be a repository name for your model or tokenizer in the Hub or a path to a local folder (in
+            which case the repository will have the name of that local folder). If not specified, will default to
+            the name given by :obj:`repo_url` and a local directory with that name will be created.
+        repo_url (:obj:`str`, `optional`):
+            Specify this in case you want to push to an existing repository in the hub. If unspecified, a new
+            repository will be created in your namespace (unless you specify an :obj:`organization`) with
+            :obj:`repo_name`.
+        commit_message (:obj:`str`, `optional`):
+            Message to commit while pushing. Will default to :obj:`"add model"`.
+        organization (:obj:`str`, `optional`):
+            Organization in which you want to push your model or tokenizer (you must be a member of this
+            organization).
+        private (:obj:`bool`, `optional`):
+            Whether or not the repository created should be private (requires a paying subscription).
+        api_endpoint (:obj:`str`, `optional`):
+            The API endpoint to use when pushing the model to the hub.
+        use_auth_token (:obj:`bool` or :obj:`str`, `optional`):
+            The token to use as HTTP bearer authorization for remote files. If :obj:`True`, will use the token
+            generated when running :obj:`transformers-cli login` (stored in :obj:`~/.huggingface`). Will default to
+            :obj:`True`.
+        git_user (``str``, `optional`):
+            will override the ``git config user.name`` for committing and pushing files to the hub.
+        git_email (``str``, `optional`):
+            will override the ``git config user.email`` for committing and pushing files to the hub.
+        config (:obj:`dict`, `optional`):
+            Configuration object to be saved alongside the model weights.
+
+    Returns:
+        The url of the commit of your model in the given repository.
+    """
+
+    if repo_path_or_name is None and repo_url is None:
+        raise ValueError("You need to specify a `repo_path_or_name` or a `repo_url`.")
+
+    if isinstance(use_auth_token, bool) and use_auth_token:
+        token = HfFolder.get_token()
+    elif isinstance(use_auth_token, str):
+        token = use_auth_token
+    else:
+        token = None
+
+    if token is None:
+        raise ValueError(
+            "You must login to the Hugging Face hub on this computer by typing `huggingface-cli login` and "
+            "entering your credentials to use `use_auth_token=True`. Alternatively, you can pass your own "
+            "token as the `use_auth_token` argument."
+        )
+
+    if repo_path_or_name is None:
+        repo_path_or_name = repo_url.split("/")[-1]
+
+    # If no URL is passed and there's no path to a directory containing files, create a repo
+    if repo_url is None and not os.path.exists(repo_path_or_name):
+        repo_name = Path(repo_path_or_name).name
+        repo_url = HfApi(endpoint=api_endpoint).create_repo(
+            token,
+            repo_name,
+            organization=organization,
+            private=private,
+            repo_type=None,
+            exist_ok=True,
+        )
+
+    repo = Repository(
+        repo_path_or_name,
+        clone_from=repo_url,
+        use_auth_token=use_auth_token,
+        git_user=git_user,
+        git_email=git_email,
+    )
+    repo.git_pull(rebase=True)
+
+    save_fastai_learner(learner, repo_path_or_name, config=config)
+
+    # Commit and push!
+    repo.git_add(auto_lfs_track=True)
+    repo.git_commit(commit_message)
+    return repo.git_push()
+
+
+
+
+class FastaiModelHubMixin(ModelHubMixin):
+    def __init__(self, *args, **kwargs):
+        """
+        Mixin class to implement model download and upload from fastai learners.
+        
+        # Downloading Learner from hf-hub:
+        Example::
+
+            >>> from huggingface_hub import from_pretrained_fastai
+            >>> model = from_pretrained_fastai("username/mymodel@main")
+
+        # TODO - Define if proceeding with a class (FastaiModelHubMixin) would be ideal for fastai and proceed with implementation
+        # otherwise, proceed with just functions.
+
+        """
+
+    def _save_pretrained(self, save_directory):
+        save_fastai_learner(self, save_directory)
+
+    @classmethod
+    def _from_pretrained(
+        cls,
+        model_id,
+        revision,
+        cache_dir,
+        force_download,
+        proxies,
+        resume_download,
+        local_files_only,
+        use_auth_token,
+        **model_kwargs,
+    ):
+        """Here we just call save_fastai_learner function so both the mixin and functional APIs stay in sync.
+
+        TODO - Some args above aren't used since we are calling snapshot_download instead of hf_hub_download.
+        """
+
+        # TODO - Figure out what to do about these config values. Config is not going to be needed to load model
+        cfg = model_kwargs.pop("config", None)
+
+        # Root is either a local filepath matching model_id or a cached snapshot
+        if not os.path.isdir(model_id):
+            storage_folder = snapshot_download(
+                repo_id=model_id, revision=revision, cache_dir=cache_dir
+            )
+        else:
+            storage_folder = model_id
+
+        # Using the pickle document in the downloaded list
+        docs = os.listdir(storage_folder)
+        for doc in docs:
+          if (doc.endswith('.pkl')): 
+            pickle = doc
+            break
+        
+        logger.info(f'Using `fastai.learner` stored in {os.path.join(model_id, pickle)}.')
+
+        model = load_learner(os.path.join(storage_folder, pickle))        
+
+        # For now, we add a new attribute, config, to store the config loaded from the hub/a local dir.
+        model.config = cfg
+
+        return model
--- a/src/huggingface_hub/fastai_mixin.py
+++ b/src/huggingface_hub/fastai_mixin.py
@@ -1,55 +1,101 @@
 ###########################################################################################################
-# Easily store and download fastai working models into the HF Hub.
+# Easily store and download `fastai>=2.4` models into the HF Hub.
 #
 # Goal:
 # (1) Add upstream support: push a fastai learner to the HF Hub. See `save_fastai_learner` and `push_to_hub_fastai`.
 # (2) Add downstream support: download a fastai learner from the hub. See `from_pretrained_fastai`.
 #
 # Limitations and next steps:
-# - Go from storing/downloading a `fastai.learner` to saving the weights directly into de hub.
+# - Possibly go from storing/downloading a `fastai.learner` to saving the weights directly into de hub.
+# - Examine whether it is worth implementing `fastai <2.4` versions.
 ###########################################################################################################
 
 import json
 import logging
 import os
-import sys
 import packaging.version
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
-# TODO - The following code to verify fastai version would be better in huggingface_hub/file_download.py.
-_PY_VERSION: str = sys.version.split()[0].rstrip("+")
-if packaging.version.Version(_PY_VERSION) < packaging.version.Version("3.8.0"):
-    import importlib_metadata
-else:
-    import importlib.metadata as importlib_metadata
+# TODO - add to huggingface_hub.constants the constant FASTAI_LEARNER_NAME: same name to all the .pkl models pushed to the hub.
+from huggingface_hub import ModelHubMixin
+from huggingface_hub.constants import CONFIG_NAME
+
+from huggingface_hub.file_download import (
+    get_fastai_version,
+    get_fastcore_version,
+)
+from huggingface_hub.hf_api import HfFolder, HfApi
+from huggingface_hub.repository import Repository
+from huggingface_hub.snapshot_download import snapshot_download
 
 # Verify if we are using the right fastai version.
-_FASTAI_VERSION: str = importlib_metadata.version("fastai")
-if packaging.version.Version(_FASTAI_VERSION) < packaging.version.Version("2.5.0"):
+if packaging.version.Version(get_fastai_version()) < packaging.version.Version("2.4"):
     raise ImportError(
-        f"You are using fastai version {_FASTAI_VERSION} which is below ^2.5.0. Run, for example, `pip install fastai==2.5.1`."
+        f"`push_to_hub_fastai` and `from_pretrained_fastai` require a fastai>=2.4 version, but you are using fastai version {get_fastai_version()} which is incompatible. Run, for example, `pip install fastai==2.5.1`."
+    )
+
+# Verify if we are using the right fastcore version.
+if packaging.version.Version(get_fastcore_version()) < packaging.version.Version(
+    "1.3.27"
+):
+    raise ImportError(
+        f"`push_to_hub_fastai` and `from_pretrained_fastai` require a fastcore>=1.3.27 version, but you are using fastcore version {get_fastcore_version()} which is incompatible. Run, for example, `pip install fastcore==1.3.27`."
     )
 
 logger = logging.getLogger(__name__)
 
-# TODO - evaluate relevance of the following check.
-# Second check to verify we are using the right fastai version.
+# Verify availability of `load_learner`.
 try:
     from fastai.learner import load_learner
 except ImportError as error:
     logger.error(
         error.__class__.__name__
-        + ": fastai version above or equal to 2.5.1 required. Run, for example, `pip install fastai==2.5.1`."
+        + f": `push_to_hub_fastai` and `from_pretrained_fastai` require a fastai>=2.4 version, but you are using fastai version {get_fastai_version()} which is incompatible. Run, for example, `pip install fastai==2.5.1`."
     )
 
-from huggingface_hub import ModelHubMixin
-from huggingface_hub.constants import CONFIG_NAME
+# Define template for a auto-generated README.md
+README_TEMPLATE = """---
+tags:
+- fastai
+---
 
-# TODO - add to huggingface_hub.constants the constant FASTAI_LEARNER_NAME: same name to all the .pkl models pushed to the hub.
-from huggingface_hub.hf_api import HfFolder, HfApi
-from huggingface_hub.repository import Repository
-from huggingface_hub.snapshot_download import snapshot_download
+# Amazing! 
+
+Congratulations on hosting your fastai model on the 🤗Hub!
+
+# Some next steps
+1. Fill out this model card with more information ([documentation here](https://huggingface.co/docs/hub/model-repos))!
+
+2. Create a demo in Gradio or Streamlit using the 🤗Spaces ([documentation here](https://huggingface.co/docs/hub/spaces)).
+
+3. Join our fastai community on the Hugging Face Discord!
+
+Greetings fellow fastlearner 🤝!
+
+"""
+
+# Define template for a auto-generated config with fastai and fastcore versions
+CONFIG_TEMPLATE = dict(
+    fastai_version=get_fastai_version(), fastcore_version=get_fastcore_version()
+)
+
+
+def _create_model_card(repo_dir: Path):
+    """Creates a model card for the repository.
+
+    repo_dir:
+        Specify directory in which you want to create a model card.
+    """
+    readme_path = repo_dir / "README.md"
+    readme = ""
+    if readme_path.exists():
+        with readme_path.open("r", encoding="utf8") as f:
+            readme = f.read()
+    else:
+        readme = README_TEMPLATE
+    with readme_path.open("w", encoding="utf-8") as f:
+        f.write(readme)
 
 
 def save_fastai_learner(
@@ -62,7 +108,7 @@ def save_fastai_learner(
     save_directory (:obj:`str`):
         Specify directory in which you want to save the fastai learner.
     config (:obj:`dict`, `optional`):
-        Configuration object with the labels of the model. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
+        Configuration object. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
 
     TODO - Save weights and model structure instead of the built learner.
     """
@@ -71,6 +117,7 @@ def save_fastai_learner(
     os.makedirs(save_directory, exist_ok=True)
 
     # saving config
+    # if user provides config then we update it with the fastai and fastcore versions in CONFIG_TEMPLATE.
     if config:
         if not isinstance(config, dict):
             raise RuntimeError(
@@ -78,7 +125,14 @@ def save_fastai_learner(
             )
         path = os.path.join(save_directory, CONFIG_NAME)
         with open(path, "w") as f:
-            json.dump(config, f)
+            json.dump({**config, **CONFIG_TEMPLATE}, f)
+    else:
+        path = os.path.join(save_directory, CONFIG_NAME)
+        with open(path, "w") as f:
+            json.dump(CONFIG_TEMPLATE, f)
+
+    # creating README.md if none exist
+    _create_model_card(Path(save_directory))
 
     # saving learner
     learner.export(os.path.join(save_directory, "model.pkl"))
@@ -181,7 +235,10 @@ def push_to_hub_fastai(
     )
     repo.git_pull(rebase=True)
 
-    save_fastai_learner(learner, repo_path_or_name, config=config)
+    if config:
+        save_fastai_learner(learner, repo_path_or_name, config=config)
+    else:
+        save_fastai_learner(learner, repo_path_or_name)
 
     # Commit and push!
     repo.git_add(auto_lfs_track=True)
@@ -247,6 +304,7 @@ class FastaiModelHubMixin(ModelHubMixin):
         logger.info(
             f"Using `fastai.learner` stored in {os.path.join(model_id, pickle)}."
         )
+        print(f"Using `fastai.learner` stored in {os.path.join(model_id, pickle)}.")
 
         model = load_learner(os.path.join(storage_folder, pickle))
 
--- a/src/huggingface_hub/fastai_mixin.py
+++ b/src/huggingface_hub/fastai_mixin.py
@@ -1,26 +1,44 @@
 import json
 import logging
 import os
+import sys
+import packaging.version
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
+# TODO - The following code to verify fastai version would be better in huggingface_hub/file_download.py.
+_PY_VERSION: str = sys.version.split()[0].rstrip("+")
+if packaging.version.Version(_PY_VERSION) < packaging.version.Version("3.8.0"):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
+
+# Verify if we are using the right fastai version.
+_FASTAI_VERSION: str = importlib_metadata.version("fastai")
+if packaging.version.Version(_FASTAI_VERSION) < packaging.version.Version("2.5.0"):
+    raise ImportError(
+        f"You are using fastai version {_FASTAI_VERSION} which is below ^2.5.0. Run, for example, `pip install fastai==2.5.1`."
+    )
+
 logger = logging.getLogger(__name__)
 
+# TODO - evaluate relevance of the following check.
+# Second check to verify we are using the right fastai version.
 try:
-  from fastai.learner import load_learner
+    from fastai.learner import load_learner
 except ImportError as error:
-        logger.error(error.__class__.__name__ + ": fastai version above or equal to 2.5.1 required. Run, for example, `pip install fastai==2.5.1`.")
-
+    logger.error(
+        error.__class__.__name__
+        + ": fastai version above or equal to 2.5.1 required. Run, for example, `pip install fastai==2.5.1`."
+    )
 
 from huggingface_hub import ModelHubMixin
 from huggingface_hub.constants import CONFIG_NAME
+
 # TODO - add to huggingface_hub.constants the constant FASTAI_LEARNER_NAME: same name to all the .pkl models pushed to the hub.
 from huggingface_hub.hf_api import HfFolder, HfApi
 from huggingface_hub.repository import Repository
 from huggingface_hub.snapshot_download import snapshot_download
-
-
-logger = logging.getLogger(__name__)
 
 
 def save_fastai_learner(
@@ -34,7 +52,7 @@ def save_fastai_learner(
         Specify directory in which you want to save the fastai learner.
     config (:obj:`dict`, `optional`):
         Configuration object with the labels of the model. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
-    
+
     TODO - Save weights and model structure instead of the built learner.
     """
 
@@ -50,9 +68,9 @@ def save_fastai_learner(
         path = os.path.join(save_directory, CONFIG_NAME)
         with open(path, "w") as f:
             json.dump(config, f)
-    
+
     # saving learner
-    learner.export(os.path.join(save_directory, 'model.pkl'))
+    learner.export(os.path.join(save_directory, "model.pkl"))
 
 
 def from_pretrained_fastai(*args, **kwargs):
@@ -78,7 +96,7 @@ def push_to_hub_fastai(
 
     Parameters:
         model:
-            The `fastai.learner' you'd like to push to the hub. 
+            The `fastai.learner' you'd like to push to the hub.
         repo_path_or_name (:obj:`str`, `optional`):
             Can either be a repository name for your model or tokenizer in the Hub or a path to a local folder (in
             which case the repository will have the name of that local folder). If not specified, will default to
@@ -160,13 +178,11 @@ def push_to_hub_fastai(
     return repo.git_push()
 
 
-
-
 class FastaiModelHubMixin(ModelHubMixin):
     def __init__(self, *args, **kwargs):
         """
         Mixin class to implement model download and upload from fastai learners.
-        
+
         # Downloading Learner from hf-hub:
         Example::
 
@@ -213,13 +229,15 @@ class FastaiModelHubMixin(ModelHubMixin):
         # Using the pickle document in the downloaded list
         docs = os.listdir(storage_folder)
         for doc in docs:
-          if (doc.endswith('.pkl')): 
-            pickle = doc
-            break
-        
-        logger.info(f'Using `fastai.learner` stored in {os.path.join(model_id, pickle)}.')
+            if doc.endswith(".pkl"):
+                pickle = doc
+                break
 
-        model = load_learner(os.path.join(storage_folder, pickle))        
+        logger.info(
+            f"Using `fastai.learner` stored in {os.path.join(model_id, pickle)}."
+        )
+
+        model = load_learner(os.path.join(storage_folder, pickle))
 
         # For now, we add a new attribute, config, to store the config loaded from the hub/a local dir.
         model.config = cfg
--- a/src/huggingface_hub/fastai_mixin.py
+++ b/src/huggingface_hub/fastai_mixin.py
@@ -1,3 +1,14 @@
+###########################################################################################################
+# Easily store and download fastai working models into the HF Hub.
+#
+# Goal:
+# (1) Add upstream support: push a fastai learner to the HF Hub. See `save_fastai_learner` and `push_to_hub_fastai`.
+# (2) Add downstream support: download a fastai learner from the hub. See `from_pretrained_fastai`.
+#
+# Limitations and next steps:
+# - Go from storing/downloading a `fastai.learner` to saving the weights directly into de hub.
+###########################################################################################################
+
 import json
 import logging
 import os
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -69,6 +69,22 @@ for package_name in _tf_candidates:
     except importlib_metadata.PackageNotFoundError:
         pass
 
+_fastai_version = "N/A"
+_fastai_available = False
+try:
+    _fastai_version: str = importlib_metadata.version("fastai")
+    _fastai_available = True
+except importlib_metadata.PackageNotFoundError:
+    pass
+
+_fastcore_version = "N/A"
+_fastcore_available = False
+try:
+    _fastcore_version: str = importlib_metadata.version("fastcore")
+    _fastcore_available = True
+except importlib_metadata.PackageNotFoundError:
+    pass
+
 
 def is_torch_available():
     return _torch_available
@@ -76,6 +92,22 @@ def is_torch_available():
 
 def is_tf_available():
     return _tf_available
+
+
+def is_fastai_available():
+    return _fastai_available
+
+
+def get_fastai_version():
+    return _fastai_version
+
+
+def is_fastcore_available():
+    return _fastcore_available
+
+
+def get_fastcore_version():
+    return _fastcore_version
 
 
 def hf_hub_url(
@@ -180,6 +212,10 @@ def http_user_agent(
         ua += "; torch/{}".format(_torch_version)
     if is_tf_available():
         ua += "; tensorflow/{}".format(_tf_version)
+    if is_fastai_available():
+        ua += "; fastai/{}".format(_fastai_version)
+    if is_fastcore_available():
+        ua += "; fastcore/{}".format(_fastcore_version)
     if isinstance(user_agent, dict):
         ua += "; " + "; ".join("{}/{}".format(k, v) for k, v in user_agent.items())
     elif isinstance(user_agent, str):


src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    else:
+        path = os.path.join(save_directory, CONFIG_NAME)
+        with open(path, "w") as f:
+            json.dump(CONFIG_TEMPLATE, f)
+
+    # creating README.md if none exist
+    _create_model_card(Path(save_directory))
+
+    # saving learner
+    learner.export(os.path.join(save_directory, "model.pkl"))
```
muellerzr, line 138: I think here it might be important to also expose the `pickle_protocol` here for folks, as I've also had to check that on occasion. Maybe even better to pass in `**kwargs` instead for folks to use that gets passed to `learn.export`?

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+def from_pretrained_fastai(*args, **kwargs):
+    return FastaiModelHubMixin.from_pretrained(*args, **kwargs)
+
+
+def push_to_hub_fastai(
+    learner,
```
muellerzr, line 146: Add a typing of `Learner` here perhaps? (Which would also need the import of `Learner`)

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    git_email: Optional[str] = None,
+    config: Optional[dict] = None,
+):
+    """
+    Upload learner checkpoint files to the 🤗 Model Hub while synchronizing a local clone of the repo in
+    :obj:`repo_path_or_name`.
+
+    Parameters:
+        model:
+            The `fastai.learner' you'd like to push to the hub.
```
muellerzr, line 164: Should be `fastai.Learner` (with a capitol L)
```suggestion
            The `fastai.Learner' you'd like to push to the hub.
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+        repo_path_or_name (:obj:`str`, `optional`):
+            Can either be a repository name for your model or tokenizer in the Hub or a path to a local folder (in
```
muellerzr, line 166: I'm not 100% sure stating `tokenizer` in the docstring makes sense here

muellerzr (changes requested):
Very nice! I just left a few nits :)

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+# Define template for a auto-generated README.md
+README_TEMPLATE = """---
+tags:
+- fastai
+---
+
+# Amazing! 
+
+Congratulations on hosting your fastai model on the 🤗Hub!
```
osanseviero, line 65:
```suggestion
Congratulations on hosting your fastai model on the Hugging Face Hub!
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+# Some next steps
+1. Fill out this model card with more information ([documentation here](https://huggingface.co/docs/hub/model-repos))!
+
+2. Create a demo in Gradio or Streamlit using the 🤗Spaces ([documentation here](https://huggingface.co/docs/hub/spaces)).
+
+3. Join our fastai community on the Hugging Face Discord!
+
+Greetings fellow fastlearner 🤝!
```
osanseviero, line 74: Consider adding placeholder sections such as in https://huggingface.co/merve/model-card-toggle-test4. Something like this

[](https://huggingface.co/merve/model-card-toggle-test4#intended-uses--limitations)[](https://huggingface.co/merve/model-card-toggle-test4#training-and-evaluation-data)Model description
More information needed

Intended uses & limitations
More information needed

Training and evaluation data
More information needed

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+    # saving learner
+    learner.export(os.path.join(save_directory, "model.pkl"))
+
+
+def from_pretrained_fastai(*args, **kwargs):
+    return FastaiModelHubMixin.from_pretrained(*args, **kwargs)
+
+
+def push_to_hub_fastai(
```
osanseviero, line 145: I would consider with a simpler interface as discussed in https://github.com/huggingface/huggingface_hub/issues/652

```
model,
repo_id: str,
use_auth_token: Optional[Union[bool, str]] = True,
```

src/huggingface_hub/fastai_mixin.py
osanseviero, line 164:
```suggestion
            The `fastai.learner' you'd like to push to the Hub.
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    if isinstance(use_auth_token, bool) and use_auth_token:
+        token = HfFolder.get_token()
+    elif isinstance(use_auth_token, str):
+        token = use_auth_token
+    else:
+        token = None
+
+    if token is None:
+        raise ValueError(
+            "You must login to the Hugging Face hub on this computer by typing `huggingface-cli login` and "
```
osanseviero, line 209:
```suggestion
            "You must login to the Hugging Face Hub on this computer by typing `huggingface-cli login` and "
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    """Saves a fastai learner to save_directory in pickle format. Use this if you're using Learners.
+
+    learner:
+        The `fastai.learner` you'd like to save.
+    save_directory (:obj:`str`):
+        Specify directory in which you want to save the fastai learner.
+    config (:obj:`dict`, `optional`):
+        Configuration object. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
+
+    TODO - Save weights and model structure instead of the built learner.
```
osanseviero, line 113: Instead of leaving TODOs in code, considering opening issues with label and assignee

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+
+    # creating path
```
osanseviero, line 116: I think you can remove this comment

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    os.makedirs(save_directory, exist_ok=True)
+
+    # saving config
+    # if user provides config then we update it with the fastai and fastcore versions in CONFIG_TEMPLATE.
+    if config:
```
osanseviero, line 121:
```suggestion
    if config is not None:
```

src/huggingface_hub/fastai_mixin.py
osanseviero: Is this possible without forcing the dependency?

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+    organization: Optional[str] = None,
+    private: Optional[bool] = None,
+    api_endpoint: Optional[str] = None,
+    use_auth_token: Optional[Union[bool, str]] = True,
+    git_user: Optional[str] = None,
+    git_email: Optional[str] = None,
+    config: Optional[dict] = None,
+):
+    """
+    Upload learner checkpoint files to the 🤗 Model Hub while synchronizing a local clone of the repo in
```
osanseviero, line 159:
```suggestion
    Upload learner checkpoint files to the Hub while synchronizing a local clone of the repo in
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+        use_auth_token=use_auth_token,
+        git_user=git_user,
+        git_email=git_email,
+    )
+    repo.git_pull(rebase=True)
+
+    if config:
+        save_fastai_learner(learner, repo_path_or_name, config=config)
+    else:
+        save_fastai_learner(learner, repo_path_or_name)
```
osanseviero, line 241: Can't you just do this instead? If config=None it will have same effect

```suggestion
save_fastai_learner(learner, repo_path_or_name, config=config)
```

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+    # Commit and push!
+    repo.git_add(auto_lfs_track=True)
+    repo.git_commit(commit_message)
+    return repo.git_push()
```
osanseviero, line 246: You can also use `repo.push_to_hub` instead

src/huggingface_hub/fastai_mixin.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class FastaiModelHubMixin(ModelHubMixin):
+    def __init__(self, *args, **kwargs):
+        """
+        Mixin class to implement model download and upload from fastai learners.
+
+        # Downloading Learner from hf-hub:
```
osanseviero, line 254:
```suggestion
        # Downloading Learner from the Hugging Face Hub:
```

osanseviero (commented):
Thank you very much for the PR! There are some important things to address, but overall is going in right direction 🚀

src/huggingface_hub/fastai_mixin.py
muellerzr: Not from what I could see sadly, so we can x this one

src/huggingface_hub/fastai_mixin.py
omarespejel: Great idea thanks! I still liked the proposed format since we can redirect model creators to our discord and create spaces. WDYT about adding the placeholder sections you mention below? In the last commit I did this way:

README_TEMPLATE = 
"""---
tags:
- fastai
---
# Amazing!
Congratulations on hosting your fastai model on the Hugging Face Hub!
# Some next steps
1. Fill out this model card with more information (template below and [documentation here](https://huggingface.co/docs/hub/model-repos))!
2. Create a demo in Gradio or Streamlit using the 🤗Spaces ([documentation here](https://huggingface.co/docs/hub/spaces)).
3. Join our fastai community on the Hugging Face Discord!
Greetings fellow fastlearner 🤝!
---
# Model card
## Model description
More information needed
## Intended uses & limitations
More information needed
## Training and evaluation data
More information needed
"""

omarespejel:
Hi, @osanseviero @muellerzr thank you for your kind reviews guys! All feedback was applied and tested. 

Main changes:
- **Changed name**. From `fastai_mixin.py` to `fastai_utils.py`.
- **No longer a mixin**. Eliminated `FastaiModelHubMixin` class and only leaved three main public functions: `from_pretrained_fastai`, `push_to_hub_fastai`, and `save_fastai_learner`. This helps with simplicity.
- **Imported `fastai_utils.py` public functions** in `huggingface_hub/__init__.py`.
- **Failing test**. Build (3.6) and Build (3.9) actions failing only due to the `test_inference_with_audio` test. Will review the reason.
- **Created `check_fastai_fastcore_versions` function** to not force fastai imports from the beggining.

Will be here :D

osanseviero:
The failing tests are unrelated. They are flaky and likely not related to this PR, so don't worry about it

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+    **kwargs,
+):
+    """
+    Saves a fastai learner to save_directory in pickle format. Use this if you're using Learners.
+
+    Parameters:
+        learner (:obj:`Learner`):
+            The `fastai.Learner` you'd like to save.
+        save_directory (:obj:`str`):
+            Specify directory in which you want to save the fastai learner.
```
osanseviero, line 133:
```suggestion
            Specific directory in which you want to save the fastai learner.
```

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+        config (:obj:`dict`, `optional`):
+            Configuration object. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
+
+    Keyword Parameters:
+        pickle_protocol (:obj:`int`, `optional`):
```
osanseviero, line 138: Why not pass this as a normal parameter? This is similar to what is done with `include_optimizer` in Keras https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/keras_mixin.py#L24

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+        model_id (:obj:`str`):
+            The model id where the pickled fastai.Learner is. Example: 'espejelomar/fastai-pet-breeds-classification'.
+
+    Keyword Parameters:
+        config (:obj:`dict`, `optional`):
+            Configuration object.
+        revision (:obj:`str`, `optional`):
+            Revision at which the repo's files are downloaded. See documentation of `snapshot_download`.
+        cache_dir (:obj:`str`, `Path`, `optional`):
+            Path to cache directory.
```
osanseviero, line 191: Same here. I would use normal parameters since it's significantly more explicit.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+    else:
+        storage_folder = model_id
+
+    # Using the pickle document in the downloaded list
+    docs = os.listdir(storage_folder)
+    for doc in docs:
+        if doc.endswith(".pkl"):
+            pickle = doc
+            break
```
osanseviero, line 220: We usually push for same naming convention. Since your save models with name `model.pkl`, you can just assume the model name is `model.pkl` here as well.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+    logger.info(f"Using `fastai.Learner` stored in {os.path.join(model_id, pickle)}.")
+    print(f"Using `fastai.Learner` stored in {os.path.join(model_id, pickle)}.")
```
osanseviero, line 222: Should this be removed?

osanseviero (commented):
Looking good! I think it's in a good state to start adding some tests :smile:

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,337 @@
...
+    model = load_learner(os.path.join(storage_folder, pickle))
+    model.config = config
+    return model
+
+
+def push_to_hub_fastai(
+    learner,
```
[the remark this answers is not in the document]
omarespejel: Sorry I forgot to apply the `repo_id` part in the last commit.

omarespejel:
Hi guys! Thanks again for the kind reviews. 

Relevant changes:
- `push_to_hub_fastai` was simplified and now accepts organization or username in `repo_id` (it was not possible before).
- Fixed a bug in `save_fastai_learner` that would make `learner.export ` unable to update cloned models.

IMO this could be MVProduct. Will continue testing in `fastai` models (available [here](https://huggingface.co/fastai)). And go on engaging the fastai discord community :)

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    # creating README.md if none exist
+    _create_model_card(Path(save_directory))
+
+    # saving learner
+    # learner.export saves the model in `self.path/save_directory` and this folder should exist.
+    # We create this folder and call it `fastai_path`
+    fastai_path = os.path.join(learner.path, save_directory)
+    os.makedirs(fastai_path, exist_ok=True)
+    learner.export(
+        fname=os.path.join(save_directory, "model.pkl"), pickle_protocol=pickle_protocol
```
osanseviero, line 168: Is specifying different pickle protocols something that will be needed @muellerzr ?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+
+    # We move the model from `self.path/save_directory/model.pkl` to `save_directory/model.pkl`.
+    os.rename(
```
osanseviero, line 172: Why is this needed? Isn't the model directly saved in `save_directory`?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    # Import `load_learner` from `fastai.learner`.
+    from fastai.learner import load_learner
+
+    # Root is either a local filepath matching model_id or a cached snapshot
+    if not os.path.isdir(model_id):
+        storage_folder = snapshot_download(
+            repo_id=model_id, revision=revision, cache_dir=cache_dir
+        )
+    else:
+        storage_folder = model_id
```
osanseviero, line 212: Why not always using?

```
storage_folder = snapshot_download(
      repo_id=model_id, revision=revision, cache_dir=cache_dir
  )
```

`from_pretrained_fastai` should be to load models from the HUb (or cached), not any local model imo

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Parameters:
+        model_id (:obj:`str`):
+            The model id where the pickled fastai.Learner is. Example: 'espejelomar/fastai-pet-breeds-classification'.
+        config (:obj:`dict`, `optional`):
+            Configuration object.
+        revision (:obj:`str`, `optional`):
+            Revision at which the repo's files are downloaded. See documentation of `snapshot_download`.
+        cache_dir (:obj:`str`, `Path`, `optional`):
+            Path to cache directory.
```
osanseviero, line 195: Maybe we can just use the default here? This will yield a simpler interface

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    revision: Optional[str] = None,
+    cache_dir: Optional[Union[str, Path]] = None,
+):
+    """
+    Load `model_id` files from the Hub.
+
+    Parameters:
+        model_id (:obj:`str`):
+            The model id where the pickled fastai.Learner is. Example: 'espejelomar/fastai-pet-breeds-classification'.
+        config (:obj:`dict`, `optional`):
```
osanseviero, line 190: What's the use case of passing a config here?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Returns:
+        The url of the commit of your model in the given repository.
+    """
+
+    # Check that fastai and fastcore versions are supported.
+    check_fastai_fastcore_versions()
+
+    # Unpacking **kwargs
+    organization: str = kwargs.get("organization", None)
```
osanseviero, line 271: This should be removed I think

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+    Upload learner checkpoint files to the Hub while synchronizing a local clone of the repo in
+    :obj:`repo_id`.
+
+    Parameters:
+        learner (:obj:`Learner`):
+            The `fastai.Learner' you'd like to push to the Hub.
+        repo_id (:obj:`str`):
+            Can either be a repository name for your model in the Hub or a path to a local folder (in
+            which case the repository will have the name of that local folder). The user can be either your individual account (e.g. 'espejelomar/sentece-embeddings-BETO') or an organization you have write access to (e.g. 'stanfordnlp/stanza-de').
```
osanseviero, line 241: One idea for simplifying the interface is that `repo_id` will just be a repo name of the model in the Hub. I don't think supporting local folder is an important use case, and it's more intuitive.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    organization: str = kwargs.get("organization", None)
+    api_endpoint: str = kwargs.get("api_endpoint", None)
+    git_user: str = kwargs.get("git_user", None)
+    git_email: str = kwargs.get("git_email", None)
+    pickle_protocol: int = kwargs.get("pickle_protocol", 2)
+
+    # Split `repo_id` into organization/user and repo
+    temp = repo_id.split("/")
+    organization = temp[0]
+    repo_name = temp[1]
```
osanseviero, line 280:
```suggestion
    organization, repo_name = repo_id.split("/")
```

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+    repo_url = None
+    if not os.path.exists(repo_id):
+        repo_url = HfApi(endpoint=api_endpoint).create_repo(
+            repo_name,
+            token=token,
+            organization=organization,
+            private=private,
+            repo_type=None,
+            exist_ok=True,
+        )
```
osanseviero, line 307: Since we are always referring to a remote repo, it can be 
```suggestion
    repo_url = HfApi(endpoint=api_endpoint).create_repo(
            repo_name,
            token=token,
            organization=organization,
            private=private,
            repo_type=None,
            exist_ok=True,
        )
```

osanseviero (commented):
This is looking good! I left a couple of questions to understand some decisions and some suggestions for simplifying them even further the interface. 

It would be great if you can add tests, that will allow you to find any bugs and makes it easier to review the PR

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,324 @@
...
+            Revision at which the repo's files are downloaded. See documentation of `snapshot_download`.
+        cache_dir (:obj:`str`, `Path`, `optional`):
+            Path to cache directory.
+
+    Returns:
+        The `fastai.Learner` model in the `model_id` repo.
+    """
+    # Check that fastai and fastcore versions are supported.
+    check_fastai_fastcore_versions()
+
```
osanseviero, line 202: Shouldn't this also load `config.json` and check if the user version matches and give a warning if not?

src/huggingface_hub/fastai_utils.py
muellerzr: Agreed not 100% sure, so would like an explanation :)

src/huggingface_hub/fastai_utils.py
muellerzr: Thinking on it more, it *shouldn't* be. (Talked with Jeremy about this a moment ago as well). Reason being if not passed in, it should pull the default for Python.

This is still a fix that needs to be done in fastai, so proposed solution is to set it to `None` when being passed to `export` so python takes over.

muellerzr: Or, even better: `pickle.DEFAULT_PROTOCOL`

omarespejel: Will use pickle.DEFAULT_PROTOCOL! Thanks for the suggetion :D

src/huggingface_hub/fastai_utils.py
omarespejel: +1 Zach

src/huggingface_hub/fastai_utils.py
omarespejel: Removed it. TBH it was not necessary. My mistake. 

Was something I was carrying from the first version of this PR and was inspired by keras_mixin.py (https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/keras_mixin.py#L262).

src/huggingface_hub/fastai_utils.py
osanseviero: Also as discussed we likely want to move to toml

osanseviero:
> we store the requirements in pyproject.toml here, how is it used? Is there another PR handling that issue?

Imo what we should do **when loading** with `from_pretrained_fastai` is check if versions are different. If yes, we can throw an error (or warning), and give the command for users to install dependencies themselves. I would go with a simple solution for now.

adrinjalali:
On the client side it's not too hard, my question was more on the server side. How do we decide which docker image to use to load the model? Or are we planning to install dependencies before loading each model?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,403 @@
...
+    model = load_learner(os.path.join(storage_folder, "model.pkl"))
+    return model
+
+
+def push_to_hub_fastai(
+    learner,
+    repo_id: str,
+    commit_message: Optional[str] = "Add model",
+    private: Optional[bool] = None,
+    use_auth_token: Optional[Union[bool, str]] = True,
```
osanseviero, line 317: Thinking about this, there is no scenario in which you can push without a token. Instead, I would change this parameter to be `token: Optional[str] = None`. 

Then, you can do 

```python 
if token is None:
  token = HfFolder.get_token()

if token is None:
    raise ValueError(
        "You must login to the Hugging Face Hub. There are two options:"
        "(1) Type `huggingface-cli login` in your terminal and enter your token."
        "(2) Enter your token in the `token` argument."
        "Your token is available in the Settings of your Hugging Face account."
    )
```

omarespejel: Very elegant solution! This simplifies quite a lot!

omarespejel:
Hi guys. Last changes:
- Add tests for `from_pretrained_fastai`, `push_to_hub_fastai`, `save_fastai_learner`. 
- Dear friend @muellerzr, how can you compare two Learners to assert they are the same? I am using `learn.show_training_loop()` but this is faulty. This is for testing that a pushed to the hub Learner is the same when I load it.
- Also @muellerzr, what is a way to get the weights of a Learner? I have not been successful looking for this. Thanks! :D
- Add in `hugging_face/setup.py` the import of `toml` which is required in `fastai_utils.py`. Build tests likely failing because of this.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,389 @@
+import json
+import os
+from pathlib import Path
+from pickle import DEFAULT_PROTOCOL
+from typing import Any, Dict, Optional
+
+import packaging.version
+
+import toml
```
osanseviero, line 9: We should likely move `toml` to `check_fastai_fastcore_pyproject_versions` which is the only place where it's used. And give a nice warning!

**Note**: You've probably identified a pattern by now. For  soft dependencies, we don't import them at the top of the file,  but where they are used, and provide clear error message for users

setup.py
```diff
@@ -23,10 +23,18 @@ def get_version() -> str:
 
 extras = {}
 
+extras["fastai"] = [
+    "toml",
+]
+
```
osanseviero, line 29: This is duplicated :)

setup.py
```diff
@@ -23,10 +23,18 @@ def get_version() -> str:
...
 extras["torch"] = [
     "torch",
 ]
 
+extras["fastai"] = [
+    "toml",
```
osanseviero, line 35: I would also add `fastai` here

tests/test_fastai_integration.py
```diff
@@ -0,0 +1,107 @@
+import os
```
osanseviero, line 1: As of now, these tests are not being executed. You will need to add these tests to the GitHub Actions workflow. You can base this from https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-tests.yml#L69-L83 in this same file

osanseviero (commented):
I left some suggestions to fix the GH workflow and run your tests in it.

Also, @omarespejel, do we really need to load the weights and compare the two learners. I think if you're able to check that a model can be saved and loaded that should be good to go. (maybe I'm missing something).

muellerzr:
@osanseviero and @omarespejel this is how I'd recommend checking learners because we need to ensure two things line up, model weights and the transform pipeline. (First is real code, second is pseudocode OOTOMH):

The model itself lives in `learn.model`. So save/load the weights and make sure that all the params in `learn.model.parameters()` align, etc. (Look at saving/loading in Transformers or Accelerate to see how we can do this).

For the preprocessing, we should make sure all the transforms in `learn.dls.after_batch` and `learn.dls.after_item` are the exact same. Not *too* sure how that can be pulled off easily pragmatically, but if you can't quite get there lmk Omar and we could pair program that to figure it out

omarespejel:
Hi, guys thanks for your reviews! 

- @osanseviero, the tests currently check that a model can be saved, pushed to the hub, and loaded. All passed :smile:. The only test not passing comes from `test_keras_integration.py`. 
- @muellerzr, it makes sense what you explain, thanks! I'll open an issue on this and I'll assign it to me. I will inform you of news about it. Thank you very much! :muscle: 

What do you think would be missing for a successful merge?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,396 @@
...
+    """
+    Creates a `pyproject.toml` for the repository.
+
+    Parameters:
+        repo_dir (:obj:`Path`):
+            Directory where `pyproject.toml` is created.
+    """
+    pyproject_path = repo_dir / "pyproject.toml"
+    pyproject = ""
+    if pyproject_path.exists():
```
osanseviero, line 212: Same as above

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,396 @@
...
+    readme_path = repo_dir / "README.md"
+    readme = ""
+    if readme_path.exists():
+        with readme_path.open("r", encoding="utf8") as f:
+            readme = f.read()
+    else:
+        readme = README_TEMPLATE
+    with readme_path.open("w", encoding="utf-8") as f:
+        f.write(readme)
+
```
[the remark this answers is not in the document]
omarespejel: Lol, this is true!! My mistake.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,396 @@
...
+            The minimum fastcore version supported.
+
+    Raises:
+        FileNotFoundError, ImportError
+
+    """
+
+    try:
+        import toml
+    except IndexError:
```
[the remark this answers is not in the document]
omarespejel: You are right. It was `ModuleNotFoundError`. Thanks!

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,396 @@
...
+            ]
+        )
+    except IndexError:
+        raise ImportError(
+            "The repository does not have a fastcore version specified in `pyproject.toml`."
+        )
+
+    # Get the fastai and fastcore versions in `pyproject.toml`. If there is none, it means that it defaults to the highest version.
+    fastai_version = fastai_version.partition("=")[2]
+    fastcore_version = fastcore_version.partition("=")[2]
```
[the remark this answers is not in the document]
omarespejel: TL;DR: Partition is easier to use for this particular case than split.

I chose `partition` because of the particular case when the `pyproject.toml` only indicate `fastai` (without a version like `fastai=2.0.0`). In this case, `partition` would output an empty value for position [2] of the tuple (example: `('fastai', '', '')`). If we had something like `fastai=2.0.0` then the position [2] of the tuple would be the version: `('fastai', '=', '2.0.0')`. In other words, the tuple always has three values where the postion [2] indicate the version (empty or not).

This is not the case for `split` where the output can vary in length: `['fastai', '2.0.0']` when there is an explicit version and `['fastai']` when there is not. The case you propose (`fastai_version.split("=")[1]`) would output an error when no version is explicitly mentioned because there is no element [1].

omarespejel:
Hi team! Just wanted to push this forward 🚀.

@nateraw or @muellerzr, when you have an opportunity, could you please review the 🤗 fastai integration?

Please let me know if you need anything.

adrinjalali:
Should this also add documentation on how to add `fastai` models to the hub?

osanseviero:
The docs should likely be in the end-users/library side, see:

* https://spacy.io/api/cli#huggingface-hub
* https://stable-baselines3.readthedocs.io/en/master/guide/integrations.html#hugging-face

Plus accompanying announcement (e.g. blog post, etc) from our side. 

I liked the ideas you mentioned at https://github.com/huggingface/huggingface_hub/issues/754#issuecomment-1066788626, but for the sake of not increasing the scope of this PR even more, I would go ahead with this. We could add this + all other integrations to a new special section of the `huggingface-hub` library docs once we have it.

omarespejel:
> The docs should likely be in the end-users/library side, see:
> 
> * https://spacy.io/api/cli#huggingface-hub
> * https://stable-baselines3.readthedocs.io/en/master/guide/integrations.html#hugging-face
> 
> Plus accompanying announcement (e.g. blog post, etc) from our side

I'll see how I can get organized with the fastai maintainers to add this to their documentation. This is part of the distribution process :smiley:.

omarespejel:
Hi! How about doing an alpha test? I can invite part of the community interested in using the integration to upload fastai models. Iterate from that. 

WDYT @osanseviero @muellerzr @nateraw

.github/workflows/python-tests.yml
```diff
@@ -83,6 +83,24 @@ jobs:
         pip install .[testing,tensorflow]
 
     - run: pytest -sv ./tests/test_keras*
+  
+  build_fastai:
+    runs-on: ubuntu-latest
+
```
osanseviero, line 89: You will need to add 

```
strategy:
      matrix:
        python-version: ["3.6", "3.9"]
```

as it was broken (and fixed) https://github.com/huggingface/huggingface_hub/commit/7c236b177759e523f6227367b33ac3f936b19638

osanseviero:
From my side it looks good and we can go ahead with testing this. If @nateraw could take a quick look at this it would be great, and we can merge afterwards

LysandreJik:
Removing this PR from the v0.5 milestone as version v0.5 will be released in a bit.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+    # Import `load_learner` from `fastai.learner`.
+    from fastai.learner import load_learner
+
+    model = load_learner(os.path.join(storage_folder, "model.pkl"))
+    return model
+
+
+def push_to_hub_fastai(
+    learner,
+    repo_id: str,
```
osanseviero, line 311: Here it's `repo_id` but above it's `model_id`. Let's make them consistent

.github/workflows/python-tests.yml
```diff
@@ -90,6 +90,27 @@ jobs:
         pip install .[testing,tensorflow]
 
     - run: pytest -Werror::FutureWarning -sv ./tests/test_keras*
+  
+  build_fastai:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.9"]
```
adrinjalali, line 98: does fastai not support 3.10?
```suggestion
        python-version: ["3.7", "3.10"]
```

setup.py
```diff
@@ -27,6 +27,12 @@ def get_version() -> str:
...
+    "fastai==2.5.3",
+    "fastcore==1.3.29",
```
adrinjalali, line 33: should these be pinned? Aren't we supporting multiple versions now?

src/huggingface_hub/__init__.py
```diff
@@ -30,6 +30,11 @@
     TF2_WEIGHTS_NAME,
     TF_WEIGHTS_NAME,
 )
+from .fastai_utils import (
+    from_pretrained_fastai,
+    push_to_hub_fastai,
+    save_fastai_learner,
+)
```
adrinjalali, line 37: WDYT of moving integrations out of the root namespace? Having them all in the root namespace makes `import huggingface_hub` pretty heavy.

I would suggest something like:

```
huggingface_hub.core
huggingface_hub.fastai
...
```

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+            The minimum fastcore version supported.
+
+    Raises:
+        :class: `FileNotFoundError`: if the there is no `pyproject.toml` in the repository that contains the fastai `Learner`.
+
+        :class: `ImportError`: if the `toml` module is not installed.
+
+        :class: `ImportError`: if the `pyproject.toml` does not indicate a version for fastai or fastcore.
+
+        :class: `ImportError`: if the `pyproject.toml` indicates a lower than minimum supported version of fastai or fastcore.
```
adrinjalali, line 80: do these work with `doc-build`?

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+            "There is no `pyproject.toml` in the repository that contains the fastai `Learner`. This is necessary to verify that your fastai and fastcore versions are compatible with those of the model you want to load."
+        )
+
+    # Check that `pyproject.toml` contains versions for fastai and fastcore. If there is no version available, it throws an error
+    # Then get the fastai and fastcore versions in `pyproject.toml`. If there is none, it means that it defaults to the highest version.
+    try:
+        fastai_version = str(
+            [package for package in package_versions if package.startswith("fastai")][0]
+        )
+        fastai_version = fastai_version.partition("=")[2]
```
adrinjalali, line 106: should we be parsing them using `packaging` and `packaging.markers` instead? It would also allow for more flexible version specifications.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+    Saves a fastai learner to `save_directory` in pickle format using the default pickle protocol for the version of python used.
+    Apply this if you're using Learners.
+
+    Params:
+        learner (:obj:`Learner`):
+            The `fastai.Learner` you'd like to save.
+        save_directory (:obj:`str`):
+            Specific directory in which you want to save the fastai learner.
+        config (:obj:`dict`, `optional`):
+            Configuration object. Will be uploaded as a .json file. Example: 'https://huggingface.co/espejelomar/fastai-pet-breeds-classification/blob/main/config.json'.
```
adrinjalali, line 231: we need to check if this is rendered properly and as a hyperlink. I don't see these docs in the rendered version.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+    # learner.export saves the model in `self.path/save_directory` and this folder should exist.
+    # We create this folder and call it `fastai_path`
+    fastai_path = os.path.join(learner.path, save_directory)
+    os.makedirs(fastai_path, exist_ok=True)
+    learner.export(
+        fname=os.path.join(save_directory, "model.pkl"),
+        pickle_protocol=DEFAULT_PROTOCOL,
+    )
+
+    # We move the model from `self.path/save_directory/model.pkl` to `save_directory/model.pkl`.
```
adrinjalali, line 268: should we move or copy? I'd be surprised as a user if this moves my files around.

adrinjalali (commented):
Thanks for the work @omarespejel . Left a few comments, the doc related ones also apply to other docstrings here.

@osanseviero WDYT of moving integrations to their own namespace, and them not touching `file_download.py`?

osanseviero:
> @osanseviero WDYT of moving integrations to their own namespace, and them not touching file_download.py?

In one of the comments/threads above, with @LysandreJik we were talking about moving all of this to a single file in a follow-up PR. As for namespacing, I think usually in all our libraries we have all of them at top namespace and it works well. I would go with the path that is more aligned/consistent with the rest of HF libraries. WDYT @LysandreJik?

```py
_PY_VERSION: str = sys.version.split()[0].rstrip("+")
if packaging.version.Version(_PY_VERSION) < packaging.version.Version("3.8.0"):
    import importlib_metadata
else:
    import importlib.metadata as importlib_metadata
_torch_version = "N/A"
_torch_available = False
try:
    _torch_version = importlib_metadata.version("torch")
    _torch_available = True
except importlib_metadata.PackageNotFoundError:
    pass
_pydot_available = False
try:
    _pydot_version = importlib_metadata.version("pydot")
    _pydot_available = True
except importlib_metadata.PackageNotFoundError:
    pass
def is_pydot_available():
    return _pydot_available
_graphviz_available = False
try:
    _graphviz_version = importlib_metadata.version("graphviz")
    _graphviz_available = True
except importlib_metadata.PackageNotFoundError:
    pass
def is_graphviz_available():
    return _graphviz_available
_tf_version = "N/A"
_tf_available = False
_tf_candidates = (
    "tensorflow",
    "tensorflow-cpu",
    "tensorflow-gpu",
    "tf-nightly",
    "tf-nightly-cpu",
    "tf-nightly-gpu",
    "intel-tensorflow",
    "intel-tensorflow-avx512",
    "tensorflow-rocm",
    "tensorflow-macos",
)
for package_name in _tf_candidates:
    try:
        _tf_version = importlib_metadata.version(package_name)
        _tf_available = True
        break
    except importlib_metadata.PackageNotFoundError:
        pass

_fastai_version = "N/A"
_fastai_available = False
try:
    _fastai_version: str = importlib_metadata.version("fastai")
    _fastai_available = True
except importlib_metadata.PackageNotFoundError:
    pass

_fastcore_version = "N/A"
_fastcore_available = False
try:
    _fastcore_version: str = importlib_metadata.version("fastcore")
    _fastcore_available = True
except importlib_metadata.PackageNotFoundError:
    pass


def is_torch_available():
    return _torch_available
def is_tf_available():
    return _tf_available


def is_fastai_available():
    return _fastai_available


def get_fastai_version():
    return _fastai_version


def is_fastcore_available():
    return _fastcore_available


def get_fastcore_version():
    return _fastcore_version
```

LysandreJik:
> I would go with the path that is more aligned/consistent with the rest of HF libraries. 

I would do that too, and this approach works very well with `transformers` which handles a very large amount of optional dependencies, so the system works and scales.

tests/test_fastai_integration.py
```diff
@@ -0,0 +1,123 @@
...
+        x = torch.randn(bs * n, 1)
+        return torch.cat((x, a * x + b + 0.1 * torch.randn(bs * n, 1)), 1)
+
+    ds = get_data(n)
+    dblock = DataBlock()
+    return dblock.dataloaders(ds)
+
+
+if is_fastai_available():
+    DummyModel = synth_learner(data=fake_dataloaders())
```
nateraw, line 64: If this `DummyModel` is an instance of an object, can we please name without capital letters? 😄 

`dummy_model` would be fine imo

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # creating pyproject.toml if none exist
+    _create_model_pyproject(Path(save_directory))
+
+    # saving learner
+    # learner.export saves the model in `self.path/save_directory` and this folder should exist.
+    # We create this folder and call it `fastai_path`
+    fastai_path = os.path.join(learner.path, save_directory)
+    os.makedirs(fastai_path, exist_ok=True)
+    learner.export(
```
nateraw, line 263: Note here that `lambda`s don't get saved if they were used in the dataloader, etc. Ran into this when running through the examples found [here](https://github.com/fastai/fastai/blob/master/nbs/examples/app_examples.ipynb). Is there anything we should consider doing around this? For example, catching these errors and providing guidance on how to fix them?

tests/test_fastai_integration.py
```diff
@@ -0,0 +1,123 @@
...
+            config=dummy_config,
+        )
+        model_info = self._api.model_info(
+            f"{USER}/{REPO_NAME}",
+        )
+        self.assertEqual(model_info.modelId, f"{USER}/{REPO_NAME}")
+
+        loaded_model = from_pretrained_fastai(f"{USER}/{REPO_NAME}")
+        self.assertEqual(
+            DummyModel.show_training_loop(), loaded_model.show_training_loop()
```
nateraw, line 121: Not familiar enough with fastai - does this prove the weights are the same?

.github/workflows/python-tests.yml
nateraw: If we run this on a matrix of versions, that means 3 runners will run, eh? Is this even all that necessary? I see most the other ones seem to run just on 3.8.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+        )
+        fastai_version = fastai_version.partition("=")[2]
+    except IndexError:
+        raise ImportError(
+            "The repository does not have a fastai version specified in `pyproject.toml`."
+        )
+    try:
+        fastcore_version = str(
+            [package for package in package_versions if package.startswith("fastcore")][
+                0
```
nateraw, line 114: Strange spacing here imo.

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # We move the model from `self.path/save_directory/model.pkl` to `save_directory/model.pkl`.
+    os.rename(
+        os.path.join(fastai_path, "model.pkl"),
+        os.path.join(save_directory, "model.pkl"),
+    )
+
+
+def from_pretrained_fastai(
+    model_id: str,
```
nateraw, line 276: Does this not work on local files? I believe the keras + pytorch integrations both do, which is why I suggest this one should too.

nateraw (approved):
Hey there! So I went through this, and it seems to work fine for me: [Open In Colab](https://colab.research.google.com/gist/nateraw/8a5fdf45d43fba593677d286ff3b71c3/test-fastai-integration.ipynb)

Outside of a couple of nits, this LGTM. 🚀

tests/test_fastai_integration.py
omarespejel: It is not enough indeed. We have currently issue #771 to make this comparison. It is currently out of the scope of this PR.

src/huggingface_hub/fastai_utils.py
omarespejel: Great discovery thanks! I catched the error and provided guidance to the user 🤗

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    if current_fastcore_version < packaging.version.Version(fastcore_min_version):
+        raise ImportError(
+            f"`push_to_hub_fastai` and `from_pretrained_fastai` require a fastcore>={fastcore_min_version} version, but you are using fastcore version {get_fastcore_version()} which is incompatible. Upgrade with `pip install fastcore==1.3.27`."
+        )
+
+
+def check_fastai_fastcore_pyproject_versions(
+    storage_folder: str,
+    fastai_min_version: Optional[str] = "2.4",
```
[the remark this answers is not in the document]
omarespejel: Done. I made the versions requirements more flexible in `setup.py`. Now they are: 
```
extras["fastai"] = [
    "toml",
    "fastai>=2.4",
    "fastcore>=1.3.27",
]
```

.github/workflows/python-tests.yml
osanseviero: I think running in different py versions is important here, specially as the pickling is different. Torch and TF integrations test 3.7 and 3.10, so I would do same here
https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-tests.yml#L55

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,415 @@
...
+    </Tip>
+    """
+
+    try:
+        import toml
+    except ModuleNotFoundError:
+        raise ImportError(
+            "`push_to_hub_fastai` and `from_pretrained_fastai` require the toml module. Install it with `pip install toml`."
+        )
+
```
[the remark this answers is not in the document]
omarespejel: It seems to me it is already there ([link to line](https://github.com/huggingface/huggingface_hub/pull/678/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R31)).
image

src/huggingface_hub/fastai_utils.py
omarespejel: Thanks Adrin, I tried this but have some troubles, and was taking too long. Will review it soon again. Meanwhile, I registered it in issue #848.

src/huggingface_hub/fastai_utils.py
omarespejel: Thanks again Adrin, this is simpler. Will review it soon again. Meanwhile, I registered it in issue #848.

osanseviero: I agree we should most likely copy instead, if possible in this PR

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,412 @@
...
+    Args:
+        learner (`Learner`):
+            The `fastai.Learner' you'd like to push to the Hub.
+        model_id (`str`):
+            The name of the repository for your model in the Hub. The user can be your individual account (e.g. 'espejelomar/sentece-embeddings-BETO') or an organization to which you have write access (e.g. 'stanfordnlp/stanza-de').
+        commit_message (`str`, *optional*):
+            Message to commit while pushing. Will default to :obj:`"add model"`.
+        private (`bool`, *optional*):
+            Whether or not the repository created should be private (requires a paying subscription).
```
osanseviero, line 347:
```suggestion
            Whether or not the repository created should be private.
```

osanseviero (approved):
Cool :rocket: Thanks for iterating on this

src/huggingface_hub/fastai_utils.py
```diff
@@ -0,0 +1,412 @@
...
+        package_versions = toml.load(f"{storage_folder}/pyproject.toml")[
+            "build-system"
+        ]["requires"]
+    except FileNotFoundError:
+        raise FileNotFoundError(
+            "There is no `pyproject.toml` in the repository that contains the fastai `Learner`. This is necessary to verify that your fastai and fastcore versions are compatible with those of the model you want to load."
+        )
+
+    # Check that `pyproject.toml` contains fastai and fastcore. If they are not available, it throws an error
+    # Then get the fastai and fastcore versions in `pyproject.toml`. If there is none, it means that it defaults to the highest version.
```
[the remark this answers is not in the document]
omarespejel: Currently, we are taking into account that if the package is specified but not the version (e.g. "fastai", the default versions are the highest. Thus, they would be accepted.

However, I believe this comment should have been in another section of the code. I moved and reword it to [here](https://github.com/huggingface/huggingface_hub/pull/678/files#diff-21cb622665e96594b1761f0db9fa7465e605d4e2a12d8e1c080c5de597738423R133
).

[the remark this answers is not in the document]
osanseviero: SG!

osanseviero (approved):

omarespejel:
Thank you @osanseviero @adrinjalali @muellerzr @nateraw! For your reviews and feedback 🤗🥳

Issue #628, opened 2022-01-27
Model cards for Keras models


merveenoyan:
Keras models (non-Transformers ones) don't have generated model cards.  I'd like to generate model cards for them that include:
- model hyperparameters
- metrics from training
- name of dataset and splits (can be filled by user later on as well) 
- versions of libraries (like TF model cards)

merveenoyan:
I think it would be super cool to actually have [KerasTuner](https://keras.io/keras_tuner/) outputs in the model cards as well (output of `results_summary()`)

merveenoyan:
I think it would be a big flex if we would write `model.summary()` output. This doesn't work for models with subclassing AFAIK, we can get it for Functional and Sequential models. This tends to be a large section but we can put it under toggle (pointed out by @osanseviero [in spaCy models](https://huggingface.co/spacy/fr_dep_news_trf))
The rest of the parameters I will get from model. 
We can later introduce history object (created with `history = model.fit()` that contains model training history even if the user doesn't log loss, accuracy etc, we could introduce this as another parameter in `push_to_hub_keras()` since this is something we can't obtain from the model but we need user to create history while fitting the model. 

I want to cc @Rocketknight1 since he's experienced with the TF model card.

merveenoyan:
@Rocketknight1 it currently looks very shifted when I save it with IO so I'm plotting the model instead :( 
<img width="477" alt="Ekran Resmi 2022-02-08 13 14 02" src="https://user-images.githubusercontent.com/53175384/153030055-53533d16-cabe-4ae4-abec-d38ce25c922f.png">

PR #679, opened 2022-02-10
Keras model card

merveenoyan:
This PR creates automatically generated model cards for Keras models, requested in #628 
The model card contains:
- a model plot (I first added summary then swapped it with the plot itself because it was looking bad)
- Model hyperparameters like in TF transformers 
- Model metrics like in TF transformers
- Additional sections for limitations, intended use etc. (see below)
- if user wants to push a specific task, I added another argument to `push_to_hub_keras()` called task_name

Ekran Resmi 2022-02-10 10 56 49
Ekran Resmi 2022-02-10 10 57 01


merveenoyan:
TF tests are failing because `tf.keras.utils.plot_model()` requires a package called pydot.
see docs: 

> Converts a Keras model to dot format and save to a file.

Is it worth having given it's another dependency? @osanseviero

src/huggingface_hub/keras_mixin.py
```diff
@@ -59,9 +173,13 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
+    hyperparameters = extract_hyperparameters_from_keras(model)
+    plot_network(model, save_directory)
```
osanseviero, line 177: Maybe let's have something similar to `is_tf_available` but for pydot. We can also add it as a test dependency.

src/huggingface_hub/keras_mixin.py
```diff
@@ -16,6 +16,120 @@
 
 logger = logging.getLogger(__name__)
 
+if is_tf_available():
+    import tensorflow as tf
+
+
+def extract_hyperparameters_from_keras(model):
```
osanseviero, line 23: Very cool 🔥

osanseviero, line 23: Should this and other private methods begin with `_` as in `_create_model_card`?

src/huggingface_hub/keras_mixin.py
```diff
@@ -82,6 +200,7 @@ def push_to_hub_keras(
     git_email: Optional[str] = None,
     config: Optional[dict] = None,
     include_optimizer: Optional[bool] = False,
+    task_name: Optional[str] = None,
```
osanseviero, line 203: Please add a docstring for this new param

osanseviero (commented):
Overall looks great! This is really really useful.

src/huggingface_hub/keras_mixin.py
merveenoyan: yeah I should make them private, thanks! 
I don't think people would ever call extract hyperparameter anyway given they can do it when they have the model

merveenoyan:
@osanseviero where do I add test dependencies (like requirements file) I've written the `is_pydot_available() `function

merveenoyan:
@osanseviero edgiest case ever, seems like they don't get installed for some reason

src/huggingface_hub/keras_mixin.py
```diff
@@ -59,9 +173,13 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
+    hyperparameters = _extract_hyperparameters_from_keras(model)
+    _plot_network(model, save_directory)
```
osanseviero, line 177: Same here

src/huggingface_hub/keras_mixin.py
```diff
@@ -59,9 +173,13 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
+    hyperparameters = _extract_hyperparameters_from_keras(model)
```
osanseviero, line 176: Why do we do this within `save_pretrained_keras` and return to `push_to_hub_keras`? Can this be done directly in `push_to_hub_keras`? I feel that makes more sense and the model is also available there

src/huggingface_hub/keras_mixin.py
```diff
@@ -16,6 +16,120 @@
...
+                k = "train_" + k
+            splits = k.split("_")
+            name = " ".join([part.capitalize() for part in splits])
+            values[name] = v
+        lines.append(values)
+    return lines
+
+
+def _plot_network(model, save_directory):
+    tf.keras.utils.plot_model(
```
osanseviero, line 59: Nit: as discussed let's try to use `svg` if possible

src/huggingface_hub/keras_mixin.py
merveenoyan: Sure I'll do them in `push_to_hub_keras()` instead.

src/huggingface_hub/keras_mixin.py
```diff
@@ -123,6 +242,8 @@ def push_to_hub_keras(
             Configuration object to be saved alongside the model weights.
         include_optimizer (:obj:`bool`, `optional`):
             Whether or not to include optimizer during serialization.
+        task_name (:obj:`str`, `optional`):
+            Name of the task the model was trained on.
```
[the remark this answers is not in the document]
merveenoyan: Thanks for the catch!

tests/test_keras_integration.py
```diff
@@ -34,13 +38,17 @@ def repo_name(id=uuid.uuid4().hex[:6]):
...
 
-    These tests are skipped when TensorFlow isn't installed.
+    These tests are skipped when TensorFlow, graphviz and pydot are installed.
 
     """
-    if not is_tf_available():
-        return unittest.skip("test requires Tensorflow")(test_case)
+    if (
+        not is_tf_available()
+        and not is_pydot_available()
```
nateraw, line 48: So TF tests won't run if pydot and graphviz aren't available? Shouldn't they still run if those two packages aren't available? I guess just what you're adding here doesn't need to be tested if those two aren't there.

nateraw (commented):
LGTM - I would be careful to skip tests without those two extra packages (but maybe I'm missing something?). We want to make sure the simplest form of `push_to_hub_keras` just involves having TF installed if possible.

thanks for adding this 😄

tests/test_keras_integration.py
merveenoyan: All of the TF tests are training a model and pushing to hub anyway and in the process they're plotting in the model card. So I first thought I'd split tests into two, one with TF and one with TF + plotting stuff and I thought it would be ok to have one dep checker to rule them all.

src/huggingface_hub/keras_mixin.py
```diff
@@ -292,14 +292,18 @@ def push_to_hub_keras(
...
+    save_pretrained_keras(
         model,
         repo_path_or_name,
         config=config,
         include_optimizer=include_optimizer,
         **model_save_kwargs,
     )
 
+    hyperparameters = _extract_hyperparameters_from_keras(model)
+    if is_graphviz_available and is_pydot_available:
```
osanseviero, line 304: Should this be with `()`?

src/huggingface_hub/keras_mixin.py
merveenoyan: it doesn't work :(

merveenoyan:
@osanseviero I honestly don't know what else I have to do so that it would install pydot. I'll look deeper into it.

merveenoyan:
This is so weird, the tests are failing bcs model plot is not appearing in the repository, probably because the plotting itself is skipped (see below). The test is running when graphviz and pydot are available. 
see:
```
def require_tf(test_case):
    if (
        not is_tf_available()
        and not is_pydot_available()
        and not is_graphviz_available()
    ):
        return unittest.skip("test requires Tensorflow, graphviz and pydot.")(test_case)
    else:
        return test_case
```

The tests that require pushing has to plot model, and the model plotting part is skipped likely because:
```
if is_graphviz_available() and is_pydot_available():
        _plot_network(model, repo_path_or_name)
```
But this shouldn't happen since I'm calling it inside the tests decorated with `require_tf`, meaning, graphviz and pydot should be available. 
When I test on my local `push_to_hub_keras()` works fine and it pushes the plot as well. 
@nateraw can you see if I miss something here?

src/huggingface_hub/file_download.py
```diff
@@ -48,6 +48,34 @@
 except importlib_metadata.PackageNotFoundError:
     pass
 
+_pydot_version = "N/A"
```
osanseviero, line 51: @LysandreJik It feels a bit weird to have these checks here. At the same time, having them here is consistent. Should we have them here or in the keras code?

src/huggingface_hub/keras_mixin.py
```diff
@@ -15,6 +19,119 @@
 
 
 logger = logging.getLogger(__name__)
+if is_tf_available():
+    import tensorflow as tf
+
+
+def _extract_hyperparameters_from_keras(model):
+    hyperparameters = dict()
+    if hasattr(model, "optimizer") and model.optimizer is not None:
```
osanseviero, line 28: If it's `None`, you still want to assign it, so this should work

```suggestion
    if hasattr(model, "optimizer"):
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -15,6 +19,119 @@
...
+            for key in line:
+                value = round(line[key], 3)
+                model_card += f" {value}| "
+    else:
+        model_card += "Model history needed"
+    model_card += "\n ## Model Plot\n"
+    model_card += "\n<details>"
+    model_card += "\n<summary>View Model Plot</summary>\n"
+    path_to_plot = "./model.png"
+    model_card += f"\n![Model Image]({path_to_plot})\n"
```
osanseviero, line 125: This should be skipped when `model_plot=False` no?

src/huggingface_hub/keras_mixin.py
```diff
@@ -123,6 +242,10 @@ def push_to_hub_keras(
             Configuration object to be saved alongside the model weights.
         include_optimizer (:obj:`bool`, `optional`):
             Whether or not to include optimizer during serialization.
+        task_name (:obj:`str`, `optional`):
+            Name of the task the model was trained on.
+        model_plot (:obj::obj:`bool`):
```
osanseviero, line 247:
```suggestion
        model_plot (:obj:`bool`):
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -178,6 +301,12 @@ def push_to_hub_keras(
         include_optimizer=include_optimizer,
         **model_save_kwargs,
     )
+
+    hyperparameters = _extract_hyperparameters_from_keras(model)
+    if model_plot and is_graphviz_available() and is_pydot_available():
+        _plot_network(model, repo_path_or_name)
+    lines = _parse_model_history(model)
```
osanseviero, line 308: I feel all of these should be within `_create_model_card`. It doesn't make sense to compute before the function if you're not using this info outside.

tests/test_keras_integration.py
```diff
@@ -274,6 +277,9 @@ def test_push_to_hub(self):
         REPO_NAME = repo_name("PUSH_TO_HUB")
         model = self.model_init()
         model.build((None, 2))
+        x = tf.constant([[0.44, 0.90], [0.65, 0.39]])
+        y = tf.constant([[1, 1], [0, 0]])
+        model.fit(x, y)
```
osanseviero, line 282: Maybe we should put all of this within a `self.model_init_fitted()` method so you don't need to duplicate code

tests/test_keras_integration.py
```diff
@@ -292,13 +298,41 @@ def test_push_to_hub(self):
 
         self._api.delete_repo(name=f"{REPO_NAME}", token=self._token)
 
+    def test_push_to_hub_model_card(self):
```
osanseviero, line 301: You probably want to add another test that uses `model_plot=False` and checks the image is not there

osanseviero (commented):
The whole logic looks good to me. I added some comments to improve the developer experience and tests, and afterwards it's good to go from my side

src/huggingface_hub/keras_mixin.py
merveenoyan: plot_model() doesn't accept svg.

src/huggingface_hub/keras_mixin.py
```diff
@@ -15,6 +19,128 @@
 
 
 logger = logging.getLogger(__name__)
+if is_tf_available():
```
osanseviero, line 22:
```suggestion
logger = logging.getLogger(__name__)

if is_tf_available():
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -15,6 +19,128 @@
...
+    model_plot: Optional[bool] = True,
+    task_name: Optional[str] = None,
+):
+    """
+    Creates a model card for the repository.
+    """
+    hyperparameters = _extract_hyperparameters_from_keras(model)
+    if model_plot and is_graphviz_available() and is_pydot_available():
+        _plot_network(model, repo_dir)
+    lines = _parse_model_history(model)
```
osanseviero, line 107: Hehe lines move all around :smile: 

Try to just define things where they are used. `lines` is not used anywhere in `_create_model_card` except in `_write_metrics`. If `lines` is just used there, why not just create variables directly within the function in which they are used? That way it's also easier to track what `lines` is and future Merve will appreciate when looking through the code.

tests/test_keras_integration.py
```diff
@@ -292,13 +302,60 @@ def test_push_to_hub(self):
...
+        model = self.model_fit(model)
+        push_to_hub_keras(
+            model,
+            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
+            api_endpoint=ENDPOINT_STAGING,
+            use_auth_token=self._token,
+            git_user="ci",
+            git_email="ci@dummy.com",
+            config={"num": 7, "act": "gelu_fast"},
+            include_optimizer=False,
```
osanseviero, line 318: Are these two needed for the test?

tests/test_keras_integration.py
```diff
@@ -292,13 +302,60 @@ def test_push_to_hub(self):
...
```
osanseviero, line 341: Are these two needed for the test?

src/huggingface_hub/file_download.py
```diff
@@ -48,6 +48,32 @@
...
+try:
+    _graphviz_version = importlib_metadata.version("graphviz")
+    _graphviz_available = True
+except importlib_metadata.PackageNotFoundError:
+    pass
+
+
+def is_graphviz_available():
+    return _graphviz_available
+
```
LysandreJik, line 75: Answering to your comment @osanseviero, I think it's time to have a `file_utils.py` module in which all of this should go.

LysandreJik: Doesn't have to be this PR, but would love a follow-up PR that moves this alongside other version checks to a dedicated module

LysandreJik (commented):
Looks very cool!

src/huggingface_hub/keras_mixin.py
```diff
@@ -16,6 +20,129 @@
...
+    hyperparameters[
+        "training_precision"
+    ] = tf.keras.mixed_precision.global_policy().name
+
+    return hyperparameters
+
+
+def _parse_model_history(model):
+    logs = model.history.history
+    num_epochs = len(logs["loss"])
```
nateraw, line 42: Careful here too, another assumption that caused my keras integration overview notebook to break. but this one is a bit easier to fix...my model had an empty `model.history.history` dict after training (its the same as the autoencoder example in the keras docs). Maybe I'm using a different tf version? (2.8.0)

nateraw: I noticed theres a model.history.params['epochs'], but my epochs shown here was 1, whereas I trained it for 10 epochs...weird.

tests/test_keras_integration.py
```diff
@@ -274,6 +283,7 @@ def test_push_to_hub(self):
         REPO_NAME = repo_name("PUSH_TO_HUB")
         model = self.model_init()
         model.build((None, 2))
+        self.model_fit(model)
```
nateraw, line 286: This adds in the assumption that the model was fit, eh? I think there should be a test that shows you can load a trained model from file + can push that as well without errors.

nateraw: If you require to be fit, btw, you don't need to build anymore, right? Its a meaningless step then if I remember correctly. I could be wrong!!

nateraw (changes requested):
I'm seeing some issues when running through examples in colab. The main issue here seems to be that there is an assumption now that you're fitting the model. I think we need to see two things:

1. Make sure you can load a model from a file and push that to the hub still (i.e. I fit a model earlier, saved it, and want to reload and push).
2. Make sure you can safely grab history/other keys you're accessing without issues in a reliable way. If you can't, then skip the model card process altogether or leave it blank.

Here's [a notebook](https://gist.github.com/nateraw/2ffbd8baf8caecbbcfbdf78cda3a7090) you can take a look at to see the issues. If I did something wrong here, please let me know!! 😄

src/huggingface_hub/keras_mixin.py
```diff
@@ -16,6 +20,129 @@
...
+        hyperparameters["optimizer"] = None
+    hyperparameters[
+        "training_precision"
+    ] = tf.keras.mixed_precision.global_policy().name
+
+    return hyperparameters
+
+
+def _parse_model_history(model):
+    logs = model.history.history
```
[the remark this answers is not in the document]
merveenoyan: If we still want unfitted models then sure I'll just add a couple of lines that checks if the objects that are created with fitting are None or not. Thanks a lot for breaking it!

src/huggingface_hub/keras_mixin.py
merveenoyan: I don't know about the second one, but the first one I'll just assume I don't know if the model is fitted or not.

tests/test_keras_integration.py
merveenoyan: I don't require to fit, maybe I could add another test that only pushes a built model instead.

src/huggingface_hub/keras_mixin.py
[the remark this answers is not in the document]
merveenoyan: for the [notebook](https://gist.github.com/nateraw/2ffbd8baf8caecbbcfbdf78cda3a7090) I just don't get why your model returns an empty history though, I tested it on an autoencoder and a CNN (one was sequential and another one was functional) and it was working just fine. I just don't get the case (I get the built one but others I just don't)

src/huggingface_hub/keras_mixin.py
merveenoyan: I also tested on the autoencoder that's why I'm surprised! I tested on two models, one was sequential and one was functional. 

again this

> for the [notebook](https://gist.github.com/nateraw/2ffbd8baf8caecbbcfbdf78cda3a7090) I just don't get why your model returns an empty history though, I tested it on an autoencoder and a CNN (one was sequential and another one was functional) and it was working just fine. I just don't get the case (I get the built one but others I just don't)

I don't think it's about TF version. I think I'll just try to get `history.history` and pass plotting model card. So many edge cases I could hit there.

tests/test_keras_integration.py
nateraw: that would be a good way to prove it! 😄

src/huggingface_hub/keras_mixin.py
merveenoyan: You can't dump `model.history.history` to an object and call `model.history.history` again, in case anyone wonders why it returns empty in the notebook. 
Ekran Resmi 2022-02-23 21 45 54

in `push_to_hub_keras()` we already call it once and it's dumped to logs. but still if I can't get it I won't create card, given these edge cases.

merveenoyan:
After you call model.predict() or model.evaluate() the model.history.history goes away. push_to_hub_keras() doesn't have such side effect. [Also see this issue](https://stackoverflow.com/questions/62299804/in-tensorflow-2-2-0-my-model-history-history-is-empty-after-fitting-the-data-al)
In the notebook @nateraw sent there's a prediction cell in between that's why it happens. (See below, I also saw the same behavior in another Keras example) 

Ekran Resmi 2022-02-23 23 40 07
Ekran Resmi 2022-02-23 23 38 54
Ekran Resmi 2022-02-23 23 54 38

merveenoyan:
If your model is build but not compiled it looks like [this](https://huggingface.co/merve/model-card-build-only). 
If model is build + compiled it looks like [this](https://huggingface.co/merve/model-card-build-compiled).
If model is fit it looks like [this](https://huggingface.co/merve/model-card-fit)

I'm roughly done (also with tests), I just need to solve this merge conflict. (which I'll do tomorrow)

src/huggingface_hub/keras_mixin.py
```diff
@@ -16,6 +20,129 @@
...
+                model_card += f" {value}| "
+    else:
+        model_card += "Model history needed"
+    return model_card
+
+
+def _create_model_card(
+    model,
+    repo_dir: Path,
+    model_plot: Optional[bool] = True,
```
osanseviero, line 103: nit: maybe `plot_model` makes more sense here (verb instead of noun)

merveenoyan: I thought it would be very confusing because it's a variable that user asks for model's plot and there's a function called `_plot_model()` (though not seen in the user side) are you sure?

merveenoyan: fixed it with last commit!

src/huggingface_hub/keras_mixin.py
```diff
@@ -178,6 +311,7 @@ def push_to_hub_keras(
         include_optimizer=include_optimizer,
         **model_save_kwargs,
     )
+    _create_model_card(model, repo_path_or_name, plot_model, task_name)
```
nateraw, line 314: Something sneaky here...if you call this within `push_to_hub_keras`, it won't get executed for the `KerasModelHubMixin` (for custom Keras components). If you instead stick this within the `save_pretrained_keras` function, it'll happen for the mixin as well (because the mixin calls explicitly to that).

tests/test_keras_integration.py
```diff
@@ -415,3 +487,15 @@ def test_save_pretrained(self):
         self.assertIn("saved_model.pb", files)
         self.assertIn("keras_metadata.pb", files)
         self.assertEqual(len(files), 4)
+
+    def test_save_pretrained_fit(self):
```
nateraw, line 491: thx for adding this test for sanity 😄

nateraw (commented):
Looks great! The issues I mentioned earlier have been resolved. Thanks so much for battling this PR.

There is one remaining issue, however...the `KerasModelHubMixin` will not write cards as `_create_model_card` is not called automatically by it, only `save_pretrained_keras` is.

You can either override the `push_to_hub` in the keras mixin or you can add the `create_model_card` call within `save_pretrained. Both options would require you to add those extra kwargs (`task_name` and `plot_model`) to the respective functions though...which may prove to be annoying. 

What do we want to do about this? @merveenoyan is this mixin being used a bit when there are more complicated use cases in the community? If so, we probably want to make sure to cover this issue here. Once this issue is addressed/we have conversation about it, I'm happy to approve.

src/huggingface_hub/keras_mixin.py
merveenoyan: Sure I'll do it! Thanks for the catch.

merveenoyan: I have a problem with this. ModelHubMixin class has `_save_pretrained` that only takes file path, and you can extend it with `**model_save_kwargs` during overriding for Keras mixin, but I also need to pass model card related `kwargs`, like `plot_model` and `task_name` otherwise I need to override the parent class to take two model card related kwargs. I think it's an overkill to extend parent class like that. What do you suggest? I added those arguments inside `save_pretrained_keras` but for custom components I need `model.save_pretrained`, no?

osanseviero: Afaik nobody is using the mixin atm btw :smile:

merveenoyan: @osanseviero I already did it for `save_pretrained_keras` and it works. I can push that.

merveenoyan:
- I moved model card creation under `save_pretrained_keras()`. 
- `model.save_pretrained()` calls `save_pretrained_keras()` with model card and plotting arguments with default values (`plot_model = True` and `task_name = None`).
- The tests check that model card and plot is created when model is built + fit, called under both `save_pretrained_keras()` and `push_to_hub_keras()`.

merveenoyan:
I managed to break the card.
It writes every epoch like so
```
for line in lines:
            model_card += f"\n| {lines.index(line) + 1}|"  # add values
```
Lines look like this
`[{'Train Loss': 0.15567651391029358}, {'Train Loss': 0.1543986201286316}, {'Train Loss': 0.1531306654214859}]`
which line looks like this: 
`{'Train Loss': 0.1531306654214859}`
if training loss never changes the indexes will look like 1 1 1 on the table.

nateraw (approved):
Nice, works for me in colab for all cases we discussed. thanks for moving this to `save_pretrained_keras` 😄 

Thanks for battling the reviews here.

LGTM 🚀

PR #682, opened 2022-02-10
Improve tests and logging

muellerzr:
## What:
- Separates out the very long `Repository` tests into their own CI/CD matrix
- Also tries to limit how `shutil.rmtree` is called to limit risk of OS error
- Makes all endpoints tests stable by adding a `retry_endpoint` decorator

## Why:
CI/CD can now take just as little as 6-8 minutes or so, rather than 10-15 due to how long the Repository tests take.

## What we watch for
Just like before, we want to make sure some combination of `--ignore` + `test_repository` pass at least once to consider it "stable".

cc @LysandreJik

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -39,7 +39,7 @@ jobs:
           pip install --upgrade pip
           pip install .[testing]
 
-      - run: pytest -sv ./tests/${{test_repository}}
+      - run: pytest -sv ./tests/${{ matrix.test_repository }}
 
   build_pytorch:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6", "3.9"]
+        python-version: ["3.9"]
         test_repository: [" --ignore tests/test_repository.py", "test_repository.py"]
 
     steps:
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.6", "3.9"]
-        test_repository: [" -k 'not RepositoryTest'", "test_repository.py -k RepositoryTest"]
+        test_repository: ["Repository only", "Everything else"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


muellerzr:
One interesting bit is in order to see the logs from logger, we need to run the tests with `log_cli=1`, hence why that's in the CI/CD

.github/workflows/python-tests.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.6", "3.9"]
+        test_repository: [" -k 'not RepositoryTest'", "test_repository.py -k RepositoryTest"]
```
LysandreJik, line 26: That's practical, but it results in 
image

which is a bit unreadable. Do you think this could be clarified?

tests/test_hf_api.py
```diff
@@ -901,10 +905,11 @@ def setUpClass(cls):
...
     def setUp(self):
         self.REPO_NAME_LARGE_FILE = repo_name_large_file()
-        try:
+        if os.path.exists(WORKING_REPO_DIR):
             shutil.rmtree(WORKING_REPO_DIR, onerror=set_write_permission_and_retry)
-        except FileNotFoundError:
-            pass
+        logger.info(
+            f"Does {WORKING_REPO_DIR} exist: {os.path.exists(WORKING_REPO_DIR)}"
+        )
```
LysandreJik, line 912: I fail to see why that would result in an improvement, but glad it does! Let's monitor the next pushes on `main` to ensure it works correctly.

tests/testing_utils.py
```diff
@@ -170,3 +174,31 @@ def with_production_testing(func):
...
+        number_of_tries: Number of tries to attempt a passing test
+        wait_time: Time to wait in-between attempts in seconds
+    """
+
+    def decorator(*args, **kwargs):
+        retry_count = 1
+        while retry_count < number_of_tries:
+            try:
+                return function(*args, **kwargs)
+            except Exception:
```
LysandreJik, line 195: Ideally this would still catch the fact that it's a HTTP exception. If catching the `HTTPError` doesn't work, could you try and log the error so that we may see how to handle it?

muellerzr:
I'm switching gears this week so I won't be back to this until later. Here is a TL;DR of progress:

- If we receive OS errors, these also seem to go away when we rerun the tests, so we should also add them to the try/catch of `@retry_endpoint`
- `retry_endpoint` currently now just supports and successfully catches the HTTP errors, which is good
- Any tests that use `api.create_repo` or `api.upload_file` should be given the `@retry_endpoint` helper. Most of this lives in `Repository` and `HfApi` tests as well as `setUp`

muellerzr:
@LysandreJik this should be good for a rereview now. Also made the tests more readable with some bash magic 
image

You can see where catches were found by looking at [this](https://github.com/huggingface/huggingface_hub/actions/runs/1866213087/attempts/1) run and doing CTRL+F "Retrying"

Specifically "Build 3.6, Repository Only"

LysandreJik (approved):
Thank you for working on this, @muellerzr!

PR #688, opened 2022-02-16
improve: improve support for pyannote.audio

hbredin:
pyannote.audio has two main types of "inference":

* models are regular PyTorch models that output tensors
* pipelines encapsulate one or more models, postprocess their outputs and return actual audio segmentation results (start_time, end_time, label) as pyannote.core.Annotation instances

I think both have their own use and target different users (e.g. models for researchers, pipelines for users who just want to get speaker diarization results).


osanseviero:
Hey there! Sorry for the slow response. As of now, the code snippet is based in the tag of a model. So if a model repo has `pyannote` tag, it will then show the code snippet. With this change, you're changing the expected tag.

Instead, I would suggest the two code types directly in the same snippet, although my concern would be of the sippet being too long. Maybe we can simplify and just do inference on the whole file. It's not like the transformers code snippets have a full example. WDYT?

hbredin:
Are you suggesting to only stick with the `pyannote-audio-pipeline` tag?

osanseviero:
Maybe just to take a step back and understand the usage. Can you use both `Pipeline`and `Model` with the same model repository? 

Once you have a repo with the tag `pyanote`, you will show corresponding code snippet. If you do the change you're suggesting, you will need to modify the repos to have one of the two snippets you suggest (but only one snippet will show)

hbredin:
> Maybe just to take a step back and understand the usage. Can you use both `Pipeline`and `Model` with the same model repository?

No, because a `Pipeline` may internally rely on multiple `Models`.
There is no bijection between a pipeline and a model.

For instance, `pyannote/speaker-diarization` pipeline relies on `pyannote/segmentation` model (for things like voice activity detection) and `pyannote/embedding` model (for the speaker clustering step).

> Once you have a repo with the tag `pyanote`, you will show corresponding code snippet. If you do the change you're suggesting, you will need to modify the repos to have one of the two snippets you suggest (but only one snippet will show)

Yes, that was the point of this PR :)

If that is too much trouble, we could focus on pipelines (`pyannote-audio-pipeline` tag) which is probably what beginners would use first. That being said, I will keep tagging pipelines and models with two different tags as I need an easy way to search them using the `HFApi`.

osanseviero:
Alright, much clearer now. 

My suggestion is then to do the following

1. Keep `pyannote` as the tag (key) used here. This will work as the primary tag
2. You can then have secondary tags, which can be the ones you suggest. 
3. Then, based on the secondary tag, you can show different snippets. See how it's done here: https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Libraries.ts#L174-L181

This will allow to have a single filter for all `pyannote` repos in the main libraries filter, but with the benefits that:
* The code snippets change based on the secondary tag
* Users can still filter on with the secondary tag.

Do you think this would solve the need?

hbredin:
Done.

js/src/lib/interfaces/Libraries.ts
```diff
@@ -4,25 +4,25 @@ import type { ModelData } from "./Types";
  * Add your new library here.
  */
 export enum ModelLibrary {
-	"adapter-transformers"   = "Adapter Transformers",
```
osanseviero, line 7: There are quite a few unrelated changes in the PR (maybe your code editor), do you think you could revert those?

js/src/lib/interfaces/Libraries.ts
```diff
@@ -138,18 +138,42 @@ const keras = (model: ModelData) =>
...
+# inference on the whole file
+inference("file.wav")
+
+# inference on an excerpt
+from pyannote.core import Segment
+excerpt = Segment(start=2.0, end=5.0)
+inference.crop("file.wav", excerpt)`;
+
+const pyannote_audio = (model: ModelData) => {
+	if (model.tags?.includes("pyannote-audio-pipeline")) {
```
osanseviero, line 172: Looks great!

osanseviero (approved):
Thank you very much for the PR! There are some unrelated changes, would you be able to remove them?

Afterwards we can go ahead and merge

js/src/lib/interfaces/Libraries.ts
hbredin: Should be OK now.

osanseviero (approved):
LGTM! Thanks :rocket:

hbredin:
Also, out of curiosity, when will this be deployed?

osanseviero:
Hey there! Sorry, we're doing a repo cleanup/split to make things much much much better organized, so this is taking a bit longer to get deployed. I'll update you as soon as we do this.

osanseviero:
The change was also applied in https://github.com/huggingface/hub-docs/commit/e00b7bd7c05c00d1fb90e157dc54730fce7ef788, this should be deployed soon.

osanseviero:
@hbredin the change is now deployed :rocket: thanks @Pierrci!

osanseviero:
@Narsil we'll need a deploy of this change in api-inference updating the framework name, which requires a small internal PR iirc. Could you help us with this?

PR #691, opened 2022-02-16
Adding RL Task

simoninithomas:
Okay so I've just added RL task:
- I've modified Tasks.ts
- I've added the icon to js/src/lib/components/icons
- I've modified ModelPipelineIcon.svelte

I might have missing something though?

Thanks for the feedback


js/src/lib/interfaces/Types.ts
```diff
@@ -38,6 +38,7 @@ export enum PipelineType {
 	"image-to-text"                                           = "Image-to-Text",
 	/// others
 	"structured-data-classification"                          = "Structured Data Classification",
+	"reinforcement-learning"				  = "Reinforcement Learning",
```
Pierrci, line 41: can you use spaces instead of tabs? (same elsewhere)

Pierrci (commented):
> I think the style error comes from the Icon since the HTML is in "compact mode" compared to the others icons?

You need to run prettier w/ `npm run format`, this should fix things

js/src/lib/interfaces/Types.ts
simoninithomas: Sure!

julien-c: BTW this shouldn't take engineering time from @Pierrci IMO, you should check with library maintainers @LysandreJik @osanseviero @muellerzr when CI is red on your PR

Issue #650, opened 2022-02-04
Invoke the CLI on a private hub endpoint


juliensimon:
In the context of a private hub deployment, customers would like to use the CLI on their hub, not on the public hub. This doesn't seem to be configurable at the moment and it would be nice to add this feature.

huggingface_hub.HfApi supports configurable endpoints, but this isn't implemented in the CLI, see BaseUserCommand in https://github.com/huggingface/huggingface_hub/blob/4d6b35d41afd23d991e2657aaf3113aa18aa1717/src/huggingface_hub/commands/user.py

The same issue exists with the transformers and datasets library. I'm going to create issues there as well, and I'll reference them below.

juliensimon:
For reference:
https://github.com/huggingface/transformers/issues/15514
https://github.com/huggingface/datasets/issues/3679

juliensimon:
I have a patch for this. Will keep testing a bit more before sending the PR.

PR #703, opened 2022-02-22
Add support for private hub

juliensimon:
This fixes https://github.com/huggingface/huggingface_hub/issues/650

We simply set HF_ENDPOINT to the URL of a private hub, and use the API as-is.  If the API itself has the endpoint, then the environment variable is ignored.

FYI, the same variable is used by the datasets library for a similar purpose.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -410,6 +410,8 @@ def erase_from_credential_store(username=None):
 
 class HfApi:
     def __init__(self, endpoint=None):
+        if endpoint is None:
+            endpoint = os.getenv('HF_ENDPOINT')
         self.endpoint = endpoint if endpoint is not None else ENDPOINT
 
     def login(self, username: str, password: str) -> str:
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -410,6 +410,8 @@ def erase_from_credential_store(username=None):
 
 class HfApi:
     def __init__(self, endpoint=None):
+        if endpoint is None:
+            endpoint = os.getenv('HF_ENDPOINT')
         self.endpoint = endpoint if endpoint is not None else ENDPOINT
```
julien-c, line 415: maybe more Pythonic to do those three lines in one line
```python
self.endpoint = endpoint if endpoint is not None else os.getenv('HF_ENDPOINT', ENDPOINT)
```

julien-c: other than that LGTM=)

juliensimon: ok!

julien-c (approved):

osanseviero (approved):
Looks good!

LysandreJik (approved):
Yes, LGTM! Thank you @juliensimon

juliensimon:
Thanks!

PR #704, opened 2022-02-22
Allow renaming and transfering repos programmatically

### LINKED ISSUE #694: `rename_repo` function for `huggingface_hub` library
A new feature to update a repository's name with `huggingface_hub`. The behavious could be similar to `update_repo_visibility`:

```Python
from huggingface_hub import rename_repo

rename_repo(name=REPO_NAME, org=ORG,
            new_name=NEW_REPO_NAME, new_org=NEW_ORG)
```

This would allow to programmatically change the name of several repositories on the hub with ease.

cc @osanseviero

osanseviero:
Fixes #694 

On top of implementing transfer repos, I'm refactoring some of the token validation/retrievals which is duplicated. I kept some things for the backward compatibility of some methods, but it will be simplified further once cleaned up.

I added two methods: `transfer_repo` and `rename_repo`, with the first being more consistent with the codebase by handling separately the model_repo name and its namespace. `rename_repo`, on the other hand, follows a bit more intuitive approach by simply using the `repo_id`, which is what you copy/paste from the Hub.

Example usage

```python
# Examples with rename_repo
rename_repo("osanseviero/test1", "huggingface/test")
rename_repo("osanseviero/test1", "osanseviero/test2")

# Examples with transfer_repo
transfer_repo("test1", new_name="test", new_organization="huggingface")
transfer_repo("test1", new_name="test", organization="osanseviero", new_organization="huggingface")
transfer_repo("test1", new_name="test2")
```

I don't love having two methods for the same goal though, so we can ax `rename_repo`, although I feel it's more intuitive for the users and does fewer things on the back.


src/huggingface_hub/hf_api.py
```diff
@@ -1269,6 +1252,92 @@ def update_repo_visibility(
...
+        Rename a repository from namespace1/repo_name1 to namespace2/repo_name2
+
+        Renaming can be used for three use cases
+        - Renaming a repository within same user.
+        - Renaming a repository within same organization.
+        - Move repository from user to an organization.
+
+        Renaming does not work for:
+        - Moving a repository from an organization to another user or organization.
+        - Moving a repository from user A to user B.
```
lvwerra, line 1272: Should we leave add some guidance for the case where a user wants to transfer a model from org2org or user2user?

lvwerra (approved):
Thanks a lot for adding this. Left a minor comment, otherwise looks good!

src/huggingface_hub/hf_api.py
```diff
@@ -1269,6 +1252,92 @@ def update_repo_visibility(
...
+        repo_type: Optional[str] = None,
+        token: Optional[str] = None,
+    ):
+        """
+        Rename a repository from namespace1/repo_name1 to namespace2/repo_name2
+
+        Renaming can be used for three use cases
+        - Renaming a repository within same user.
+        - Renaming a repository within same organization.
+        - Move repository from user to an organization.
```
julien-c, line 1268: are you sure we cannot move repository from organization to (currently loggedin) user? 

cc @SBrandeis @Pierrci @coyotte508 for the latest on what's planned/shipped

julien-c (commented):
code looks good but I agree we should probably pick one method and not expose two methods for the same thing

I'd agree that the `rename_repo`  API looks better/simpler

src/huggingface_hub/hf_api.py
coyotte508: > are you sure we cannot move repository from organization to (currently loggedin) user?

It's not possible at the moment, unless the one doing the move is an hf user (which may be changed by https://github.com/huggingface/moon-landing/pull/2134)

osanseviero:
> I'd agree that the rename_repo API looks better/simpler

With the big con it would be inconsistent with the rest of the methods, unfortunately.

src/huggingface_hub/hf_api.py
julien-c: (other cases can be tracked in https://github.com/huggingface/moon-landing/issues/1451)

LysandreJik:
Thanks for your PR, very clean. Throwing another idea in the pit: how about `move_repo`? To me, `move` covers both `rename` and `transfer`, while `transfer` seems like it's only for cross-organization changes and `rename` sounds to me like inner-organization changes. No strong opinion though.

And to recap what we have in the library right now:
- `hf_hub_url`/`hf_hub_download` takes `repo_id`
- `model_info` takes `repo_id`
- `list_repo_files` takes `repo_id`
- `upload_file` takes `repo_id`
- `delete_file` takes `repo_id`
- `Repository` takes `repo_id`

And

- `create_repo` takes `name`/`organization`
- `delete_repo` takes `name`/`organization`
- `update_repo_visibility` takes `name`/`organization`

So all methods that manage the repositories' settings and existence on the hub ask for an explicit `organization`. However, this is legacy from the very beginning of the library so I think we should feel free to reconsider if having `<org>/<id>` is simpler than two arguments (which I definitely think it is), especially as setting up a deprecation cycle for this should be very straightforward.

If we don't opt for updating the create/delete/update_repo_visibility methods, I would keep the `organization` parameters separate.

osanseviero:
Thanks for the feedback! 
* Given the `rename_repo` was more intuitive by passing repo_id instead of name, I kept that and removed `transfer_repo`. 
* I renamed `rename_repo` to `move_repo` as per @LysandreJik suggestion, I agree it sounds better.
* About the create/delete/update_repo_visibility methods, let's deprecate the organization parameters. I think this will be somewhat easy but we'll need to modify quite a bit of 3rd party integrations in the future (they use pinned versions so not big issue). Let's keep this PR small; are you ok if we do this in a follow up PR? (opened an [issue](https://github.com/huggingface/huggingface_hub/issues/706))

LysandreJik:
Yes, absolutely!

osanseviero:
@lvwerra about this comment

> Should we add some guidance for the case where a user wants to transfer a model from org2org or user2user?

I don't think providing this guidance should be in the client library (except a link to the docs). What we can do is expose the error message to the user when this happens so it's as confusing. It's then in the backend side to provide better error messages in the response.

Example result with existing implementation

> requests.exceptions.HTTPError: 403 Error Message: {"error":"You don't have the rights to move this model to OsanOrg2"}

lvwerra:
@osanseviero sure, I don't have a strong opinion where the message comes from. I just thought from a users perspective the current documentation alone might be a bit frustrating: I need to move a repo  org2org > I find a function that could potentially do that > function docstring (or error following message) says just no, this is not possible > what now?

I think rather than just saying it is not supported it would be nice to directly offer a solution or point to instructions.

sgugger:
I agree with all @LysandreJik said, especially the part about deprecating the `organization` parameter in create/delete/update_repo_visibility for more consistency.

(@LysandreJik I hope you realize that for backward-compatibility reasons, the first positional argument of those methods will need to support both a `repo_name` and automatically add the namespace of the user, or a full `organization/repo_name` which is something I've asked for ever and ever)

julien-c:
I'm supportive of deprecating the organization parameters and aligning to "full" repo ids everywhere.

I think third party library usage is still rather small and we can work with them to pin their dependencies and/or support them updating their code.

osanseviero:
Answering @lvwerra 

> @osanseviero sure, I don't have a strong opinion where the message comes from. I just thought from a users perspective the current documentation alone might be a bit frustrating: I need to move a repo org2org > I find a function that could potentially do that > function docstring (or error following message) says just no, this is not possible > what now?
> 
> I think rather than just saying it is not supported it would be nice to directly offer a solution or point to instructions.

After thinking about this, I don't think the limits of the server API should be documented in the library. Instead, I added a section under docs with clear documentation for this. The documentation is linked both in the docstring and in case of an error. If we do server changes, then we just need to update the documentation. Here is an example of an error when doing something invalid. 

```
requests.exceptions.HTTPError: 403 Error Message: {"error":"You don't have the rights to move this model to OsanOrg2"}. For additional documentation please see https://hf.co/docs/hub/main#how-can-i-rename-or-transfer-a-repo.
```

julien-c (approved):
lgtm!

lvwerra:
@osanseviero that's great, thanks for adding the link to the docs!

docs/hub/main.md
```diff
@@ -94,6 +94,21 @@ Here is a video to learn more about it from our [course](http://hf.co/course)!
...
+Moving can be used for three use cases ✅ 
+- Renaming a repository within same user.
+- Renaming a repository within same organization.
+- Transfering repository from user to an organization.
+
+Moving does not work for ❌
+- Transfering a repository from an organization to another user or organization.
+- Transfering a repository from user A to user B.
+
+If these are use cases you need help with, please send us an email at **website at huggingface.co**.
```
LysandreJik, line 110: Nice!

docs/hub/main.md
```diff
@@ -94,6 +94,21 @@ Here is a video to learn more about it from our [course](http://hf.co/course)!
...
 <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/rkCly_cbMBk" title="Managing a repo" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
+## How can I rename or transfer a repo?
+
+You can go to the settings page in any repo to achieve this. Note that there are certain limitations in terms of use cases.
+
+Moving can be used for three use cases ✅ 
+- Renaming a repository within same user.
+- Renaming a repository within same organization.
+- Transfering repository from user to an organization.
```
Pierrci, line 104:
```suggestion
- Transfering repository from user to an organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
```
or something similar? :) cc @SBrandeis

SBrandeis: Note that this also holds for renaming a repo inside the same organization

PR #711, opened 2022-02-23
Add RL task

simoninithomas:
Ok I've made a new PR because my former one was outdated and dirty.
Normally this time it's good :crossed_fingers:


simoninithomas:
And... there is still this prettier problem :sob: ok let's me try find a solution

PR #716, opened 2022-02-23
Misc documentation fixes

osanseviero:
Notes
* `how-to-inference` is just about the client wrapper, so I clarified it a bit since it's not just for 3rd party integration
* Fixed broken image
* Consistent formating in https://huggingface.co/docs/hub/adding-a-model which has broken sections at the moment


LysandreJik (approved):
Thank you for fixing that, @osanseviero!

docs/hub/_sections.yml
```diff
@@ -32,7 +32,7 @@
   title: How to create repositories and upload files to the Hub
 
 - local: how-to-inference
-  title: How to set up Inference API with your library
+  title: How to use the Inference API with the Hub library
```
julien-c, line 35: 👍

julien-c: `with` => `from`

docs/hub/how-to-inference.md
```diff
@@ -4,7 +4,9 @@ title: How to programmatically access the Inference API
 
 # How to programmatically access the Inference API
 
-The Inference API provides fast inference for your hosted models. The Inference API can be accessed via usual HTTP requests with your favorite programming languages, but the `huggingface_hub` library has a client wrapper to access the Inference API programmatically. This guide will show you how to make calls to the Inference API from your library. For more detailed information, refer to the [Inference API documentation](https://api-inference.huggingface.co/docs/python/html/index.html).
+The Inference API provides fast inference for your hosted models. The Inference API can be accessed via usual HTTP requests with your favorite programming languages, but the `huggingface_hub` library has a client wrapper to access the Inference API programmatically. This guide will show you how to make calls to the Inference API with the `huggingface_hub` library.
+
+**If you want to make the HTTP calls directly, please refer to [Accelerated Inference API Documentation](https://api-inference.huggingface.co/docs/python/html/index.html).**
```
julien-c, line 9:
```suggestion
**If you want to make the HTTP calls directly, please refer to [Accelerated Inference API Documentation](https://api-inference.huggingface.co/docs/python/html/index.html)** or to the sample snippets visible on every supported model page.
```

maybe with a screenshot like:
152540141-4e4c0319-55aa-42b2-8008-2fd41355de62

stevhliu (approved):
Thanks for the fixes! 

This [section](https://huggingface.co/docs/hub/adding-a-model#using-the-huggingface_hub-client-library) doesn't display the code formatting around `huggingface_hub` in the left navbar. In general, I also think we should avoid empty headings with no content. Adding a brief description provides more context for users, and makes it easier to browse. For example:

## Using the huggingface_hub client library
The `huggingface_hub` library allows users to publish and download files from the Hub. The library also offers useful tools like creating a repository and managing privacy settings.

### Installing the library and logging in

osanseviero:
> In general, I also think we should avoid empty headings with no content. Adding a brief description provides more context for users, and makes it easier to browse. For example:

At the moment the first heading is a h2 and the second is a h3, but there's an internal bug (cc @gary149 https://github.com/huggingface/moon-landing/issues/799) that displays both similarly.

PR #720, opened 2022-02-24
Bump fairseq version to avoid bug

osanseviero:
[facebook/tts_transformer-en-ljspeech](https://huggingface.co/facebook/tts_transformer-en-ljspeech?text=Hello%2C+this+is+a+test+run.) inference is broken, but this is fixed in a later `fairseq` version (this has not made to a release yet). I could add this to a unit test, but the library failing for a certain type of model is not something we should have in our test coverage imo

### DIFF
--- a/api-inference-community/docker_images/fairseq/requirements.txt
+++ b/api-inference-community/docker_images/fairseq/requirements.txt
@@ -5,4 +5,4 @@ phonemizer==2.2.1
 librosa==0.8.1
 hanziconv==0.3.2
 sentencepiece==0.1.91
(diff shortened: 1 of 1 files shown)


osanseviero:
The GH Actions tests seems to have been failing for quite some time, so I think we could move ahead with this PR if you're ok
https://github.com/huggingface/huggingface_hub/actions/workflows/python-api-fairseq.yaml

Narsil:
Running tests manually to check the new behavior since the test is not cooperative.
(The tests run all kind of timeouts so it could be a false positive)

Narsil (approved):
LGTM !

Overriding the test which fails only becaus it's a large model, but caching does work.

PR #723, opened 2022-02-28
Add autoreview capability

muellerzr:
Somewhat linked to https://github.com/huggingface/huggingface_hub/pull/692, adds a CODEOWNERS file to automatically assign reviewers based on what sections of the code is setup.

Let me know if it all seems fine or if there should be more adjustments to be made!


.github/CODEOWNERS
```diff
@@ -0,0 +1,16 @@
+# See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+# These owners will be the default owners for everything in the repo. Unless a later match takes precedence, @jph00 will 
+# be requested for review when someone opens a pull request.
+
+# Overall repository reviewers added to every PR
+* @LysandreJik @osanseviero
```
osanseviero, line 6: My understanding from "These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence," is that since we have (near) full coverage, neither Lysandre nor me will be added to every PR. 

Also I don't think @LysandreJik wants to review everything in her, I think he should have same owners as you, since he mostly works in the `huggingface_hub` part of this repo

.github/workflows/reviewers.yml
```diff
@@ -0,0 +1,39 @@
...
+    core-library:
+      - muellerzr
+    documentation:
+      - stevhliu
+    community-inference:
+      - Narsil
+    community-tasks:
+      - merveenoyan
+    widgets:
+      - mishig25
```
osanseviero, line 18: My understanding is that this specifies that every time someone from `key` opens a PR, the values will review. For example, for people from the `core-library` group, `muellerzr` would review. But I don't think there is `core-library` group and afaik we don't need to auto-assign reviewers based on author  

https://github.com/marketplace/actions/auto-request-review#auto-assign-reviewers-based-on-the-author

osanseviero: So if I understand correctly `per_author` should likely be `groups` instead, no?

.github/workflows/reviewers.yml
```diff
@@ -0,0 +1,39 @@
+reviewers:
+  # The default reviewers
+  defaults:
+    - LysandreJik
+    - osanseviero
```
osanseviero, line 5: Just for clarification, default reviewers are only assigned if no other reviewers are matched. Is that the expectation?

osanseviero (changes requested):

.github/workflows/reviewers.yml
muellerzr: Correct! Thanks for missing that. Will fix nit

.github/workflows/reviewers.yml
muellerzr: That's not, so we can turn defaults into a group, and add that group to each PR. (Something we *couldn't* do well w/o the action)

osanseviero (approved):
Looks good! Thanks for the PR! :rocket:

Issue #706, opened 2022-02-22
Deprecate organization parameter in create/delete/update_repo_visibility


osanseviero:
Related discussion: https://github.com/huggingface/huggingface_hub/pull/704#issuecomment-1048260407

PR #733, opened 2022-03-02
MNT deprecate name and organization in favor of repo_id

adrinjalali:
Fixes https://github.com/huggingface/huggingface_hub/issues/706

Tests and deprecations on other methods to be added if we're okay with the approach.

cc @LysandreJik @osanseviero


src/huggingface_hub/hf_api.py
```diff
@@ -1177,18 +1177,55 @@ def delete_repo(
...
     ) -> Dict[str, bool]:
-        """
-        Update the visibility setting of a repository.
+        """Update the visibility setting of a repository.
+
+        Args:
+            repo_id: A user or an organization name and a repo name seperated by a
+                ``/``.
+
+                .. versionadded: 0.4.0
```
osanseviero, line 1193: I've never seen this before. Not sure if we want it, but if it's useful this is cool :rocket:

src/huggingface_hub/hf_api.py
```diff
@@ -1177,18 +1177,55 @@ def delete_repo(
...
+
+            private: Whether the model repo should be private (requires a paid
+                huggingface.co account)
+
+            token: A token to be used for the download.
+                - If ``True``, the token is read from the HuggingFace config
+                folder.
+                - If a string, it's used as the authentication token.
```
osanseviero, line 1201: `token` should always be a string, so I would keep that. If not specified, it retrieves the stored token.

src/huggingface_hub/hf_api.py
```diff
@@ -1177,18 +1177,55 @@ def delete_repo(
...
+            The HTTP response in json.
         """
         if repo_type not in REPO_TYPES:
             raise ValueError("Invalid repo type")
 
+        if repo_id and name:
+            raise ValueError(
+                "Only pass `repo_id` and leave deprecated `name` to be None."
+            )
+        elif name:
```
osanseviero, line 1217: Should we also show this warning if user specified `organization`?

osanseviero (commented):
I like this approach for deprecating! I'm all in favour of updating the other params as well.

Thank you very much for this PR! :rocket: :fire:

src/huggingface_hub/hf_api.py
adrinjalali: IMO this is very important for third party developers who need to know which versions of the library they're compatible with. This is how `versionadded` and `versionchanged` can be rendered: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html#sklearn.ensemble.HistGradientBoostingRegressor

src/huggingface_hub/hf_api.py
adrinjalali: `_validate_or_retrieve_token` seems to disagree with you I think. Although in this case you're right I guess? I'm not sure if we're being consistent in treatment of `token` and `use_auth_token`  across this file.

adrinjalali:
This is now ready for another review!

src/huggingface_hub/hf_api.py
```diff
@@ -1082,8 +1112,8 @@ def create_repo(
                 "Ignoring provided space_sdk because repo_type is not 'space'."
             )
 
-        if lfsmultipartthresh is not None:
-            json["lfsmultipartthresh"] = lfsmultipartthresh
+        if getattr(self, "_lfsmultipartthresh", None):
+            json["lfsmultipartthresh"] = self._lfsmultipartthresh
```
osanseviero, line 1116: Why was this changed?

src/huggingface_hub/hf_api.py
```diff
@@ -992,34 +1014,42 @@ def dataset_info(
...
+        Args:
+            repo_id: A namespace (user or an organization) and a repo name
+                seperated by a ``/``.
 
-        Params:
-            private: Whether the model repo should be private (requires a paid huggingface.co account)
+                .. versionadded: 0.4.0
 
-            repo_type: Set to :obj:`"dataset"` or :obj:`"space"` if uploading to a dataset or space, :obj:`None` or :obj:`"model"` if uploading to a model. Default is :obj:`None`.
+            token: An authentication token.
```
osanseviero, line 1034: Maybe here and in the others, let's add a link to https://huggingface.co/settings/tokens so users know where to get the token

osanseviero (approved):
This looks good to me!

src/huggingface_hub/hf_api.py
adrinjalali: done

src/huggingface_hub/hf_api.py
adrinjalali: The docstring of this parameter says:

    lfsmultipartthresh: Optional: internal param for testing purposes.

And it doesn't seem to affect the logic of the code.

It doesn't seem like a parameter which should be a part of the public API and exposed to the user, and it's a trivial fix in the tests to make it work this way. Happy to move it to a separate PR if you think it needs to be separate.

tests/test_hf_api.py
```diff
@@ -1002,11 +1057,12 @@ def setup_local_clone(self, REMOTE_URL):
 
     @retry_endpoint
     def test_end_to_end_thresh_6M(self):
+        self._api._lfsmultipartthresh = 6 * 10**6
```
adrinjalali, line 1060: around here is where lfsmultipartthresh is used in the tests.

adrinjalali:
I'm not sure if the failed CI is related to this PR.

LysandreJik (commented):
Thanks a lot for solving the issue!

You're solving the issue by adding a new argument `repo_id`, which is fine, but I think it's misleading in some cases, as I would expect the `name` to work correctly as a positional argument. I personally would use the method as follows:

```py
from huggingface_hub import create_repo

create_repo("my-brand-new-repo")
create_repo("my-brand-new-repo", organization="huggingface")
```
However with this approach this results in a warning that may be a bit arcane:

```py
>>> create_repo('<repo>')
```
```
/.../huggingface_hub/src/huggingface_hub/hf_api.py:65: FutureWarning: `name` and `organization` input arguments are deprecated and will be removed in v0.7. Pass `repo_id` instead.
  FutureWarning,
'https://huggingface.co/<user>/<repo>'
```

I don't really know what's `name` and why I'm being prompted to change it here, given I didn't pass it with the keyword.

Furthermore, it makes the following API usage deprecated, even though it is both documented [here](https://github.com/huggingface/huggingface_hub/blame/6dac5f45ac7c60ad04c83e4af7b02de63306bd89/docs/hub/how-to-upstream.md#L78-L82) and feels like the API we're trying to push forward.

```py
>>> create_repo('<org>/<repo>')
```
```
/.../huggingface_hub/src/huggingface_hub/hf_api.py:65: FutureWarning: `name` and `organization` input arguments are deprecated and will be removed in v0.7. Pass `repo_id` instead.
  FutureWarning,
'https://huggingface.co/<org>/<repo>'
```

---

I would prefer doing the following:
1. Deprecating the `organization` argument
2. Replacing the `name` argument by `repo_id`
3. Adding `name` as an additional kwarg at the bottom, instead of the `repo_id` you've added.

This should enable the following usage:
- `create_repo("<repo>")`
- `create_repo("<org>/<repo>")`
- `create_repo(repo_id="<repo>")`
- `create_repo(repo_id="<org>/<repo>")`
- `create_repo("<name>", organization="<org>")` # Deprecation warning!
- `create_repo(name="<name>", organization="<org>")` # Deprecation warning!

What do you think?

adrinjalali:
Test failing with

```
E           requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://moon-staging.huggingface.co/api/repos/create - You don't have the rights to create a model under this namespace
```

on:

```py
    def test_push_to_hub(self):
        REPO_NAME = repo_name("PUSH_TO_HUB")
        model = DummyModel()
        model.push_to_hub(
            repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
            api_endpoint=ENDPOINT_STAGING,
            use_auth_token=self._token,
            git_user="ci",
            git_email="ci@dummy.com",
            config={"num": 7, "act": "gelu_fast"},
        )
```

I'm not sure how to fix this. @julien-c seems to have written the test.

src/huggingface_hub/hf_api.py
```diff
@@ -50,6 +50,28 @@
...
+        warnings.warn(
+            "`name` and `organization` input arguments are deprecated and "
+            "will be removed in v0.7. Pass `repo_id` instead.",
+            FutureWarning,
+        )
+    else:
+        if "/" in repo_id:
+            name, organization = repo_id.split("/")
+        else:
+            name, organization = repo_id, None
```
LysandreJik, line 71: Current CI failure comes from here, the `name` and `organization` are switched: we're expecting `<org>/<repo_id>` rather than the other way around.

```suggestion
        if "/" in repo_id:
            organization, name = repo_id.split("/")
        else:
            organization, name = None, repo_id
```

src/huggingface_hub/commands/user.py
```diff
@@ -255,9 +255,10 @@ def run(self):
                 exit()
         try:
             url = self._api.create_repo(
-                self.args.name,
+                repo_id=self.args.name
+                if self.args.organization is None
+                else f"{self.args.organization}/{self.args.name}",
```
LysandreJik, line 260: (nit) can the if/else statement be taken out of the method call? I would find it easier to parse

src/huggingface_hub/hf_api.py
```diff
@@ -992,34 +1014,47 @@ def dataset_info(
...
 
         Returns:
             URL to the newly created repo.
+
+        References:
+            .. [1] https://huggingface.co/settings/tokens
+
+        Reference
         """
+        name, organization = _validate_repo_id_deprecation(repo_id, name, organization)
```
LysandreJik, line 1056: I would add a check to ensure that at least `repo_id` or `name` is passed, since they are now both optional arguments:
```
>>> from huggingface_hub import create_repo
>>> create_repo()
  File "/usr/lib/python3.6/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/home/lysandre/Workspaces/Python/huggingface_hub/src/huggingface_hub/hf_api.py", line 1056, in create_repo
    name, organization = _validate_repo_id_deprecation(repo_id, name, organization)
  File "/home/lysandre/Workspaces/Python/huggingface_hub/src/huggingface_hub/hf_api.py", line 68, in _validate_repo_id_deprecation
    if "/" in repo_id:
TypeError: argument of type 'NoneType' is not iterable
```

Something like the following
```suggestion
        if name is None and repo_id is None:
            raise ValueError("Please pass either `name` or `repo_id`")
            
        name, organization = _validate_repo_id_deprecation(repo_id, name, organization)
```

src/huggingface_hub/hf_api.py
LysandreJik: Ideally this method would be unit-tested so as to raise an error from the method itself rather than for another test to raise a seemingly unrelated error

LysandreJik (commented):
This looks good! Only left a comment regarding a switch between the org and name, it should solve the CI issue.

src/huggingface_hub/hf_api.py
osanseviero: +1, but since name is deprecated, I would just ask to specify `repo_id`.

LysandreJik: Indeed, good call!

adrinjalali:
@LysandreJik if you're happy with the changes, please feel free to merge.

LysandreJik (approved):
LGTM, great work!

LysandreJik:
Resolved conflicts, will merge when tests pass

lhoestq:
would love to be pinged if you do changes if `push_to_hub` in the future !

PR #734, opened 2022-03-02
Add ModelEvalResults.svelte

mishig25:


Pierrci:
Is this still needed @mishig25?(I think it was for the course?)

mishig25:
We are no longer adding it for 3 reasons:
1. We have decided [to use gradio iframes](https://huggingface.slack.com/archives/C01SJL5C5LK/p1647929911960239) instead of InferenceWidgets. Therefore, besides ModelEvalResults.svelte, there is no other reason to add huggingface_hub submoodule to doc-builder.
2. Moving ModelEvalResults.svelte from moon-landing to huggignface_hub would be complicated because its [type/interface](https://github.com/huggingface/moon-landing/blob/524c3c1e8c59bf6a913d4228fc6ff276b82a1872/server/lib/FrontData.ts#L52) depends on a lot of other types/interfaces that are not present in huggingface_hub.
3. It is only being used 3 times in the course

Therefore, I've just switched to using screenshots (a href img that take you to model card):
Screenshot 2022-03-24 at 09 32 42

julien-c:
yes a screenshot is perfectly fine i think

PR #735, opened 2022-03-02
Try with write-all permissions

muellerzr:
Updates CI to work with write-all, which should bypass the selection issue


osanseviero (approved):
Thanks!

PR #742, opened 2022-03-04
Update rename/transfer repo rule docs

allendorf:
Updated to reflect new move repo authorizations in https://github.com/huggingface/moon-landing/pull/2204 (and fixed a few typos)

### DIFF
--- a/docs/hub/main.md
+++ b/docs/hub/main.md
@@ -98,14 +98,16 @@ Here is a video to learn more about it from our [course](http://hf.co/course)!
 
 You can go to the settings page in any repo to achieve this. Note that there are certain limitations in terms of use cases.
 
-Moving can be used for three use cases ✅ 
+Moving can be used in these use cases ✅ 
 - Renaming a repository within same user.
 - Renaming a repository within same organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
-- Transfering repository from user to an organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
+- Transferring repository from user to an organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
+- Transferring a repository from an organization to yourself. You must be part of the organization, and have "admin" rights in the organization.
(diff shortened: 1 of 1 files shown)


coyotte508 (approved):

docs/hub/main.md
```diff
@@ -98,14 +98,16 @@ Here is a video to learn more about it from our [course](http://hf.co/course)!
...
 - Renaming a repository within same organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
-- Transfering repository from user to an organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
+- Transferring repository from user to an organization. The user must be part of the organization and have "write" or "admin" rights in the organization.
+- Transferring a repository from an organization to yourself. You must be part of the organization, and have "admin" rights in the organization.
+- Transferring a repository from an organization to another organization. The user must have "admin" rights in the source organization and either "write" or "admin" rights in the target organization. 
 
 Moving does not work for ❌
-- Transfering a repository from an organization to another user or organization.
-- Transfering a repository from user A to user B.
+- Transferring a repository from an organization to another user or organization.
```
stevhliu, line 109: This line is a bit confusing since the use case above says you can do an org-to-org transfer if you have the appropriate rights, but here it says you can't. I think it'd be clearer here if we explicitly say you can't do an org-to-org transfer if you aren't a member of the target organization or have the rights, instead of letting the reader infer that themselves?

stevhliu (approved):
Thanks for updating the docs with the authorizations! I left one comment on clarifying when you can transfer a repository, otherwise everything else looks good 😊

docs/hub/main.md
allendorf: good catch! updated here: a1893e10e8857de9cff665b514a541a7abe27ae0 let me know what you think!

stevhliu: Looks great, thank you!!

Pierrci (approved):
LGTM too! 👍

osanseviero (approved):
Thanks! Looks good :rocket:

julien-c (approved):

LysandreJik (approved):

PR #748, opened 2022-03-08
Logging fix for `hf_api`, logging documentation

LysandreJik:
This fixes an issue with the logging of `hf_api`, replaces `logger.error` for deprecation warnings by a `FutureWarning`, and adds a small document regarding how to use logging with `huggingface_hub`.

cc @FrancescoSaverioZuppichini as you have raised the issue with `hf_api` this morning.

### DIFF
--- a/src/huggingface_hub/README.md
+++ b/src/huggingface_hub/README.md
@@ -372,3 +372,37 @@ API.
 ```python
 inference = InferenceApi("bert-base-uncased", task="feature-extraction", token=API_TOKEN)
 ```
+
+## Controlling the logging of `huggingface_hub`
+
+The `huggingface_hub` package exposes a `logging` utility to control the logging level of the package itself.
+You can import it as such:
+
+```py
+from huggingface_hub import logging
+```
+
+Then, you may define the verbosity in order to update the amount of logs you'll see:
+
+```python
+from huggingface_hub import logging
+
+logging.set_verbosity_error()
+logging.set_verbosity_warning()
+logging.set_verbosity_info()
+logging.set_verbosity_debug()
+
+logging.set_verbosity(...)
+```
+
+The levels should be understood as follows:
+
+- `error`: this will only show critical logs about usage which may result in an error or unexpected behavior.
+- `warning`: this will show logs which aren't critical, about usage which may result in unintended behavior.
+  Additionally, important informative logs may be shown.
+- `info`: this will show most logs, including some verbose logging regarding what is happening under the hood.
+  If something is behaving in an unexpected manner, we recommend switching the verbosity level to this in order
+  to get more information.
+- `debug`: this shows all logs, including some internal logs which may be used to track exactly what's happening
+  under the hood.
+
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -424,8 +424,9 @@ class HfApi:
 
         Throws: requests.exceptions.HTTPError if credentials are invalid
         """
-        logger.error(
-            "HfApi.login: This method is deprecated in favor of `set_access_token`."
+        warnings.warn(
+            "HfApi.login: This method is deprecated in favor of `set_access_token`.",
+            FutureWarning,
         )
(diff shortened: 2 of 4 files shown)


muellerzr (approved):
LG2M, should we also include a note on how to expose the logging functionality when running tests?

E.g.

```bash
pytest -sv tests --log-cli-level=INFO
```

osanseviero (approved):
LG!

src/huggingface_hub/README.md
```diff
@@ -371,4 +371,38 @@ API.
 
 ```python
 inference = InferenceApi("bert-base-uncased", task="feature-extraction", token=API_TOKEN)
-```
\ No newline at end of file
+```
+
+## Controlling the logging of `huggingface_hub`
```
adrinjalali, line 376: Should this information be in README? How are we going to create the documentation for this library? We probably need quite a few user guides and I think this should go in one of those, which probably should be located in the `/doc` folder.

src/huggingface_hub/hf_api.py
```diff
@@ -422,8 +424,9 @@ def login(self, username: str, password: str) -> str:
 
         Throws: requests.exceptions.HTTPError if credentials are invalid
         """
-        logging.error(
-            "HfApi.login: This method is deprecated in favor of `set_access_token`."
+        warnings.warn(
+            "HfApi.login: This method is deprecated in favor of `set_access_token`.",
```
adrinjalali, line 428: should we mention when it will be removed? Do we have a deprecation cycle policy?

src/huggingface_hub/hf_api.py
```diff
@@ -492,7 +495,9 @@ def logout(self, token: Optional[str] = None) -> None:
             token (``str``, `optional`):
                 Hugging Face token. Will default to the locally saved token if not provided.
         """
-        logging.error("This method is deprecated in favor of `unset_access_token`.")
+        warnings.warn(
```
adrinjalali, line 498: These warnings are not tested I think. Could you please add tests, similar to the one here: 


https://github.com/huggingface/huggingface_hub/pull/733/files#diff-5e02bb367c95034caad81bbd7a52410b9a05cb29878494288f74ef7805822679R153-R172

src/huggingface_hub/README.md
LysandreJik: I'm putting this here for now, but this should definitely be in a separate guide as outlined in https://github.com/huggingface/huggingface_hub/issues/754

src/huggingface_hub/hf_api.py
LysandreJik: I'm adding a mention to `v0.7`: we should remove this as soon as possible, but there are downstream libraries using it still, so it's important to give them time to switch.

There's no written deprecation cycle policy, but I agree it would be welcome.

src/huggingface_hub/hf_api.py
LysandreJik: Thanks for the link, added!

tests/test_hf_api.py
```diff
@@ -139,6 +139,25 @@ def test_login_cli(self):
             read_from_credential_store(USERNAME_PLACEHOLDER), (None, None)
         )
 
+    def test_login_deprecation_error(self):
+        with pytest.warns(
+            FutureWarning,
+            match=r"HfApi.login: This method is deprecated in favor of "
+            r"`set_access_token` and will be removed in v0.7.",
```
adrinjalali, line 146: I think this is not black formatted.

adrinjalali (approved):
Thanks @LysandreJik , other than the formatting LGTM.

tests/test_hf_api.py
LysandreJik: Actually, it is! I've split it in two to respect the 88 line char limit, black seems okay with it

LysandreJik: Black unindented it, the second line was previously indented at the same level as the `|` in `match=|`

adrinjalali: OK

Issue #332, opened 2021-09-09
Tracking integration for Object Detection


osanseviero:
## Tracking integration of task - Object Detection

Note that you're not expected to do all of the following steps. This PR helps track all the steps required to get a new task fully supported in the Hub 🔥

- [ ] Integration with Inference API. Select at least one of the following:
  - [ ] Added a `transformers` `pipeline`
  - [ ] Added to Community Inference API for 3rd party library
  - [ ] Added to Community Inference API for generic
- [ ] Added basic UI elements (icon, order specification, etc)
- [ ] Added a widget #74

Integration guide: https://hf.co/docs/hub/adding-a-task

osanseviero:
@Narsil would it be possible to enable this in the Inference API? The pipeline is submitted. 

cc @mishig25

mishig25:
https://github.com/huggingface/huggingface_hub/pull/247 should be almost ready. Once the inference api working for obj det, I'll make this from draft PR to regular one

mishig25:
HuggingPics did an awesome job of showcasing ViT and transformers image classification pipeline.
Should we do the equivalent of it for Detr and ovject detection? HuggingDetections ?
cc: @nateraw @julien-c

mishig25:
Now that Nicolas added `obj det` pipeline to new pipeline design, `obj det` task should be good from `transformers` side.
Any chance we can add a third-part obj det library to the hub? @osanseviero @nateraw 
From [github search](https://github.com/topics/object-detection), [open-mmlab/mmdetection](https://github.com/open-mmlab/mmdetection) looks particularly interesting

osanseviero:
As discussed in chat, for community we could begin adding `object-detection` to generic and try with the following libraries/models:
* https://github.com/ultralytics/yolov5 
* https://github.com/facebookresearch/detectron2
* LayoutParser

mishig25:
API-inference is working for obj-det now; thanks @Narsil!
The PR merged from `draft` to `regular`

mishig25:
Widget is live on the hub!

osanseviero:
That's great. I think the only point left over is integrating to community Inference API

PR #749, opened 2022-03-08
Add object-detection to community Inference API

osanseviero:
This is required to unblock #747 and as part of #332.


Narsil (approved):
LGTM

PR #756, opened 2022-03-09
[doc] org-cards

julien-c:


osanseviero:
Isn't `docs/assets/hub/org-card.png` too large? (1.6MB)

julien-c:
Oops yeah you're probably right. I shouldn't have rushed... 🙈

julien-c:
sorry i rushed this and merged, still interested in a quick review though

stevhliu:
Thanks for adding this! Two suggestions:

1. Would it be better if we showed an image of a more complete org card - like [Amazon's](https://huggingface.co/amazon) - to showcase what you can do in the org card (i.e., have links to a video, blog, and docs)? 
2. I think maybe we should also mention that you need to know some HTML to style the org card so users know what to expect ahead of time.

Otherwise, it looks great! I love the button on the top right of the org card that links to the docs 👍

julien-c:
my hunch is that most users are just going to write some markdown text and not really bother with complicated HTML, wdyt @osanseviero? (that's why I added a simpler example as the first example)

But yeah you're right we can add a sentence or paragraph about being able to write HTML inside the .md file, do you want to open a PR?

osanseviero:
(Note that Amazon's card does not really play the video, it just redirects to YT.) 

re: used image. I think as-is is fine. Note we link to the more complex org cards right below, so it's up to the users if they want to investigate more complex org cards.

I agree some additional info on writing HTML would be useful!

PR #757, opened 2022-03-09
New release (requirements are still too strict.)

Narsil:
Missed the real `requirements.txt` earlier.

Fixes #758

### DIFF
--- a/api-inference-community/requirements.txt
+++ b/api-inference-community/requirements.txt
@@ -1,7 +1,7 @@
 starlette>=0.14.2
 numpy>=1.18.0
-pydantic==1.8.1
+pydantic>=1.8.1
 parameterized>=0.8.1
 pillow>=8.2.0
 huggingface_hub>=0.0.12
-datasets==1.10
+datasets>=1.10
--- a/api-inference-community/requirements.txt
+++ b/api-inference-community/requirements.txt
@@ -1,7 +1,7 @@
-starlette==0.14.2
+starlette>=0.14.2
 numpy>=1.18.0
 pydantic==1.8.1
-parameterized==0.8.1
-pillow==8.2.0
+parameterized>=0.8.1
+pillow>=8.2.0
 huggingface_hub>=0.0.12
 datasets==1.10
--- a/api-inference-community/setup.py
+++ b/api-inference-community/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup
 
 setup(
     name="api_inference_community",
-    version="0.0.22",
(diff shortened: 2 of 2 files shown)


api-inference-community/setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name="api_inference_community",
-    version="0.0.21",
+    version="0.0.23",
```
frgfm, line 6: Shouldn't it be `0.0.22` ?

api-inference-community/requirements.txt
```diff
@@ -1,7 +1,7 @@
...
 numpy>=1.18.0
-pydantic==1.8.1
-parameterized==0.8.1
-pillow==8.2.0
+pydantic>=1.8.1
+parameterized>=0.8.1
+pillow>=8.2.0
 huggingface_hub>=0.0.12
-datasets==1.10
+datasets>=1.10
```
frgfm, line 7: apart from pillow, are the other losse specifiers required?

frgfm (commented):
Thanks for iterating on #747 :pray: Quick comments on the PR :point_down:

api-inference-community/setup.py
Narsil: No it's 0.0.23 Since it's 0.0 I publish a bit faster than necessary :)

api-inference-community/requirements.txt
Narsil: well this is a library, so everything should be smooth to allow users to get whatever they want, it's in the application layer you want to pin specific versions to get reproducible builds.

Technically to be 100% safe you should always do =>Major,<Major+1 to avoid breaking changes when they are released.

But since we use such a shallow surface I think we're ok not safeguarding us from this.

Narsil: Just to emphasize, the real culprit here was not really Pillow==8.2.0 (only to some extent).

The biggest culprit is `datasets==1.10` because datasets itself restricts `huggingface_hub<0.1` which is now at `0.4`  So This `huggingface_hub` incompatibility forces `pip` to look for earlier versions of packages(not entirely sure about the traversal order). Pip will attempt a bunch of versions before dying on `Pillow` strict requirements, but modifying only `Pillow` doesn't fix the incompatibility issues, it would just send the incompatibility somewhere else. We could have fixed them one by one until fixing `datasets` in the end (it's the first issue reported by pip when doing the install).

osanseviero (approved):
Looks good!

PR #760, opened 2022-03-10
Fix pytorch and tensorflow python matrix

osanseviero:
The GH Actions workflow makes reference to the python-version but there is no matrix being set. I'm not sure what's the behavior right now, but I think this is a bug

cc @merveenoyan @nateraw @muellerzr

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -50,6 +50,9 @@ jobs:
 
   build_pytorch:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.6", "3.9"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


muellerzr:
Makes sense to me, great catch!

Do we think git-lfs needs a min/max python version check? Or not so much since that's more an apt package

osanseviero:
I agree with :smile: 

> not so much since that's more an apt package

LysandreJik:
Looks good, but I think there's quite a few failures in Python 3.6 for TensorFlow

osanseviero:
@merveenoyan, I think the parameter `show_layer_activations` in `plot_model` is kinda recent (August), so it won't work with older versions of TensorFlow and Python. Is that expected?

LysandreJik (approved):
LGTM :)

PR #761, opened 2022-03-10
Added tempfile for tensorboard logs in tensorboard tests in test_keras_integration.py

merveenoyan:
TensorBoard logs are in parent directory of the repository so they weren't removed along with repository, I added lines to remove them so that they don't bug tensorboard tests.

### DIFF
--- a/tests/test_keras_integration.py
+++ b/tests/test_keras_integration.py
@@ -370,8 +370,8 @@ class HubKerasSequentialTest(HubMixingTestKeras):
 
     @retry_endpoint
     def test_push_to_hub_tensorboard(self):
-        os.makedirs(f"{WORKING_REPO_DIR}/tb_log_dir")
-        with open(f"{WORKING_REPO_DIR}/tb_log_dir/tensorboard.txt", "w") as fp:
+        os.makedirs(f"{WORKING_REPO_DIR}/log_dir")
+        with open(f"{WORKING_REPO_DIR}/log_dir/tensorboard.txt", "w") as fp:
             fp.write("Keras FTW")
         REPO_NAME = repo_name("PUSH_TO_HUB")
         model = self.model_init()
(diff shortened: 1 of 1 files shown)


osanseviero:
Let us know whenever it's ready :smile: also try to re-run the workflow few times to make sure it works.

merveenoyan:
@osanseviero I created tensorboard log directories under the parent working directory of the repository and not inside the repository itself, and AFAIK only repo gets removed so the logs remained and hence the error about how the log directory already exists. When I do makedirs (inside working directory/repo_name) before I call push_to_hub_keras the repo is not empty so it doesn't clone there.

adrinjalali:
I guess here, and in a few other places in the test suite, we actually create files/folders under the working directory. Wouldn't it better to do all of those operations under `/tmp` and some temporary file/folder names, that never clash between tests?

merveenoyan:
@adrinjalali I actually saw couple of files that exist under tests/fixtures that were left (I don't know if it's on purpose). I removed them directly instead, I can have another temporary directory if you want me to.

adrinjalali:
Yes I've also seen files under `fixture`. And they also sometimes end up with very odd permissions. I think the test suite shouldn't create/touch any files under working folder.

Probably a better solution is to create everything (cloning repo, creating tensorboard logs, etc) under a folder you can get from python's tempdir module: https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

merveenoyan:
something along the lines of:

```
with tempfile.TemporaryDirectory() as tmpdirname:
        
            with open(f"{tmpdirname}/tensorboard.txt", "w") as fp:
                fp.write("Keras FTW")
```

osanseviero (approved):
Looks great! Thanks for the PR :rocket: 

Would you like to rename the PR?

PR #763, opened 2022-03-10
De-emphasize Streamlit version from here?

julien-c:


docs/hub/spaces.md
```diff
@@ -28,7 +28,7 @@ Each Spaces environment is limited to 16GB RAM and 8 CPU cores. Organization [su
 
 ## Streamlit and Gradio
 
-Spaces support [Streamlit](https://streamlit.io/) and [Gradio](https://gradio.app/) for quickly building apps in Python. The default environment runs version `"1.0.0"` of Streamlit and the latest version of Gradio. We recommend you try both because they're really awesome! Here are some of our thoughts on Streamlit and Gradio:
+Spaces support [Streamlit](https://streamlit.io/) and [Gradio](https://gradio.app/) for quickly building apps in Python. The default environment runs whichever latest version of Streamlit and Gradio is the current one at the time you create your Space repo. We recommend you try both because they're really awesome! Here are some of our thoughts on Streamlit and Gradio:
```
adrinjalali, line 31: do they get set at the time of creating the repo, or whenever a new docker image is built (which can be after every commit?). Also, can they set the version if they have a specific requirement?

(nit: wrapping the lines around instead of a single line paragraph would also be nice :P )

julien-c: set at repo creation

and users can override it with whatever version is supported (i think this is already documented, @cbensimon?)

osanseviero (approved):
Seems good!

docs/hub/spaces.md
stevhliu, line 31: We have it documented [here](https://huggingface.co/docs/hub/spaces#troubleshoot) (only for Streamlit), and then in the Reference below, we list all the supported versions. Small suggestion, otherwise looks good!

```suggestion
Spaces support [Streamlit](https://streamlit.io/) and [Gradio](https://gradio.app/) for quickly building apps in Python. The default environment runs the latest version of Streamlit and Gradio at the time the Space repo is created. We recommend you try both because they're really awesome! Here are some of our thoughts on Streamlit and Gradio:
```

stevhliu (approved):
Looks good! 👍

docs/hub/spaces.md
cbensimon: Yes, this is exact. And yes, already documented but only for Streamlit currently. I'll update it

LysandreJik (approved):

PR #765, opened 2022-03-10
HTML in org card

stevhliu:
Lets users know they can write simple Markdown text or more complicated HTML to style their org card if they want to (#756 ).


osanseviero (approved):
Thanks!

julien-c (approved):

PR #780, opened 2022-03-18
Disable logging in with organization token

merveenoyan:
Hello 👋  I dived into code to see where the user might login and where tokens are checked. I've seen three cases:
- When the token in the folder is retrieved (when user doesn't give it)
- When user logs in through CLI or notebook (_login)
- When user passes the token directly (which are checked inside repository related functions that are used in places like mixins) like so: 
`if token is None:
            token = self._validate_or_retrieve_token()
        elif isinstance(token, str):
            if token.startswith("api_org"):
                raise ValueError("You must use your personal account token for login.")
            elif not self._is_valid_token(token):
                if self._is_valid_token(name):
                    warnings.warn(
                        "`create_repo` now takes `token` as an optional positional argument. "
                        "Be sure to adapt your code!",
                        FutureWarning,
                    )
                    token, name = name, token
                else:
                    raise ValueError("Invalid token passed!")`

So I added checks there and wrote tests for those three cases too.
cc: @osanseviero

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -477,7 +477,6 @@ class HfApi:
             raise ValueError(
                 "You need to pass a valid `token` or login by using `huggingface-cli login`"
             )
-
         path = f"{self.endpoint}/api/whoami-v2"
         r = requests.get(path, headers={"authorization": f"Bearer {token}"})
         try:
@@ -510,8 +509,11 @@ class HfApi:
                     "You need to provide a `token` or be logged in to Hugging Face with "
                     "`huggingface-cli login`."
                 )
-        elif not self._is_valid_token(token):
-            raise ValueError("Invalid token passed!")
+        elif isinstance(token, str):
+            if token.startswith("api_org"):
+                raise ValueError("You must use your personal account token for login.")
+            elif not self._is_valid_token(token):
+                raise ValueError("Invalid token passed!")
         return token
 
     def logout(self, token: Optional[str] = None) -> None:
@@ -1069,16 +1071,19 @@ class HfApi:
         path = f"{self.endpoint}/api/repos/create"
         if token is None:
             token = self._validate_or_retrieve_token()
-        elif not self._is_valid_token(token):
-            if self._is_valid_token(name):
-                warnings.warn(
-                    "`create_repo` now takes `token` as an optional positional argument. "
-                    "Be sure to adapt your code!",
-                    FutureWarning,
-                )
-                token, name = name, token
-            else:
-                raise ValueError("Invalid token passed!")
+        elif isinstance(token, str):
+            if token.startswith("api_org"):
+                raise ValueError("You must use your personal account token for login.")
+            elif not self._is_valid_token(token):
+                if self._is_valid_token(name):
+                    warnings.warn(
+                        "`create_repo` now takes `token` as an optional positional argument. "
+                        "Be sure to adapt your code!",
+                        FutureWarning,
+                    )
+                    token, name = name, token
+                else:
+                    raise ValueError("Invalid token passed!")
 
         checked_name = repo_type_and_id_from_hf_id(name)
 
@@ -1188,16 +1193,19 @@ class HfApi:
         path = f"{self.endpoint}/api/repos/delete"
         if token is None:
             token = self._validate_or_retrieve_token()
-        elif not self._is_valid_token(token):
-            if self._is_valid_token(name):
-                warnings.warn(
-                    "`delete_repo` now takes `token` as an optional positional argument. "
-                    "Be sure to adapt your code!",
-                    FutureWarning,
-                )
-                token, name = name, token
-            else:
-                raise ValueError("Invalid token passed!")
+        elif isinstance(token, str):
+            if token.startswith("api_org"):
+                raise ValueError("You must use your personal account token for login.")
+            elif not self._is_valid_token(token):
+                if self._is_valid_token(name):
+                    warnings.warn(
+                        "`delete_repo` now takes `token` as an optional positional argument. "
+                        "Be sure to adapt your code!",
+                        FutureWarning,
+                    )
+                    token, name = name, token
+                else:
+                    raise ValueError("Invalid token passed!")
 
         checked_name = repo_type_and_id_from_hf_id(name)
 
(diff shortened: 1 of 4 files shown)


adrinjalali (changes requested):
Thanks for the fix @merveenoyan . 

Now that you're touching these bits, I realize there's quite a few repetitions. Could you maybe refactor those bits by creating a single `_get_token` utility function which us used throughout the codebase instead of having the same logic multiple times?

I'm not attached to the `_get_token` function name though.

merveenoyan:
I refactored & tested but will check if there's nits to do on Monday 🙂

src/huggingface_hub/hf_api.py
```diff
@@ -499,20 +498,36 @@ def _is_valid_token(self, token: str):
...
+            if self._is_valid_token(name):
+                warnings.warn(
+                    f"`{function_name}` now takes `token` as an optional positional argument. "
+                    "Be sure to adapt your code!",
+                    FutureWarning,
+                )
+                token, name = name, token
+        if isinstance(token, str):
+            if token.startswith("api_org"):
+                raise ValueError("You must use your personal account token for login.")
```
osanseviero, line 524: This is not just for login, since users might be passing the token as a param

src/huggingface_hub/hf_api.py
```diff
@@ -499,20 +498,36 @@ def _is_valid_token(self, token: str):
...
+            if not self._is_valid_token(token):
+                raise ValueError("Invalid token passed!")
+        if name is not None:
+            return token, name
+        else:
+            return token
```
osanseviero, line 530: I think you can always return token, name. It would then return None for the second case for name which is fine and will make  usage easier

src/huggingface_hub/hf_api.py
```diff
@@ -1068,17 +1083,11 @@ def create_repo(
...
-                    FutureWarning,
-                )
-                token, name = name, token
-            else:
-                raise ValueError("Invalid token passed!")
+            token = self._validate_or_retrieve_token(function_name="create_repo")
+        elif isinstance(token, str):
+            token, name = self._validate_or_retrieve_token(
+                token, name, function_name="create_repo"
+            )
```
osanseviero, line 1090: I think this can be simplified even further by just doing this:
```
token, name = self._validate_or_retrieve_token(
                token, name, function_name="create_repo"
            )
```
`_validate_or_retrieve_token` already handles case in which token is None or string, so no need to do the check here. Same applies for other functions

src/huggingface_hub/hf_api.py
```diff
@@ -1423,22 +1422,25 @@ def upload_file(
...
-                )
-                token, path_or_fileobj, path_in_repo, repo_id = (
-                    path_or_fileobj,
-                    path_in_repo,
-                    repo_id,
-                    token,
-                )
-            else:
-                raise ValueError("Invalid token passed!")
+            token = self._validate_or_retrieve_token(function_name="upload_file")
```
osanseviero, line 1425: Should this be updated since the logic below is in `validate_or_retrieve_token`?

src/huggingface_hub/hf_api.py
adrinjalali: I think we can move `_validate_or_retrieve_token` outside this class, move it to a utils module and use it also in the other file.

merveenoyan: @adrinjalali I recently pushed a new change but didn't do this yet, there's a repetition in `user.py` if we don't do this, I agree. Will push that soon.

merveenoyan: @adrinjalali it doesn't make any sense to put `_validate_or_retrieve_token` under endpoint helpers, the function depends heavily on other classes and functions in `hf_api.py`, so instead I called it in `user.py` like `hf_api._validate_or_retrieve_token`.

merveenoyan:
I don't see anything else to refactor. I just have one question.

What is the reason behind this in upload_file? (below is initial version)
```
elif not self._is_valid_token(token):
 if self._is_valid_token(path_or_fileobj):
```
Do people pass tokens in paths or files? I refactored assuming that. (it didn't make much sense to me though) @osanseviero

adrinjalali:
> What is the reason behind this in upload_file? (below is initial version)

I believe this is there only for backward compatibility. At some point the order of args were changed, and therefore this makes sure the older order is still accepted.

src/huggingface_hub/hf_api.py
```diff
@@ -499,20 +498,34 @@ def _is_valid_token(self, token: str):
...
-        """
-        Either retrieves stored token or validates passed token.
-        """
+    def _validate_or_retrieve_token(
+        self,
+        token: Optional[str] = None,
+        name: Optional[str] = None,
+        function_name: Optional[str] = None,
+    ):
         if token is None or token is True:
```
adrinjalali, line 507: Could you please add a docstring here with args and return values explain?

src/huggingface_hub/hf_api.py
```diff
@@ -499,20 +498,34 @@ def _is_valid_token(self, token: str):
...
                 raise EnvironmentError(
                     "You need to provide a `token` or be logged in to Hugging Face with "
                     "`huggingface-cli login`."
                 )
-        elif not self._is_valid_token(token):
-            raise ValueError("Invalid token passed!")
-        return token
+        if name is not None:
+            if self._is_valid_token(name):
+                warnings.warn(
```
adrinjalali, line 516: this needs to be removed in the future.

```suggestion
                # TODO(0.6) REMOVE
                warnings.warn(
```

src/huggingface_hub/hf_api.py
```diff
@@ -499,20 +498,34 @@ def _is_valid_token(self, token: str):
...
+                    FutureWarning,
+                )
+                token, name = name, token
+        if isinstance(token, str):
+            if token.startswith("api_org"):
+                raise ValueError("You must use your personal account token.")
+            if not self._is_valid_token(token):
+                raise ValueError("Invalid token passed!")
+
+        return token, name
```
adrinjalali, line 528: We only need to return name if name is provided

```suggestion
        return token if name is None else (token, name)
```

adrinjalali (commented):
Otherwise LGTM

src/huggingface_hub/hf_api.py
osanseviero: I think the existing solution simplifies things since then the function will always return a tuple. This consistency will reduce the amount of handling needed in downstream functions

src/huggingface_hub/hf_api.py
```diff
@@ -553,19 +552,21 @@ def _is_valid_token(self, token: str):
...
-            `str`: The valid token
-
-        Raises:
-            `ValueError`: if the token is invalid.
+            token (``str``, `optional`):
+                Hugging Face token. Will default to the locally saved token if not provided.
+            name (``str``, `optional`):
+                Name of the repository.
+            function_name (``str``, `optional`):
+                If called from a function, name of that function for deprecation warning.
```
adrinjalali, line 569: Missing `Returns` and `Raises` sections.

It would also be nice to explain why `name` and `function_name` are required and that they'd be removed in a future version (in v0.7 maybe?)

merveenoyan: Done!

adrinjalali (approved):
This ended up being quite a bit more than the original issue (it's probably my fault 🙈 ). But I really like the improvement here. Thanks @merveenoyan

adrinjalali:
One could argue that the refactoring could or should have been a separate PR. But I'm glad this is done now.

osanseviero:
Note that this PR reverts docstrings changes needed for the new documentation (e.g. using `*optional*`). You can see it in the diff

PR #782, opened 2022-03-18
API reference documentation

LysandreJik:
Work in progress aiming to supply an API reference for the huggingface_hub python library.

Done from a feature branch as many may collaborate on the branch and it will be easier to setup the GitHub Actions CI in a first step.


adrinjalali:
I would appreciate if we setup the CI such that it doesn't require a branch from the same repo, and having the initial PR from a separate repo would make sure the CI runs fine.

If we expect this to be something that multiple people work one, we could have a branch in the repo for it, and have PRs to that branch instead of `main`, and once we're comfortable with the branch, we can have a single PR to merge it with `main`. This way multiple people could also send separate PRs to the same branch and there wouldn't be conflicts between their work for working on the same branch.

LysandreJik:
Regarding the CI: yes, that is the goal, but it is not ready right now and needs iteration, which I hope to do right here.

Regarding the second point, I'm not sure I understand; that is absolutely the goal as well. I am creating an initial commit on a feature branch, which takes care of most of the work. As you can see, it is not yet ready as the CI did not run, a very important step in order to collaborate on this specific branch.

We can then work on that branch through PRs as you mention, once I set this branch as ready for contributions.

How would you have managed this differently?

adrinjalali:
> We should be good to go to work on this feature branch now. 

Not sure if I understand. So you mean we can now send PRs to this branch?

LysandreJik:
Yes, that's what I mean!

LysandreJik:
Opened a first PR here which takes care of all the remaining methods mentioned above: https://github.com/huggingface/huggingface_hub/pull/799

docs/source/package_reference/hf_api.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+[[autodoc]] HfApi.list_datasets
+
+[[autodoc]] HfApi.dataset_info
+
+[[autodoc]] HfApi.list_metrics
+
+[[autodoc]] HfApi.get_model_tags
+
+[[autodoc]] HfApi.get_dataset_tags
```
adrinjalali, line 23: does every method of a class need to be listed her? I would rather if we could just list the class and all public methods are then rendered.

docs/source/package_reference/hf_api.mdx
```diff
@@ -0,0 +1,47 @@
...
+[[autodoc]] HfApi.delete_file
+
+[[autodoc]] HfApi.get_full_repo_name
+
+## Hugging Face local storage
+
+`huggingface_hub` stores the authentication information locally so that it may be re-used in subsequent
+methods.
+
+It does this using the `HfFolder` utility, which saves data at the root of the user.
```
adrinjalali, line 44: hyperlink to `HfFolder`?

adrinjalali (commented):
Same with long lines and hyperlink issues here.

docs/source/package_reference/hf_api.mdx
LysandreJik: Good catch, I actually just wanted to list the methods accessible at the root, not the `HfApi` itself. Will adapt!

docs/source/package_reference/mixins.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+[[autodoc]] ModelHubMixin
+
+### Keras
+
+[[autodoc]] from_pretrained_keras
+
+[[autodoc]] push_to_hub_keras
+
+[[autodoc]] save_pretrained_keras
```
osanseviero, line 18: I think this should also have `KerasModelHubMixin` for users using the subclassing TF API

src/huggingface_hub/hf_api.py
```diff
@@ -582,91 +662,121 @@ def list_models(
         Get the public list of all the models on huggingface.co
 
         Args:
-            filter (:class:`ModelFilter` or :obj:`str` or :class:`Iterable`, `optional`):
-                A string or `ModelFilter` which can be used to identify models on the hub.
+            filter ([`ModelFilter`] or `str` or `Iterable`, *optional*):
+                A string or [`ModelFilter`] which can be used to identify models
+                on the hub.
```
osanseviero, line 667:
```suggestion
                on the Hub.
```

LysandreJik:
@stevhliu, @osanseviero, could you take a look at this PR? Would love to hear your feedback about the content.

This will likely be put at huggingface.co/docs/python-wrapper

julien-c:
> This will likely be put at https://huggingface.co/docs/python-wrapper

For consistency I would rather serve the doc at `https://huggingface.co/docs/huggingface_hub`, but add "Client library" or "Python wrapper" to the titles etc so it's not confusing. Thoughts?

LysandreJik:
This PR should be ready to be merged.

When merged, I would like to open issues for the following:
- On `doc-builder`, open issues to ask for support of a few items:
    - Different colors for `<Tip>`, which should have different severity levels
    - Mention that in-parameter-docstrings doesn't support code samples
    - Raises and Yields are not rendered correctly
    - `versionadded` and other directives that would have been useful
- On `huggingface_hub`, open issues to track the following:
    - Lack of images right now, and definition of a way to provide images to the hub

If this looks good to you, @osanseviero @stevhliu, I'll merge this PR.

docs/source/_toctree.yml
```diff
@@ -0,0 +1,24 @@
+- sections:
```
coyotte508, line 1: I think a `title` should also be added here (like below with `title: "Reference`).

image

https://moon-ci-docs.huggingface.co/docs/huggingface_hub/pr_782/en/index

LysandreJik: Indeed, good catch! Will add this.

adrinjalali (approved):
Happy to merge and improve on smaller PRs. This is already too large.

LysandreJik:
Thanks all for the reviews, all green, merging this PR! Will open a few relevant issues as discussed above.

PR #787, opened 2022-03-22
CI Error on FutureWarning

adrinjalali:
This PR makes the CI to fail on `FutureWarning` so that our test suite won't use our own deprecated API and our codebase use new API from our dependencies.

### DIFF
--- a/tests/test_snapshot_download.py
+++ b/tests/test_snapshot_download.py
@@ -171,7 +171,7 @@ class SnapshotDownloadTests(unittest.TestCase):
             self.assertTrue(self.second_commit_hash in storage_folder)
 
         self._api.update_repo_visibility(
-            token=self._token, name=REPO_NAME, private=False
+            token=self._token, repo_id=REPO_NAME, private=False
         )
 
     def test_download_model_local_only(self):
--- a/tests/test_snapshot_download.py
+++ b/tests/test_snapshot_download.py
@@ -111,7 +111,7 @@ class SnapshotDownloadTests(unittest.TestCase):
 
     def test_download_private_model(self):
         self._api.update_repo_visibility(
-            token=self._token, name=REPO_NAME, private=True
+            token=self._token, repo_id=REPO_NAME, private=True
         )
 
         # Test download fails without token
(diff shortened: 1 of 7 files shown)


adrinjalali:
Would be nice to have this in so that we can merge https://github.com/huggingface/huggingface_hub/pull/745 after

LysandreJik (approved):
This looks good to me, thanks for working on it @adrinjalali!

tests/test_snapshot_download.py
```diff
@@ -27,7 +27,8 @@ def setUpClass(cls):
         """
         Share this valid token in all tests below.
         """
-        cls._token = cls._api.login(username=USER, password=PASS)
+        cls._token = TOKEN
+        cls._api.set_access_token(TOKEN)
```
osanseviero, line 31: Do we really want to do this? This means that if someone runs the test locally, they will be changing their credential store no?

osanseviero (commented):
Looks great, thanks for this change! I left a small question

tests/test_snapshot_download.py
julien-c: i think that's already the case anyways for other tests no?

adrinjalali: Yes it does, but that's already the case (which I think shouldn't be). I'll open a separate issue to discuss that.

osanseviero: I don't think that's true. `login` does not update the git credential store, it just gives you the token.

adrinjalali: `login` stores `username` and `password`, just not the token. I don't think either of them should: https://github.com/huggingface/huggingface_hub/issues/796

```python
    def login(self, username: str, password: str) -> str:
        """
        Call HF API to sign in a user and get a token if credentials are valid.

        Outputs: token if credentials are valid

        Throws: requests.exceptions.HTTPError if credentials are invalid
        """
        warnings.warn(
            "HfApi.login: This method is deprecated in favor of `set_access_token`"
            " and will be removed in v0.7.",
            FutureWarning,
        )
        path = f"{self.endpoint}/api/login"
        r = requests.post(path, json={"username": username, "password": password})
        r.raise_for_status()
        d = r.json()

        write_to_credential_store(username, password)
        return d["token"]

    def set_access_token(access_token: str):
        write_to_credential_store(USERNAME_PLACEHOLDER, access_token)

```

osanseviero: Ah thank you for pointing it out!

PR #788, opened 2022-03-23
Test fixes for audio inference API call

osanseviero:
The test was failing as the model called through the inference API was not pinned, hence it was taking some time to load and returning the load message. This change uses a pinned test model.


adrinjalali (approved):
Thanks @osanseviero

PR #783, opened 2022-03-18
MNT more informative message on error in `Hf.Api.delete_repo`

adrinjalali:
Fixes https://github.com/huggingface/huggingface_hub/issues/751

cc @coyotte508 @LysandreJik

PR #792, opened 2022-03-23
Uniformize `hf_api` a bit and add support for Spaces

julien-c:


adrinjalali (commented):
Some of the changes here are breaking changes. I think we should introduce a deprecation cycle like the other PRs if we're breaking API.

julien-c:
@adrinjalali I think only two lines in this diff are potential breaking changes, so I'll revert them. Other than that i don't think there any BC, but I'll check

src/huggingface_hub/constants.py
```diff
@@ -30,6 +30,7 @@
 
 
 HUGGINGFACE_CO_URL_TEMPLATE = ENDPOINT + "/{repo_id}/resolve/{revision}/{filename}"
+HUGGINGFACE_HEADER_X_REPO_COMMIT = "X-Repo-Commit"
```
LysandreJik, line 33: Could you add a comment specifying what that will be used for?

src/huggingface_hub/hf_api.py
```diff
@@ -972,56 +1004,103 @@ def model_info(
...
         path = (
-            f"{self.endpoint}/api/datasets/{repo_id}"
+            f"{self.endpoint}/api/spaces/{repo_id}"
             if revision is None
-            else f"{self.endpoint}/api/datasets/{repo_id}/revision/{revision}"
+            else f"{self.endpoint}/api/spaces/{repo_id}/revision/{revision}"
         )
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        params = {"full": "true"}
-        r = requests.get(path, headers=headers, params=params, timeout=timeout)
```
LysandreJik, line 1021: The `full: true` param is not needed anymore?

src/huggingface_hub/hf_api.py
```diff
@@ -972,56 +1004,103 @@ def model_info(
...
+        r = requests.get(path, headers=headers, timeout=timeout)
         r.raise_for_status()
         d = r.json()
-        return DatasetInfo(**d)
+        return SpaceInfo(**d)
+
+    def repo_info(
```
LysandreJik, line 1059: Cool addition!

src/huggingface_hub/hf_api.py
```diff
@@ -142,39 +143,33 @@ def __repr__(self):
         return f"{self.__class__.__name__}({', '.join(items)})"
 
 
-class ModelFile:
+class ModelFile(RepoFile):
     """
-    Data structure that represents a public file inside a model, accessible from huggingface.co
+    Data structure that represents a public file inside a repo, accessible from huggingface.co
```
LysandreJik, line 148: If you're updating the docstring from `model` to `repo`, you can remove this docstring altogether as it'll re-use the inherited `RepoFile`. You could also revert the `model` -> `repo` change to have a more explicit docstring.

LysandreJik: (especially since `dataset` and `Space` are kept in the respective `XXXFile` classes defined below)

LysandreJik (approved):
Other than the `sha` addition as a kwarg to `DatasetInfo` to correspond better to the `ModelInfo`, I don't see a breaking change; did I miss anything @adrinjalali?

src/huggingface_hub/hf_api.py
```diff
@@ -128,12 +128,13 @@ def repo_type_and_id_from_hf_id(hf_id: str):
     return repo_type, namespace, repo_id
 
 
-class RepoObj:
+class RepoFile:
```
adrinjalali, line 131: This is a breaking change, the statement `import RepoObj from huggingface_hub.hf_api` has no indication that `RepoObj` is a private class and it's being renamed here.

If users are not supposed to import from `huggingface_hub.hf_api`, then we should rename this file to `_hf_api.py` and deprecate the old import pattern and let users import only from `huggingface_hub` root module.

If you think this class is never used by any of our users then sure, we can rename it, but we should then rename the file to make sure people import things that we think are *public* API.

src/huggingface_hub/hf_api.py
```diff
@@ -142,39 +143,33 @@ def __repr__(self):
...
-            setattr(self, k, v)
+    pass
 
-    def __repr__(self):
-        items = (f"{k}='{v}'" for k, v in self.__dict__.items())
-        return f"{self.__class__.__name__}({', '.join(items)})"
+
+class SpaceFile(RepoFile):
+    """
+    Data structure that represents a public file inside a Space, accessible from huggingface.co
```
adrinjalali, line 164: could we document what this data structure includes and what they mean?

src/huggingface_hub/hf_api.py
```diff
@@ -264,6 +261,41 @@ def __str__(self):
         return r
 
 
+class SpaceInfo:
+    """
+    Info about a Space accessible from huggingface.co
+    """
```
adrinjalali, line 267: this docstring needs to be expanded to document the `__init__` args.

src/huggingface_hub/hf_api.py
```diff
@@ -972,56 +1004,103 @@ def model_info(
...
+        self,
+        repo_id: str,
+        revision: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        token: Optional[str] = None,
+        timeout: Optional[float] = None,
+    ) -> List[str]:
+        """
+        Get the list of files in a given repo.
+        """
```
adrinjalali, line 1095: same here with documentation.

adrinjalali (commented):
We also need to document the new methods in the README here.

src/huggingface_hub/hf_api.py
julien-c: yes I'm 99.9% sure no user ever used this, it hasn't been used for a long time.

That being said, no objection to renaming the file to `_hf_api.py` if it's the canonical way to make a internal file "private" (not consistent w/ what we do in other libraries but no big deal)

src/huggingface_hub/hf_api.py
```diff
@@ -972,56 +1004,103 @@ def model_info(
...
+        r = requests.get(path, headers=headers, timeout=timeout)
         r.raise_for_status()
         d = r.json()
-        return DatasetInfo(**d)
+        return SpaceInfo(**d)
+
+    def repo_info(
+        self,
+        repo_id: str,
+        revision: Optional[str] = None,
```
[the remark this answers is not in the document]
julien-c: Cool!

src/huggingface_hub/hf_api.py
julien-c: it wasn't used for this endpoint on the server side

src/huggingface_hub/constants.py
julien-c: actually let's just remove it from this PR and i'll add to another one

julien-c, line 33:
```suggestion
```

src/huggingface_hub/hf_api.py
```diff
@@ -972,56 +1004,103 @@ def model_info(
...
+        repo_type: Optional[str] = None,
+        token: Optional[str] = None,
+        timeout: Optional[float] = None,
+    ) -> Union[ModelInfo, DatasetInfo, SpaceInfo]:
+        """
+        Get the info object for a given repo of a given type.
+        """
```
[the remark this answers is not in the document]
julien-c: Ok will do

src/huggingface_hub/hf_api.py
julien-c: This one is really an internal "dataclass" like feature, almost like a dict. Do you think it's still useful to document the attributes in the docstring for this?

src/huggingface_hub/hf_api.py
adrinjalali: Cool, we can leave the renaming of the file to a separate PR.

src/huggingface_hub/hf_api.py
adrinjalali: It's not just for our end users. Things become more maintainable for other contributors if they're better documented. That's why I ended up writing documentation for `_request_with_retry` [here](https://github.com/huggingface/huggingface_hub/pull/731/files#diff-765507f4602931b08aa20d0177611d5da7d5c7805882b0a79c81b5ed72c9472aR301) for example.

src/huggingface_hub/hf_api.py
LysandreJik: Since this was just moved around, this is handled by this PR already: https://github.com/huggingface/huggingface_hub/pull/782/files#diff-71a3e23532e1355e79edd51b351bd514f0aa55b946e4c530b0c95f647773ac5dR1097

LysandreJik:
Added documentation for the methods that weren't touched by https://github.com/huggingface/huggingface_hub/pull/782 so that everything is in the same format.

julien-c:
BTW feel free to rebase this on `main` if you'd like!

LysandreJik:
Yes, will now rebase on `main` as the big doc PR was just merged.

src/huggingface_hub/hf_api.py
```diff
@@ -131,45 +131,20 @@ def repo_type_and_id_from_hf_id(hf_id: str):
...
+        rfilename (str):
+            file name, relative to the repo root. This is the only attribute
+            that's guaranteed to be here, but under certain conditions there can
+            certain other stuff.
     """
 
-    def __init__(self, rfilename: str, **kwargs):
-        self.rfilename = rfilename  # filename relative to the dataset root
+    def __init__(self, *, rfilename: str, **kwargs):
+        self.rfilename = rfilename  # filename relative to the repo root
```
LysandreJik, line 147: I think here we shouldn't put the `*` as:
- There is no decorator
- `rfilename` doesn't have a default so isn't a kwarg

Ok for you if I remove it?

src/huggingface_hub/hf_api.py
```diff
@@ -274,6 +255,57 @@ def __str__(self):
...
+            list of files that constitute the Space
+        private (`bool`, *optional*):
+            is the repo private
+        author (`str`, *optional*):
+            repo author
+    """
+
+    def __init__(
+        self,
+        *,
```
LysandreJik, line 282: Should have the decorator as well, will add it

src/huggingface_hub/hf_api.py
LysandreJik: (Removed in e31072b, let me know if you want to keep it)

src/huggingface_hub/hf_api.py
LysandreJik: (added in e31072b)

src/huggingface_hub/hf_api.py
```diff
@@ -180,7 +155,7 @@ def __repr__(self):
 
 class ModelInfo:
     """
-    Info about a public model accessible from huggingface.co
+    Info about a model accessible from huggingface.co
```
adrinjalali, line 158: Since we're touching this class, I think it'd be nice to add the info about init args here.

src/huggingface_hub/hf_api.py
```diff
@@ -225,27 +204,29 @@ def __str__(self):
 
 class DatasetInfo:
     """
-    Info about a public dataset accessible from huggingface.co
+    Info about a dataset accessible from huggingface.co
```
adrinjalali, line 207: same here

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
         *,
         revision: Optional[str] = None,
-        repo_type: Optional[str] = None,
         token: Optional[str] = None,
         timeout: Optional[float] = None,
-    ) -> List[str]:
+    ) -> DatasetInfo:
         """
-        Get the list of files in a given repo.
+        Get info on one specific dataset on huggingface.co
```
adrinjalali, line 1139:
```suggestion
        Get info on one specific dataset on huggingface.co.
```

I think the standard requires that the sentence finishes with a dot. I know numpydoc tests for it.

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
             revision (`str`, *optional*):
-                The revision of the model repository from which to get the
+                The revision of the dataset repository from which to get the
                 information.
-            repo_type (`str`, *optional*):
-                Set to `"dataset"` or `"space"` if uploading to a dataset or
-                space, `None` or `"model"` if uploading to a model. Default is
-                `None`.
             token (`str`, *optional*):
                 An authentication token [1]_.
```
adrinjalali, line 1151: This reference is not rendered properly ([link](https://moon-ci-docs.huggingface.co/docs/huggingface_hub/pr_792/en/package_reference/hf_api#huggingface_hub.HfApi.dataset_info) ). How do we do references in mdx?

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
             timeout (`float`, *optional*):
                 Whether to set a timeout for the request to the Hub.
 
         Returns:
-            `List[str]`: the list of files in a given repository.
+            [`DatasetInfo`]: The dataset repository information.
```
adrinjalali, line 1156: How can we make `DatasetInfo` to be a link to the corresponding API docs?

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
-                repo_id=repo_id, revision=revision, token=token, timeout=timeout
-            )
-        elif repo_type == "dataset":
-            info = self.dataset_info(
-                repo_id=repo_id, revision=revision, token=token, timeout=timeout
-            )
-        else:
-            raise ValueError("Spaces are not available yet.")
+        if token is None:
+            token = HfFolder.get_token()
```
adrinjalali, line 1163: should we use `_validate_or_retrieve_token` instead?

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
-        return [f.rfilename for f in info.siblings]
+        path = (
+            f"{self.endpoint}/api/datasets/{repo_id}"
+            if revision is None
+            else f"{self.endpoint}/api/datasets/{repo_id}/revision/{revision}"
+        )
+        headers = {"authorization": f"Bearer {token}"} if token is not None else None
+        r = requests.get(path, headers=headers, timeout=timeout)
+        r.raise_for_status()
+        d = r.json()
```
adrinjalali, line 1173: do we need to do the same as we did in https://github.com/huggingface/huggingface_hub/pull/783 here? (we could have a utility function to avoid repeating code if that's the case)

src/huggingface_hub/hf_api.py
```diff
@@ -1094,82 +1127,80 @@ def model_info(
...
         repo_id: str,
         *,
         revision: Optional[str] = None,
         token: Optional[str] = None,
         timeout: Optional[float] = None,
-    ) -> DatasetInfo:
+    ) -> SpaceInfo:
         """
-        Get info on one specific dataset on huggingface.co
+        Get info on one specific Space on huggingface.co
```
adrinjalali, line 1186:
```suggestion
        Get info on one specific Space on huggingface.co.
```

adrinjalali (commented):
Should the new methods also be included in a user guide?

src/huggingface_hub/hf_api.py
LysandreJik: I don't think there's a mechanism in place yet, I've replaced occurrences with a link directly.

src/huggingface_hub/hf_api.py
LysandreJik: I'm not sure that's the case, IIUC it was mostly linked to `delete_repo` having different errors when they should have been the same?

LysandreJik:
I should have addressed all comments.

> Should the new methods also be included in a user guide?

They'll be in the API reference on merge, which I think is a good first step.

LysandreJik:
Need to patch the failure.

src/huggingface_hub/hf_api.py
adrinjalali: opened an issue for this: https://github.com/huggingface/doc-builder/issues/181

src/huggingface_hub/hf_api.py
adrinjalali: opened an issue for this: https://github.com/huggingface/doc-builder/issues/181

src/huggingface_hub/hf_api.py
```diff
@@ -192,22 +167,42 @@ def __repr__(self):
...
+        pipeline_tag (`str`, *optional*):
+            Pipeline tag to identify the correct widget.
+        siblings (`List[Dict]`, *optional*):
+            list of files that constitute the Space
+        private (`bool`, *optional*):
+            is the repo private
+        author (`str`, *optional*):
+            repo author
+        config (`Dict`, *optional*):
+            Model configuration information
```
adrinjalali, line 190: missing `**kwargs`

src/huggingface_hub/hf_api.py
```diff
@@ -237,27 +234,49 @@ def __str__(self):
...
+        private (`bool`, *optional*):
+            is the repo private
+        author (`str`, *optional*):
+            repo author
+        description (`str`, *optional*):
+            Description of the dataset
+        citation (`str`, *optional*):
+            Dataset citation
+        cardData (`Dict`, *optional*):
+            Metadata of the model card as a dictionary.
```
adrinjalali, line 259: missing `**kwargs`

src/huggingface_hub/hf_api.py
```diff
@@ -286,6 +305,58 @@ def __str__(self):
...
+            repo sha at this particular revision
+        lastModified (`str`, *optional*):
+            date of last commit to repo
+        siblings (`List[Dict]`, *optional*):
+            list of files that constitute the Space
+        private (`bool`, *optional*):
+            is the repo private
+        author (`str`, *optional*):
+            repo author
+    """
```
adrinjalali, line 328: missing`**kwargs`

src/huggingface_hub/hf_api.py
```diff
@@ -286,6 +305,58 @@ def __str__(self):
...
+
+    @_deprecate_positional_args
```
adrinjalali, line 330: we've released, so the default is now too early, WDYT?

```suggestion
    @_deprecate_positional_args(version="0.8")
```

src/huggingface_hub/hf_api.py
```diff
@@ -1123,101 +1190,175 @@ def model_info(
...
         headers = {"authorization": f"Bearer {token}"} if token is not None else None
-        params = {"full": "true"}
-        r = requests.get(path, headers=headers, params=params, timeout=timeout)
+        r = requests.get(path, headers=headers, timeout=timeout)
         r.raise_for_status()
         d = r.json()
-        return DatasetInfo(**d)
+        return SpaceInfo(**d)
+
+    @_deprecate_positional_args
```
adrinjalali, line 1278:
```suggestion
    @_deprecate_positional_args(version="0.8")
```

src/huggingface_hub/hf_api.py
```diff
@@ -1123,101 +1190,175 @@ def model_info(
...
+                repo_id, revision=revision, token=token, timeout=timeout
+            )
+        elif repo_type == "space":
+            return self.space_info(
+                repo_id, revision=revision, token=token, timeout=timeout
+            )
+        else:
+            raise ValueError("Unsupported repo type.")
+
+    @_deprecate_positional_args
```
adrinjalali, line 1322:
```suggestion
    @_deprecate_positional_args(version="0.8")
```

adrinjalali (commented):
mostly cosmetic comments, except `HfFolder.get_token()`. I think if a new method is failing on using `validate_or_retrieve_token`, it's in the scope of the same PR to see why it'd fail. We have 9 instances of using `_validate_or_retrieve_token` and only two `HfFolder.get_token()`. We don't have to remove the existing ones here, but new code shouldn't be use it.

LysandreJik:
Thanks for your feedback. I've taken care of all your comments, except regarding the `_validate_or_retrieve_token`, as I stand by what I said in the previous comment. I understand how

> I think if a new method is failing on using `_validate_or_retrieve_token`, it's in the scope of the same PR to see why it'd fail.

that would be sensible, but the two methods are not designed to work together. The `_validate_or_retrieve_token` method exists to either validate the passed token, or retrieve the logged-in token, before validating it. It cannot be used as a no-op in case the passed token is `None` *and* the user isn't logged-in.

That's exactly what `model_info` (from which a direct copy is made for both `dataset_info` and `space_info`) is advertised to do, though:
```
Get info on one specific model on huggingface.co

Model can be private if you pass an acceptable token or are logged in.
```
In the case where no acceptable token is passed and the user isn't logged-in, then it should still work. Updating to use `_validate_or_retrieve_token` breaks that, as it raises an error.

Furthermore, I think that since `model_info` aims to be the exact same as `dataset_info` and `space_info`, but for different repository types, they should not use different approaches to retrieve the token, as situations like the one above can arise, which results in a desynchronisation of the three methods and an API behaving a bit differently acorss seemingly identical methods. I think in this case, if it is out of scope to refactor `model_info`, then `dataset_info` and `space_info` ought to be designed exactly like `model_info` is on `main`.

LysandreJik:
Rebasing on `main` to fix the tests.

adrinjalali:
Ok, opened https://github.com/huggingface/huggingface_hub/issues/837 to discuss the token issue. Happy to merge this as is and fix that issue later.

adrinjalali (approved):

PR #799, opened 2022-03-24
Finish converting docstrings

LysandreJik:
This PR aims to wrap up the docstring conversion.

### DIFF
--- a/docs/source/package_reference/hf_api.mdx
+++ b/docs/source/package_reference/hf_api.mdx
@@ -2,52 +2,52 @@
 
 [[autodoc]] login
 
-[[autodoc]] HfApi.logout
+[[autodoc]] logout
 
-[[autodoc]] HfApi.whoami
+[[autodoc]] whoami
 
-[[autodoc]] HfApi.list_models
+[[autodoc]] list_models
 
-[[autodoc]] HfApi.model_info
+[[autodoc]] model_info
 
-[[autodoc]] HfApi.list_repo_files
+[[autodoc]] list_repo_files
 
-[[autodoc]] HfApi.list_datasets
+[[autodoc]] list_datasets
 
-[[autodoc]] HfApi.dataset_info
+[[autodoc]] dataset_info
 
-[[autodoc]] HfApi.list_metrics
+[[autodoc]] list_metrics
 
-[[autodoc]] HfApi.get_model_tags
+[[autodoc]] get_model_tags
 
-[[autodoc]] HfApi.get_dataset_tags
+[[autodoc]] get_dataset_tags
 
-[[autodoc]] HfApi.create_repo
+[[autodoc]] create_repo
 
-[[autodoc]] HfApi.delete_repo
+[[autodoc]] delete_repo
 
-[[autodoc]] HfApi.update_repo_visibility
+[[autodoc]] update_repo_visibility
 
-[[autodoc]] HfApi.move_repo
+[[autodoc]] move_repo
 
-[[autodoc]] HfApi.upload_file
+[[autodoc]] upload_file
 
-[[autodoc]] HfApi.delete_file
+[[autodoc]] delete_file
 
-[[autodoc]] HfApi.get_full_repo_name
+[[autodoc]] get_full_repo_name
 
 ## Hugging Face local storage
 
 `huggingface_hub` stores the authentication information locally so that it may be re-used in subsequent
 methods.
 
-It does this using the `HfFolder` utility, which saves data at the root of the user.
+It does this using the [`HfFolder`] utility, which saves data at the root of the user.
 
 [[autodoc]] HfFolder
 
 ## Filtering helpers
 
-Some helpers to filter repositories on the hub are available in the `huggingface_hub` repository.
+Some helpers to filter repositories on the hub are available in the `huggingface_hub` package.
 
 [[autodoc]] DatasetFilter
 
--- a/docs/source/package_reference/hf_api.mdx
+++ b/docs/source/package_reference/hf_api.mdx
@@ -45,3 +45,10 @@ It does this using the `HfFolder` utility, which saves data at the root of the u
 
 [[autodoc]] HfFolder
 
+## Filtering helpers
+
+Some helpers to filter repositories on the hub are available in the `huggingface_hub` repository.
+
+[[autodoc]] DatasetFilter
+
+[[autodoc]] ModelFilter
--- a/docs/source/package_reference/hf_api.mdx
+++ b/docs/source/package_reference/hf_api.mdx
@@ -1,40 +1,31 @@
 # Hugging Face Hub API
 
-[[autodoc]] login
+Below is the documentation for the `HfApi` class, which serves as a Python wrapper for the Hugging Face
+Hub's API.
 
-[[autodoc]] logout
+All methods from the `HfApi` are also accessible from the package's root directly, both approaches are detailed
+below.
 
-[[autodoc]] whoami
+The following approach uses the method from the root of the package:
 
-[[autodoc]] list_models
+```python
+from huggingface_hub import list_models
 
-[[autodoc]] model_info
+models = list_models()
+```
 
-[[autodoc]] list_repo_files
+The following approach uses the `HfApi` class:
 
-[[autodoc]] list_datasets
+```python
+from huggingface_hub import HfApi
 
-[[autodoc]] dataset_info
+hf_api = HfApi()
+models = hf_api.list_models()
+```
 
-[[autodoc]] list_metrics
+Using the `HfApi` class directly enables you to set a different endpoint to that of the Hugging Face's Hub.
 
-[[autodoc]] get_model_tags
-
-[[autodoc]] get_dataset_tags
-
-[[autodoc]] create_repo
-
-[[autodoc]] delete_repo
-
-[[autodoc]] update_repo_visibility
-
-[[autodoc]] move_repo
-
-[[autodoc]] upload_file
-
-[[autodoc]] delete_file
-
-[[autodoc]] get_full_repo_name
+[[autodoc]] HfApi
 
 ## Hugging Face local storage
 
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -464,7 +464,8 @@ class HfApi:
 
         <Tip>
 
-        Warning: Deprecated, will be removed in v0.7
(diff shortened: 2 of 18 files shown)


docs/source/package_reference/hf_api.mdx
```diff
@@ -45,3 +45,10 @@ It does this using the `HfFolder` utility, which saves data at the root of the u
 
 [[autodoc]] HfFolder
 
+## Filtering helpers
+
+Some helpers to filter repositories on the hub are available in the `huggingface_hub` repository.
```
adrinjalali, line 50: I think in the documentation we should refer to `huggingface_hub` as a package, and not a repo. Users should be reading the docs and installing the package, so from their point of view the repo doesn't matter much.

src/huggingface_hub/hf_api.py
```diff
@@ -633,7 +633,7 @@ def list_models(
         Get the public list of all the models on huggingface.co
 
         Args:
-            filter (:class:`ModelFilter` or :obj:`str` or :class:`Iterable`, *optional*):
+            filter (`ModelFilter` or `str` or `Iterable`, *optional*):
```
adrinjalali, line 636: In the rendered version `ModelFilter` seems to be a simple text and not a hyperlink to `ModelFilter`'s documentation. I think we should keep those link with this migration (`:class:` does that in rst)

src/huggingface_hub/hf_api.py
```diff
@@ -846,7 +846,7 @@ def list_datasets(
         Get the public list of all the datasets on huggingface.co
 
         Args:
-            filter (:class:`DatasetFilter` or :obj:`str` or :class:`Iterable`, *optional*):
+            filter (`DatasetFilter` or `str` or `Iterable`, *optional*):
```
adrinjalali, line 849: same here

src/huggingface_hub/hf_api.py
```diff
@@ -1437,35 +1515,33 @@ def upload_file(
...
 
-        Raises:
-            :class:`ValueError`: if some parameter value is invalid
+        <Tip>
+
+        Raises the following errors:
 
-            :class:`requests.HTTPError`: if the HuggingFace API returned an error
+            - `HTTPError` if the HuggingFace API returned an error
+            - `ValueError` if some parameter value is invalid
```
adrinjalali, line 1542: and here. These should be links to the corresponding docs.

src/huggingface_hub/hf_api.py
```diff
@@ -1565,24 +1641,24 @@ def delete_file(
...
+                Relative filepath in the repo, for example: `"checkpoints/1fec34a/weights.bin"`
             repo_id (`str`):
-                The repository from which the file will be deleted, for example: :obj:`"username/custom_transformers"`
-
-            token (`str`):
+                The repository from which the file will be deleted, for example: `"username/custom_transformers"`
+            token (`str`, *optional*):
                 Authentication token, obtained with :function:`HfApi.login` method. Will default to the stored token.
+            repo_type (`str`, *optional*):
+                Set to `"dataset"` or `"space"` if the file is in a dataset or space, `None` or `"model"` if in a model. Default is `None`.
```
adrinjalali, line 1650: very long line.

src/huggingface_hub/hf_api.py
```diff
@@ -1565,24 +1641,24 @@ def delete_file(
...
+            revision (`str`, *optional*):
+                The git revision to commit from. Defaults to the `"main"` branch.
```
adrinjalali, line 1652: maybe?

```suggestion
                The git revision to commit from. Defaults to the `HEAD` of the
                `"main"` branch.```

File: src/huggingface_hub/inference_api.py [hunk 7]
```diff
@@ -47,28 +47,30 @@ class InferenceApi:
...
+    >>> inputs = {"question":"What's my name?", "context":"My name is Clara and I live in Berkeley."}
+    >>> inference(inputs)
+    >>> >> {'score': 0.9326569437980652, 'start': 11, 'end': 16, 'answer': 'Clara'}
+
+    >>> # Zero-shot example
+    >>> inference = InferenceApi("typeform/distilbert-base-uncased-mnli")
+    >>> inputs = "Hi, I recently bought a device from your company but it is not working as advertised and I would like to get reimbursed!"
+    >>> params = {"candidate_labels":["refund", "legal", "faq"]}
+    >>> inference(inputs, params)
+    >>> >> {'sequence': 'Hi, I recently bought a device from your company but it is not working as advertised and I would like to get reimbursed!', 'labels': ['refund', 'faq', 'legal'], 'scores': [0.9378499388694763, 0.04914155602455139, 0.013008488342165947]}
```
adrinjalali, line 69: break these lines?

src/huggingface_hub/keras_mixin.py
```diff
@@ -152,22 +152,24 @@ def save_pretrained_keras(
...
+        save_directory (`str`):
+            Specify directory in which you want to save the Keras model.
+        config (`dict`, *optional*):
+            Configuration object to be saved alongside the model weights.
+        include_optimizer(`bool`, *optional*, defaults to `False`):
+            Whether or not to include optimizer in serialization.
+        plot_model (`bool`, *optional*, defaults to `True`):
+            Setting this to `True` will plot the model and put it in the model card. Requires graphviz and pydot to be installed.
+        task_name (`str`, *optional*):
+            Name of the task the model was trained on. See the available tasks at https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts.
```
adrinjalali, line 170: the link could be a `Reference`, but it should also be in the docs and not on github I think.

src/huggingface_hub/utils/logging.py
```diff
@@ -88,53 +102,75 @@ def get_logger(name: Optional[str] = None) -> logging.Logger:
...
-            Logging level, e.g., ``huggingface_hub.logging.DEBUG`` and ``huggingface_hub.logging.INFO``.
+        verbosity (`int`):
+            Logging level, e.g., `huggingface_hub.logging.DEBUG` and `huggingface_hub.logging.INFO`.
     """
     _get_library_root_logger().setLevel(verbosity)
 
 
 def set_verbosity_info():
+    """
+    Sets the verbosity to `logging.INFO`.
```
adrinjalali, line 137: missing `Returns` sections around here.

adrinjalali (commented):
Nice! Thanks @LysandreJik , this is definitely quite a bit of an improvement.

There are quite a few very long lines, it'd be nice to break them.

The links are also missing, as in, they're not just a simple `code text` rather than a hyperlink to corresponding classes and methods.

adrinjalali:
Do we really want to render type hints? It feels quite bloated and not very readable.

Screenshot from 2022-03-25 11-00-01

src/huggingface_hub/hf_api.py
LysandreJik: Great idea, much better this way!

src/huggingface_hub/utils/logging.py
LysandreJik: The `set` don't return anything, so I haven't added returns for all methods returning nothing (or `None`). Would you find it clearer if it was specified?

src/huggingface_hub/keras_mixin.py
```diff
@@ -152,22 +152,24 @@ def save_pretrained_keras(
...
-        Setting this to `True` will plot the model and put it in the model card. Requires graphviz and pydot to be installed.
-    task_name (`str`, *optional*):
-        Name of the task the model was trained on. See the available tasks at https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts.
-    model_save_kwargs(`dict`, *optional*):
-        model_save_kwargs will be passed to tf.keras.models.save_model().
+    """
+    Saves a Keras model to save_directory in SavedModel format. Use this if you're using the Functional or Sequential APIs.
+
+    Args:
+        model (`Keras.Model`):
```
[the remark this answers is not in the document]
LysandreJik: I think this is a limitation of the `doc-builder`, I've added a link to the object in the description, let me know if it works. Will open an issue in doc-builder.

LysandreJik:
The comments should have been taken care of. If this works for you, @adrinjalali, I'd like to merge this PR in the feature branch and acentralize the discussion about next steps there.

docs/source/package_reference/hf_api.mdx
```diff
@@ -2,46 +2,53 @@
 
 [[autodoc]] login
 
-[[autodoc]] HfApi.logout
+[[autodoc]] logout
```
adrinjalali, line 5: Are we suggesting people to use these methods instead of `HfApi`? I would think exposing `HfApi` as the primary point for users would make more sense though. Especially if we go with not storing the token automatically and storing it in the instance.

docs/source/package_reference/hf_api.mdx
```diff
@@ -2,46 +2,53 @@
 
 [[autodoc]] login
```
adrinjalali, line 3: This should be removed, it's deprecated.

src/huggingface_hub/utils/logging.py
adrinjalali: I see, in that case I think these functions shouldn't be `return`ing the output of `set_verbosity` which is `None` (doesn't have to be fixed in this PR, you could open another good first issue for this one).

adrinjalali (commented):
Love this!

src/huggingface_hub/utils/logging.py
LysandreJik: Very good point

docs/source/package_reference/hf_api.mdx
LysandreJik: I've added only `HfApi` for now, with an introduction of how each method can still be imported from the root - this was the preferred way up to now as simpler, but it might not be if we start storing the token automatically, so I've added a mention of both. Let me know if that works for you.

docs/source/package_reference/hf_api.mdx
LysandreJik: I've manually added deprecation warnings with the expected version removal; as those methods are points of entry I fear removing the methods themselves would lead to users Ctrl+F and not finding what they're looking for. I've added a redirection to the methods that should now be used instead.

Let me know if that works for you.

adrinjalali (approved):
LGTM. Happy to see this merged.

Didn't see an answer to this one: https://github.com/huggingface/huggingface_hub/pull/799#issuecomment-1078851127

Issue #35, opened 2021-05-05
[Issue]: Too long file-paths for Windows


nreimers:
Hi,
I was running the following code:

``` 
from huggingface_hub import snapshot_download, hf_hub_url, cached_download
model_name = 'osanseviero/full-sentence-distillroberta2'
url = hf_hub_url(model_name, 'modules.json')
print(url)
path = cached_download(url, cache_dir='local_cache')
print("file downloaded")
```

The call to cached_download() freezes on my Windows machine.

The issue is in this line:
https://github.com/huggingface/huggingface_hub/blob/67198a0f2e2716e952b7cae0ef0cca6900bded39/src/huggingface_hub/file_download.py#L412

The lock_path was:
```
D:\some\long\subfolder\dir\i\am\using\....\local_cache\1547cad195ef32a189cf8d0de51c99ab3c5b767f1c64e01a5711d3a6af9dfdb1.ab4f0ee45ce35db416c1d6d811634418f90db5d7a46bb8ff552d13c6468ec38e.lock
```

In total the path had 275 characters. However, on Windows only file-paths up to 255 characters are allowed. The generated filename (1547[....].lock) has 135 characters, only leaving at max 120 characters for the folder path. 

If you want to download to any path with more than 120 characters, the process will freeze as the FileLock cannot be acquired.

**How to fix it:**
- Not sure what a good fix would be. It would be nice if FileLock would raise an Exception, that the path is too long. But this is sadly not the case. Is there a way an exception could be thrown here?
- I was able to get around the issue adding the force_filename and to use a shorter file name:
```
path = cached_download(url, cache_dir=f'local_cache', force_filename='modules.json')
``` 
- Does it make sense to use shorter than the 135 character file-name?
- Is there some way to check first if the lock can be created, before we acquire it for the download? This way we would hopefully note that the lock cannot be created.

LysandreJik:
Hi, thanks for opening the issue and for the very detailed bug report! 

This is an annoying error indeed. Have you never encountered it before with `transformers`? It's based on the same approach, so it should result in the same hang. 

See here for example: https://github.com/huggingface/transformers/blob/f594090a936a06524d2df77c0eb65a42bf694d6b/src/transformers/file_utils.py#L1504

By adding a print statement above that line:
```py
[...]
    # Prevent parallel downloads of the same file with a lock.
    lock_path = cache_path + ".lock"

    print("lock path", lock_path)
    with FileLock(lock_path):
[...]
```
[...]

PR #801, opened 2022-03-25
New git-aware cache file layout

julien-c:
## A new way to layout cached files on disk, unifying what we do for `snapshot_download` and `cached_download`, and laying the ground for new features (ability to query the disk to figure out which local files/revisions we have)

See old comment from last year inside `snapshot_download`:

```python
# Note: at some point maybe this format of storage should actually replace
# the flat storage structure we've used so far (initially from allennlp
# if I remember correctly).
```

This new layout will be git-aware (compatible with versioning)

### One picture is worth 1,000 words

Here's a screenshot of the cache's file tree generated by a few file downloads from a model repo (sample code below)

image

```python
import torch
from huggingface_hub.file_download import hf_hub_download

OLDER_REVISION = "bbc77c8132af1cc5cf678da3f1ddf2de43606d48"

hf_hub_download("julien-c/EsperBERTo-small", filename="README.md")

hf_hub_download("julien-c/EsperBERTo-small", filename="pytorch_model.bin")

hf_hub_download(
    "julien-c/EsperBERTo-small", filename="README.md", revision=OLDER_REVISION
)

weights_file = hf_hub_download(
    "julien-c/EsperBERTo-small", filename="pytorch_model.bin", revision=OLDER_REVISION
)

w = torch.load(weights_file, map_location=torch.device("cpu"))
# Yay it works! just loaded a torch file from a symlink
```

### Preliminary mini-spec

- The cache directory contains one subfolder per repo_id (namespaced by repo type)
- inside each repo folder:
  - `refs` is a list of the latest known `revision => commit_hash` pairs
  - `blobs` contains the actual file blobs (identified by their `git-sha` or `sha256`, depending on whether they're LFS files or not)
  - `snapshots` contains one subfolder per commit, each "commit" contains the subset of the files that have been resolved at that particular commit. Each filename is a symlink to the blob at that particular commit.

### Explaining the result

In the sample code I'm downloading the same two filenames at two different commits. 
- The `README.md` is different at both commits so there's both versions of it in `blobs`.
- The `pytorch_model.bin` however, is the same at both commits, so it's not downloaded again, and the disk space is shared between the two snapshots. 🎉 🎉


src/huggingface_hub/snapshot_download.py
```diff
@@ -178,58 +187,51 @@ def snapshot_download(
...
-        # if we have internet connection we retrieve the correct folder name from the huggingface api
-        _api = HfApi()
-        model_info = _api.model_info(repo_id=repo_id, revision=revision, token=token)
+        return storage_folder
+
+    # if we have internet connection we retrieve the correct folder name from the huggingface api
+    _api = HfApi()
+    repo_info = _api.repo_info(
+        repo_id=repo_id, repo_type=repo_type, revision=revision, token=token
+    )
```
julien-c, line 199: This uses the new `repo_info()` function from #792

julien-c:
Some additional context for this is in https://github.com/huggingface/transformers/issues/15927#issuecomment-1060664149

julien-c:
## Ok this should be ready for a deeper review @LysandreJik  🎉 

`snapshot_download` is way simpler now and only makes one networking call if you already have the repo snapshot locally 🎉  

Also cc @patrickvonplaten and @sgugger

patrickvonplaten:
Wow this is super cool! No more re-downloading all files when just the README is changed :heart:

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
-    # remove all `/` occurrences to correctly convert repo to directory name
-    repo_id_flattened = repo_id.replace("/", REPO_ID_SEPARATOR)
-
-    # if we have no internet connection we will look for the
-    # last modified folder in the cache
-    if local_files_only:
-        # possible repos have <path/to/cache_dir>/<flatten_repo_id> prefix
-        repo_folders_prefix = os.path.join(cache_dir, repo_id_flattened)
-
-        # list all possible folders that can correspond to the repo_id
```
patrickvonplaten, line 132: All that beautiful code is gone :cry:

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
-
-        # Create potential nested dir
-        nested_dirname = os.path.dirname(
-            os.path.join(storage_folder, relative_filepath)
+    # if we have no internet connection we will look for an
+    # appropriate folder in the cache
+    # If the specified revision is a commit hash, look inside "snapshots".
+    # If the specified revision is a branch or tag, look inside "refs".
+    if local_files_only:
+        if REGEX_COMMIT_HASH.match(revision):
```
patrickvonplaten, line 158: (nit) might be easier to understand

```suggestion
        commit_hash = revision
        if not REGEX_COMMIT_HASH.match(commit_hash):
            # rertieve commit_hash from file
            ref_path = os.path.join(storage_folder, "refs", revision)
            with open(ref_path) as f:
                commit_hash = f.read()
        
        snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)
        if os.path.exists(snapshot_folder):
            return snapshot_folder
```

            
            
```

File: src/huggingface_hub/snapshot_download.py [hunk 4]
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
+        else:
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                commit_hash = f.read()
+            snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)
+            if os.path.exists(snapshot_folder):
+                return snapshot_folder
+
+        raise ValueError(
+            "Cannot find an appropriate cached folder for the specified revision on the"
```
patrickvonplaten, line 171:
```suggestion
            "Cannot find an appropriate cached snapshot folder for the specified revision on the"
```

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
-                f" from the branch {revision} in offline mode might lead to unexpected"
-                " behavior by not taking into account the latest commits."
-            )
-
-        # find last modified folder
-        storage_folder = max(repo_folders, key=os.path.getmtime)
+    if repo_type is None:
+        repo_type = "model"
+    if repo_type not in REPO_TYPES:
+        raise ValueError("Invalid repo type")
```
patrickvonplaten, line 147:
```suggestion
        raise ValueError(f"Invalid repo type. {repo_type} does not exist. Please select one of {REPO_TYPES}")
```

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
-            url,
-            cache_dir=storage_folder,
-            force_filename=relative_filepath,
+    # if we have internet connection we retrieve the correct folder name from the huggingface api
+    _api = HfApi()
+    repo_info = _api.repo_info(
+        repo_id=repo_id, repo_type=repo_type, revision=revision, token=token
+    )
+    filtered_repo_files = _filter_repo_files(
+        repo_files=[f.rfilename for f in repo_info.siblings],
```
patrickvonplaten, line 182: would this also work with three layers deep?

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
+    # If the specified revision is a branch or tag, look inside "refs".
+    if local_files_only:
+        if REGEX_COMMIT_HASH.match(revision):
+            snapshot_folder = os.path.join(storage_folder, "snapshots", revision)
+            if os.path.exists(snapshot_folder):
+                return snapshot_folder
+        else:
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                commit_hash = f.read()
```
patrickvonplaten, line 165: that's a very nice mechanism!

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +845,260 @@ def hf_hub_download(
...
+            pointer_path = os.path.join(
+                storage_folder, "snapshots", revision, relative_filename
+            )
+            if os.path.exists(pointer_path):
+                return pointer_path
+        else:
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                commit_hash = f.read()
+            pointer_path = os.path.join(
```
patrickvonplaten, line 993: (nit) could maybe simplify the if-else structure here, to just have:

```py
commit_hash = revision
if not REGEX_COMMIT_HASH.match(revision):
            with open(ref_path) as f:
                commit_hash = f.read()

pointer_path = os.path.join(
    storage_folder, "snapshots", revision, relative_filename
)
if os.path.exists(pointer_path):
    return pointer_path
```

patrickvonplaten (approved):
Super excited about this feature. We can advertise it well for all Wav2Vec2 + ngram !

Left only some nits, can all be disregarded

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +845,260 @@ def hf_hub_download(
...
+            etag = _normalize_etag(etag)
+            # In case of a redirect,
+            # save an extra redirect on the request.get call,
+            # and ensure we download the exact atomic version even if it changed
+            # between the HEAD and the GET (unlikely, but hey).
+            if 300 <= r.status_code <= 399:
+                url_to_download = r.headers["Location"]
+        except (requests.exceptions.SSLError, requests.exceptions.ProxyError):
+            # Actually raise for those subclasses of ConnectionError
+            raise
```
sgugger, line 962: On the Transformers side, we also raise the `RepositoryNotFoundError`, `EntryNotFoundError` and `RevisionNotFoundError` here (which are raised by some code in the try block). There are different from connection/timeout errors and give explicit information to the user about what's wrong, so would be great to integrate those here too.

(Can go in a separate PR if it makes more sense.)

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +845,260 @@ def hf_hub_download(
...
+        logger.info("storing %s in cache at %s", url, blob_path)
+        os.replace(temp_file.name, blob_path)
+
+        logger.info("creating pointer to %s from %s", blob_path, pointer_path)
+        _create_relative_symlink(blob_path, pointer_path)
+
+    try:
+        os.remove(lock_path)
+    except OSError:
+        pass
```
sgugger, line 1102: Should this be in a `finally` block so it's always executed, even if the previous code failed?

sgugger (approved):
Super nice new system! Can't wait to be able to use in Transformers :-)

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +845,260 @@ def hf_hub_download(
...
+        with temp_file_manager() as temp_file:
+            logger.info("downloading %s to %s", url, temp_file.name)
+
+            http_get(
+                url_to_download,
+                temp_file,
+                proxies=proxies,
+                resume_size=resume_size,
+                headers=headers,
+            )
```
LysandreJik, line 1091: In the connection errors I had, this failed and ended up leaving a ~100MB temporary file at the root of the cache; not sure if we can do something about that though

src/huggingface_hub/file_download.py
```diff
@@ -142,6 +147,9 @@ def get_fastcore_version():
     return _fastcore_version
 
 
+REGEX_COMMIT_HASH = re.compile(r"^[0-9a-f]{5,40}$")
```
LysandreJik, line 150: Theoretically it could be 4 instead of 5 here, as hf.co resolves correctly down to as low as 4 characters
```suggestion
REGEX_COMMIT_HASH = re.compile(r"^[0-9a-f]{4,40}$")
```

src/huggingface_hub/file_download.py
```diff
@@ -689,6 +706,61 @@ def _resumable_file_manager() -> "io.BufferedWriter":
...
+def _create_relative_symlink(src: str, dst: str) -> None:
+    """Create a symbolic link named dst pointing to src as a relative path to dst.
+
+    The relative part is mostly because it seems more elegant to the author.
+
+    The result layout looks something like
+        └── [ 128]  snapshots
+            ├── [ 128]  2439f60ef33a0d46d85da5001d52aeda5b00ce9f
+            │   ├── [  52]  README.md -> ../../blobs/d7edf6bd2a681fb0175f7735299831ee1b22b812
+            │   └── [  76]  pytorch_model.bin -> ../../blobs/403450e234d65943a7dcf7e05a771ce3c92faa84dd07db4ac20f592037a1e4bd
```
LysandreJik, line 736: Showcasing this is elegant, and I think showcasing the full layout somewhere is warranted as it makes understanding it super simple. Happy to provide it!

LysandreJik (commented):
Thanks for working on it, this is super nice and clean.

It downloads files under `~/.cache/huggingface/hub`. Is this the expected place for them to be downloaded at? Should we expect other libraries (transformers, datasets) to go tap in that folder instead of the current transformers/datasets folders?

PS: When playing around with it, I had several connection errors that I didn't seem to have previously when using `snapshot_download`. Might be an isolated incident as I don't think this should put any more stress on the backend (quite the contrary).

src/huggingface_hub/file_download.py
LysandreJik: (right now putting 4 characters means that we're skipping the cache check and going straight for `hf_hub_url`)

julien-c: oh i actually meant to only allow passing a full commit (i.e. just `{40}`), to make sure no false positive are detected (like naming a branch `deadbeef` etc)

Do you see an issue in accepting only full commit hashes?

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,69 @@ def snapshot_download(
...
+    # we pass the commit_hash to hf_hub_download
+    # so no network call happens if we already
+    # have the file locally.
+
+    for repo_file in filtered_repo_files:
+        _ = hf_hub_download(
+            repo_id,
+            filename=repo_file,
+            repo_type=repo_type,
+            revision=commit_hash,
```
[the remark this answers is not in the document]
julien-c: Yes i think i forgot to store the `ref` in `snapshot_download`, I realized that later. Will add it

julien-c:
> It downloads files under `~/.cache/huggingface/hub`. Is this the expected place for them to be downloaded at? Should we expect other libraries (transformers, datasets) to go tap in that folder instead of the current transformers/datasets folders?

Yes, that was my idea – or we could also just go up one level and store in `~/.cache/huggingface`, considering that this is going to be library-agnostic and huggingface-wide anyways

julien-c:
> PS: When playing around with it, I had several connection errors that I didn't seem to have previously when using `snapshot_download`. Might be an isolated incident

The API calls themselves are the same as before, so it shouldn't change much – let's investigate if it continues

src/huggingface_hub/snapshot_download.py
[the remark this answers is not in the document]
julien-c: and actually your code suggestion is perfect i think

src/huggingface_hub/snapshot_download.py
julien-c: what do you mean?

 it should support arbitrary nesting, but i haven't extensively tested yet :)

src/huggingface_hub/file_download.py
julien-c: yes 👍 on adding them

No strong opinion on whether we want to add them here or in another PR. I'd say another PR given we might want to also update `cached_download`, which this PR does not touch. WDYT @LysandreJik?

src/huggingface_hub/snapshot_download.py
julien-c: no strong opinion either way! i'll let @LysandreJik decide =)

src/huggingface_hub/file_download.py
julien-c: took a stab at it in 81b75678b3556544f6a493c110057c8746a41c78

Did that using `tree -h`

src/huggingface_hub/file_download.py
julien-c: i'm not touching that particular line of code (just "unwrapping" it) so I'd rather keep it w/ same behavior as before

julien-c: (to minimize scope of changes)

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +871,263 @@ def hf_hub_download(
...
+    # Prevent parallel downloads of the same file with a lock.
+    lock_path = blob_path + ".lock"
+
+    # Some Windows versions do not allow for paths longer than 255 characters.
+    # In this case, we must specify it is an extended path by using the "\\?\" prefix.
+    if os.name == "nt" and len(os.path.abspath(lock_path)) > 255:
+        lock_path = "\\\\?\\" + os.path.abspath(lock_path)
+
+    if os.name == "nt" and len(os.path.abspath(blob_path)) > 255:
+        blob_path = "\\\\?\\" + os.path.abspath(blob_path)
```
julien-c, line 1081: I'm not entirely certain this is still useful, TBH.... In any case, we're not doing it for other files or symlinks in this PR, so it's unlikely to still work as is

I'd consider removing it altogether @LysandreJik 

Slightly related poll on Windows older versions support: https://twitter.com/julien_c/status/1522958087883309059

src/huggingface_hub/snapshot_download.py
LysandreJik: Sure, I like this proposal! I find it improves readability, thank you @patrickvonplaten.

src/huggingface_hub/file_download.py
LysandreJik: Handled in https://github.com/huggingface/huggingface_hub/pull/801/commits/d786693779f587c903b53adffe2b81734594edfd

src/huggingface_hub/file_download.py
LysandreJik: Taking care of that in https://github.com/huggingface/huggingface_hub/pull/878

src/huggingface_hub/file_download.py
LysandreJik: This was setup for very recent windows versions as well. Will test on Windows to see if it can be removed.

It was introduced in https://github.com/huggingface/huggingface_hub/pull/40 due to https://github.com/huggingface/huggingface_hub/issues/35

LysandreJik:
Remaining steps before merging:

- [ ] Write tests for the new cache layout
- [ ] Verify it works correctly on Windows
- [ ] Ensure backwards compatibility
- [ ] Document the layout

src/huggingface_hub/snapshot_download.py
```diff
@@ -120,133 +141,90 @@ def snapshot_download(
...
+    # appropriate folder in the cache
+    # If the specified revision is a commit hash, look inside "snapshots".
+    # If the specified revision is a branch or tag, look inside "refs".
+    if local_files_only:
+        commit_hash = revision
+        if not REGEX_COMMIT_HASH.match(commit_hash):
+            # rertieve commit_hash from file
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                commit_hash = f.read()
```
julien-c, line 166:
```suggestion
        def resolve_ref(revision) -> str:
            # retrieve commit_hash from file
            ref_path = os.path.join(storage_folder, "refs", revision)
            with open(ref_path) as f:
                return f.read()

        commit_hash = revision 
            if REGEX_COMMIT_HASH.match(revision)
            else resolve_ref(revision)
```

I think that maybe it is cleaner to use a ternary then @LysandreJik @patrickvonplaten WDYT?

thomwolf:
I noticed a bug here when using `subfolder`:

If I have two files with the same name in different subfolders and I'm downloading them subsequently using the `hf_hub_download` method, the second file will not be downloaded because it's seen as already existing.

A space to test the issue: these two files are different (see the end of the gltf files):
- https://huggingface.co/spaces/simenv-tests/Box/blob/main/glTF-Embedded/Box.gltf
- https://huggingface.co/spaces/simenv-tests/Box/blob/main/glTF/Box.gltf

Proposal to solve the issue: add the `subfolder` in the path used to create the cache file in the snapshot folder so that `if os.path.exists(pointer_path)` fails instead or giving true.

Current workaround I'll use: set `force_download` to True

julien-c:
can you post the snippet of code you use?

julien-c:
I think there's a bug in this line:
```python
# cross platform transcription of filename, to be used as a local file path.
relative_filename = os.path.join(*filename.split("/"))
```
where it should take into account the potential `subfolder` param

thomwolf:
Code snippet could be for instance:
```
first_file = hf_hub_download(repo_id='simenv-tests/Box',
            filename='Box.gltf,
            subfolder='glTF',
            repo_type="space",
        )
print(first_file)
second_file = hf_hub_download(repo_id='simenv-tests/Box',
            filename='Box.gltf,
            subfolder='glTF-Embedded',
            repo_type="space",
        )
print(second_file)
```
Manual inspection of the first and second files show they are identical (see in particular the `"buffers"` section of both JSON)

If you remove the `subfolder` parameter, what would you replace it with?

julien-c:
> If you remove the `subfolder` parameter, what would you replace it with?

just the filename param, should be the full nested filename: `glTF/Box.gltf `

tests/test_cache_layout.py
```diff
@@ -0,0 +1,385 @@
...
+        repo_id = f"{USER}/hfh-cache-layout"
+        create_repo(repo_id, token=self._token, exist_ok=True)
+
+        try:
+            upload_file(
+                path_or_fileobj=BytesIO(b"Some string"),
+                path_in_repo="file.txt",
+                repo_id=repo_id,
+                token=self._token,
+            )
```
julien-c, line 324: this method is soon to be deprecated but we can update it then (cc @SBrandeis)

tests/test_cache_layout.py
```diff
@@ -0,0 +1,385 @@
...
+    def setUpClass(cls):
+        """
+        Share this valid token in all tests below.
+        """
+        cls._token = TOKEN
+        cls._api.set_access_token(TOKEN)
+
+    def test_update_reference(self):
+        repo_id = f"{USER}/hfh-cache-layout"
+        create_repo(repo_id, token=self._token, exist_ok=True)
```
julien-c, line 316: not sure about this test: it's not going to be able to run it multiple times concurrently

Is the plan to have a uuid in the repo name like in other "integration tests"?

tests/test_cache_layout.py
```diff
@@ -0,0 +1,385 @@
...
+            # The snapshot content should link to a blob
+            self.assertTrue(os.path.islink(snapshot_content_path))
+
+            resolved_blob_relative = os.readlink(snapshot_content_path)
+            resolved_blob_absolute = os.path.normpath(
+                os.path.join(snapshot_path, resolved_blob_relative)
+            )
+
+            with open(resolved_blob_absolute) as f:
+                blob_contents = f.readline().strip()
```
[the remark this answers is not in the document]
LysandreJik: Because the file contains a single line! Reading the whole file would be better, indeed. Thanks!

tests/test_cache_layout.py
LysandreJik: Yes, you're correct. Updating to use the uuid.

src/huggingface_hub/file_download.py
```diff
@@ -514,6 +526,11 @@ def cached_download(
 
     </Tip>
     """
+    warnings.warn(
+        "`cached_download` is the legacy way to download files from the HF hub, please"
+        " consider upgrading to `hf_hub_download`",
+        FutureWarning,
+    )
```
LysandreJik, line 533: @julien-c, `cached_download` implements features that are not available with `hf_hub_download` (namely, caching from a URL). Is this a feature that you think should be abandonned? I think there is room for both methods to exist, but this warning says otherwise.

julien-c: > Is this a feature that you think should be abandonned?

I think it's a feature that does not really make sense in the context of the HF hub, and a HF Hub client library.

My concern is that if we don't add this warning, users will not upgrade to the new way.

As a middleground, maybe we could add a `legacy_cache_layout` flag that just disables the warning? and mention in the warning "If you want to cache from an arbitrary URL instead, pass legacy_cache_layout=True"

WDYT?

src/huggingface_hub/_snapshot_download.py
```diff
@@ -120,133 +141,90 @@ def snapshot_download(
...
+    # appropriate folder in the cache
+    # If the specified revision is a commit hash, look inside "snapshots".
+    # If the specified revision is a branch or tag, look inside "refs".
+    if local_files_only:
+        commit_hash = revision
+        if not REGEX_COMMIT_HASH.match(commit_hash):
+            # rertieve commit_hash from file
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                commit_hash = f.read()
```
julien-c, line 166: still need to update this one no?

src/huggingface_hub/file_download.py
LysandreJik: Yes, sounds good to me. I'll proceed this way.

src/huggingface_hub/_snapshot_download.py
```diff
@@ -158,26 +158,20 @@ def snapshot_download(
...
             ref_path = os.path.join(storage_folder, "refs", revision)
             with open(ref_path) as f:
-                commit_hash = f.read()
+                return f.read()
 
+        commit_hash = (
+            revision if REGEX_COMMIT_HASH.match(revision) else resolve_ref(revision)
+        )
         snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)
+
```
LysandreJik, line 172: See side by side comparison of before/after @julien-c:


<table>
<tr>
<td>After</td>
<td>Before</td>
</tr>

<tr>
<td>

```py
if local_files_only:

    def resolve_ref(revision) -> str:
        # retrieve commit_hash from file
        with open(ref_path) as f:
            return f.read()

    commit_hash = (
        revision if REGEX_COMMIT_HASH.match(revision) else resolve_ref(revision)
    )
    snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)

    if os.path.exists(snapshot_folder):
        return snapshot_folder

    raise ValueError(
        "Cannot find an appropriate cached snapshot folder for the specified"
        " revision on the local disk and outgoing traffic has been disabled. To"
        " enable repo look-ups and downloads online, set 'local_files_only' to"
        " False."
    )
```

</td>
<td>

```py
if local_files_only:
    if REGEX_COMMIT_HASH.match(revision):
        snapshot_folder = os.path.join(storage_folder, "snapshots", revision)
        if os.path.exists(snapshot_folder):
            return snapshot_folder
    else:
        with open(ref_path) as f:
            commit_hash = f.read()
        snapshot_folder = os.path.join(storage_folder, "snapshots", commit_hash)
        if os.path.exists(snapshot_folder):
            return snapshot_folder
     raise ValueError(
        "Cannot find an appropriate cached folder for the specified revision on the"
        " local disk and outgoing traffic has been disabled. To enable repo"
        " look-ups and downloads online, set 'local_files_only' to False."
    )
```

</td>
</tr>

</table>

julien-c: Yep, looks great to me!

docs/source/package_reference/file_download.mdx
```diff
@@ -8,3 +8,84 @@
...
+
+The caching system is designed as follows:
+
+```
+<CACHE_DIR>
+├─ <MODELS>
+├─ <DATASETS>
+├─ <SPACES>
+```
+
```
sgugger, line 25: Maybe remind the user the default for `CACHE_DIR` and how it can be customized?

docs/source/package_reference/file_download.mdx
```diff
@@ -8,3 +8,84 @@
...
+Each folder is designed to contain the following:
+
+### Refs
+
+The `refs` folder contains files which indicates the latest revision of the given reference. For example,
+if we have previously fetched a file from the `main` branch of a repository, the `refs`
+folder will contain a file named `main`, which will itself contain the commit identifier of the current head.
+
+If the latest commit of `main` had `aaaaaa` as identifier, then it would contain `aaaaaa`.
```
sgugger, line 63:
```suggestion
If the latest commit of `main` has `aaaaaa` as identifier, then it will contain `aaaaaa`.
```

docs/source/package_reference/file_download.mdx
```diff
@@ -8,3 +8,84 @@
...
+if we had downloaded the `READMD.md` file at revision `aaaaaa`, we would have the following path:
+
+```
+<CACHE_DIR>/<REPO_NAME>/snapshots/aaaaaa/README.md
+```
+
+That `README.md` file is actually a symlink linking to the blob that has the hash of the file.
+
+Creating the skeleton this way means opens up the mechanism to file sharing: if the same file was fetched in
+revision `bbbbbb`, it would have the same hash and the file would not need to be redownloaded.
```
sgugger, line 91: I think it would be helpful to include the example @julien-c put in the description to illustrate all the theoretical doc.

src/huggingface_hub/_snapshot_download.py
```diff
@@ -120,133 +141,84 @@ def snapshot_download(
...
-        # if there's an allowlist, skip download if file does not match any regex
-        if allow_regex is not None and not any(
-            fnmatch(model_file, r) for r in allow_regex
-        ):
-            continue
+        def resolve_ref(revision) -> str:
+            # retrieve commit_hash from file
+            ref_path = os.path.join(storage_folder, "refs", revision)
+            with open(ref_path) as f:
+                return f.read()
```
sgugger, line 166: This function is used once so I don't get why we refactor the three lines of code in it. E.g.,
```py
if REGEX_COMMIT_HASH.match(revision):
    commit_hash = revision
else:
    # retrieve commit_hash from file
    with open(ref_path) as f:
        commit_hash = f.read()
```

src/huggingface_hub/file_download.py
```diff
@@ -689,6 +730,73 @@ def _resumable_file_manager() -> "io.BufferedWriter":
...
+            )
+        else:
+            raise
+
+
+def repo_folder_name(
+    *,
+    repo_id: str,
+    repo_type: str,
+) -> str:
```
sgugger, line 786: Think even with a 80 char limit this fits in one line ;-)

src/huggingface_hub/file_download.py
```diff
@@ -689,6 +730,73 @@ def _resumable_file_manager() -> "io.BufferedWriter":
...
+    """Return a serialized version of a hf.co repo name and type, safe for disk storage
+    as a single non-nested folder.
+
+    Example: models--julien-c--EsperBERTo-small
+    """
+    # remove all `/` occurrences to correctly convert repo to directory name
+    parts = [
+        f"{repo_type}s",
+        *repo_id.split("/"),
+    ]
```
sgugger, line 796: Same here.

src/huggingface_hub/file_download.py
```diff
@@ -753,6 +886,10 @@ def hf_hub_download(
         local_files_only (`bool`, *optional*, defaults to `False`):
             If `True`, avoid downloading the file and return the path to the
             local cached file if it exists.
+        legacy_cache_layout (`bool`, *optional*, defaults to `False`):
+            If `True`, uses the legacy file cache layout i.e. just call `hf_hub_url`
```
sgugger, line 890:
```suggestion
            If `True`, uses the legacy file cache layout i.e. just call [`hf_hub_url`]
```
Link to `hf_hub_url`?

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +908,269 @@ def hf_hub_download(
...
+
+    if cache_dir is None:
+        cache_dir = HUGGINGFACE_HUB_CACHE
+    if revision is None:
+        revision = DEFAULT_REVISION
+    if isinstance(cache_dir, Path):
+        cache_dir = str(cache_dir)
+
+    if subfolder is not None:
+        filename = f"{subfolder}/{filename}"
```
sgugger, line 952: Maybe put a comment here that we use `/` since it is used for URLs or local paths, and that we will deal with / on Windows afterward?
(I did a comment on Windows that I deleted after seeing line 967 below.)

src/huggingface_hub/file_download.py
```diff
@@ -771,21 +908,269 @@ def hf_hub_download(
...
+            r.raise_for_status()
+            commit_hash = r.headers[HUGGINGFACE_HEADER_X_REPO_COMMIT]
+            if commit_hash is None:
+                raise OSError(
+                    "Distant resource does not seem to be on huggingface.co (missing"
+                    " commit header)."
+                )
+            etag = r.headers.get(HUGGINGFACE_HEADER_X_LINKED_ETAG) or r.headers.get(
+                "ETag"
+            )
```
sgugger, line 1021:
```suggestion
            etag = r.headers.get(HUGGINGFACE_HEADER_X_LINKED_ETAG)
            etag = etag or r.headers.get("ETag")
```
Probably a matter of personal preference, but I find it easier to read like this.

sgugger (approved):
Left some nits, feel free to ignore.
Very nice tests!

src/huggingface_hub/file_download.py
julien-c: you get the huge performance hit of doing two var assignments instead of one, and it's a bit less "functional" (well, it's Python, so...) 😅

(no strong opinion though)

LysandreJik:
Thank you, @patrickvonplaten, @sgugger, @julien-c and @thomwolf for the reviews! I'll go ahead and merge this now so that we can test it extensively while on `main` to ensure nothing breaks.

PR #804, opened 2022-03-28
TST calling delete_repo under tempfile for fixing the test

merveenoyan:
I thought I pushed this change but it wasn't pushed and then I saw the other branch was removed because it was merged.

I couldn't reproduce the error on my local, my guess is that I was calling `delete_repo` out of `tempfile` which caused the repository not found error. Hopefully this fix will work.


adrinjalali:
Oh interesting. Yeah this makes sense. Let's see what the CI says.

PR #806, opened 2022-03-29
Introducing list of tags to Keras model card

### LINKED ISSUE #800: Allow specifying multiple tags with the Keras mixin
As a user, I might want to not just specify the task name, but also other tags. Since tags with a task name are automatically identified as a task, we can simply accept a list of tags .

For example 
```
save_pretrained_keras(..., tags=["huggan", "gan", "text-to-image"])
```

cc @merveenoyan @NielsRogge @nateraw

merveenoyan:
This PR closes #800 
Resulting model card looks like [this](https://huggingface.co/merve/model_card_tags) when given the tags list ["audio", "speech"].


julien-c:
BTW (commenting here because i just glanced at the code): maybe more future proof to create the yaml part of the model card using `pyyaml` (which is already a dependency of this lib anyways)

I remember that's what we did in `transformers`

osanseviero:
Actually we already have `repocard.py` which could make things easier for this

https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/repocard.py

merveenoyan:
@julien-c thanks for the tip, it looks much cleaner!

osanseviero:
I think using `pyyaml` like this works, I will add some additional suggestions to this PR a bit later

src/huggingface_hub/keras_mixin.py
```diff
@@ -215,7 +218,7 @@ def push_to_hub_keras(
     git_email: Optional[str] = None,
     config: Optional[dict] = None,
     include_optimizer: Optional[bool] = False,
-    task_name: Optional[str] = None,
+    tags: Optional[list] = None,
```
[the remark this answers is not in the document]
merveenoyan: sure let me bring it back, thanks for the heads up!

src/huggingface_hub/keras_mixin.py
```diff
@@ -149,33 +159,28 @@ def save_pretrained_keras(
...
-        plot_model (`bool`, *optional*, defaults to `True`):
-            Setting this to `True` will plot the model and put it in the model
-            card. Requires graphviz and pydot to be installed.
-        task_name (`str`, *optional*):
-            Name of the task the model was trained on. Available tasks
-            [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts).
-        model_save_kwargs(`dict`, *optional*):
-            model_save_kwargs will be passed to
-            [`tf.keras.models.save_model()`](https://www.tensorflow.org/api_docs/python/tf/keras/models/save_model).
+    """Saves a Keras model to save_directory in SavedModel format. Use this if you're using the Functional or Sequential APIs.
```
osanseviero, line 166: The docstring format changed a bit in head for the new documentation, it would be great if you could align with https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/keras_mixin.py#L155 so there is no unrelated diff here

merveenoyan:
Merged main into this branch for other docs related changes and aligned docstring with main.

osanseviero:
> is there any formatter or style checker that can indicate if a docstring is fit? @osanseviero

Not sure, but as you can see at https://github.com/huggingface/huggingface_hub/pull/806#issuecomment-1084464233, the docs are actually built for you :fire:

src/huggingface_hub/keras_mixin.py
```diff
@@ -100,6 +102,7 @@ def _create_model_card(
     model,
     repo_dir: Path,
     plot_model: Optional[bool] = True,
+    tags: Optional[Union[list, str]] = None,
     task_name: Optional[str] = None,
```
osanseviero, line 106: Since this is a private method (as denoted by the `_`), I think you can freely remove `task_name` from here

src/huggingface_hub/keras_mixin.py
```diff
@@ -109,10 +112,17 @@ def _create_model_card(
...
-    model_card = "---\n"
+    metadata = {}
+    if isinstance(tags, list):
+        metadata["tags"] = tags
+    elif isinstance(tags, str):
+        metadata["tags"] = [tags]
     if task_name is not None:
-        model_card += f"tags:\n- {task_name}\n"
-    model_card += "library_name: keras\n---\n"
+        metadata["tags"].append(task_name)
```
osanseviero, line 121: And I would instead do this in the upstream methods (e.g. `save_pretrained_keras`)

src/huggingface_hub/keras_mixin.py
```diff
@@ -149,13 +159,13 @@ def save_pretrained_keras(
     config: Optional[Dict[str, Any]] = None,
     include_optimizer: Optional[bool] = False,
     plot_model: Optional[bool] = True,
+    tags: Optional[Union[list, str]] = None,
```
osanseviero, line 162: The issue with this approach is that for people that were using the `**model_save_kwargs`, this will break as the function would be expecting task_name now. 

@adrinjalali do you have any suggestions here? The goal is to deprecate `task_name` without making a breaking change, but since there is `**model_save_kwargs` it might be a bit more complex.

src/huggingface_hub/keras_mixin.py
```diff
@@ -149,13 +159,13 @@ def save_pretrained_keras(
...
     task_name: Optional[str] = None,
     **model_save_kwargs,
 ):
     """
     Saves a Keras model to save_directory in SavedModel format. Use this if
     you're using the Functional or Sequential APIs.
-
```
osanseviero, line 158: please remove unrelated change

src/huggingface_hub/keras_mixin.py
```diff
@@ -164,15 +174,19 @@ def save_pretrained_keras(
         save_directory (`str`):
             Specify directory in which you want to save the Keras model.
         config (`dict`, *optional*):
-            Configuration object to be saved alongside the model weights.
+                Configuration object to be saved alongside the model weights.
```
osanseviero, line 177: Please remove unneeded extra tab

src/huggingface_hub/keras_mixin.py
```diff
@@ -164,15 +174,19 @@ def save_pretrained_keras(
...
-            Setting this to `True` will plot the model and put it in the model
-            card. Requires graphviz and pydot to be installed.
+        tags (`dict`, *optional*):
+            List of tags that are related to model or string of a single tag. See example tags
+            [here](https://github.com/huggingface/hub-docs/blame/main/modelcard.md).
         task_name (`str`, *optional*):
             Name of the task the model was trained on. Available tasks
             [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts).
+            This is deprecated in favor of `tags` and will be removed in v0.7.
+        plot_model (`bool`, *optional*, defaults to `True`):
```
osanseviero, line 187: `plot_model` is before `tags` in the params I think

tests/test_keras_integration.py
```diff
@@ -300,7 +300,7 @@ def test_abs_path_from_pretrained(self):
             f"{WORKING_REPO_DIR}/{REPO_NAME}",
             config={"num": 10, "act": "gelu_fast"},
             plot_model=True,
-            task_name=None,
+            tags=None,
```
osanseviero, line 303: You should likely also add a test that asserts that a warning is raised when people use `task_name`

osanseviero (changes requested):
Thanks!

src/huggingface_hub/keras_mixin.py
adrinjalali: re backward compatibility, since these are positional args, you need to change the order between `tags` and `task_name` for people who were passing things positionally.

re `**model_save_kwargs`, after deprecation cycle is over, in this function we can have something like: ```if 'task_name' in model_save_kwargs: raise TypeError('You should be using `tags` instead of `task_name`')``` since `task_name` is not an arg to `save_model`.

adrinjalali: That said, I think backward compatibility constraints that we want to impose on `huggingface_hub` core parts don't necessarily apply here since the users here are mostly end users rather than library developers.

osanseviero: > re backward compatibility, since these are positional args, you need to change the order between tags and task_name for people who were passing things positionally.

With this change `tags` support both a list and an individual string, so this would still support positional params just fine. That is, the value os `task_name` is a valid value in `tags` and it has the same effect. 

Given your idea, I think we could entirely remove `task_name` from the param in this PR? We could just get it (if it is there) from `model_save_kwargs`. WDYT?

adrinjalali: that's true. Sounds good to me, we manually raise a warning now, and later a `TypeError`.

osanseviero: You could delete `task_name` from  `model_save_kwargs` after checking if it's there (or just pop it)

merveenoyan:
I left metadata part as dictionary for convenience.
So if the `task_name` is seen, user receives warning, it's popped from kwargs and passed to metadata. I've also put a test to see if the warning is raised.

```
save_pretrained_keras(
                model,
                f"{WORKING_REPO_DIR}/{REPO_NAME}",
                tags=["test"],
                task_name="test",
                save_traces=True,
            )
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -170,9 +172,9 @@ def save_pretrained_keras(
         plot_model (`bool`, *optional*, defaults to `True`):
             Setting this to `True` will plot the model and put it in the model
             card. Requires graphviz and pydot to be installed.
-        task_name (`str`, *optional*):
-            Name of the task the model was trained on. Available tasks
-            [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts).
+        tags (`dict`, *optional*):
```
osanseviero, line 175: The type here seems wrong, this should be a Union of `List` and `str` I think.

src/huggingface_hub/keras_mixin.py
```diff
@@ -199,7 +207,18 @@ def save_pretrained_keras(
         with open(path, "w") as f:
             json.dump(config, f)
 
-    _create_model_card(model, save_directory, plot_model, task_name)
+    metadata = {}
+    if isinstance(tags, list):
+        metadata["tags"] = tags
+    elif isinstance(tags, str):
+        metadata["tags"] = [tags]
+    if "task_name" in locals():
```
osanseviero, line 215: In lines 194 to 199 you can maybe do 

```py
task_name = model_save_kwargs.pop("task_name", None) # This means the default will be None
if task_name is not None:
   warnings.warn(
            "`task_name` input argument is removed. Pass `tags` instead.",
            FutureWarning,
        )
```

and similarly, here you won't need to check `locals()`, which seems a bit hacky, and you can just do
`if task_name is not None`.

src/huggingface_hub/keras_mixin.py
```diff
@@ -278,8 +297,8 @@ def push_to_hub_keras(
     git_email: Optional[str] = None,
     config: Optional[dict] = None,
     include_optimizer: Optional[bool] = False,
-    task_name: Optional[str] = None,
     plot_model: Optional[bool] = True,
+    tags: Optional[Union[list, str]] = None,
```
osanseviero, line 301: nit: you can keep `tags` before `plot_model` so we have backwards compatibility (in same place as `task_name` was). That means that if someone (unlikely) passes all arguments, this would still work with same code as before.

tests/test_keras_integration.py
```diff
@@ -272,6 +272,23 @@ def test_from_pretrained_weights(self):
             .item()
         )
 
+    def test_save_pretrained_task_name_deprecation(self):
```
osanseviero, line 275: nice!

osanseviero (commented):
Cool! This is looking safer and very nice! I left a couple of cleanup nits but almost ready to go :rocket:

src/huggingface_hub/keras_mixin.py
merveenoyan: it doesn't work, it says "`task_name` local variable mentioned before assignment" that's why I had to do it. I'll check if I mention it anywhere else.

src/huggingface_hub/keras_mixin.py
merveenoyan: I changed this

src/huggingface_hub/keras_mixin.py
merveenoyan: I changed this.

src/huggingface_hub/keras_mixin.py
osanseviero: Yes, you get the error because `task_name` is created within an if. My suggestion is to have this outside all the ifs. 

```
os.makedirs(save_directory, exist_ok=True)

task_name = model_save_kwargs.pop("task_name", None) # This means the default will be None
if task_name is not None:
   warnings.warn(
            "`task_name` input argument is removed. Pass `tags` instead.",
            FutureWarning,
        )
...
```

src/huggingface_hub/keras_mixin.py
```diff
@@ -328,9 +347,9 @@ def push_to_hub_keras(
             Configuration object to be saved alongside the model weights.
         include_optimizer (`bool`, *optional*, defaults to `False`):
             Whether or not to include optimizer during serialization.
-        task_name (`str`, *optional*):
-            Name of the task the model was trained on. Available tasks
-            [here](https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Types.ts).
+        tags (`dict`, *optional*):
```
osanseviero, line 350: The type here is off

src/huggingface_hub/keras_mixin.py
```diff
@@ -170,9 +172,9 @@ def save_pretrained_keras(
         plot_model (`bool`, *optional*, defaults to `True`):
             Setting this to `True` will plot the model and put it in the model
             card. Requires graphviz and pydot to be installed.
-        task_name (`str`, *optional*):
-            Name of the task the model was trained on. Available tasks
-            [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts).
+        tags (`list`, *optional*):
```
osanseviero, line 175: The type here is also off I think

merveenoyan:
@osanseviero my stupidity, I did fix them after I did a git add and I didn't add changes again apparently 😅😂 I was like "I could swear I changed them previously!!! am I hallucinating? 🥲" very sorry, I do listen to your comments before I resolve them, just saying.

src/huggingface_hub/keras_mixin.py
merveenoyan: ah thanks!

src/huggingface_hub/keras_mixin.py
```diff
@@ -199,7 +201,24 @@ def save_pretrained_keras(
...
+    metadata = {}
+    if isinstance(tags, list):
+        metadata["tags"] = tags
+    elif isinstance(tags, str):
+        metadata["tags"] = [tags]
+
+    task_name = model_save_kwargs.pop("task_name", None)
+    if task_name is not None:
+        warnings.warn(
+            "`task_name` input argument is removed. Pass `tags` instead.",
```
osanseviero, line 213:
```suggestion
            "`task_name` input argument is deprecated. Pass `tags` instead.",
```

osanseviero (approved):
Very nice! Thank you so much for this PR and iterating on it! :rocket:

adrinjalali:
Also, note that we're trying to merge other people's PRs instead of our own on this repo :) 

(I should get on writing our policies and their variations for our repos)

PR #809, opened 2022-03-30
Enable saving, loading and pushing for multiple Keras models in one repository

merveenoyan:
I started working on this previously, so it was another branch, will definitely fork for all my next PRs. This PR solves #533 

`push_to_hub_keras()` and `save_pretrained_keras()` now supports multiple models through passing them via a dictionary, like so:
(sharing only the parts affected)
```
push_to_hub_keras(
    {"model_1": model_1, "model_2": model_2},
    repo_path_or_name=f"{WORKING_REPO_DIR}/{REPO_NAME}",
    config={
                "model_1": {"num": 7, "act": "gelu_fast"},
                "model_2": {"num": 7, "act": "gelu_fast"},
            },
    multiple_model_save_args={
        "model_1": {"save_traces": False},
        "model_2": {"save_traces": True},
    },
)
```

So model can be either a Keras model or dictionary of Keras models. You can apply different save args for different models through `multiple_model_save_args` (which is completely optional) and `config` works the same as model dict.

This creates a directory like so:
Ekran Resmi 2022-03-30 17 11 54

You can load with `from_pretrained_keras` and it returns a dict of models if there are multiple models in the repository: 

`models_dict = from_pretrained_keras(dir)`
Which looks like the one we've pushed:
`{"generator": some keras model, "discriminator": other keras model}`

Model card has a shared part at the top:
Ekran Resmi 2022-03-30 17 12 30
And model specific parts at the bottom: 
Ekran Resmi 2022-03-30 17 13 13

See [this example repository](https://huggingface.co/merve/multiple-models-readme).

I wrote four tests: `from_pretrained_keras` that checks if the values in returned dictionary are Keras model classes, `save_pretrained_keras` that checks if the model folders are there and the required files are inside model folders, and two tests for`push_to_hub_keras` that checks if the pushed models are there and another one for if the kwargs are injected into `tf.keras.models.save_model()` for each model.

There seem to be many commits because it was another branch based on model card PR, but there actually is two commits on top of that.

There's no deprecation whatsoever but I might add some warnings for when people are pushing only one model and are passing values to `multiple_model_save_args` and others.


merveenoyan:
@nateraw I saw you noted down to get rid of config for `from_pretrained_keras()` would you like me to do that? I think it doesn't make any sense and not used anyway.

merveenoyan:
"I love resolving merge conflicts" said by no one. ever.

merveenoyan:
I wonder if we can use [builder](https://sbcode.net/python/builder/) to get rid of dict of dicts.

src/huggingface_hub/keras_mixin.py
```diff
@@ -173,6 +201,9 @@ def save_pretrained_keras(
         task_name (`str`, *optional*):
             Name of the task the model was trained on. Available tasks
             [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts).
+        multiple_model_save_args (`dict`, *optional*):
```
nateraw, line 204: This param seems very confusing, since its essentially the same thing as `model_save_kwargs`, but its just multiple...so then its slightly unclear to a user which one they should use.

nateraw (changes requested):
As we discussed offline, I am not sure I am in favor of these changes. However, I would like to start a discussion with others about this feature so we can address the problems that it is trying to solve. 

For others - context here is that some models have more than one component that you would want to push separately (think Generator and Discriminator in a GAN). _In these cases, each separate model would have its own saved files and associated args/config._

There are two approaches to solve this problem off the top of my head:
1. Push up models in separate subdirs, as this PR intends to do 
2. Push up the separate model components to separate repositories (this one would require no code changes at the expense of having two potentially out-of-sync repos). 

The reason specifically why I'm not too sure about the changes here is because it seems we're moving away from the generic API put forward by the generic PushToHub mixin, and sort of hacking around things for Keras. would be nice to think about this problem from a higher level to see what we want to do across the board. 

CC @osanseviero @LysandreJik

src/huggingface_hub/keras_mixin.py
merveenoyan: I was thinking of having this in kwargs as a dictionary and popping it in case multiple models are pushed, but not sure. 🤔

merveenoyan:
@osanseviero don't you need discriminator if you'd like to train again, I never needed to do that so I thought I'd ask. I don't know why but as of now, `tf.keras.models.save_model` doesn't support serial graph saving that's why I did that. If you want I can look for a fix for this.

osanseviero:
Yes, I've seen people saving a model as a single one with TF before, so I would maybe try that way.

osanseviero:
https://stackoverflow.com/questions/57812324/saving-and-loading-multiple-models-with-the-same-graph-in-tensorflow-functional for example

merveenoyan:
pinging @ariG23498 here 🤩

ariG23498:
What is the update on this? I am sorry I could not reply any sooner.

PR #810, opened 2022-03-31
MNT change dev version to 0.5, 0.4 is already released

adrinjalali:
We're releasing 0.5, not 0.4.


LysandreJik (approved):
LGTM, thanks for the change @adrinjalali!

LysandreJik:
Re-running the tests now that the inference-api rate limit was fixed.

adrinjalali:
```python-traceback
self = <tests.test_hf_api.HfApiPublicTest testMethod=test_model_info_with_security>

    @with_production_testing
    def test_model_info_with_security(self):
        _api = HfApi()
        model = _api.model_info(
            repo_id=DUMMY_MODEL_ID,
            revision=DUMMY_MODEL_ID_REVISION_ONE_SPECIFIC_COMMIT,
            securityStatus=True,
        )
        self.assertEqual(
>           getattr(model, "securityStatus"),
            {"containsInfected": False, "infectionTypes": []},
        )
E       AttributeError: 'ModelInfo' object has no attribute 'securityStatus'
```

we've seen this before, didn't we fix it?

adrinjalali:
O changed the version to `0.5` instead of `0.5.0` since minor bugfix number is irrelevant for the purpose of people knowing when something was added or changed (I think)

LysandreJik:
Sounds good, LGTM. Merging!

cc @muellerzr I believe you had encountered this issue in the past?

muellerzr:
Yeah, it just needs to be reran (not a fan of that of course). It's an issue on the db side IIRC.

julien-c:
yes i'm aware of this server-side issue (flakyness on `securityStatus`, I'll create an action item to address that (cc @McPatate)

Issue #811, opened 2022-03-31
[API Reference docs] The `build_documentation` Action does not run successfully


LysandreJik:
The `build_documentation` action fails as it's unable to find the workflow:

https://github.com/huggingface/huggingface_hub/actions/runs/2070320057

PR #813, opened 2022-03-31
[API Reference docs] Fix `build_documentation` Action

LysandreJik:
This PR fixes the `build_documentation` action by removing the git reference used previously. This was a reference to a branch that has since been merged in the default branch.

Closes https://github.com/huggingface/huggingface_hub/issues/811

### DIFF
--- a/.github/workflows/build_documentation.yaml
+++ b/.github/workflows/build_documentation.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
    build:
-    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@workflow_templates
+    uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml
     with:
       commit_sha: ${{ github.sha }}
       package: huggingface_hub


adrinjalali:
@LysandreJik can we replace the comment left by @HuggingFaceDocBuilderDev by a github action whose `Details` would link to the artifacts? It's adding too much noise to the PRs IMO.

adrinjalali:
Should the reference be `main` or the PR? Ideally we'd like the CI on `main` to build and publish the docs, and the PRs to build the docs for the PR (ideally only the changed files)

LysandreJik:
> @LysandreJik can we replace the comment left by @HuggingFaceDocBuilderDev by a github action whose Details would link to the artifacts? It's adding too much noise to the PRs IMO.

I'd rather we stay coherent with the rest of the HF libraries which all provide the docs in a very accessible manner. Could you show me what you have in mind with a GitHub Action whose details would link to the artifacts? I believe we had something similar with CircleCI in the past, but none of the users were aware of it even if mentioned in the contributing guide.

> Should the reference be main or the PR? Ideally we'd like the CI on main to build and publish the docs, and the PRs to build the docs for the PR (ideally only the changed files)

The reference specified here is the reference to `doc-builder`. Here it will use the workflows defined on the `main` branch of doc-builder.

The CI currently does the following:
- On the `main` branch, it builds and publishes the docs to the `main` URL (which will be visible at https://huggingface.co/docs/huggingface_hub/main/en/index)
- On pull requests, builds and publishes the docs to a short-lived environment that will persist while the PR is open (and will reinstantiate when the PR reopens), which links to https://moon-ci-docs.huggingface.co/docs/huggingface_hub/pr_<PR_NUMBER>
- On releases, publishes to the URL with the appropriat version. The latest version released becomes the default version shown.

adrinjalali:
This is how it could look like, which you can see in this PR (https://github.com/scikit-learn/scikit-learn/pull/23002), and it only compiles the changed documents, with links to the PR version vs the main (dev) version vs the latest stable release (we really don't have to implement all of that here). The `Details` on the `ci/circleci: doc artifact` action directly points to where the built documentation sits:

Screenshot from 2022-03-31 14-10-44

The solution right now adds a comment (at least one I guess) to every PR, that's a ton of noise. HuggingFace is not a single community, and I think the community of each project should decide how to configure the tools they use rather than it being one solution for all.

LysandreJik:
I understand. Would you be open to discussing this in a separate issue, as it's not linked to this PR; and to merge this PR so that we may have the docs deployed?

adrinjalali (approved):

PR #815, opened 2022-03-31
👨‍💻 Configure HF Hub URL with environment variable

### LINKED ISSUE #808: `HfApi.endpoint` should be configurable in `Repository`
https://github.com/huggingface/huggingface_hub/blob/c2ffc48cd4d9c0271c0cc68d17a11e1088d3fa1a/src/huggingface_hub/repository.py#L434

We should be able to pass arguments to `HfApi()` from `Repository`

SBrandeis:
Closes #808

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ def get_version() -> str:
 
 install_requires = [
     "filelock",
-    "requests",
+    "requests>=2.27",
     "tqdm",
     "pyyaml",
     "typing-extensions>=3.7.4.3",  # to be able to import TypeAlias
@@ -27,11 +27,7 @@ extras["torch"] = [
     "torch",
 ]
 
-extras["tensorflow"] = [
-    "tensorflow",
-    "pydot",
-    "graphviz"
-]
+extras["tensorflow"] = ["tensorflow", "pydot", "graphviz"]
 
 extras["testing"] = [
     "pytest",
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -81,7 +81,7 @@ def _validate_repo_id_deprecation(repo_id, name, organization):
     return name, organization
 
 
-def repo_type_and_id_from_hf_id(hf_id: str):
+def repo_type_and_id_from_hf_id(hf_id: str, _hf_co_domain: Optional[str] = None):
     """
     Returns the repo type and ID from a huggingface.co URL linking to a
     repository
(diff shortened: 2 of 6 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -80,7 +81,7 @@ def _validate_repo_id_deprecation(repo_id, name, organization):
     return name, organization
 
 
-def repo_type_and_id_from_hf_id(hf_id: str):
+def repo_type_and_id_from_hf_id(hf_id: str, _hf_co_domain: Optional[str] = None):
```
adrinjalali, line 84: why a private-ish input arg? We can have it public, and maybe call it `hub_url`? Also needs to be documented in the docstring.

adrinjalali (commented):
This is certainly an interesting solution to the issue.

I think it's an improvement to what we have, so I'd be happy to merge, but I was thinking of a solution where `HfApi` accepts a hub url, and that one could pass a `HfApi` instance to `Repository`. The latter solution would allow for further customization of `HfApi` used in `Repository`.

But I'm happy with this solution anyway.

SBrandeis:
Is [d2dba15](https://github.com/huggingface/huggingface_hub/pull/815/commits/d2dba15246690c8099c4eeeae293b50ac02327cc) more aligned with what you had in mind ?

setup.py
```diff
@@ -13,7 +13,7 @@ def get_version() -> str:
 
 install_requires = [
     "filelock",
-    "requests",
+    "requests>=2.27",
```
adrinjalali, line 16: What makes us require a very recent version of `requests`? This potentially creates dependency conflicts for users trying to install our library alongside other packages which potentially don't support latest `requests` versions. What's the oldest version we can work with here?

adrinjalali (commented):
This looks great now.

setup.py
SBrandeis: Minimal required version of `requests` is 2.27.0 because [we use `JSONDecodeError`](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L24), and it was introduced only in `2.27.0`: https://docs.python-requests.org/en/latest/community/updates/#id2

LysandreJik (approved):
Thanks for working on it, @SBrandeis! 

[Out of scope for this PR, but some thoughts about passing a `HfApi` argument to other classes]
I didn't expect `HfApi` to be an argument to `Repository`. I think it is an elegant solution to the problem, but I wonder if the name that is given to its instance, `hf_api`, is the best to describe what it actually does.

To me `hf_api` describes an API that has methods you can call. But here, we're using it as a stateful client instance to the API, which contains configured values which may be leveraged by other classes. To that end, I wonder if `client` or something along those lines wouldn't be a better variable name after an `HfApi` instantiation.

setup.py
julien-c: When was `2.27.0` released? and do we know of libraries that pin an upper bound on the version of `requests` they use?

SBrandeis: `2.27.0`  was released on 2022-03-01

SBrandeis: `requests.JSONDecodeError` was introduced by https://github.com/psf/requests/pull/5856

julien-c: Ok, in my opinion this is ok to ship this version pin, given that most libraries would probably not pin an upper bound on the version of `requests` they use. And we can always relax this constraint if we realize it's affecting users?

adrinjalali: Let's merge this with this pinned version, I'll open a PR today to fix it and remove the constraint before we release a new version. The pinned version is way too recent. cc @LysandreJik

src/huggingface_hub/repository.py
```diff
@@ -441,6 +442,7 @@ def __init__(
         revision: Optional[str] = None,
         private: bool = False,
         skip_lfs_files: bool = False,
+        hf_api: Optional[HfApi] = None,
```
adrinjalali, line 445: I agree with @LysandreJik that `client` would be a much better name here for the argument.

adrinjalali (commented):
Other than renaming the argument, LGTM.

adrinjalali (approved):

adrinjalali:
Thanks for the work @SBrandeis. For future PRs, it'd be nice if you could send them from your own fork rather than creating feature branches on the main repo :) We're also trying to have a reviewer merge the PR rather than the author of the PR here :)

SBrandeis:
OK, noted !

PR #817, opened 2022-04-04
MNT support oder requests versions

### LINKED ISSUE #816: Relax version constraint on `requests`
#815 introduces a requirement on the minimal version of `requests` (`>=2.27`)

This requirement should be relaxed because 2.27 is way too recent

> Let's merge this with this pinned version, I'll open a PR today to fix it and remove the constraint before we release a new version. The pinned version is way too recent. cc @LysandreJik

_Originally posted by @adrinjalali in https://github.com/huggingface/huggingface_hub/pull/815#discussion_r841533043_

adrinjalali:
I've tested locally for all possibilities of installed packages (simplejson, json, old and new requests).

I don't think we need to add CI to test for old requests though.

Fixes #816

### DIFF
--- a/.github/conda/meta.yaml
+++ b/.github/conda/meta.yaml
@@ -1,4 +1,4 @@
-{% set name = "huggingface_hub" %}  # noqa
+{% set name = "huggingface_hub" %}
 
 package:
   name: "{{ name|lower }}"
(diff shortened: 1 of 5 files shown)


LysandreJik (approved):
Thanks for the effort, @adrinjalali!

julien-c (commented):
cool!

PR #819, opened 2022-04-05
Rename the env variable `HF_ENDPOINT`.

### LINKED ISSUE #798: Is it possible to make `HF_ENDPOINT` endpoint overridable in environment variable ?

 We want to update the endpoint for this library (actually for `datasets`) for hub reasons.
AFAIK, it's actually also necessary for China.

Something like in `transformers` : https://github.com/huggingface/transformers/blob/main/src/transformers/utils/hub.py#L110
Having the same defaults, just make it overridable with an env variable (other mecanisms are fine, just seems env variable are consistent and easy)

Also `ENDPOINT` and `HUGGINGFACE_CO_URL_HOME` seem to be duplicate names.

Current fix, is by monkey patching the relevant variables from external code, which is not ideal for maintenance reasons.

@LysandreJik @osanseviero  ? 

This could be a good opportunity to use the same variables across HF libraries in datasets, I rechecked and it's using `HF_ENDPOINT` actually.

Narsil:
This improves consistency across HF ecosystem.
https://github.com/huggingface/transformers/blob/c595b6e6a90c29242c9dbfeecc5b132fe2e3d1f4/src/transformers/utils/hub.py#L110
https://github.com/huggingface/datasets/blob/8c5f27889657a1a6a7203b2f054bd0ceffce2178/src/datasets/config.py#L24

Fixes #798

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -467,7 +467,7 @@ def erase_from_credential_store(username=None):
 class HfApi:
     def __init__(self, endpoint=None):
         self.endpoint = (
-            endpoint if endpoint is not None else os.getenv("HF_ENDPOINT", ENDPOINT)
+            endpoint if endpoint is not None else ENDPOINT
         )
 
     def login(self, username: str, password: str) -> str:
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -466,9 +466,7 @@ def erase_from_credential_store(username=None):
 
 class HfApi:
     def __init__(self, endpoint=None):
-        self.endpoint = (
-            endpoint if endpoint is not None else ENDPOINT
-        )
+        self.endpoint = endpoint if endpoint is not None else ENDPOINT
 
     def login(self, username: str, password: str) -> str:
         """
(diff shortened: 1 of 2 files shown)


Narsil:
Pinging @SBrandeis since your made the initial modification so you are aware of this name change.

Is it OK to change the name here ?

SBrandeis:
Yes @Narsil that is fine

src/huggingface_hub/hf_api.py
```diff
@@ -467,7 +467,7 @@ def erase_from_credential_store(username=None):
 class HfApi:
     def __init__(self, endpoint=None):
         self.endpoint = (
-            endpoint if endpoint is not None else os.getenv("HF_ENDPOINT", ENDPOINT)
+            endpoint if endpoint is not None else ENDPOINT
```
Narsil, line 470: I removed this, since `ENDPOINT` is already inheriting from the same env variable, to reduce the potential from divergence.

osanseviero (approved):
Looks great! Thank you!

adrinjalali (approved):

adrinjalali:
Thanks for the fix @Narsil . It'd be nice if for future PRs you'd create the feature branches on your fork rather than the main repo 😊

Narsil:
> Thanks for the fix @Narsil . It'd be nice if for future PRs you'd create the feature branches on your fork rather than the main repo blush

Ok, thanks for the heads up.

PR #820, opened 2022-04-05
Release guide

LysandreJik:
This PR adds a small release guide to the `setup.py` for releases.

I added it here as it's the same location as for the `transformers` library and it seems adequate, but happy to move it somewhere else more appropriate.


LysandreJik:
Following the guide to do a patch, I realize there's some info missing. Will complete once I'm done with the patch.

setup.py
```diff
@@ -1,4 +1,60 @@
...
+   to the version you're releasing:
+   ```
+   __version__ = "<VERSION>"
+   ```
+   
+4. Make sure that the conda build works correctly by building it locally:
+   ```
+   conda install -c defaults anaconda-client conda-build
+   HUB_VERSION=<VERSION> conda-build .github/conda
+   ```
```
adrinjalali, line 24: do we want to maintain conda here? `transformers` doesn't seem to be uploading to our channel anymore, and we have our package on conda-forge. That recipe is more up-to-date.

setup.py
```diff
@@ -1,4 +1,60 @@
...
+   python setup.py sdist bdist_wheel
+   pip install dist/huggingface_hub-<VERSION>-py3-none-any.whl
+   ```
+   
+6. Commit, tag, and push the branch:
+   ```
+   git commit -am "Release: v<VERSION>"
+   git tag v<VERSION> -m "Adds tag v<VERSION> for pypi and conda"
+   git push -u --tags origin v<MINOR-VERSION>-release
+   ```
```
adrinjalali, line 38: to give visibility to others, this can be a PR to the release branch instead of direct commit.

adrinjalali: It would also help with checking the doc build on that PR.

setup.py
```diff
@@ -1,4 +1,60 @@
...
+
+7. Verify that the docs have been built correctly. You can check that on the following link:
+   https://huggingface.co/docs/huggingface_hub/v<VERSION>
+   
+8. Checkout main once again to update the version in the `__init__.py` file:
+   ```
+   git checkout main
+   ```
```
adrinjalali, line 46: this can also be a PR

adrinjalali (commented):
Thanks for the PR @LysandreJik . I think it'd be nice to move this to `doc/dev/release.mdx` rather than having it in `setup.py`.

setup.py
LysandreJik: Right now the build works well and is automatic; I'd like to draft a small message to share with the rest of the team regarding conda support and whether we want to continue to support it through our own channel or not. In the meantime, I'd like to keep it here.

setup.py
LysandreJik: How would you name the branch from which that PR is made? Others will get visibility with the release already, I am not 100% sure I see why the PR is necessary here

adrinjalali: that PR comes from the maintainer's own fork, doesn't matter how it's named. E.g: https://github.com/scikit-learn/scikit-learn/pull/21120

In most cases it'll be a simple PR, but sometimes you might want to include more commits from the `main` branch.

adrinjalali (approved):
This is a good start anyway. I'm happy for this to be merged, updated or not :)

LysandreJik:
I'm having a hard time finding coherent names for the branches as I haven't followed that workflow previously; I'll merge this as it is, but very open if you want to open another PR to complete the existing doc with the two additional branches :)

Issue #821, opened 2022-04-06
HfApi.create_repo fails in 0.5 if the org name is passed in `repo_id` and `organization` arguments


lhoestq:
As reported in https://github.com/huggingface/datasets/issues/4105 if you pass a `repo_id` (containing the org name or not), and also the `organization` parameter to `HfApi.create_repo`, then you get this error:
```python
~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/huggingface_hub/utils/_deprecation.py in inner_f(*args, **kwargs)
     44             )
     45             kwargs.update(zip(sig.parameters, args))
---> 46             return f(**kwargs)
     47 
     48         return inner_f

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/huggingface_hub/hf_api.py in create_repo(self, repo_id, token, organization, private, repo_type, exist_ok, space_sdk, name)
   1260         - [1] https://huggingface.co/settings/tokens
   1261         """
-> 1262         name, organization = _validate_repo_id_deprecation(repo_id, name, organization)
   1263 
   1264         path = f"{self.endpoint}/api/repos/create"

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/huggingface_hub/hf_api.py in _validate_repo_id_deprecation(repo_id, name, organization)
     66     if repo_id and (name or organization):
     67         raise ValueError(
---> 68             "Only pass `repo_id` and leave deprecated `name` and "
     69             "`organization` to be None."
     70         )

ValueError: Only pass `repo_id` and leave deprecated `name` and `organization` to be None.
```

This used to be allowed in 0.4 and is a breaking change that makes `push_to_hub` fail for all versions of `datasets`.

As a general rule, deprecations should also log warnings rather than fail.

adrinjalali:
PR with the hotfix: https://github.com/huggingface/huggingface_hub/pull/822

PR #822, opened 2022-04-06
FIX don't raise if name/organizaiton are passed postionally

adrinjalali:
Fixes https://github.com/huggingface/huggingface_hub/issues/821

If user passes `name` as a positional argument and pass `organization` as well, we fail, whereas we should raise a `FutureWarning`.

cc @LysandreJik @osanseviero 

This needs to be backported to 0.5.1 for a patch release.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -59,7 +59,7 @@ logger = logging.get_logger(__name__)
 def _validate_repo_id_deprecation(repo_id, name, organization):
     """Returns (name, organization) from the input."""
     if repo_id and not name and organization:
-        # this means the user pad passed name as positional, now mapped to
+        # this means the user had passed name as positional, now mapped to
         # repo_id and is passing organization as well. This wouldn't be an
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -58,6 +58,13 @@
 # TODO: remove after deprecation period is over (v0.7)
 def _validate_repo_id_deprecation(repo_id, name, organization):
     """Returns (name, organization) from the input."""
+    if repo_id and not name and organization:
+        # this means the user pad passed name as positional, now mapped to
```
lhoestq, line 62:
```suggestion
        # this means the user had passed name as positional, now mapped to
```

lhoestq (approved):
Thanks ! Just tested with `datasets` and it fixes `push_to_hub` :)

osanseviero (approved):
Thanks! 👍

LysandreJik (approved):
LGTM, thanks @adrinjalali

PR #790, opened 2022-03-23
CI support python 3.7-3.10 - remove 3.6 support

adrinjalali:
This PR removes tests on 3.6 and adds 3.10 support on the CI.

Fixes #786

PR #823, opened 2022-04-06
Allow installation on Python 3.6

albertvillanova:
This PR unrolls a change introduced by:
- #790

That change prevents the installation of `huggingface_hub` on Python 3.6 (used by `datasets` CI).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
             "huggingface-cli=huggingface_hub.commands.huggingface_cli:main"
         ]
     },
-    python_requires=">=3.7.0",
+    python_requires=">=3.6.0",
     install_requires=install_requires,
     classifiers=[
         "Intended Audience :: Developers",


albertvillanova:
According to my understanding, I thought the main target of PR #790 was to add CI testing for Python 3.10; and eventually, start a deprecation cycle for Python 3.6; but not immediately stop support for Python 3.6.

adrinjalali:
If you need to support python=3.6, I suggest supporting an older version of `huggingface_hub`. Python=3.6 support is removed already by many libraries (numpy, scipy included), and we rather not have support for it here.

albertvillanova:
If you think it is OK to stop supporting Python 3.6 without a deprecation cycle...

adrinjalali:
Python=3.7 also gives us new features that soon we can start using (like dataclasses), so supporting 3.6 would really be a hassle.

Also, please create PRs from your own fork rather than a branch on the main repo :)

adrinjalali:
> If you think it is OK to stop supporting Python 3.6 without a deprecation cycle...

This is a good point, however, if we start having deprecation cycles for versions we'd like to stop supporting, users of the lowest python version would pretty much constantly be seeing those warnings.

This change for the user means if they're on py=3.6 and they install datasets, they'll get the older version of datasets with an older version of huggingface_hub. I don't think that's a big issue.

Issue #825, opened 2022-04-08
push_to_hub_keras rejected because it contains binary files


nateraw:
I noticed I can't push the LFS files associated with Keras models anymore. What's interesting is that I'm experiencing the same issue on older versions of `huggingface_hub` (0.4.0) as well as new versions (0.5.1). 

Here is a minimal example to recreate the issue:

```python
import tensorflow as tf
from tensorflow.keras import layers, Sequential

from huggingface_hub import push_to_hub_keras

model = Sequential(
    [
        layers.Dense(1000, activation="relu", name="layer1"),
        layers.Dense(1000, activation="relu", name="layer2"),
        layers.Dense(1000, name="layer3"),
    ]
)
model.build((None, 2))

push_to_hub_keras(model, 'test-save-keras-sequential')
```

You can also try it yourself in Colab: [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/gist/nateraw/dcb52b7b9846a41006000709d8c40159/keras-push-to-hub-rejected.ipynb)

When you run the above, you get this...I redacted my token which was being printed here.

```
remote: -------------------------------------------------------------------------        
remote: Your push was rejected because it contains binary files.        
remote: Please use https://git-lfs.github.com/ to store binary files.        
remote: See also: https://hf.co/docs/hub/adding-a-model#uploading-your-files        
remote: -------------------------------------------------------------------------        
remote: Offending files:        
remote:  - variables/variables.data-00000-of-00001 (ref: refs/heads/main)        
To https://huggingface.co/nateraw/test-save-keras-sequential
 ! [remote rejected] main -> main (pre-receive hook declined)
error: failed to push some refs to 'https://user:<REDACTED TOKEN>@huggingface.co/nateraw/test-save-keras-sequential'
```

LysandreJik:
Linked to https://github.com/huggingface/huggingface_hub/issues/687

merveenoyan:
The issue linked by @LysandreJik is for git lfs not tracking small files, what Dimitre pushed was a GAN model and I doubt that's a small file.

LysandreJik:
The issue linked above is due to binary files not being automatically tracked by git-lfs; in this instance, the push was rejected because it contains binary file. I believe having automatic tracking via git-lfs for binary files would solve the issue shown by @nateraw's code sample above.

LysandreJik:
[...]

Issue #687, opened 2022-02-16
Git add with auto_lfs_track should track binary files


osanseviero:
When pushing binary files, users get the following error

```
remote: Your push was rejected because it contains binary files.
remote: Please use https://git-lfs.github.com/ to store binary files.
```

With auto LFS track this should be tracked by default.

WDYT @LysandreJik @julien-c ?

julien-c:
i don't remember the `auto_lfs_track` implementation, is there a file size threshold somewhere? We changed the file size threshold for binary files only to `1MB` on the hub side recently. Non binary files still `10MB`

osanseviero:
Yes, we set a 10mb auto track at https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/repository.py#L849-L853

We could do some similar check for binary files for 1MB?

LysandreJik:
Hey @osanseviero, indeed, happy to do that. Do you have a reproducible code example I can play with?

osanseviero:
Yes @LysandreJik. 

```python
from huggingface_hub import Repository, HfApi
import os

repo_url = HfApi().create_repo("test-bin-bug")
repo = Repository("local_repo", clone_from=repo_url)
with open(os.path.join("local_repo", "file"), "wb") as out:
    out.truncate(1024*1024)
repo.push_to_hub("Commit #1")
```

Pierrci:
@LysandreJik you can check https://github.com/huggingface/moon-landing/blob/5fa4de96cb48dc0613933d2aa7e3311bbd0abaac/server/scripts/git/pre-receive#L143 for what we use server-side to detect binary files (cc @coyotte508 who implemented it)

LysandreJik:
Will be closed by https://github.com/huggingface/huggingface_hub/pull/828

PR #828, opened 2022-04-08
Auto track binary files

LysandreJik:
Closes https://github.com/huggingface/huggingface_hub/issues/825.
Closed https://github.com/huggingface/huggingface_hub/issues/687.

This adds the option to automatically track binary files. It follows the following two assumptions:
- Binary files usually cannot be read with the `open` file object reader, which raises a unicode error.
- The file can sometimes be read even with the `open` keyword; in that case, we search for the null character which should only be present in binary files, and which seems to be what `grep` uses in order to identify binary files (and we use `grep` in the backend to look for these binary files)

This PR adds a new method, `auto_track_binary_files`, which automatically tracks binary files. 

A change could be seen as a breaking change, which is the `auto_lfs_track` keyword of the `git_add` method. Previously, it only tracked large files, it now tracks both large files and binary files.
I left it as I believe it's more of a bugfix than a breaking change, as all binary files would have been previously rejected, resulting in errors. 

If this is not deemed acceptable, the deprecation cycle would be to:
- add the `auto_track_large_files` and `auto_track_binary_files` keyword arguments to `git_add`, to toggle manually, and deprecate `auto_lfs_track` to mention that it will soon handle both large and binary files.

The tests have been added directly to the others that test for large files, as this results in a significant speedup vs splitting these over two tests.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -226,6 +226,26 @@ def is_git_ignored(filename: Union[str, Path]) -> bool:
     return is_ignored
 
 
+def is_binary_file(filename: Union[str, Path]) -> bool:
+    """
+    Check if file is a binary file.
+
+    Args:
+        filename (`str` or `Path`):
+            The filename to check.
+
+    Returns:
+        `bool`: `True` if the file passed is a binary file, `False`
+        otherwise.
+    """
+    try:
+        with open(filename) as f:
+            f.read()
+        return False
+    except UnicodeDecodeError:
+        return True
+
+
 def files_to_be_staged(pattern: str, folder: Union[str, Path]) -> List[str]:
     """
     Returns a list of filenames that are to be staged.
@@ -981,6 +1001,41 @@ class Repository:
         except subprocess.CalledProcessError as exc:
             raise EnvironmentError(exc.stderr)
 
+    def auto_track_binary_files(self, pattern: Optional[str] = ".") -> List[str]:
+        """
+        Automatically track binary files with git-lfs.
+
+        Args:
+            pattern (`str`, *optional*, defaults to "."):
+                The pattern with which to track files that are above 10MBs.
+
+        Returns:
+            `List[str]`: List of filenames that are now tracked due to being binary files
+        """
+        files_to_be_tracked_with_lfs = []
+
+        deleted_files = self.list_deleted_files()
+
+        for filename in files_to_be_staged(pattern, folder=self.local_dir):
+            if filename in deleted_files:
+                continue
+
+            path_to_file = os.path.join(os.getcwd(), self.local_dir, filename)
+            is_binary = is_binary_file(path_to_file)
+
+            if (
+                    is_binary
+                    and not is_tracked_with_lfs(path_to_file)
+                    and not is_git_ignored(path_to_file)
+            ):
+                self.lfs_track(filename)
+                files_to_be_tracked_with_lfs.append(filename)
+
+        # Cleanup the .gitattributes if files were deleted
+        self.lfs_untrack(deleted_files)
+
+        return files_to_be_tracked_with_lfs
+
     def auto_track_large_files(self, pattern: Optional[str] = ".") -> List[str]:
         """
         Automatically track large files (files that weigh more than 10MBs) with
@@ -1094,7 +1149,10 @@ class Repository:
                 file over 10MB in size will be automatically tracked.
         """
         if auto_lfs_track:
-            tracked_files = self.auto_track_large_files(pattern)
+            tracked_files = [
+                *self.auto_track_large_files(pattern),
+                *self.auto_track_binary_files(pattern)
+            ]
             if tracked_files:
                 logger.warning(
                     f"Adding files tracked by Git LFS: {tracked_files}. This may take a bit of time if the files are large."
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -239,7 +239,7 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
     """
     try:
         with open(filename) as f:
-            content = f.read()
+            content = f.read(512)
 
         # Check for the presence of the null character in the string
         return "\x00" in content
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -238,11 +238,15 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
         `bool`: `True` if the file passed is a binary file, `False` otherwise.
     """
     try:
-        with open(filename) as f:
+        with open(filename, "rb") as f:
             content = f.read()
 
-        # Check for the presence of the null character in the string
-        return "\x00" in content
+        # Code sample taken from the following stack overflow thread
+        # https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python/7392391#7392391
+        text_chars = bytearray(
+            {7, 8, 9, 10, 12, 13, 27} | set(range(0x20, 0x100)) - {0x7F}
+        )
+        return bool(content.translate(None, text_chars))
     except UnicodeDecodeError:
         return True
 
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -235,15 +235,14 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
             The filename to check.
 
     Returns:
-        `bool`: `True` if the file passed is a binary file, `False`
-        otherwise.
+        `bool`: `True` if the file passed is a binary file, `False` otherwise.
     """
     try:
         with open(filename) as f:
             content = f.read()
 
         # Check for the presence of the null character in the string
-        return '\x00' in content
+        return "\x00" in content
     except UnicodeDecodeError:
         return True
 
@@ -507,8 +506,8 @@ class Repository:
             skip_lfs_files (`bool`, *optional*, defaults to `False`):
                 whether to skip git-LFS files or not.
             client (`HfApi`, *optional*):
-                Instance of HfApi to use when calling the HF Hub API.
-                A new instance will be created if this is left to `None`.
+                Instance of HfApi to use when calling the HF Hub API. A new
+                instance will be created if this is left to `None`.
         """
 
         os.makedirs(local_dir, exist_ok=True)
@@ -1012,7 +1011,8 @@ class Repository:
                 The pattern with which to track files that are above 10MBs.
 
         Returns:
-            `List[str]`: List of filenames that are now tracked due to being binary files
+            `List[str]`: List of filenames that are now tracked due to being
+            binary files
         """
         files_to_be_tracked_with_lfs = []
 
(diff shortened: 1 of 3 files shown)


LysandreJik:
This seems like a partial fix to the issue, as this code sample by @osanseviero will still fail:
```py
from huggingface_hub import Repository, HfApi
import os

repo_url = HfApi().create_repo("test-bin-bug-4")
repo = Repository("local_repo", clone_from=repo_url)
with open(os.path.join("local_repo", "file"), "wb") as out:
    out.truncate(1024*1024)
repo.push_to_hub("Commit #1")
```

Trying to find a cross-platform approach that manages both.

LysandreJik:
Checking for the null character seems to do the trick.

coyotte508 (commented):
The git hook serverside to check if a file is binary is:

```bash
git cat-file blob ${objectsha} | grep --quiet --binary-files=without-match . && echo Text || echo Binary
```

If it's not possible for cross-platform reasons, the check on the API (when receiving files to upload) is on the first 512 bytes of data with this code: https://github.com/gjtorikian/isBinaryFile/blob/main/src/index.ts#L158-L256

nateraw:
Thanks for the quick fix! This solved my issue here -  #825 . This issue was brought up in our HugGAN event, and its effecting other users, so I'd love to get this merged as soon as we can 😄

julien-c:
This proposed fix makes sense to me!

Yes looking for a null byte character is a common (simple) heuristic for binary files and AFAIK is used in text editors etc.

I agree with @coyotte508's other points

src/huggingface_hub/repository.py
```diff
@@ -981,6 +1002,42 @@ def lfs_enable_largefiles(self):
         except subprocess.CalledProcessError as exc:
             raise EnvironmentError(exc.stderr)
 
+    def auto_track_binary_files(self, pattern: Optional[str] = ".") -> List[str]:
+        """
+        Automatically track binary files with git-lfs.
+
+        Args:
+            pattern (`str`, *optional*, defaults to "."):
+                The pattern with which to track files that are above 10MBs.
```
osanseviero, line 1011:
```suggestion
                The pattern with which to track files that are binary.
```

tests/test_repository.py
```diff
@@ -1113,35 +1129,48 @@ def test_auto_track_large_files(self):
         # This content is 20MB (over 10MB)
         large_file = [100] * int(4e6)
 
+        # This content is binary (contains the null character)
+        binary_file = "\x00\x00\x00\x00"
```
osanseviero, line 1133: I would consider moving this to another test to keep this test focused in large files (or rename this test, although I think different tests would be better)

osanseviero (approved):
Thanks for this PR!! :rocket: :rocket:

src/huggingface_hub/repository.py
```diff
@@ -239,7 +239,7 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
     """
     try:
         with open(filename) as f:
-            content = f.read()
+            content = f.read(512)
```
coyotte508, line 242: 512 bytes is a low number for the heuristic of only checking for the null character (on the API the heuristic is more complex).

If bytes are uniformly distributed, you will have 13% failure (used [this calculator](https://www.omnicalculator.com/statistics/probability)).

Also, with reading 512 bytes, I'm not sure if it can create a `UnicodeDecodeError` in a text file if cutting in the middle of a character at the end? I don't know enough about python to answer that...

Reading the whole file is ok with me, but in the `auto_track_binary_files` function, the `is_binary` check should be done _after_ checking whether the file is already tracked by LFS or in the gitignore.

That way, we'd ensure that at most 10MB are loaded in memory, since otherwise the `auto_track_large_files` would have already marked it LFS.

LysandreJik: Good feedback, thanks @coyotte508. Will adapt.

src/huggingface_hub/repository.py
```diff
@@ -1023,15 +1023,22 @@ def auto_track_binary_files(self, pattern: Optional[str] = ".") -> List[str]:
...
-            is_binary = is_binary_file(path_to_file)
 
-            if (
-                is_binary
-                and not is_tracked_with_lfs(path_to_file)
-                and not is_git_ignored(path_to_file)
-            ):
-                self.lfs_track(filename)
-                files_to_be_tracked_with_lfs.append(filename)
+            if not (is_tracked_with_lfs(path_to_file) or is_git_ignored(path_to_file)):
```
coyotte508, line 1027:
```suggestion
            if not (is_tracked_with_lfs(path_to_file) and not is_git_ignored(path_to_file)):
```

coyotte508 (approved):
Ok for me =)

src/huggingface_hub/repository.py
coyotte508: yes good future proofing :)

src/huggingface_hub/repository.py
LysandreJik: Here the initial `not` is outside of the statement so it encompasses both; I'm happy to follow your proposal but the parenthesis around the two statements should be removed. Would you find that clearer?

(it's on me, I updated that in the latest commit so that it fits on the same line, but it seems to hinder understanding)

coyotte508: oh my bad - I read wrongly

src/huggingface_hub/repository.py
```diff
@@ -226,6 +226,27 @@ def is_git_ignored(filename: Union[str, Path]) -> bool:
     return is_ignored
 
 
+def is_binary_file(filename: Union[str, Path]) -> bool:
```
adrinjalali, line 229: Other than reading the first few bytes, I think this answer gives a more accurate way of checking if the file is binary or not: https://stackoverflow.com/a/7392391/2536294

src/huggingface_hub/repository.py
```diff
@@ -981,6 +1002,49 @@ def lfs_enable_largefiles(self):
         except subprocess.CalledProcessError as exc:
             raise EnvironmentError(exc.stderr)
 
+    def auto_track_binary_files(self, pattern: Optional[str] = ".") -> List[str]:
```
adrinjalali, line 1005: this probably needs to be a private method

```suggestion
    def _auto_track_binary_files(self, pattern: Optional[str] = ".") -> List[str]:
```

tests/test_repository.py
```diff
@@ -1128,6 +1128,30 @@ def test_auto_track_large_files(self):
...
+    def test_auto_track_binary_files(self):
+        repo = Repository(WORKING_REPO_DIR)
+
+        # This content is non-binary
+        non_binary_file = [100] * int(1e6)
+
+        # This content is binary (contains the null character)
+        binary_file = "\x00\x00\x00\x00"
+
+        with open(f"{WORKING_REPO_DIR}/non_binary_file.txt", "w+") as f:
```
adrinjalali, line 1140: we shouldn't write on the working dir, we should work with python's temp files and folders.

tests/test_repository.py
```diff
@@ -1175,6 +1200,54 @@ def test_auto_track_large_files_ignored_with_gitignore(self):
...
         )
 
+    def test_auto_track_binary_files_ignored_with_gitignore(self):
+        repo = Repository(WORKING_REPO_DIR)
+
+        # This content is binary (contains the null character)
+        binary_file = "\x00\x00\x00\x00"
+
+        # Test nested gitignores
+        os.makedirs(f"{WORKING_REPO_DIR}/directory")
```
adrinjalali, line 1210: same here

adrinjalali (commented):
Thanks for the PR @LysandreJik

src/huggingface_hub/repository.py
LysandreJik: This should actually be a public method, same as `auto_track_large_files`. I think it's viable for a user to do `repo.auto_track_large_files` and `auto_track_binary_files` independently of `git_add`!

tests/test_repository.py
LysandreJik: The `WORKING_REPO_DIR` is a folder that is created on the start of the test, and which is reset at every beginning of every test, see here: https://github.com/huggingface/huggingface_hub/blob/6d27a15cb3b8f9d6b6b87075701192569055136f/tests/test_repository.py#L76-L91

This isn't the working directory in which the test is run, but a folder nested in the `fixtures` folder:

https://github.com/huggingface/huggingface_hub/blob/6d27a15cb3b8f9d6b6b87075701192569055136f/tests/test_repository.py#L50-L52

tests/test_repository.py
LysandreJik: Same answer as above :)

src/huggingface_hub/repository.py
LysandreJik: I implemented what you propose in 9fe6bb6, let me know if that's what you had in mind. Cc @coyotte508

tests/test_repository.py
osanseviero: Good point! Let's keep as is then :smile:

src/huggingface_hub/repository.py
```diff
@@ -226,6 +226,31 @@ def is_git_ignored(filename: Union[str, Path]) -> bool:
...
+    Args:
+        filename (`str` or `Path`):
+            The filename to check.
+
+    Returns:
+        `bool`: `True` if the file passed is a binary file, `False` otherwise.
+    """
+    try:
+        with open(filename, "rb") as f:
+            content = f.read()
```
adrinjalali, line 242:
```suggestion
            content = f.read(1024)  # or 512 if we want to be consistent with the backend
```

adrinjalali (approved):
Otherwise LGTM.

I'll open a separate issue for the `fixtures` folder issue.

src/huggingface_hub/repository.py
LysandreJik: This was discussed above here: https://github.com/huggingface/huggingface_hub/pull/828#discussion_r848732391

Do you disagree with the conclusion?

LysandreJik: The solution currently implemented loads a maximum of 10MB in memory when calling `git_add`: it  tracks [large files](https://github.com/huggingface/huggingface_hub/blob/9fe6bb6f303d0b642bfed753866bf5aab1378376/src/huggingface_hub/repository.py#L1167) before tracking [binary files](https://github.com/huggingface/huggingface_hub/blob/9fe6bb6f303d0b642bfed753866bf5aab1378376/src/huggingface_hub/repository.py#L1170).

When tracking binary files, it looks at files [which are not yet tracked with lfs](https://github.com/huggingface/huggingface_hub/blob/9fe6bb6f303d0b642bfed753866bf5aab1378376/src/huggingface_hub/repository.py#L1031), eliminating large files.

Instead of the 1MB limit that you propose here, we could instead put a max of 10MB [here](https://github.com/huggingface/huggingface_hub/blob/9fe6bb6f303d0b642bfed753866bf5aab1378376/src/huggingface_hub/repository.py#L242), which will only be triggered when `auto_track_binary_files` is called independently of `git_add` (which is a possibility!).

adrinjalali: yes, but you also want to have this method public, which means people can call it before having tracked large files.

I'm happy with your suggestion of 10MB.

LysandreJik: Sounds good, thanks for your review. Addressed in [cbfdce5](https://github.com/huggingface/huggingface_hub/pull/828/commits/cbfdce5d3c5af3cfefc493e2111c63e382cb2f3d)

adrinjalali (approved):

PR #833, opened 2022-04-11
Remove docs, tasks and inference API from huggingface_hub

osanseviero:
This PR removes content under `docs`, `tasks`, and `api-inference-community` since there is no active development in this repo anymore for that. The PR does not remove `js` since there are a couple of PRs open for it still.


julien-c (approved):
Yeah! this is going to be way cleaner!!

osanseviero:
cc @NimaBoscarino

README.md
```diff
@@ -1,43 +1,18 @@
...
-The `huggingface_hub` repository is currently being split in three repositories as mentioned in the following
-[issue](https://github.com/huggingface/huggingface_hub/issues/744).
+## Welcome to the huggingface_hub library
 
-As such, the `js`, `docs`, `tasks`, and `api-inference-community` folders are currently frozen and we instead ask
-you to redirect your issues/PRs to the following repositories:
 
-- for `js`, `docs` and `tasks`: [huggingface/hub-docs](https://github.com/huggingface/hub-docs)
-- for `api-inference-community`: [huggingface/api-inference-community](https://github.com/huggingface/api-inference-community)
+The `huggingface_hub` is a client library to interact with the Hugging Face Hub. The Hugging Face Hub is a platform with over 30K models, 3K datasets, and 2K demos in which people can easily collaborate in their ML workflows. The Hub works as a central place where anyone can share, explore, discover, and experiment with open-source Machine Learning.
```
LysandreJik, line 7: If we want to nitpick :smile: 
```suggestion
The `huggingface_hub` is a client library to interact with the Hugging Face Hub. The Hugging Face Hub is a platform with over 35K models, 4K datasets, and 2K demos in which people can easily collaborate in their ML workflows. The Hub works as a central place where anyone can share, explore, discover, and experiment with open-source Machine Learning.
```

LysandreJik (approved):
Impressive diff! Thanks for taking care of that, @osanseviero.

The edit of the contributing guide & readme look good to me. Will be super clean like that!

(out of scope) I wonder if the readme shouldn't be beefed up a bit now to provide code samples, examples etc. I think the README is a big source of traffic, so making it as welcoming/easy to get started as possible would be nice.

osanseviero:
> (out of scope) I wonder if the readme shouldn't be beefed up a bit now to provide code samples, examples etc. I think the README is a big source of traffic, so making it as welcoming/easy to get started as possible would be nice.

Yes, totally. As mentioned above, "Updates the README. It has lots of improvement headroom, but the proposal is better than what is over there right now I think", so for sure we can significantly improve things, but I would like to merge this for now since this PR is already better than what exists now. WDYT? (alternatively I could also move content from the README under huggingface_hub directory, WDYT?)

LysandreJik:
Fine for me to merge like this yes, and let's revisit the README in another PR. Having a separate README in the `src/huggingface_hub` directory indeed doesn't make a lot of sense anymore.

PR #838, opened 2022-04-14
Add metadata fields to model-index for reproducibility

lewtun:
This PR adds new fields to the `model-index` object of model cards to enable fine-grained reproducibility / comparison with evaluation results. This will also allow us to provide fine-grained filters on leaderboards so that people can distinguish e.g. evaluations on the validation vs test sets, different configs etc.

Note that I've added a `config` field as my understanding is that people currently use the `name` field as a free text entry that is displayed on the frontend (see screenshot)

Internal Slack thread: https://huggingface.slack.com/archives/C035S5G2J3D/p1649837714080049

cc @lvwerra @TristanThrush @sashavor 

Screen Shot 2022-04-14 at 17 52 14


modelcard.md
```diff
@@ -22,9 +22,12 @@ model-index:
...
     dataset:
-      type: {dataset_type}  # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
-      name: {dataset_name}  # Required. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}         # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}         # Required. Example: Common Voice zh-CN
+      config: {dataset_config}     # Optional. Example: zh-CN
+      split: {dataset_split}       # Optional. Example: test
+      revision: {dataset_revision} # Optional. Example: main
+      args: {arg_0}                # Optional. Example: zh-CN
```
lewtun, line 30: I see there is some redundancy between the `config` and `args` fields. 

As discussed offline with @lhoestq it makes sense to treat the core parameters of `load_dataset()` at the same level as `type` and bundle all dataset-specific args into `args`

osanseviero:
Would it be possible to share an example with all the fields? It's a bit unclear to me how `args` and `config` would look like. even after going through the thread.

sashavor:
Can we use some kind of hashes or IDs? seems like the name itself may get too long..

lhoestq:
> Would it be possible to share an example with all the fields? It's a bit unclear to me how args and config would look like. even after going through the thread.

Here is an example with config

```yaml
    dataset:
      type: common_voice
      name: Common Voice (English subset)
      config: en
      split: test
      revision: 5503434ddd753f426f4b38109466949a1217c2bb
```

and here an example with args

```yaml
    dataset:
      type: wikipedia
      name: Wikipedia (English dump from March 2022)
      revision: 1b33219ef036fff346b7fe5dc002256b2d6bf60f
      args:
        language: en
        date: 20220301
```

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
 model-index:
 - name: {model_id}
   results:
-  - task: 
-      type: {task_type}  # Required. Example: automatic-speech-recognition
-      name: {task_name}  # Optional. Example: Speech Recognition
+  - task:
+      type: {task_type}             # Required. Example: automatic-speech-recognition
+      name: {task_name}             # Optional. Example: Speech Recognition
+      verified: {evaluation_type}   # Optional. Indicates if evaluation self-reported or generated by Hugging Face. Example: false
```
lewtun, line 24: I've added this new field to distinguish evaluation results that are generated by users vs those that will come from our evaluation service. Internal Slack thread: https://huggingface.slack.com/archives/C035S5G2J3D/p1650448897906539?thread_ts=1650438388.214969&cid=C035S5G2J3D

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
-      type: {dataset_type}  # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
-      name: {dataset_name}  # Required. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}          # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}          # Required. Example: Common Voice (French)
+      config: {dataset_config}      # Optional. Example: fr
+      split: {dataset_split}        # Optional. Example: test
+      revision: {dataset_revision}  # Optional. Example: 5503434ddd753f426f4b38109466949a1217c2bb
+      args: 
+        {arg_0: value_0}            # Optional. Example for wikipedia: language: en
```
lewtun, line 31: Not sure if I'm using the brackets {} correctly here, but the aim is to indicate that `args` is a dict of key-value pairs. Following @lhoestq's example, I've used `wikipedia` here

julien-c (commented):
this is not breaking our PWC integration, right? (haven't checked in detail but you should if possible)

lewtun:
> this is not breaking our PWC integration, right? (haven't checked in detail but you should if possible)

Here's a model card which contains these new fields and the [PwC link](https://paperswithcode.com/sota/text-classification-on-clinc-oos) still works: https://huggingface.co/lewtun/MiniLMv2-L12-H384-distilled-finetuned-clinc/blob/main/README.md#L8

However, I discovered that the `name` field has to equal the dataset ID on PwC - changing this to a "pretty name" like "CLINC150" broke the integration. 

My hunch is that `name` has to be the same as the `paperswithcode_id` that we store on _dataset cards_, but this seems to be in conflict with its use on the frontend as a means to display a "pretty" dataset name. 

I'm happy to update the description of the `name` field, but could use some advice on how we should indicate its meaning / usage to users.

Screen Shot 2022-04-25 at 16 14 10

julien-c:
i think the best would be to check with Rob and other PWC team members in `#hf-and-paperwithcode`

anton-l (commented):
Accidentally stumbled upon this PR, this is very timely, thanks @lewtun! 
Because of metadata inconsistencies and the users' confusion I've started working on a metrics editor for speech models (https://huggingface.co/spaces/anton-l/speech-bench-metrics-editor) so these updates should make the process much smoother :partying_face:

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
+      type: {task_type}             # Required. Example: automatic-speech-recognition
+      name: {task_name}             # Optional. Example: Speech Recognition
     dataset:
-      type: {dataset_type}  # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
-      name: {dataset_name}  # Required. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}          # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}          # Required. Example: Common Voice (French)
+      config: {dataset_config}      # Optional. Example: fr
+      split: {dataset_split}        # Optional. Example: test
```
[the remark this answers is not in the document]
lewtun: Interesting idea :) 

Are you aware of any examples where one trains a model on a dataset config, but evaluates on a different one? For the splits, I guess the simplest change would be to have something like:

```
  splits:
    train_split: train
    eval_split: test
```

This would then align with the `train-eval-index` metadata that we're planning to add to dataset cards in https://github.com/huggingface/autonlp-backend/issues/414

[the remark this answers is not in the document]
anton-l: One example that comes to mind is LibriSpeech that has a `clean` config for training/finetuning, and `clean`+`other` configs (with `clean-dev`/`clean-test`/`other-dev`/`other-test` splits) for evaluation, but it's particularly messy in itself IMO.
So the addition of training splits should be enough I think!

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
     metrics:
-      - type: {metric_type}    # Required. Example: wer
-        value: {metric_value}  # Required. Example: 20.90
-        name: {metric_name}    # Optional. Example: Test WER
-        args: {arg_0}          # Optional. Example for BLEU: max_order
+      - type: {metric_type}         # Required. Example: wer
+        value: {metric_value}       # Required. Example: 20.90
+        name: {metric_name}         # Optional. Example: Test WER
+        args: {arg_0}               # Optional. Example for BLEU: max_order
+        verified: {evaluation_type} # Optional. Indicates if evaluation self-reported or generated by Hugging Face. Example: false
```
julien-c, line 38:
```suggestion
        verified: true     # Optional. If true, indicates that evaluation was generated by Hugging Face (vs. self-reported).
```

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
+      revision: {dataset_revision}  # Optional. Example: 5503434ddd753f426f4b38109466949a1217c2bb
+      args: 
+        {arg_0: value_0}            # Optional. Example for wikipedia: language: en
+        {arg_1: value_1}            # Optional. Example for wikipedia: date: 20220301
     metrics:
-      - type: {metric_type}    # Required. Example: wer
-        value: {metric_value}  # Required. Example: 20.90
-        name: {metric_name}    # Optional. Example: Test WER
-        args: {arg_0}          # Optional. Example for BLEU: max_order
+      - type: {metric_type}         # Required. Example: wer
```
julien-c, line 34:
```suggestion
      - type: {metric_type}         # Required. Example: wer. Use metric id from https://hf.co/metrics
```

modelcard.md
julien-c, line 31:
```suggestion
        {arg_0}: {value_0}            # Optional. Example for wikipedia: language: en
```

modelcard.md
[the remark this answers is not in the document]
julien-c: `model-index` is for eval results, are we sure we want to also add training info in there? might be a bit confusing

The alternative would be to come up with a different training-specific ontology (even though they might look very similar)

julien-c (commented):
left a few suggestions/questions but looks good to me

Maybe now's a good time to move that PR over to `hub-docs` (where that file now lives)?

modelcard.md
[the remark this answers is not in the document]
lewtun: > `model-index` is for eval results, are we sure we want to also add training info in there? might be a bit confusing

I agree, and think it makes most sense to just report the split that the model was evaluated on.

julien-c:
Actually @osanseviero @LysandreJik i have a doubt: where should this file live? here or `hub-docs`? (it's in both repos currently)

osanseviero:
I think this should be part of the Hugging Face Hub documentation as it's part of good practices of model repos product, not this library. (so `hub-docs`)

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
-      name: {dataset_name}  # Required. Example: Common Voice zh-CN
-      args: {arg_0}         # Optional. Example: zh-CN
+      type: {dataset_type}          # Required. Example: common_voice. Use dataset id from https://hf.co/datasets
+      name: {dataset_name}          # Required. Example: Common Voice (French)
+      config: {dataset_config}      # Optional. Example: fr
+      split: {dataset_split}        # Optional. Example: test
+      revision: {dataset_revision}  # Optional. Example: 5503434ddd753f426f4b38109466949a1217c2bb
+      args: 
+        {arg_0}: {value_0}            # Optional. Additional arguments to `load_dataset()`. Example for wikipedia: language: en
+        {arg_1: value_1}            # Optional. Example for wikipedia: date: 20220301
```
julien-c, line 32:
```suggestion
        {arg_0}: {value_0}          # Optional. Additional arguments to `load_dataset()`. Example for wikipedia: language: en
        {arg_1}: {value_1}            # Optional. Example for wikipedia: date: 20220301
```

modelcard.md
```diff
@@ -18,20 +18,26 @@ metrics:
...
     metrics:
-      - type: {metric_type}    # Required. Example: wer
-        value: {metric_value}  # Required. Example: 20.90
-        name: {metric_name}    # Optional. Example: Test WER
-        args: {arg_0}          # Optional. Example for BLEU: max_order
+      - type: {metric_type}         # Required. Example: wer. Use metric id from https://hf.co/metrics
+        value: {metric_value}       # Required. Example: 20.90
+        name: {metric_name}         # Optional. Example: Test WER
+        args: {arg_0}               # Optional. Example for BLEU: max_order
+        verified: true     # Optional. If true, indicates that evaluation was generated by Hugging Face (vs. self-reported).
```
julien-c, line 38:
```suggestion
        args: {arg_0}               # Optional. Example for BLEU: max_order
        verified: true                # Optional. If true, indicates that evaluation was generated by Hugging Face (vs. self-reported).
```

lewtun:
Closing this in favour of https://github.com/huggingface/hub-docs/pull/136

PR #844, opened 2022-04-19
Add update metadata

### LINKED ISSUE #835: Expose a function to update the meta data in the readme
As discussed in https://github.com/huggingface/evaluate/issues/6 we would like a function to update the meta data in the readme of a repo:

```python
huggingface_hub.metadata_update(
  repo_id="lvwerra/my_model",
  update={"metrics": new_value}
)
```

This would enable to programmatically update meta data such as metrics from the `evaluate`, `datasets`, or `transformers` libraries.

There are two use-cases:
1. The key does not exist yet so we add an entry.
2. The key does already exist but needs to be updated - in that case we might want a `overwrite=True` option as safety. 

What do you think @julien-c, @osanseviero? I can have a look at this if you want.

lvwerra:
This PR adds a `metadata_update` function that allows the user to update the metadata in a repository on the hub. The function accepts a dict with metadata (following the same pattern as the YAML in the README) and behaves as follows for all top level fields except `model-index`:

- if field in existing README does not exist it is added.
- if it exists an error is thrown except if `overwrite=True` is passed as a safety guard.

For `model-index` the behaviour is more nuanced:

- if an entry with the same `task` and `dataset` exist, then
    - if the same metric type/name does not exist the metric is appended to the list
    - if the same metric type/name exists the value is overwritten (given `overwrite=True`)
- if an entry with the same `task` and `dataset` does not exist, the result is appended to the results

For reference this is an example of a model's metadata structure as a dictionary:
```Python
{'datasets': ['lvwerra/codeparrot-clean-train'],
 'language': 'code',
 'model-index': [{'name': 'codeparrot',
                  'results': [{'dataset': {'name': 'HumanEval',
                                           'type': 'openai_humaneval'},
                               'metrics': [{'name': 'pass@1',
                                            'type': 'code_eval',
                                            'value': 3.99},
                                           {'name': 'pass@10',
                                            'type': 'code_eval',
                                            'value': 8.69},
                                           {'name': 'pass@100',
                                            'type': 'code_eval',
                                            'value': 17.88}],
                               'task': {'name': 'Code Generation',
                                        'type': 'code-generation'}}]}],
 'tags': ['code', 'gpt2', 'generation'],
 'widget': [{'example_title': 'Transformers',
             'text': 'from transformer import'},
            {'example_title': 'Hello World!',
             'text': 'def print_hello_world():\n\t'},
            {'example_title': 'File size',
             'text': 'def get_file_size(filepath):'},
            {'example_title': 'Numpy', 'text': 'import numpy as'}]}
```

One minor issue I found is that I need to use `force_download=True` for the tests to pass as otherwise the `hf_hub_download` uses the cached but outdated version of the README even if the README has been updated on the remote. cc @LysandreJik 
https://github.com/huggingface/huggingface_hub/blob/a191950c1b21c0f709cfe012e283d5371c4ce9eb/src/huggingface_hub/repocard.py#L137
This feature will be used for https://github.com/huggingface/evaluate/issues/6 and closes #835.

### DIFF
--- a/src/huggingface_hub/repocard.py
+++ b/src/huggingface_hub/repocard.py
@@ -110,6 +110,25 @@ def metadata_update(
     overwrite: bool = False,
     token: str = None,
 ) -> None:
+    """
+    Updates the metadata in the README.md of a repository on the Hugging Face Hub.
+
+    Args:
+        repo_id (`str`):
+            The name of the repository.
+        metadata (`dict`):
+            A dictionary containing the metadata to be updated.
+        repo_type (`str`, *optional*):
+            Set to `"dataset"` or `"space"` if updating to a dataset or space,
+            `None` or `"model"` if updating to a model. Default is `None`.
+        overwrite (`bool`, *optional*):
+            If set to `True` an existing field can be overwritten, otherwise
+            attempting to overwrite an existing field will cause an error.
+            Default is `False`.
+        token (`str`, *optional*):
+            The Hugging Face authentication token
+    """
+
     filepath = hf_hub_download(
         repo_id,
         filename="README.md",
--- a/src/huggingface_hub/repocard.py
+++ b/src/huggingface_hub/repocard.py
@@ -5,6 +5,8 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 import yaml
+from huggingface_hub.file_download import hf_hub_download
+from huggingface_hub.hf_api import HfApi
 from huggingface_hub.repocard_types import (
     ModelIndex,
     SingleMetric,
@@ -99,3 +101,98 @@ def metadata_eval_result(
         model_index, dict_factory=lambda x: {k: v for (k, v) in x if v is not None}
     )
     return {"model-index": [data]}
+
+
+def metadata_update(
+    repo_id: str,
+    metadata: Dict,
+    repo_type: str = None,
+    overwrite: bool = False,
+    token: str = None,
+) -> None:
+    filepath = hf_hub_download(
+        repo_id,
+        filename="README.md",
+        repo_type=repo_type,
+        use_auth_token=token,
+        force_download=True,
+    )
+    existing_metadata = metadata_load(filepath)
+
+    for key in metadata:
+        # update all fields except model index
+        if not key == "model-index":
+            if key in existing_metadata and not overwrite:
+                if existing_metadata[key] != metadata[key]:
+                    raise ValueError(
+                        f"""You passed a new value for the existing meta data field '{key}'. Set `overwrite=True` to overwrite existing metadata."""
+                    )
+            else:
+                existing_metadata[key] = metadata[key]
+        # update model index containing the evaluation results
+        else:
+            if "model-index" not in existing_metadata:
+                existing_metadata["model-index"] = metadata["model-index"]
+            else:
+                existing_metadata["model-index"][0][
+                    "results"
+                ] = _update_metadata_model_index(
+                    existing_metadata["model-index"][0]["results"],
+                    metadata["model-index"][0]["results"],
+                    overwrite=overwrite,
+                )
+
+    # save and push to hub
+    metadata_save(filepath, existing_metadata)
+
+    HfApi().upload_file(
+        path_or_fileobj=filepath,
+        path_in_repo="README.md",
+        repo_id=repo_id,
+        repo_type=repo_type,
+        identical_ok=True,
+        token=token,
+    )
+
+
+def _update_metadata_model_index(existing_results, new_results, overwrite=False):
+    for new_result in new_results:
+        result_found = False
+        for existing_result_index, existing_result in enumerate(existing_results):
+            if (
+                new_result["dataset"] == existing_result["dataset"]
+                and new_result["task"] == existing_result["task"]
+            ):
+                result_found = True
+                for new_metric in new_result["metrics"]:
+                    metric_exists = False
+                    for existing_metric_index, existing_metric in enumerate(
+                        existing_result["metrics"]
+                    ):
+                        if (
+                            new_metric["name"] == existing_metric["name"]
+                            and new_metric["type"] == existing_metric["type"]
+                        ):
+                            if overwrite:
+                                existing_results[existing_result_index]["metrics"][
+                                    existing_metric_index
+                                ]["value"] = new_metric["value"]
+                            else:
+                                # if metric exists and value is not the same throw an error without overwrite flag
+                                if (
+                                    existing_results[existing_result_index]["metrics"][
+                                        existing_metric_index
+                                    ]["value"]
+                                    != new_metric["value"]
+                                ):
+                                    raise ValueError(
+                                        f"""You passed a new value for the existing metric '{new_metric["name"]}'. Set `overwrite=True` to overwrite existing metrics."""
+                                    )
+                            metric_exists = True
+                    if not metric_exists:
+                        existing_results[existing_result_index]["metrics"].append(
+                            new_metric
+                        )
+        if not result_found:
+            existing_results.append(new_result)
+    return existing_results
--- a/tests/test_repocard.py
+++ b/tests/test_repocard.py
@@ -13,14 +13,26 @@
 # limitations under the License.
 import os
 import shutil
+import time
 import unittest
 from pathlib import Path
 
+import pytest
+
+import yaml
 from huggingface_hub.constants import REPOCARD_NAME
-from huggingface_hub.repocard import metadata_eval_result, metadata_load, metadata_save
+from huggingface_hub.hf_api import HfApi
+from huggingface_hub.repocard import (
+    metadata_eval_result,
+    metadata_load,
+    metadata_save,
+    metadata_update,
+)
+from huggingface_hub.repository import Repository
 from huggingface_hub.utils import logging
 
-from .testing_utils import set_write_permission_and_retry
+from .testing_constants import ENDPOINT_STAGING, TOKEN, USER
+from .testing_utils import retry_endpoint, set_write_permission_and_retry
 
 
 ROUND_TRIP_MODELCARD_CASE = """
@@ -97,6 +109,8 @@ REPOCARD_DIR = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "fixtures/repocard"
 )
 
+REPO_NAME = "dummy-hf-hub-{}".format(int(time.time() * 10e3))
+
 
 class RepocardTest(unittest.TestCase):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


julien-c:
the behavior sounds reasonable to me, but isn't it weird in terms of API design that there's a special case for `model-index`? (maybe not)

lvwerra:
The main reason to do it different for `model-index` is usability: if we take the same approach as for the other fields one would need to always pass all the existing results/metrics too, even if for just updating/adding a single entry.

Alternatively, we could outsource that logic to a helper function that grabs the existing `model-index` and updates it with the new results and then one can pass the complete (existing+new) `model-index` to `metadata_update`.

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +101,117 @@ def metadata_eval_result(
...
+        path_or_fileobj=filepath,
+        path_in_repo="README.md",
+        repo_id=repo_id,
+        repo_type=repo_type,
+        identical_ok=True,
+        token=token,
+    )
+
+
+def _update_metadata_model_index(existing_results, new_results, overwrite=False):
```
adrinjalali, line 177: to better review this function, and to be able to maintain it nicely in the future, we need:
- add documentation on what the `results` look like, and how they're updated
- probably move the third and fourth `for` loops into their own function

as a reviewer, it would be nice if I could see an example covering all the control branches in this function.

adrinjalali (commented):
Thanks for the PR @lvwerra .

In the future, please send PRs from your own branch. We're trying to cleanup branches on the main repo. The CI works here on PRs from fork repositories.

src/huggingface_hub/repocard.py
lvwerra: That's a good idea, I'll refactor the function. I'll also make an example for each case.

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,156 @@ def metadata_eval_result(
...
+                result_found = True
+                existing_results[existing_result_index][
+                    "metrics"
+                ] = _update_metadata_results_metric(
+                    new_result["metrics"],
+                    existing_result["metrics"],
+                    overwrite=overwrite,
+                )
+        if not result_found:
+            existing_results.append(new_result)
```
adrinjalali, line 214: Since dictionaries are mutable, you could also write this as:

```py
        try:
            existing_result = next(
                x
                for x in existing_results
                if all(x[feat] == new_result[feat] for feat in UNIQUE_RESULT_FEATURES)
            )
            existing_result["metrics"] = _update_metadata_results_metric(
                new_result["metrics"],
                existing_result["metrics"],
                overwrite=overwrite,
            )
        except StopIteration:
            existing_results.append(new_result)
```

which should be faster since it avoids one slow for loop.

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,156 @@ def metadata_eval_result(
...
+    return existing_results
+
+
+def _update_metadata_results_metric(new_metrics, existing_metrics, overwrite=False):
+    """
+    Updates the metrics list of a result in the metadata. If metrics with same unique
+    features exist they values are updated, else a new metric is appended. Updating
```
adrinjalali, line 221:
```suggestion
    features exist their values are updated, else a new metric is appended. Updating
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,156 @@ def metadata_eval_result(
...
+                    if (
+                        existing_metrics[existing_metric_index]["value"]
+                        != new_metric["value"]
+                    ):
+                        raise ValueError(
+                            f"""You passed a new value for the existing metric '{new_metric["name"]}'. Set `overwrite=True` to overwrite existing metrics."""
+                        )
+                metric_exists = True
+        if not metric_exists:
+            existing_metrics.append(new_metric)
```
adrinjalali, line 258: Could do the same here:

```py
    for new_metric in new_metrics:
        try:
            existing_metric = next(
                x
                for x in existing_metrics
                if all(x[feat] == new_metric[feat] for feat in UNIQUE_METRIC_FEATURES)
            )
            if overwrite:
                existing_metric["value"] = new_metric["value"]
            else:
                # if metric exists and value is not the same throw an error
                # without overwrite flag
                if existing_metric["value"] != new_metric["value"]:
                    existing_str = ",".join(
                        new_metric[feat] for feat in UNIQUE_METRIC_FEATURES
                    )
                    raise ValueError(
                        "You passed a new value for the existing metric"
                        f" '{existing_str}'. Set `overwrite=True` to overwrite existing"
                        " metrics."
                    )
        except StopIteration:
            existing_metrics.append(new_metric)
```

tests/test_repocard.py
```diff
@@ -97,6 +109,8 @@
     os.path.dirname(os.path.abspath(__file__)), "fixtures/repocard"
 )
 
+REPO_NAME = "dummy-hf-hub-{}".format(int(time.time() * 10e3))
```
adrinjalali, line 112: Probably better to use uuid here:

```suggestion
REPO_NAME = "dummy-hf-hub-{str(uuid.uuid4())}"
```

tests/test_repocard.py
```diff
@@ -159,3 +173,142 @@ def test_metadata_eval_result(self):
...
+        """
+        Share this valid token in all tests below.
+        """
+        cls._token = TOKEN
+        cls._api.set_access_token(TOKEN)
+
+    @retry_endpoint
+    def setUp(self) -> None:
+        if os.path.exists(REPO_NAME):
+            shutil.rmtree(REPO_NAME, onerror=set_write_permission_and_retry)
```
adrinjalali, line 192: could we clone the repo under a `tempfile.mkdtemp`?

tests/test_repocard.py
```diff
@@ -159,3 +173,142 @@ def test_metadata_eval_result(self):
...
+        self._api.delete_repo(repo_id=f"{REPO_NAME}", token=self._token)  # {USER}/
+        shutil.rmtree(REPO_NAME)
+
+    def test_update_dataset_name(self):
+        new_datasets_data = {"datasets": "['test/test_dataset']"}
+        metadata_update(f"{USER}/{REPO_NAME}", new_datasets_data, token=self._token)
+
+        self.repo.git_pull()
+        updated_metadata = metadata_load(f"{REPO_NAME}/README.md")
+        expected_metadata = self.existing_metadata.copy()
```
adrinjalali, line 220:
```suggestion
        expected_metadata = copy.deepcopy(self.existing_metadata)
```

to copy nested objects in a dict. Otherwise:

```py
>>> a = {1: {2: 3}}
>>> b = a.copy()
>>> b[1][2] = 4
>>> b
{1: {2: 4}}
>>> a
{1: {2: 4}}
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,156 @@ def metadata_eval_result(
...
+                        f"""You passed a new value for the existing meta data field '{key}'. Set `overwrite=True` to overwrite existing metadata."""
+                    )
+            else:
+                existing_metadata[key] = metadata[key]
+        # update model index containing the evaluation results
+        else:
+            if "model-index" not in existing_metadata:
+                existing_metadata["model-index"] = metadata["model-index"]
+            else:
+                # the model-index contains a list of results as used by PwC but only has one element thus we take the first one
```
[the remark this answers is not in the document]
lvwerra: Papers with Code :)

src/huggingface_hub/repocard.py
lvwerra: This as well :)

adrinjalali (commented):
LGTM, could you please merge with the latest `main` and run black on these files again? Otherwise LGTM.

lvwerra:
Done - thanks for reviewing  and the helpful suggestions @adrinjalali!

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+        filename="README.md",
+        repo_type=repo_type,
+        use_auth_token=token,
+        force_download=True,
+    )
+    existing_metadata = metadata_load(filepath)
+
+    for key in metadata:
+        # update all fields except model index
+        if key != "model-index":
```
osanseviero, line 145: Nit: I would have the logic the other way around. 

First `if key == "model-index", then else. 

It makes readability easier

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+
+    # save and push to hub
+    metadata_save(filepath, existing_metadata)
+
+    HfApi().upload_file(
+        path_or_fileobj=filepath,
+        path_in_repo="README.md",
+        repo_id=repo_id,
+        repo_type=repo_type,
+        identical_ok=True,
```
osanseviero, line 175: Do we really want `identical_ok`? This creates empty commits. https://huggingface.co/osanseviero/test_metrics/commit/024a4657d24cfef7ce7992779ab2f0efbece1968

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+    features exist they are updated, else a new result is appended. Updating existing
+    values is only possible if `overwrite=True`.
+
+    Args:
+        new_metrics (`list`):
+            List of new metadata results
+        existing_metrics (`list`):
+            List of existing metadata results
+        overwrite (`bool`, *optional*, defaults to `False`):
+            If set to `True` an existing metric values can be overwritten, otherwise
```
osanseviero, line 192:
```suggestion
            If set to `True`, an existing metric values can be overwritten, otherwise
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+        metadata (`dict`):
+            A dictionary containing the metadata to be updated.
+        repo_type (`str`, *optional*):
+            Set to `"dataset"` or `"space"` if updating to a dataset or space,
+            `None` or `"model"` if updating to a model. Default is `None`.
+        overwrite (`bool`, *optional*, defaults to `False`):
+            If set to `True` an existing field can be overwritten, otherwise
+            attempting to overwrite an existing field will cause an error.
+        token (`str`, *optional*):
+            The Hugging Face authentication token
```
osanseviero, line 131:
```suggestion
            The Hugging Face authentication token.
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+
+def _update_metadata_model_index(existing_results, new_results, overwrite=False):
+    """
+    Updates the model-index fields in the metadata. If results with same unique
+    features exist they are updated, else a new result is appended. Updating existing
+    values is only possible if `overwrite=True`.
+
+    Args:
+        new_metrics (`list`):
+            List of new metadata results
```
osanseviero, line 188:
```suggestion
            List of new metadata results.
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+        existing_metrics (`list`):
+            List of existing metadata results
```
osanseviero, line 190:
```suggestion
            List of existing metadata results.
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+                for x in existing_results
+                if all(x[feat] == new_result[feat] for feat in UNIQUE_RESULT_FEATURES)
+            )
+            existing_result["metrics"] = _update_metadata_results_metric(
+                new_result["metrics"],
+                existing_result["metrics"],
+                overwrite=overwrite,
+            )
+        except StopIteration:
+            existing_results.append(new_result)
```
osanseviero, line 211: Is this really needed? I think try/except to do this is a bit less readable and could be more prone to introduce bugs. Maybe the for loop could be more explicit. I think that was how it was before, I personally find that more readable and easier to maintain.

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+
+def _update_metadata_results_metric(new_metrics, existing_metrics, overwrite=False):
+    """
+    Updates the metrics list of a result in the metadata. If metrics with same unique
+    features exist their values are updated, else a new metric is appended. Updating
+    existing values is only possible if `overwrite=True`.
+
+    Args:
+        new_metrics (`list`):
+            List of new metrics
```
osanseviero, line 223:
```suggestion
            List of new metrics.
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+        existing_metrics (`list`):
+            List of existing metrics
```
osanseviero, line 225:
```suggestion
            List of existing metrics.
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+        overwrite (`bool`, *optional*, defaults to `False`):
+            If set to `True` an existing metric values can be overwritten, otherwise
```
osanseviero, line 227:
```suggestion
            If set to `True`, an existing metric values can be overwritten, otherwise
```

osanseviero (commented):
This looks very neat! I left some minor suggestions and a couple of questions. Thanks for this PR :fire: :fire: 

I wrote a colab while I was exploring this https://colab.research.google.com/drive/1fG8OWYTnI6ucnafYKtrf-HwxtWbVBVQh?usp=sharing

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,153 @@ def metadata_eval_result(
...
+
+    Returns:
+        `list`: List of updated metadata results
+    """
+    for new_result in new_results:
+        try:
+            existing_result = next(
+                x
+                for x in existing_results
+                if all(x[feat] == new_result[feat] for feat in UNIQUE_RESULT_FEATURES)
```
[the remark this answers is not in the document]
lvwerra: Since we pull the metadata from the Hub can't we assume that it is valid? In your experiment pushing invalid metadata was rejected, right? So it should not be there in the first place if it is invalid.

src/huggingface_hub/repocard.py
lvwerra: I have no strong opinion here - happy to revert. @adrinjalali?

src/huggingface_hub/repocard.py
lvwerra: Good point, I'll change that. BTW can't see your repos (private?).

src/huggingface_hub/repocard.py
adrinjalali: The for loop is slower, and try/except clauses are quite pythonic. I do think the current state is quite better than a for loop. One can always add a bit of comment on when the exception is raised if that helps with readability.

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
osanseviero: Hmm @julien-c WDYT? https://huggingface.co/osanseviero/llama-horse-zebra 

The result has task and metrics, but no associated dataset. This is not rejected by the server, and the evaluation results are nicely shown, with the only con that there is no associated dataset, so there's no leaderboard.

IMO this is still valid metadata which is just incomplete. Most spaCy models are like this https://huggingface.co/models?library=spacy

src/huggingface_hub/repocard.py
osanseviero: Now it's public

src/huggingface_hub/repocard.py
lvwerra: I don't think speed is important here as we iterate through a dictionary with a handful of entries while pulling/pushing to the hub in the same function which is probably 1000x slower.

adrinjalali: I do find the for/loop implementation less readable. Removing the outer loop would make it quite complicated and not very readable, that's why I didn't suggest to remove it.

As for consistency, I don't think it's about staying consistent with the rest of the code. Sometimes a for loop makes sense, sometimes it doesn't.  And as a team when we work on a codebase, sometimes we find better ways to do things, and that's fine, and I don't think we should hold back because we have done things in a certain way so far. To me it's okay to look at a codebase and notice old stuff vs new stuff. The way people do things changes and new code can look different than the old code and that's fine to me.

But if both of you think the for loop is more readable than this, then sure, change it. To me it's the other way around.

src/huggingface_hub/repocard.py
lvwerra: done

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
julien-c: yes we try to still display the eval results in this case even though it's not perfect

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
         model_index, dict_factory=lambda x: {k: v for (k, v) in x if v is not None}
     )
     return {"model-index": [data]}
+
+
+def metadata_update(
+    repo_id: str,
+    metadata: Dict,
```
julien-c, line 111:
```suggestion
    repo_id: str,
    *,
    metadata: Dict,
```

kwarg only to prevent API breakage is we need to re-order params later on? WDYT?

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+        overwrite (`bool`, *optional*, defaults to `False`):
+            If set to `True` an existing field can be overwritten, otherwise
+            attempting to overwrite an existing field will cause an error.
+        token (`str`, *optional*):
+            The Hugging Face authentication token.
+    """
+
+    filepath = hf_hub_download(
+        repo_id,
+        filename="README.md",
```
julien-c, line 136:
```suggestion
        filename=REPOCARD_NAME,
```
(nit, feel free to disregard)

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+                    )
+            else:
+                existing_metadata[key] = metadata[key]
+
+    # save and push to hub
+    metadata_save(filepath, existing_metadata)
+
+    HfApi().upload_file(
+        path_or_fileobj=filepath,
+        path_in_repo="README.md",
```
julien-c, line 172:
```suggestion
        path_in_repo=REPOCARD_NAME,
```
(nit, feel free to disregard)

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+                    raise ValueError(
+                        f"""You passed a new value for the existing meta data field '{key}'. Set `overwrite=True` to overwrite existing metadata."""
+                    )
+            else:
+                existing_metadata[key] = metadata[key]
+
+    # save and push to hub
+    metadata_save(filepath, existing_metadata)
+
+    HfApi().upload_file(
```
julien-c, line 170:
```suggestion
    return HfApi().upload_file(
```
return, so this whole function returns the commit URL

julien-c (commented):
left a few small nits, other than that looks good to me!

src/huggingface_hub/repocard.py
LysandreJik: `metadata` would need to have a default value in that case. May I offer the option of having `metadata` be before the `*`?
```suggestion
    repo_id: str,
    metadata: Dict,
    *
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
     return {"model-index": [data]}
+
+
+def metadata_update(
+    repo_id: str,
+    metadata: Dict,
+    repo_type: str = None,
+    overwrite: bool = False,
+    token: str = None,
+) -> None:
```
LysandreJik, line 115: If taking into account the `return` proposed by Julien:
```suggestion
) -> str:
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+            A dictionary containing the metadata to be updated.
+        repo_type (`str`, *optional*):
+            Set to `"dataset"` or `"space"` if updating to a dataset or space,
+            `None` or `"model"` if updating to a model. Default is `None`.
+        overwrite (`bool`, *optional*, defaults to `False`):
+            If set to `True` an existing field can be overwritten, otherwise
+            attempting to overwrite an existing field will cause an error.
+        token (`str`, *optional*):
+            The Hugging Face authentication token.
+    """
```
LysandreJik, line 132:
```suggestion

    Returns:
        `str`: URL of the commit which updated the card metadata.
    """
```

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+
+def metadata_update(
+    repo_id: str,
+    metadata: Dict,
+    repo_type: str = None,
+    overwrite: bool = False,
+    token: str = None,
+) -> None:
+    """
+    Updates the metadata in the README.md of a repository on the Hugging Face Hub.
```
LysandreJik, line 117: Could you provide an example? It would make it simpler for users to understand the format of the metadata.

src/huggingface_hub/repocard.py
```diff
@@ -99,3 +104,162 @@ def metadata_eval_result(
...
+def _update_metadata_model_index(existing_results, new_results, overwrite=False):
+    """
+    Updates the model-index fields in the metadata. If results with same unique
+    features exist they are updated, else a new result is appended. Updating existing
+    values is only possible if `overwrite=True`.
+
+    Args:
+        new_metrics (`list`):
+            List of new metadata results.
+        existing_metrics (`list`):
```
LysandreJik, line 189: Usually we put the type of the items in the lists, such as `List[str]`.

LysandreJik (approved):
Looks good! Only left nits.

src/huggingface_hub/repocard.py
lvwerra: Done

LysandreJik:
Thank you for addressing all comments! Merging.

src/huggingface_hub/repocard.py
lvwerra: I also thought they couldn't but it turns out they can have no default value:
https://stackoverflow.com/a/29037261

julien-c: yes it's just that they're then required (which is fine here IMO)

PR #849, opened 2022-04-21
MNT enable preview on black

adrinjalali:
This PR adds `--preview` to `black` which handles strings much nicer.

cc @LysandreJik @osanseviero


osanseviero:
No strong opinion from my side, but I usually like cross-library consistency if possible, but we already diverged in line length if I recall correctly (we use 88 here (default) vs transformers which uses 119), so really no strong opinion.

PR #852, opened 2022-04-22
add `space_sdk` to `Repository`

lvwerra:
This PR adds the `space_sdk` keyword argument to `Repository` such that the `clone_from` succeeds for spaces repositories. Previously this failed as a `repo_type=="space"` requires a SDK when `create_repo` is called.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -468,6 +468,7 @@ class Repository:
         private: bool = False,
         skip_lfs_files: bool = False,
         client: Optional[HfApi] = None,
+        space_sdk: Optional[str] = None,
     ):
         """
         Instantiate a local clone of a git repo.
@@ -512,6 +513,9 @@ class Repository:
             client (`HfApi`, *optional*):
                 Instance of HfApi to use when calling the HF Hub API. A new
                 instance will be created if this is left to `None`.
+            space_sdk (`str`, *optional*):
+                Choice of SDK to use if repo_type is "space". Can be
+                "streamlit", "gradio", or "static".
         """
 
         os.makedirs(local_dir, exist_ok=True)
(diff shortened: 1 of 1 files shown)


osanseviero:
This fixes #672

julien-c:
FYI with upcoming Spaces templates there won't be a notion of `space_sdk` anymore

Not a huge fan of this PR but i guess if it fixes things in the short term, why not

adrinjalali:
If we're going to have to deprecate `space_sdk`, would it not make sense to have the original issue as "being fixed in the backend" and not fix it here?

lvwerra:
I am fine either way. I wanted to use `Repository` to create the `evaluate` widgets automatically. What's the ETA on the new templates?

julien-c:
> What's the ETA on the new templates?

i would say some time in the coming month or so, WDYT @cbensimon ?

lvwerra:
In the meantime, this also works as a workaround:

```Python
repo_url = f"https://huggingface.co/spaces/{namespace}/{repo_name}"
create_repo(namespace + "/" + repo_name, repo_type="space", space_sdk="gradio")

subprocess.run(
    f"git clone {repo_url}".split(),
    stderr=subprocess.PIPE,
    stdout=subprocess.PIPE,
    check=True,
    encoding="utf-8",
    cwd=output_dir,
    env=os.environ.copy(),
    )

repo = Repository(
    local_dir=output_dir/repo_name, 
    )
```

lvwerra:
Sounds good to me and I can live with the workaround in the meantime. Closing this.

osanseviero:
@julien-c @cbensimon is there an internal issue we could link to #672?

julien-c:
nope no internal issue yet but will ping you when we have one

PR #858, opened 2022-04-29
Add skip_lfs_files to mixins' push_to_hub

nateraw:
I know #847 is going to change a lot with the mixins, but this is a really helpful simple change that we leveraged in the HugGAN sprint that I figured we could add for now. Its nice to be able to skip LFS files when you're using a temp dir as your save dir (which you might be doing in a for loop, such as a training loop).

### DIFF
--- a/src/huggingface_hub/hub_mixin.py
+++ b/src/huggingface_hub/hub_mixin.py
@@ -220,6 +220,7 @@ class ModelHubMixin:
         git_user: Optional[str] = None,
         git_email: Optional[str] = None,
         config: Optional[dict] = None,
+        skip_lfs_files: bool = False,
     ) -> str:
         """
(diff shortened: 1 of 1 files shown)


osanseviero:
Just to see if I understand correctly, this will make the pulling within the `push_to_hub` avoid pulling the large files, but it would still push the lfs files. Is that correct?

osanseviero:
A bit unrelated to this PR, but it would be nice to change

```py
repo.git_add()
repo.git_commit(commit_message)
repo.git_push()
```

to 

```py
repo.push_to_hub()
```

adrinjalali (approved):
LGTM as a temporary change. But we may change it before the next release anyway.

adrinjalali:
@osanseviero I kinda do appreciate the methods being explicit about whether they do git or they work with the REST endpoints.

PR #860, opened 2022-05-03
Enabling Keras download stats

merveenoyan:
This PR adds `library_name` and `library_version` to `snapshot_download` in Keras mixin.

I think existing `from_pretrained_keras` tests should cover this so I didn't write an additional test.

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -122,6 +122,10 @@ def is_tf_available():
     return _tf_available
 
 
+def get_tf_version():
+    return _tf_version
+
+
 def is_fastai_available():
     return _fastai_available
 
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Optional, Union
 import yaml
 from huggingface_hub import ModelHubMixin
 from huggingface_hub.file_download import (
+    get_tf_version,
     is_graphviz_available,
     is_pydot_available,
     is_tf_available,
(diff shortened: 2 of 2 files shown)


merveenoyan:
I'm checking out why `test_filter_models_with_complex_query` and `test_tags` are failing.

adrinjalali:
What is the consequence of this change?

merveenoyan:
(Leaving notes to myself)
I don't think it's related to my PR honestly but:
```
(Pdb) api.get_model_tags()
*** KeyError: 'language'
```
(called from `__init__ `of `ModelSearchArguments()` that is called inside the test)
but I do have `"language"` in the response returned from `get_model_tags()` 🧐

merveenoyan:
@adrinjalali when we add library name and version to snapshot download it sends a request with the info inside the header that gives us statistics about Keras downloads.

adrinjalali:
Ah I see, so it doesn't really change anything on the client side. Got it.

merveenoyan:
There's a problem with dataset tags. I'm fixing it.

merveenoyan:
There's a dataset that is empty after the dataset `batterypapers`, (this is the dict {'id':` 'dataset:', 'label': '', 'type': 'dataset'},`).
It breaks `_unpack_and_assign_dictionary()` which causes a bug in the tests (and initialization of DatasetTags class in general).

adrinjalali:
Should we be ignoring those datasets then? As in, if there's an issue creating the class, ignore the entry.

merveenoyan:
@adrinjalali if @osanseviero approves. Zach said he reached out to owners of dataset to remove them.

osanseviero:
This internal PR from @julien-c will prevent this situation in the future https://github.com/huggingface/moon-landing/pull/2822

merveenoyan:
@osanseviero I re-ran and problems seem to be gone.

osanseviero:
Yes, we chatted with the model authors and fixed in the backend. Thanks!

osanseviero (approved):
Looks good! Thanks :fire:

PR #861, opened 2022-05-05
Keras: Saving history in a JSON file

merveenoyan:
This is a minor change that saves history as a JSON file, if history exists.
We check if history exists or not during parsing it to a markdown in `_parse_model_history()` so I handled it there not to write repetitive checks. It can be further improved inside, just wanted to keep them separately for the sake of readability. (will later refactor the whole mixin and card creation after sprint is done) This might not be the best choice ever so feel free to criticize.
Wrote a test to check if the file is there.

### DIFF
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -44,8 +44,8 @@ def _parse_model_history(model, save_directory):
     if model.history is not None:
         if model.history.history != {}:
             path = os.path.join(save_directory, "history.json")
-            with open(path, "w") as f:
-                json.dump(model.history.history, f)
+            with open(path, "w", encoding="utf-8") as f:
+                json.dump(model.history.history, f, indent=2, sort_keys=True)
             lines = []
             logs = model.history.history
             num_epochs = len(logs["loss"])
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -39,10 +39,13 @@ def _extract_hyperparameters_from_keras(model):
     return hyperparameters
 
 
-def _parse_model_history(model):
+def _parse_model_history(model, save_directory):
     lines = None
     if model.history is not None:
         if model.history.history != {}:
+            path = os.path.join(save_directory, "history.json")
+            with open(path, "w") as f:
+                json.dump(model.history.history, f)
             lines = []
             logs = model.history.history
             num_epochs = len(logs["loss"])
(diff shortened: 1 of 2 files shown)


osanseviero (approved):
The PR Looks fine to me, thanks Merve for this PR!! @nateraw could you take a second look please?

> (will later refactor the whole mixin and card creation after sprint is done)

Note from internal discussions, #847 and other things going on that refactoring the mixin and card creation is overlapping with other efforts (but also totally unrelated to this PR I think :smile: )

merveenoyan:
@osanseviero I find model card code not readable, and I feel like functions can be tweaked for better readability that's what the refactor I'm planning is about. But it's not prioritized anyway 🤗

src/huggingface_hub/keras_mixin.py
```diff
@@ -39,10 +39,13 @@ def _extract_hyperparameters_from_keras(model):
...
 
 
-def _parse_model_history(model):
+def _parse_model_history(model, save_directory):
     lines = None
     if model.history is not None:
         if model.history.history != {}:
+            path = os.path.join(save_directory, "history.json")
+            with open(path, "w") as f:
+                json.dump(model.history.history, f)
```
nateraw, line 48: Small nits here for Hugging Face style points. 🤗

```suggestion
            with open(path, "w", encoding="utf-8") as f:
                json.dump(model.history.history, f, indent=2, sort_keys=True)
```

For reference...

[Here's](https://github.com/huggingface/transformers/blob/31ec2cb2badfbdd4c1ac9c6c9b8a74e974984206/src/transformers/configuration_utils.py#L792) where I got the style from


And [here's](https://github.com/huggingface/transformers/blob/31ec2cb2badfbdd4c1ac9c6c9b8a74e974984206/src/transformers/configuration_utils.py#L805-L806) why I went with specifying the encoding explicitly (I really dunno if this is necessary, but it doesn't hurt I suppose 😅 )

Feel free to ignore if you disagree...I just think its easier to read on the Hub when its formatted all pretty 💄

src/huggingface_hub/keras_mixin.py
```diff
@@ -39,10 +39,13 @@ def _extract_hyperparameters_from_keras(model):
     return hyperparameters
 
 
-def _parse_model_history(model):
+def _parse_model_history(model, save_directory):
```
nateraw, line 42: I'm not in love with the fact this history writing is happening so deep in the logic here. `save_pretrained_keras` -> `_create_model_card` -> `_write_metrics` -> `_parse_model_history` is so deep in the code to just save `model.history.history` as a json file. 

I don't want to block on this, but just wanted to point that out so we remember to put this in a nicer spot in future refactors.

nateraw (approved):
LGTM 🤗. In future PRs, let's try to simplify here, as this keras mixin file is getting a little hard to read haha. 

Thanks for this one @merveenoyan! ❤️

src/huggingface_hub/keras_mixin.py
julien-c: Totally agree, git diffs will also be better if JSON is pretty formatted. Good catch @nateraw

osanseviero (approved):
Thanks for the PR! :rocket:

Issue #497, opened 2021-11-23
Add ability to upload folder with upload_file


nateraw:
**Is your feature request related to a problem? Please describe.**
In the spirit of moving away from forcing users to deal with git, it would be nice to allow folks to upload entire folders using `upload_file`, instead of having them glob/walk through all the files and call `upload_file` each time.

**Describe the solution you'd like**

```python
from huggingface_hub import upload_file

# Uploads contents of `images/` to `nateraw/some-repo`
upload_file('nateraw/some-repo', './images/')
```
**Describe alternatives you've considered**

Bonus points if we have "compression" options as well (zip, tar, gz, etc). If we do that though, we should be mindful of potentially including decompression options in the proposed `download_file` function #460 

**Additional context**
Wanted to upload a few small folders and found myself having to write more code than I wanted to in order to loop over all my local files and call `upload_file`.

osanseviero:
IMO instead of supporting uploading folders with `upload_file`, I would rather have a separate method `upload_folder` or `upload_files` which allows receiving a directory (or file pattern even!) to upload multiple files.

WDYT @LysandreJik?

LysandreJik:
Why not, if it helps! I've tried to parallelize uploads in the past but it has always resulted in 500 errors from the backend.

Feel free to open a PR @osanseviero!

osanseviero:
Yes, the backend fails as discussed in #707. A v0 could push sequentially, which would be a bit of a mess and slow, but still solve the problem. The discussion in https://github.com/huggingface/moon-landing/issues/2260 might unblock this use case though.

Pierrci:
@osanseviero yes, I think it's probably simpler to wait for the revamp of the file upload system

nateraw:
Closed with #888

PR #864, opened 2022-05-06
Add basic `upload_folder` feature

nateraw:
As discussed in #497, here's a quick draft of a basic `upload_folder` that will sequentially upload files within a folder. It's not ideal by any means, but it could get the job done for many simple cases.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -17,9 +17,9 @@ import re
 import subprocess
 import sys
 import warnings
-from pathlib import Path
 from io import BufferedIOBase, RawIOBase
 from os.path import expanduser
+from pathlib import Path
 from typing import IO, Dict, Iterable, List, Optional, Tuple, Union
 
 import requests
@@ -1880,7 +1880,7 @@ class HfApi:
         repo_type: Optional[str] = None,
         revision: Optional[str] = None,
         identical_ok: bool = True,
-        glob_pattern: Optional[str] = '**/*',
+        glob_pattern: Optional[str] = "**/*",
     ):
         path = Path(path)
 
@@ -1894,7 +1894,7 @@ class HfApi:
                 continue
 
             filepath_in_repo = Path(path_in_repo) / file.relative_to(path)
-        
+
             upload_file(
                 path_or_fileobj=str(file),
                 path_in_repo=filepath_in_repo.as_posix(),
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -17,6 +17,7 @@ import re
 import subprocess
 import sys
 import warnings
+from pathlib import Path
 from io import BufferedIOBase, RawIOBase
 from os.path import expanduser
 from typing import IO, Dict, Iterable, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


osanseviero:
Sorry, I just saw this is a draft PR. Let us know when we can review :hugs:

julien-c:
i think the ability server-side to upload multiple files into one commit via the API is sufficiently near that maybe it doesn't make a lot of sense to include this in the library just yet – maybe more like sample code provided alongside (can even live in the repo)

Pinging @SBrandeis @coyotte508 @Pierrci for feedback on this though

coyotte508:
Precision: There is an [ongoing PR](https://github.com/huggingface/moon-landing/pull/2803) that will support uploading multiple files. 

It's designed for upload of LFS files, so it only supports LFS files and text files. It doesn't support non-LFS binary files. And the total non-LFS payload allowed is currently 250kB (could be upped to a few MB).

The endpoints could be updated to have better support for non-LFS files, in another PR, but for now it's uploading one or multiple LFS files in a single commit.

coyotte508:
Actually - The API will allow arbitrary commits (with a payload up to 20 MB, in addition to LFS data). So it should fit your needs :smile_cat:

coyotte508:
It's up - the payload of the commit API has been upped to 100MB.

The endpoint is `/api/{models,datasets,spaces}/:namespace/:repo/commit/:branch`.

The payload looks like that in TypeScript (question mark = optional):

```ts
export interface CommitPayload {
	/** 
	 * List of added / edited files. Encoding is "utf-8" if not specified. Use "base64" for files 
	 * containing non-utf8 characters, or when in doubt.
	 */
	files?:        Array<{ content: string; path: string; encoding?: "utf-8" | "base64"; }>;
	deletedFiles?: Array<{ path: string; }>;
	/**
	 * New files to track with LFS, if not already tracked.
	 * Do this through the backend as QOL.
	 */
	lfsFiles?:     Array<{ path: string; algo: "sha256"; oid: string; }>;
	/** Commit summary */
	summary:       string;
	/** Commit description */
	description?:   string;
}
```

You do not need to manually edit the `.gitattributes` file when adding LFS files: just specifiy the filenames & hashes of lfs files in the `lfsFiles` field.

---

With this API, you need to preupload LFS files. It's basically using the native git lfs API: a [batch](https://github.com/git-lfs/git-lfs/blob/main/docs/api/batch.md) call first, followed by either a [basic](https://github.com/git-lfs/git-lfs/blob/main/docs/api/basic-transfers.md) upload or multipart upload (with our custom protocol). 

You can see the TS implementation [here](https://github.com/huggingface/moon-landing/commit/ecf3ca63057cd14c68ed07e7458a081b496d4eef#diff-527c7c9195e0de01a94e1fedbc0858e534c9e7e582e8e76f5fa8c300d018b2b4R278).

Definitely available on slack to talk about it.

---

Other note: a `preupload` endpoint has also been added, to know if a file should be uploaded through LFS or not, by sending the file path, file size and a sample of the data (512 bytes), as well as an optional `gitAttributes` file. You don't have to use it if local detection is preferrable ^^ cc @LysandreJik

LysandreJik:
Thanks for the heads-up, @coyotte508! Will take a look at this and see how it could simplify the `upload_file`/`upload_folder` that @nateraw offers and simplify the #847.

julien-c:
let's maybe close this PR as @SBrandeis is offering to work on this and the implem will be very different

Still would love @nateraw (and others) review though

PR #866, opened 2022-05-06
Clickable link in huggingface-cli 

leondz:
When invoking huggingface-cli login without active authentication, a helpful feedback message containing a link to the tokens page is given. The link however is followed immediately by a full stop . so that it lands on a 404 when clicked (macOS / "hyper" shell). This PR adds one space to avoid the false 404 and land on the right page.


LysandreJik (approved):
Thank you, @leondz!

Issue #869, opened 2022-05-11
Remove history from Keras model card


merveenoyan:
This is a minor change, I would like to get rid of history in model card before next sprint (which I’m planning to have on first week of june) given we save it as a JSON now (see #861 ) , because for models with more epochs it will look bad on the card as it will log every epoch, and we already have tensorboard traces and history JSON, wanted to be transparent before I move on with the decision. 
Pinging @nateraw @osanseviero

merveenoyan:
@LysandreJik I'd love to get this in next release :)

LysandreJik:
Sounds good, adding it to the milestone!

PR #872, opened 2022-05-12
ENH remove history writing into model card, convert hyperparameters from a dict into table in model card, removed a header

merveenoyan:
This PR removes the history part in model card and turns hyperparameters dictionary into table in the model card like below. 

It was previously looking like this:
Ekran Resmi 2022-05-16 17 10 30

I removed the metrics (aka history) from the card itself, and turned hyperparameters into a table. History is already kept in a json file and when people train a model with 1000 epochs it will not look good on the card that's why I did it. 

Ekran Resmi 2022-05-12 17 48 24

I also removed the header "Training Procedure" because I felt like it looked redundant.

this solves #869

### DIFF
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -27,7 +27,7 @@ if is_tf_available():
     import tensorflow as tf
 
 
-def _extract_hyperparameters_from_keras(model):
+def _create_hyperparameters_table(model):
     if model.optimizer is not None:
         optimizer_params = model.optimizer.get_config()
         optimizer_params[
@@ -73,7 +73,7 @@ def _create_model_card(
     """
     Creates a model card for the repository.
     """
-    hyperparameters = _extract_hyperparameters_from_keras(model)
+    hyperparameters = _create_hyperparameters_table(model)
     if plot_model and is_graphviz_available() and is_pydot_available():
         _plot_network(model, repo_dir)
     readme_path = f"{repo_dir}/README.md"
@@ -185,7 +185,7 @@ def save_pretrained_keras(
         if model.history.history != {}:
             path = os.path.join(save_directory, "history.json")
             with open(path, "w") as f:
-                json.dump(model.history.history, f)
+                json.dump(model.history.history, f, indent=2, sort_keys=True)
 
     _create_model_card(model, save_directory, plot_model, metadata)
     tf.keras.models.save_model(
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -50,35 +50,6 @@ def _extract_hyperparameters_from_keras(model):
     return table
 
 
-def _parse_model_history(model, save_directory):
-    lines = None
-    if model.history is not None:
-        if model.history.history != {}:
-            path = os.path.join(save_directory, "history.json")
-            with open(path, "w", encoding="utf-8") as f:
-                json.dump(model.history.history, f, indent=2, sort_keys=True)
-            lines = []
-            logs = model.history.history
-            num_epochs = len(logs["loss"])
-
-            for value in range(num_epochs):
-                epoch_dict = {
-                    log_key: log_value_list[value]
-                    for log_key, log_value_list in logs.items()
-                }
-                values = dict()
-                for k, v in epoch_dict.items():
-                    if k.startswith("val_"):
-                        k = "validation_" + k[4:]
-                    elif k != "epoch":
-                        k = "train_" + k
-                    splits = k.split("_")
-                    name = " ".join([part.capitalize() for part in splits])
-                    values[name] = v
-                lines.append(values)
-    return lines
-
-
 def _plot_network(model, save_directory):
     tf.keras.utils.plot_model(
         model,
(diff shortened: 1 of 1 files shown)


adrinjalali:
@merveenoyan could you write a test for this so that we could see how it affects the results?

osanseviero (commented):
The PR name says "ENH remove history writing into model card and convert hyperparameters into table in model card for Keras #872", but i feel there are more things going around. Improving a bit the PR description and name would be very useful here. 

Some general thoughts/questions though:
* Creating a markdown table in a function called `_extract_hyperparameters_from_keras` seems misleading, should this be renamed?
* Why was "training procedure" h2 removed?
* Why were the metric writing removed entirely?

merveenoyan:
@osanseviero I will check and address your comments :)

merveenoyan:
> The PR name says "ENH remove history writing into model card and convert hyperparameters into table in model card for Keras #872", but i feel there are more things going around. Improving a bit the PR description and name would be very useful here.
> 
> Some general thoughts/questions though:
> 
> * Creating a markdown table in a function called `_extract_hyperparameters_from_keras` seems misleading, should this be renamed?
> * Why was "training procedure" h2 removed?
> * Why were the metric writing removed entirely?

I removed metric writing in the card only, previously we put it inside a JSON file so it's redundant to have it in the card and looks bad when user trains a model with 100 epochs (in the card).
I also thought h2 was redundant, I can bring it back though. 
I changed the name of the function, thanks for the heads up :)

src/huggingface_hub/keras_mixin.py
```diff
@@ -223,6 +181,12 @@ def save_pretrained_keras(
         else:
             metadata["tags"] = [task_name]
 
+    if model.history is not None:
+        if model.history.history != {}:
+            path = os.path.join(save_directory, "history.json")
+            with open(path, "w") as f:
+                json.dump(model.history.history, f, indent=2, sort_keys=True)
```
nateraw, line 188: Thank you for making sure the format is in line with HF stuff.

src/huggingface_hub/keras_mixin.py
```diff
@@ -27,45 +27,27 @@
     import tensorflow as tf
 
 
-def _extract_hyperparameters_from_keras(model):
+def _create_hyperparameters_table(model):
```
nateraw, line 30: A docstring here would be helpful to explain what's going on.

src/huggingface_hub/keras_mixin.py
```diff
@@ -123,15 +85,11 @@ def _create_model_card(
     model_card += "\n## Intended uses & limitations\n\nMore information needed\n"
     model_card += "\n## Training and evaluation data\n\nMore information needed\n"
     if hyperparameters is not None:
-        model_card += "\n## Training procedure\n"
```
nateraw, line 126: Why is this removed? I don't know if it needs to be there or not, but just curious as to why its gone now.

src/huggingface_hub/keras_mixin.py
```diff
@@ -27,45 +27,27 @@
...
-
-def _parse_model_history(model, save_directory):
-    lines = None
-    if model.history is not None:
-        if model.history.history != {}:
-            path = os.path.join(save_directory, "history.json")
-            with open(path, "w", encoding="utf-8") as f:
-                json.dump(model.history.history, f, indent=2, sort_keys=True)
-            lines = []
-            logs = model.history.history
```
nateraw, line 50: Wait so here the training metrics section is gone too? No TensorBoard logs either?

nateraw (commented):
Ok so I see a few things going on here, not just hparams related stuff. 

- I see training metrics being removed entirely? (Maybe I looked too quickly?)
- I don't see tensorboard in your [second model repo](https://huggingface.co/merve/model-card-pr) you posted above that is the result of this PR. Is this intentional? 
- I would really like to see some comments/docstrings added so folks who want to contribute can have better understanding of what's going on
- The code has changed quite a bit but the tests haven't. Should we be adding in tests for this? (I think probably yes?)

---

Also, if we're going to add hparams to a file, might I suggest we do so with `hparams.yaml` so we can get the added benefit of seeing hparams in tensorboard if the user pushes tensorboard? This would take some additional logic + testing I'm sure, but I'm pretty sure if you put hparams.yaml in the same dir as your tensorboard logs, they should surface up in the `hparams` tab in TensorBoard, which can be really helpful. Not a dealbreaker, but an idea that could be fruitful to some.

osanseviero:
Multiple small PRs, each one with individual changes, tends to be better than a larger PR with multiple changes impacting different completely unrelated things, it's making it a bit harder for 3 people to go through the PR, understand what is being changed, and clearly is creating a bit of confusion 😅

src/huggingface_hub/keras_mixin.py
merveenoyan: I thought it was redundant. I can bring it back.

src/huggingface_hub/keras_mixin.py
merveenoyan: TensorBoard logs don't leverage this anyway.
see https://github.com/huggingface/huggingface_hub/blob/65d3c88a24b648dd65ad76d78f11d76c45a7083c/src/huggingface_hub/keras_mixin.py#L422

merveenoyan:
I will break this PR up.

osanseviero:
Closing this as it's now split.

PR #879, opened 2022-05-19
FIX make import machinary private

adrinjalali:
This is quite minor, one review should be enough.

cc @osanseviero @LysandreJik


osanseviero (approved):
LG, thanks!

PR #880, opened 2022-05-23
MNT deprecate imports from snapshot_download

adrinjalali:
In https://github.com/huggingface/huggingface_hub/pull/874 we had to rename the `snapshot_download.py` file which makes `from huggingface_hub.snapshot_download import blah` fail.

This PR deprecated such imports and warns the user instead of failing.

Ref: https://github.com/scientific-python/lazy_loader/pull/9

cc @LysandreJik @osanseviero 

P.S.: I won't be available to apply suggestions here this week, but I think we should have this in the release. If you think anything needs to be changed in this PR, please apply the suggestions and then merge :)

### DIFF
--- a/tests/test_snapshot_download.py
+++ b/tests/test_snapshot_download.py
@@ -4,6 +4,8 @@ import tempfile
 import time
 import unittest
 
+import pytest
+
 import requests
 from huggingface_hub import HfApi, Repository, snapshot_download
 from huggingface_hub.hf_api import HfFolder
@@ -357,3 +359,8 @@ class SnapshotDownloadTests(unittest.TestCase):
 
     def test_download_model_with_ignore_regex_list(self):
         self.check_download_model_with_regex(["*.git*", "*.pt"], allow=False)
+
+
+def test_snapshot_download_import():
+    with pytest.warns(FutureWarning, match="has been made private"):
+        from huggingface_hub.snapshot_download import snapshot_download  # noqa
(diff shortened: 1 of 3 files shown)


osanseviero:
It seems `test_snapshot_download_import` is failing with `ModuleNotFoundError` https://github.com/huggingface/huggingface_hub/runs/6551770715?check_suite_focus=true

LysandreJik:
That looks very nice

tests/test_snapshot_download.py
```diff
@@ -357,3 +359,8 @@ def test_download_model_with_ignore_regex(self):
 
     def test_download_model_with_ignore_regex_list(self):
         self.check_download_model_with_regex(["*.git*", "*.pt"], allow=False)
+
+
+def test_snapshot_download_import():
+    with pytest.warns(FutureWarning, match="has been made private"):
+        from huggingface_hub.snapshot_download import snapshot_download  # noqa
```
LysandreJik, line 366: Could we also check that the method is the same as the one imported from `from huggingface_hub import snapshot_download`?

LysandreJik (approved):
Looks great, thanks for working on it @adrinjalali! Only left a nit regarding testing.

tests/test_snapshot_download.py
adrinjalali: Done

LysandreJik:
Perfect, thank you @adrinjalali!

PR #884, opened 2022-05-25
Extend `upload_file()` and `metadata_update()` to support Hub pull requests

lewtun:
## ⚠️ this PR is only for temporary use before the more future-proof #888 is merged

This PR extends the `upload_file()`  and `metadata_update()` functions to enable Hub PRs to be created programatically. The corresponding endpoint logic on `moon-landing` can be found [here](https://github.com/huggingface/moon-landing/blob/430b47c3c6432bcf73dddf4d5f8ee3823fc59997/server/app/gitViewerRoutes.ts#L970-L976).

Here's two examples to show what's involved:

**Hub PR via `upload_file()`**

This is the most common usage and involves:

* Downloading a file from a repo
* Modifying its contents
* Saving the file
* Pushing to the Hub 

```python
import json

from huggingface_hub import upload_file

# Download and open file
repo_id = "transformersbook/distilbert-base-uncased-finetuned-emotion"
filepath = hf_hub_download(repo_id=repo_id, filename="config.json")
with open(filepath, "r") as f:
    config = json.load(f)

# Modify contents and save
id2label = {
    "0": "sadness",
    "1": "joy",
    "2": "love",
    "3": "anger",
    "4": "fear",
    "5": "surprise",
}
label2id = {v: k for k, v in id2label.items()}
config["id2label"] = id2label
config["label2id"] = label2id
with open("config.json", "w") as f:
    json.dump(config, f, indent=2)

# Open Hub PR 
# Returns 'https://huggingface.co/transformersbook/distilbert-base-uncased-finetuned-emotion/discussions/PR_NUMBER'
hub_pr_url = upload_file(
    path_or_fileobj="config.json",
    path_in_repo="config.json",
    repo_id=repo_id,
    create_pr=True,
    pr_commit_summary="Fix label mappings",
    pr_commit_description="This PR fixes the label mappings to match the dataset.",
)
```

**Hub PR via `metadata_update()`**

This use case concerns scenarios where one wants to update the evaluation metadata on a model card. In this case, no files need to be manually downloaded since the evaluation results can be passed as a `dict`:

```python
from huggingface_hub.repocard import metadata_update

model_id = "lewtun/test-hub-pr-1"
eval_results = {
    "model-index": [
        {
            "name": "lewtun/gpt-4-finetuned-emotion",
            "results": [
                {
                    "task": {
                        "name": "Multi-class Classification",
                        "type": "text-classification",
                    },
                    "dataset": {"type": "emotion", "name": "Emotion"},
[...]

PR #888, opened 2022-05-30
✨ New `create_commit` API

SBrandeis:
# What's new
- `HfApi.create_commit` method to upload / delete several files at once to the HF Hub
- `HfApi.upload_folder` method: convenience wrapper around the previous method to upload a folder to the HF Hub, preserving its structure
- `commit_api` module with utility functions handling upload of large files via the git LFS protocol

# TODO
- [ ] Progress reporting with TQDM (probably in another PR)
- [x] Test the `create_pr` functionality

cc @coyotte508 @nateraw @lewtun @julien-c @osanseviero


src/huggingface_hub/hf_api.py
```diff
@@ -1868,6 +1875,109 @@ def upload_file(
...
+        self,
+        *,
+        repo_id: str,
+        files: Iterable[FileUploadInput],
+        commit_summary: str,
+        commit_description: Optional[str] = None,
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+    ):
```
julien-c, line 1888: shouldn't we do a more general `create_commit` which lets one delete and add (or update) files in the same call?

julien-c: like `file_changes: FileChanges` with `FileChanges` is `{ additions: Obj[]; deletions: Obj[] }`

or `files: Iterable[FileUploadInput] = []` AND `deleted_files: Iterable[str] = []` which throws if both arrays are empty?

coyotte508:
IMO `upload_file` should also use the new APIs (as well as `delete_file`)

src/huggingface_hub/hf_api.py
```diff
@@ -1724,17 +1735,78 @@ def move_repo(
...
+        self,
+        *,
+        repo_id: str,
+        operations: Iterable[CommitOperation],
+        commit_summary: str,
+        commit_description: Optional[str] = None,
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        create_pr: Optional[bool] = None,
```
SBrandeis, line 1748: cc @lewtun (untested atm)

src/huggingface_hub/hf_api.py
```diff
@@ -1803,70 +1880,82 @@ def upload_file(
...
-
-        try:
-            r.raise_for_status()
-        except HTTPError as err:
-            if identical_ok and err.response.status_code == 409:
-                from .file_download import hf_hub_url
-
-                return hf_hub_url(
-                    repo_id, path_in_repo, revision=revision, repo_type=repo_type
+    def upload_folder(
```
SBrandeis, line 1907: cc @lhoestq 

Only additive for now - meaning it does not delete files that are already present on the repo and not locally

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def iter_content(
+        self, chunk_size: Optional[int] = None
+    ) -> Generator[bytes, None, None]:
+        if isinstance(self.path_or_fileobj, bytes):
+            if chunk_size is None:
+                yield self.path_or_fileobj
+            else:
+                idx = 0
+                while chunk := self.path_or_fileobj[idx : idx + chunk_size]:
```
SBrandeis, line 54: Walrus operator has been introduced in Python 3.8

Will need to rewrite this function to be compatible with 3.7

src/huggingface_hub/hf_api.py
```diff
@@ -1726,17 +1737,78 @@ def move_repo(
...
+        )
+        commit_resp.raise_for_status()
+        return commit_resp.json().get("pullRequestUrl", None)
+
     @_deprecate_positional_args
     def upload_file(
         self,
         *,
-        path_or_fileobj: Union[str, bytes, IO],
+        path_or_fileobj: Union[str, bytes, BufferedIOBase],
```
SBrandeis, line 1802: Restrains the type to `BufferedIOBase` subclasses

src/huggingface_hub/hf_api.py
```diff
@@ -1805,70 +1878,82 @@ def upload_file(
...
+            token=token,
+            revision=revision,
+            create_pr=create_pr,
+        )
+        return (
+            pr_url
+            if pr_url is not None
+            else hf_hub_url(
+                repo_id, path_in_repo, revision=revision, repo_type=repo_type
             )
```
julien-c, line 1906: maybe the server should return the commit URL (or the blob URL?) in the JSON if not in PR mode?

and it's consistent with the old endpoint:

[internal link] 
https://github.com/huggingface/moon-landing/blob/cb42e3de2011d92905a15bf584f3a6906ca52f18/server/app/gitViewerRoutes.ts#L1309-L1312

Bonus is that then you don't need to move `hf_hub_url` which would keep this diff smaller =)

julien-c: (my opinion is that we can change the semantic of the function to return the commit URL instead of the blob URL)

src/huggingface_hub/hf_api.py
```diff
@@ -1724,17 +1735,78 @@ def move_repo(
...
+        *,
+        repo_id: str,
+        operations: Iterable[CommitOperation],
+        commit_summary: str,
+        commit_description: Optional[str] = None,
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        create_pr: Optional[bool] = None,
+    ) -> Optional[str]:
```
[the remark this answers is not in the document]
julien-c: like it!

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,444 @@
...
+        elif isinstance(self.path_or_fileobj, str):
+            with open(self.path_or_fileobj, "rb") as reader:
+                while chunk := reader.read(chunk_size):
+                    yield chunk
+        else:
+            while chunk := self.path_or_fileobj.read(chunk_size or -1):
+                yield chunk
+
+
+CommitOperation = Union[CommitOperationAdd, CommitOperationDelete]
```
julien-c, line 65: nice to see some good typings in a Python codebase 🎉 =)

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,444 @@
...
+            for file in additions
+            if file.upload_mode == "regular"
+        ],
+        "lfsFiles": [
+            {"path": file.path_in_repo, "algo": "sha256", "oid": file.sha}
+            for file in additions
+            if file.upload_mode == "lfs"
+        ],
+        "deletedFiles": [{"path": deletion.path_in_repo} for deletion in deletions],
+    }
```
julien-c, line 444: off-topic, but did you read the code in `commands/lfs.py` @SBrandeis? (i wrote it but don't remember any of it 😱 would be good for you to read if you haven't already, out of curiosity – re. lfs protocol)

src/huggingface_hub/hf_api.py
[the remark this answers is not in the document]
lewtun: looks nice! cc @gante who was looking for this kind of functionality too

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,444 @@
...
+class CommitOperationDelete:
+    path_in_repo: str
+
+
+@dataclass
+class CommitOperationAdd:
+    path_in_repo: str
+    path_or_fileobj: Union[str, bytes, io.BufferedIOBase]
+
+    def __post_init__(self):
```
adrinjalali, line 34: validation should happen when we need to use the data. A user can construct a class with some data, and then change it later in which case `__post_init__` won't be triggered.

src/huggingface_hub/hf_api.py
```diff
@@ -1726,17 +1737,78 @@ def move_repo(
...
         *,
-        path_or_fileobj: Union[str, bytes, IO],
+        path_or_fileobj: Union[str, bytes, BufferedIOBase],
         path_in_repo: str,
         repo_id: str,
         token: Optional[str] = None,
         repo_type: Optional[str] = None,
         revision: Optional[str] = None,
         identical_ok: bool = True,
+        commit_summary: Optional[str] = None,
```
lewtun, line 1809: Could we also include these args in the `metadata_update()` function as well, e.g. how I did it here: https://github.com/huggingface/huggingface_hub/pull/884/files#diff-52ca35d2851516051ee1658be7c52fedba606b08b94bdc3ac2bb67f54e0d0a0bR198-R200

adrinjalali (commented):
Thanks for the PR @SBrandeis . There's quite a bit to unpack here.

For us to better be able to review this code, it would be nice to have a section in the user guides explaining the new API.

src/huggingface_hub/hf_api.py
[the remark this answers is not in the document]
SBrandeis: > I'm also not sure if create_pr should be a part of this method. Creating a PR seems to be more of its own operation to me

Good point, although passing a `create_pr` argument is close to the flow on the website (see the screenshot below)

image

On the other hand, we could have the following API (which is closer to the "advanced" PR workflow)

```python
from huggingface_hub import create_pull_request, create_commit

revision = create_pull_request(...)
create_commit(..., revision=revision)
```

SBrandeis:
Also cc @Wauplin 🤗

src/huggingface_hub/hf_api.py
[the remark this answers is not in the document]
adrinjalali: Trying the website's flow, I find it quite confusing, since I expect to follow the same PR/MR workflow as GH or GitLab, but we seem to be forcing people on a branch name that we choose, and they then have to fetch that locally and then apply their changes.

The usual flow is that users create their branch locally, push, and then create their PR, and I'm not sure if we want to make people get used to a very different flow than what they already know.

Regardless, for users who are happy with their python code or terminal, it would make more sense to give them an interface which is more familiar to what they do in other projects/platforms.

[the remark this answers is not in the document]
coyotte508: >  since I expect to follow the same PR/MR workflow as GH or GitLab

>  I'm not sure if we want to make people get used to a very different flow than what they already know.

> it would make more sense to give them an interface which is more familiar to what they do in other projects/platforms.

The reason functions such as `upload_file` exit in the hub library is because data scientists may not be familiar with the git ecosystem.

HF targets data scientists, whereas Github/Gitlab target software engineers.

We made the choice to be separate from Github/Gitlab (no "fork" button) because the use case is different. We just want the hub library to be easy for data scientists, with the least friction. It should make all the git commands (creating a PR, adding new stuff to a PR, ...) as simple as possible.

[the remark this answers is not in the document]
adrinjalali: Using git workflow has become quite normal even among data scientists in the industry. It might not be the case in academia, but in the industry that's how people work since they work with other dev teams which use the same workflow. I don't mind how it's implemented in the UI (although I don't like it), but in the library it should be something that more advanced users don't find unintuitive and confusing.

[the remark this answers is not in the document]
julien-c: We could maybe define a `create_pull_request` method that would be calling exactly `create_commit` with the `create_pr=True` param, what do you think?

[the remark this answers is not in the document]
adrinjalali: What's the input to `create_pull_request`? A target (and maybe source) branch name? Can we enable this workflow:

- `git checkout -b feature_branch`
- `create_commit`
- `create_pr(target='main')`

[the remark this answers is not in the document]
julien-c: no `create_pull_request` would be the exact same params as `create_commit` ie. a commit message and a changeset. 

No, we do not want to support PRs from/to arbitrary branches. It's a design decision, we can reassess in a year but I think it'll be a good design decision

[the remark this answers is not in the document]
adrinjalali: why?

SBrandeis:
I implemented a very naive multi-threaded upload in 05a040e

It's not very elegant, so let me know if you want to handle that in a different PR @adrinjalali @LysandreJik @nateraw @lewtun 

cc @julien-c, @lhoestq @Wauplin who might have more experience with multi-threading in Python

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,392 @@
...
+                ]
+            )
+            raise ValueError(f"LFS batch endpoint returned errors:\n{message}")
+
+        upload_func = partial(_upload_lfs_object, oid2addop=oid2addop)
+        if num_threads is None:
+            [upload_func(obj) for obj in objects]
+        else:
+            with ThreadPoolExecutor(num_threads) as pool:
+                wait([pool.submit(upload_func, obj) for obj in objects])
```
Wauplin, line 239: > I implemented a very naive multi-threaded upload in https://github.com/huggingface/huggingface_hub/commit/05a040e00dd78c9136494c7de76b062d62af111b
> It's not very elegant, so let me know if you want to handle that in a different PR

Hey @SBrandeis, I'm following up on your last message about multi-threading. 

IMHO, it is a good use case for using `asyncio` here which offers more flexibility on how to handle the futures. However, to do so you need `_upload_lfs_object` to be async, which means `upload_lfs_file` requires an implementation with async calls instead of the  `request.put(...)`. Since `upload_lfs_file` seems to be a public API (please confirm), the decision to switch it to an async method is more of a design decision. A solution to use async here could be to have an internal `_async_upload_lfs_file` that does the job and the public `upload_lfs_file` would just be calling this internal one (and you can reuse the async one in commit_api).


Otherwise, staying with a `ThreadPoolExecutor` like in your solution can do the job. I would be more careful about exceptions. For now if a thread throws an exception, it will be silently ignored. What I would do is:
1. Keep a list of your futures:
```py
futures = [pool.submit(upload_func, obj) for obj in objects]
```
2.  Wait on the futures but only until first exception (see https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait)
```py
completed, pending = wait(futures, return_when=ALL_COMPLETED)
```

3. Cancel pending futures (if any). If a future is pending it means another future raised an exception. Here it's just to avoid uploading other big files if you know at least 1 failed. :warning: Problem is that you can't cancel a task that is already running. IDK a good way to stop a running thread (maybe look into https://superfastpython.com/threadpoolexecutor-stop-tasks/ ?)
```py
for future in pending:
    future.cancel()
```

4. Check that all futures are completed and raise exception if necessary. Idk how HF hub is working if some files have been uploaded but not all (should it delete them ?). I guess that's more a question on the backend side than in this lib/sdk ?
```py
for future in futures:
    future.result() # <- will raise if the future threw an exception
```


Hope this can help. I'd be glad to have other opinions about how to properly handle that as well.

julien-c: I really don't know a lot about concurrency in Python (the last time I used it, 2 years ago, I wanted to kill myself)
but two high-level questions:
- shouldn't we use httpx or similar library?
- are we sure we do want to implement concurrent uploads? (some people will be constrained by total local upload bandwidth, i.e. the speedup of concurrency is a bit hard for me to predict)

LysandreJik (commented):
Thank you for working on this very impressive PR, Simon! I think most of the behavior is in place.

IMO before we merge there are a few things that we have to fix still:
- There is one backward-compatibility issue regarding `identical_ok`.
- Some objects do not have documentation, or documentation that needs to be updated (`CommitOperationDelete`, `CommitOperationAdd`, `base_url`, others)
- I fear that the documentation template you have taken was the old one, and it needs to be updated. The good news is that switching to the new format will be painless, it should only be a matter of `s/``/`/`: we now use single backticks instead of double backticks for variables.

This PR contains API reference but this is not linked anywhere. I think we should either add a file in `docs` that introduces the commit API, or we should modify an existing one. However, this PR is already huge and it would be better to do it in a second PR.

I would personally add comments at quite a few places to explain the logic that's happening.

I ran into an error when trying to upload an LFS file with the most recent version. Here's a reproducible code example:

```py
import pathlib
import tempfile

with tempfile.TemporaryDirectory() as tmp:
    path_to_dir = pathlib.Path(tmp)

    with open(path_to_dir / 'file_0.txt', 'w+') as f:
        f.write("Hey")

    with open(path_to_dir / 'file_1.txt', 'wb') as f:
        f.truncate(1024 * 1024)


    upload_folder(
        repo_id='<org>/<model_id>',
        folder_path=tmp,
        path_in_repo="dummy-dir",
    )
```

src/huggingface_hub/constants.py
```diff
@@ -25,7 +25,7 @@
 )
 
 ENDPOINT = os.getenv("HF_ENDPOINT") or (
-    "https://moon-staging.huggingface.co" if _staging_mode else "https://huggingface.co"
+    "https://hub-ci.huggingface.co" if _staging_mode else "https://huggingface.co"
```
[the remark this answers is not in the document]
SBrandeis: Yes ! @XciD set it up - it should be up to date with moonlanding main branch, and uses Gitaly

src/huggingface_hub/commit_api.py
lhoestq: > shouldn't we use httpx or similar library?

a similar library is aiohttp, and is already a dependency of `datasets` (we use it for streaming)

> some people will be constrained by total local upload bandwidth, i.e. the speedup of concurrency is a bit hard for me to predict

True, but it helps if you have thousands of small files for example (e.g. folder of CSV, PNG, WAV etc.)

src/huggingface_hub/constants.py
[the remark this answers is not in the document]
XciD: Not yet in full CI/CD but will be soon

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,415 @@
...
+    endpoint = endpoint if endpoint is not None else ENDPOINT
+    headers = {"authorization": f"Bearer {token}"} if token is not None else None
+    payload = {"files": [_preupload_payload(op) for op in additions]}
+
+    resp = requests.post(
+        f"{endpoint}/api/{repo_type}s/{repo_id}/preupload/{revision}",
+        json=payload,
+        headers=headers,
+    )
+    resp.raise_for_status()
```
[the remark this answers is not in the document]
SBrandeis: That's the way we do it in the website

In my opinion the proposed change is better because we don't have to edit this code if the logic ever changes

src/huggingface_hub/hf_api.py
```diff
@@ -1726,17 +1736,124 @@ def move_repo(
             " completed."
         )
 
+    def create_commit(
+        self,
+        *,
+        repo_id: str,
+        operations: Iterable[CommitOperation],
```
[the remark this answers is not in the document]
SBrandeis: Yes, agreed !

src/huggingface_hub/commit_api.py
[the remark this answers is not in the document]
anton-l: This case is fine IMO, the head requests have only been an issue during mass-downloads of READMEs from the hub (for metadata fixes), which I can totally live with :)

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+logger = logging.get_logger(__name__)
+
+
+UploadMode = Literal["lfs", "regular"]
+
+
+@dataclass
+class CommitOperationDelete:
+    """
+    Data strcture holding necessary info to delete
```
osanseviero, line 37:
```suggestion
    Data structure holding necessary info to delete
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+    a file from a repository on the HF Hub
```
osanseviero, line 38:
```suggestion
    a file from a repository on the Hub
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+
+    Args:
+        path_in_repo (`str`):
+            Path of the file to delete in the repo
```
osanseviero, line 42: Maybe for consistency with  https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L1750-L1752

```suggestion
        path_in_repo (`str`):
           Relative filepath in the repo, for example:
            `"checkpoints/1fec34a/weights.bin"`
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+
+    path_in_repo: str
+
+
+@dataclass
+class CommitOperationAdd:
+    """
+    Data structire holding necessary info to upload a file
```
osanseviero, line 51:
```suggestion
    Data structure holding necessary info to upload a file
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+
+    path_in_repo: str
+    path_or_fileobj: Union[str, bytes, BinaryIO]
+
+    _upload_info: Optional[UploadInfo] = field(default=None, init=False)
+
+    def validate(self):
+        """
+        Ensures `path_or_fileobj` is valid
```
osanseviero, line 72:
```suggestion
        Ensures `path_or_fileobj` is valid.
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+def fetch_upload_modes(
+    additions: Iterable[CommitOperationAdd],
+    repo_type: str,
+    repo_id: str,
+    token: str,
+    revision: str,
+    endpoint: Optional[str] = None,
+) -> List[Tuple[CommitOperationAdd, UploadMode]]:
+    """
+    Requests the HF Hub to determine wether each input file should be
```
osanseviero, line 316:
```suggestion
    Requests the Hub to determine wether each input file should be
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+def prepare_commit_payload(
+    additions: Iterable[Tuple[CommitOperationAdd, UploadMode]],
+    deletions: Iterable[CommitOperationDelete],
+    commit_summary: str,
+    commit_description: Optional[str] = None,
+):
+    """
+    Builds the payload to pass to the `commit` API of the HF Hub
```
osanseviero, line 378:
```suggestion
    Builds the payload to pass to the `commit` API of the Hub
```

src/huggingface_hub/hf_api.py
```diff
@@ -1726,20 +1736,138 @@ def move_repo(
...
+
+                    - ``CommitOperationAdd`` to upload a file
+                    - ``CommitOperationDelete`` to delete a file
+
+            commit_summary (``str``):
+                The summary of the commit that will be created
+
+            commit_description (``str``, *optional*):
+                The description of the commit that will be created
+
```
osanseviero, line 1771: The difference between these two is not very clear to me. Is `summary` first line and description the rest? Should we rename one of these to `message` for consistency with `git commit -m...`?

osanseviero (commented):
Looks very cool! I just did a light pass and left some minor suggestions/questions

tests/test_hf_api.py
```diff
@@ -452,47 +453,37 @@ def test_move_repo(self):
...
+        self.nested_tmp_file = os.path.join(self.tmp_dir, "nested", "file")
+        with open(self.nested_tmp_file, "w+") as f:
+            f.write(self.tmp_file_content)
+
         self.addCleanup(
             lambda: shutil.rmtree(self.tmp_dir, onerror=set_write_permission_and_retry)
         )
 
-    @retry_endpoint
-    def test_upload_file_validation(self):
```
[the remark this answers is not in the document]
SBrandeis: re-added it !

src/huggingface_hub/hf_api.py
```diff
@@ -1764,10 +1893,17 @@ def upload_file(
                 The git revision to commit from. Defaults to the head of the
                 `"main"` branch.
             identical_ok (`bool`, *optional*, defaults to `True`):
-                When set to false, will raise an [HTTPError](
-                https://2.python-requests.org/en/master/api/#requests.HTTPError)
+                (Deprecated, will be removed in v0.3.0) When set to false, will
+                raise an [HTTPError](https://requests.readthedocs.io/en/latest/api/#requests.HTTPError)
                 when the file you're trying to upload already exists on the hub
                 and its content did not change.
```
SBrandeis, line 1899: @LysandreJik @adrinjalali lmk if there's anything more I should do

src/huggingface_hub/repocard.py
```diff
@@ -192,6 +210,10 @@ def metadata_update(
         repo_type=repo_type,
         identical_ok=False,
```
SBrandeis, line 211: There will be a breaking change when we remove `identical_ok`

src/huggingface_hub/hf_api.py
LysandreJik: I will add that it will be removed in v0.10.0 rather than v0.3.0 (we'll be releasing v0.8.0 next week)

LysandreJik (approved):
Thanks for going the extra mile for the deprecation! LGTM!

src/huggingface_hub/file_download.py
```diff
@@ -957,6 +957,9 @@ def hf_hub_download(
             # between the HEAD and the GET (unlikely, but hey).
             if 300 <= r.status_code <= 399:
                 url_to_download = r.headers["Location"]
+                if "lfs.huggingface.co" in url_to_download or "lfs-staging.huggingface.co" in url_to_download:
+                    # Remove authorization header when downloading a LFS blob
+                    headers.pop("authorization", None)
```
SBrandeis, line 965: Had to remove the `authorization` header when requesting a LFS file, otherwise S3 throws a 400 with:
```xml
<Error>
   <Code>InvalidArgument</Code>
   <Message>Unsupported Authorization Type</Message>
   <ArgumentName>Authorization</ArgumentName>
   <ArgumentValue>Bearer XXX</ArgumentValue>
   <RequestId>...</RequestId> 
  <HostId>...</HostId>
</Error>
```

Not sure if this is new ? cc @julien-c @huggingface/moon-landing-back

XciD: related to the CDN changes I've made for filename

src/huggingface_hub/hf_api.py
```diff
@@ -1761,6 +1897,67 @@ def upload_file(
...
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        identical_ok: bool = True,
+        commit_message: Optional[str] = None,
+        commit_description: Optional[str] = None,
+        create_pr: Optional[bool] = None,
+    ) -> str:
+        """
+        Deprecated - will be removed in 0.10.0
```
Pierrci, line 1957: This seems a bit inconsistent with `identical_ok` which is planned to be removed in 0.11.0, and whose value is used to determine whether or not to use this method in `upload_file`?

src/huggingface_hub/hf_api.py
```diff
@@ -1761,6 +1897,67 @@ def upload_file(
...
+        path_or_fileobj: Union[str, bytes, BinaryIO],
+        path_in_repo: str,
+        repo_id: str,
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        identical_ok: bool = True,
+        commit_message: Optional[str] = None,
+        commit_description: Optional[str] = None,
+        create_pr: Optional[bool] = None,
```
coyotte508, line 1954: Note that this won't work with https://github.com/huggingface/moon-landing/pull/3144

The `create_pr` param is new so it should be ok to not have it in `legacy_upload_file` ?

julien-c: agreed

src/huggingface_hub/repocard.py
```diff
@@ -146,10 +150,25 @@ def metadata_update(
...
-
+        commit_message (`str`, *optional*):
+            The summary / title / first line of the generated commit. Defaults to
+            `f"Update metdata with huggingface_hub"`
+        commit_description (`str` *optional*)
+            The description of the generated commit
+        revision (`str`, *optional*):
+            The git revision to commit from. Defaults to the head of the
+            `"main"` branch.
+        create_pr (`boolean`, *optional*):
```
nateraw, line 161: What happens when I pass `create_pr=True` with `revision=None`?

SBrandeis: It creates a PR against the `main` branch

src/huggingface_hub/hf_api.py
SBrandeis: Adressed in 2aa1ef8

I removed `_legacy_upload_file`

SBrandeis:
@adrinjalali @LysandreJik can this be merged ?

src/huggingface_hub/__init__.py
```diff
@@ -141,12 +142,13 @@ def __dir__():
             "from_pretrained_fastai",
             "push_to_hub_fastai",
         ],
-        "file_download": ["cached_download", "hf_hub_download", "hf_hub_url"],
```
adrinjalali, line 144: this is a backward incompatible change. We should either not move it, or if we're moving it, there should be a proxy function in this file to warn the users that it's been moved.

src/huggingface_hub/commands/lfs.py
```diff
@@ -119,43 +118,6 @@ def read_msg() -> Optional[Dict]:
     return msg
 
 
-class FileSlice(AbstractContextManager):
```
adrinjalali, line 122: Same here. If we're almost certain that people are not using this class in other libraries, we can remove it, but this might be breaking people's code. It'd be nice if @osanseviero and @LysandreJik could confirm this is not used in other places.

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
+"""
```
adrinjalali, line 1: if we have this file as `_commit_api.py`, we won't need to deal with backward compatible changes if we move methods around which users shouldn't be using.

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from typing import BinaryIO, Dict, Iterable, List, Optional, Tuple, Union
+
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
adrinjalali, line 19: Not really important, but curious, which one do you prefer? What's here or this suggestion?

```suggestion
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+
+    path_in_repo: str
+    path_or_fileobj: Union[str, bytes, BinaryIO]
+
+    _upload_info: Optional[UploadInfo] = field(default=None, init=False)
+
+    def validate(self):
+        """
+        Ensures `path_or_fileobj` is valid.
```
adrinjalali, line 74: Could you please also explain what a valid `path_or_fileobj` is?

Also, I wonder why we don't use the `file` pattern which is common in python, to be a file object or a path anyway, while having a much shorter parameter name.

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
...
+                self.path_or_fileobj.seek(0, os.SEEK_CUR)
+            except (OSError, AttributeError) as exc:
+                raise ValueError(
+                    "path_or_fileobj is a file-like object but does not implement"
+                    " seek() and tell()"
+                ) from exc
+
+    def upload_info(self) -> UploadInfo:
+        """
+        Computes and caches UploadInfo for the underlying data behind `path_or_fileobj`
```
adrinjalali, line 104: would be nice to have hyperlinks to objects when we refer to them in the docs (`UploadInfo`)

Also, I'm not sure if I understand what this method does if I was reading the docstring out of context. Is this a method which is supposed to be called by users? i.e. should it be public or private?

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
...
+                raise RuntimeError(
+                    f"Error while uploading {operation.path_in_repo} to the Hub"
+                ) from exc
+
+
+def _upload_lfs_object(
+    operation: CommitOperationAdd, lfs_batch_action: dict, token: str
+):
+    """
+    Handles uploading a given object to the Hub with the LFS protocol.
```
adrinjalali, line 270:
```suggestion
    """Handles uploading a given object to the Hub with the LFS protocol.
```

src/huggingface_hub/commit_api.py
```diff
@@ -0,0 +1,424 @@
...
+
+    Defers to `~.utils.lfs.lfs_upload` for the actual upload logic.
```
adrinjalali, line 272: This pattern (`~.utils.lfs.lfs_upload`) understood by doc builder?

src/huggingface_hub/hf_api.py
```diff
@@ -1674,19 +1684,137 @@ def move_repo(
...
+        self,
+        repo_id: str,
+        operations: Iterable[CommitOperation],
+        *,
+        commit_message: str,
+        commit_description: Optional[str] = None,
+        token: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        create_pr: Optional[bool] = None,
```
adrinjalali, line 1697: why not use the actual default value here?

src/huggingface_hub/hf_api.py
```diff
@@ -1674,19 +1684,137 @@ def move_repo(
...
+        commit_url = f"{self.endpoint}/api/{repo_type}s/{repo_id}/commit/{revision}"
+
+        commit_resp = requests.post(
+            url=commit_url,
+            headers={"Authorization": f"Bearer {token}"},
+            json=commit_payload,
+            params={"create_pr": 1} if create_pr else None,
+        )
+        _raise_for_status(commit_resp)
+        return commit_resp.json().get("pullRequestUrl", None)
```
adrinjalali, line 1800: what's the output if it's not a PR?

src/huggingface_hub/hf_api.py
```diff
@@ -1674,19 +1684,137 @@ def move_repo(
...
+                space, `None` or `"model"` if uploading to a model. Default is
+                `None`.
+
+            revision (`str`, *optional*):
+                The git revision to commit from. Defaults to the head of the
+                `"main"` branch.
+
+            create_pr (`boolean`, *optional*):
+                Whether or not to create a Pull Request from `revision` with that commit.
+                Defaults to `False`.
```
adrinjalali, line 1735: should also explain how it affects the return value.

src/huggingface_hub/hf_api.py
```diff
@@ -1674,19 +1684,137 @@ def move_repo(
...
+
+            num_threads (`int`, *optional*):
+                Number of concurrent threads for uploading files. Defaults to 5.
+                Setting it to 2 means at most 2 files will be uploaded concurrently.
+        """
```
adrinjalali, line 1740: missing Returns section

src/huggingface_hub/hf_api.py
```diff
@@ -1713,6 +1841,13 @@ def upload_file(
...
+            commit_message (`str`, *optional*):
+                The summary / title / first line of the generated commit
+            commit_description (`str` *optional*)
+                The description of the generated commit
+            create_pr (`boolean`, *optional*):
+                Whether or not to create a Pull Request from `revision` with that commit.
+                Defaults to `False`.
 
         Returns:
             `str`: The URL to visualize the uploaded file on the hub
```
adrinjalali, line 1853: I think we're changing the return value here? And it depends on `create_pr`?

adrinjalali (commented):
Thanks for all the work so far @SBrandeis . I tried not to be too picky with the review, but there are a few aspects which we need to improve here.

There are some parts flagged by codecov which are not tested, and I didn't check if the rest of the logic and execution paths are covered.

There are a few backward incompatibility issues which are quite major, related to commonly used methods.

This PR also needs to touch the user guides and explain how users can create PRs and use the output of the new methods.

src/huggingface_hub/commit_api.py
nateraw: I prefer the `sys.version_info` solution to the try/except. idk just feels cleaner to me imo

SBrandeis: No preference for me, though the `sys.version_info` solution is consistent with the rest of the codebase

src/huggingface_hub/commit_api.py
SBrandeis: > I wonder why we don't use the file pattern

Usually, a `file` argument is expected to have a `write` method, no? See https://docs.python.org/fr/3/library/functions.html#print for example

Here we also accept a file path (`str`), hence `path_or_fileobj`

SBrandeis: From a personal POV, I like the explicitness of `path_or_fileobj`, which is clearly either a path or a [file object](https://docs.python.org/3/glossary.html#term-file-object)

src/huggingface_hub/commit_api.py
SBrandeis: Made it private

src/huggingface_hub/hf_api.py
SBrandeis: `None`

src/huggingface_hub/commands/lfs.py
SBrandeis: Addressed in f419000

src/huggingface_hub/hf_api.py
SBrandeis: Addressed in 552bfb6

src/huggingface_hub/commit_api.py
SBrandeis: Not sure, I switched to `[~huggingface_hub.XXX]` as elsewhere

src/huggingface_hub/commands/lfs.py
julien-c: > If we're almost certain that people are not using this class in other libraries

We are, it was just an internal obj for the LFS-specific part of the upload script. Very little chance that anyone ever used it. I would suggest to just remove

julien-c (approved):
LGTM, let's merge on monday? 🚢

adrinjalali (commented):
I think this is a major enough of a feature that we'd need a separate documentation page under `https://github.com/huggingface/huggingface_hub/tree/main/docs/source` for it, and also touch the quick start guide to link to the right part of the docs.

Otherwise I think we can then merge this and let it be tested before the next release.

coyotte508:
https://github.com/huggingface/moon-landing/pull/3144 introduced an optional `parentCommit` feature that adds safety when committing - it rejects when the latest commit on the branch does not match `parentCommit`.

It'd be nice to have that soon in the API, either allowing to specify the parent commit directly on the hub library side or checking the local state (if we still use git repos?)

In another PR of course.

btw cc @lhoestq for when you want to implement the commit API on datasets.

SBrandeis:
Merging this, given comments on Slack: https://huggingface.slack.com/archives/C02V5EA0A95/p1655717469105299

adrinjalali:
Since we are not going to have this in the current release, a few more hours wouldn't change anything, and @TristanThrush had explicitly mentioned the request (I have no idea of the reasoning behind it), and that it's the middle of the night for him right now, it would have been nice to wait for him to have a chance to respond. We also really try to not merge our own PRs @SBrandeis .

SBrandeis:
> We also really try to not merge our own PRs @SBrandeis .

Sorry, @adrinjalali 😓 I'll keep this in mind next time

PR #889, opened 2022-05-30
Add list_spaces to HfApi

### LINKED ISSUE #881: Add list_spaces
Allow users listing spaces similar to how they can list models and datasets

We can use something like this for example

```
HF_LIST_ENDPOINT = HF_ENDPOINT + "/api/spaces?filter={type}"
r = requests.get(HF_LIST_ENDPOINT.format(type=type))
r.raise_for_status()
d = r.json()
```

cc @lvwerra

cakiki:
First try; fixes #881.  :-) 
Reused the `SpaceInfo` class, but not sure that was the way to go because of unused fields:
```
SpaceInfo: {
	id: BigSalmon/Streamlit_Headline_Writer
	sha: None
	lastModified: None
	siblings: None
	private: None
	author: None
	sdk: streamlit
}
```

cc @osanseviero @LysandreJik


src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,34 @@ def list_metrics(self) -> List[MetricInfo]:
         d = r.json()
         return [MetricInfo(**x) for x in d]
 
+    def list_spaces(self, space_sdk: Optional[str] = None) -> List[SpaceInfo]:
```
osanseviero, line 1151: Consider using parameters similar to the ones in https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L763-L776. I think here it would be nice to have `use_auth_token` so people can filter for private Spaces.

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,34 @@ def list_metrics(self) -> List[MetricInfo]:
...
+        """
+        if space_sdk is None:
+            path = f"{self.endpoint}/api/spaces"
+        else:
+            if space_sdk not in SPACES_SDK_TYPES:
+                raise ValueError(
+                    f"Invalid SDK type {space_sdk}. Has to be one of"
+                    f" {SPACES_SDK_TYPES}."
+                )
+            path = f"{self.endpoint}/api/spaces?filter={space_sdk}"
```
osanseviero, line 1172: `filter=` can actually be used for other things apart from SDK. For example, if you go to https://huggingface.co/spaces?filter=metric you see all Spaces that have tag metric in the README (https://huggingface.co/spaces/evaluate-metric/trec_eval/blob/main/README.md#L7)

osanseviero (commented):
Overall this looks to go in the right direction! I would suggest to add a test for this :smile:

src/huggingface_hub/hf_api.py
julien-c: yep!

cakiki:
Should I be defining a new `SpacesFilter` class like this one: https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/utils/endpoint_helpers.py#L67 ?

julien-c:
@osanseviero @LysandreJik can chime in but i don't think so

LysandreJik:
It's useful, but definitely not necessary. Having `list_spaces` is already a very nice first step!

cakiki:
`datasets=True` and `models=True` doesn't seem to work (returns an empty list). Not sure if that's an implementation error on my side or an endpoint issue. Passing a dataset id or a model id works fine.

julien-c (commented):
looks good to me from a quick glance!

Thank you @cakiki!!

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+        self,
+        *,
+        filter: Union[str, Iterable[str], None] = None,
+        author: Optional[str] = None,
+        search: Optional[str] = None,
+        sort: Union[Literal["lastModified"], str, None] = None,
+        direction: Optional[Literal[-1]] = None,
+        limit: Optional[int] = None,
+        cardData: Optional[bool] = None,
+        datasets: Union[bool, str, Iterable[str], None] = None,
```
osanseviero, line 1162: So indeed this does not work

```py
path = "https://huggingface.co/api/spaces"
params = {"models":True}

r = requests.get(path, params=params)

len(r.json())
``` 
I would keep datasets and models parameters to be just string and list of strings. That way you just use them to specify which datasets/models you want linked. We discussed internally and we think we can remove the boolean

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+        linked: Optional[bool] = None,
+        full: Optional[bool] = None,
+        use_auth_token: Optional[str] = None,
+    ) -> List[SpaceInfo]:
+        """
+        Get the public list of all Spaces on huggingface.co
+
+        Args:
+            filter `str` or `Iterable`, *optional*):
+                A string or which can be used to identify Spaces on the hub.
```
osanseviero, line 1173:
```suggestion
                A string tag or list of tags that can be used to identify Spaces on the Hub.
```

Maybe I would also mention it corresponds to the `tags` and `sdk` field in https://huggingface.co/spaces/evaluate-metric/precision/blob/main/README.md#L10

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+            author (`str`, *optional*):
+                A string which identify the author of the returned Spaces
+            search (`str`, *optional*):
+                A string that will be contained in the returned spaces.
```
osanseviero, line 1177:
```suggestion
                A string that will be contained in the returned Spaces.
```

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+        use_auth_token: Optional[str] = None,
+    ) -> List[SpaceInfo]:
+        """
+        Get the public list of all Spaces on huggingface.co
+
+        Args:
+            filter `str` or `Iterable`, *optional*):
+                A string or which can be used to identify Spaces on the hub.
+            author (`str`, *optional*):
+                A string which identify the author of the returned Spaces
```
osanseviero, line 1175:
```suggestion
                A string which identify the author of the returned Spaces.
```

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+            sort (`Literal["lastModified"]` or `str`, *optional*):
+                The key with which to sort the resulting Spaces. Possible
+                values are the properties of the `SpaceInfo` class.
+            direction (`Literal[-1]` or `int`, *optional*):
+                Direction in which to sort. The value `-1` sorts by descending
+                order while all other values sort by ascending order.
+            limit (`int`, *optional*):
+                The limit on the number of Spaces fetched. Leaving this option
+                to `None` fetches all Spaces.
+            cardData (`bool`, *optional*):
```
osanseviero, line 1187: I tried this but didn't give me anything new (no tags, etc). Is that expected @julien-c ?

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+                Whether to grab the metadata for the Space as well.
+            datasets (`bool` or `str` or `Iterable`, *optional*):
+                Whether to return Spaces that make use of a dataset.
+                The name of a specific dataset can be passed as a string.
+            models (`bool` or `str` or `Iterable`, *optional*):
+                Whether to return Spaces that make use of a model.
+                The name of a specific model can be passed as a string.
```
osanseviero, line 1194: As discussed we can remove the boolean option

julien-c:
hmm so @osanseviero made me realize i was mistaken, `datasets (boolean)` or `models (boolean)` are not supported

I only meant that if you pass models or datasets in your query (as string or array of string), then they will also be returned as fields in the API output.

Sorry about the incorrect information!

src/huggingface_hub/hf_api.py
```diff
@@ -1148,6 +1148,100 @@ def list_metrics(self) -> List[MetricInfo]:
...
+            linked (`bool` or `str` or `Iterable`, *optional*):
```
osanseviero, line 1195: linked is only boolean :)

src/huggingface_hub/hf_api.py
cakiki: Gotcha; I was mirroring the behavior of `list_datasets`; is it implemented there, or is that an error in the code that needs to be fixed there?

https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L1013-L1018

osanseviero: No, that code is ok because what they do is set `full` to `True` when there is `cardData`, but here it would not make sense I think. I would remove this parameter :)

osanseviero (approved):
Looks good! I left some comments on the tests but otherwise looks almost ready to merge

src/huggingface_hub/hf_api.py
```diff
@@ -1076,6 +1076,94 @@ def list_metrics(self) -> List[MetricInfo]:
...
+            params.update({"limit": limit})
+        if full is not None:
+            if full:
+                params.update({"full": True})
+        if linked is not None:
+            if linked:
+                params.update({"linked": True})
+        if datasets is not None:
+            if datasets:
+                params.update({"datasets": datasets})
```
osanseviero, line 1157: Why not

```suggestion
        if datasets is not None:
          params.update({"datasets": datasets})
```
? Same below

cakiki: I'm not sure tbh; I think I was emulating this part:

https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L1013-L1015

osanseviero:
Thank you @cakiki for the contribution!! :fire: :fire:

julien-c:
Yay it's very cool!! thanks

Issue #805, opened 2022-03-28
push_to_hub_keras API dose not support sequential GANs models. 


kingabzpro:
**Is your feature request related to a problem? Please describe.**
I face error when I run 
```python
push_to_hub_keras(model = GAN, repo_url = "https://huggingface.co/kingabzpro/CELEB-GANs")

ValueError: Expected a Trackable object for export, got <keras.engine.sequential.Sequential object at 0x7f84d97320d0>
```

**Describe the solution you'd like**
I think some sequential models dosen't support `tf.keras.models.save_model(GAN, "./CELEB-GANs/")`, instead we can use `GAN.save("model.h5")` 

**Describe alternatives you've considered**
The complete code is from [Kaggle notebook](https://www.kaggle.com/code/kingabzpro/fake-faces-with-dcgans?scriptVersionId=91480379). 

**For a Reference**
I have added .h5 model manually: https://huggingface.co/kingabzpro/CELEB-GANs/tree/main

kingabzpro:
Can you check again now? https://www.kaggle.com/code/kingabzpro/fake-faces-with-dcgans

PR #890, opened 2022-05-31
Fixes: #805 Added h5 save format compatibility

rushic24:
```
from huggingface_hub import push_to_hub_keras
push_to_hub_keras(transformer, repo_url = "rushic24/keras-english-to-spanish", include_optimizer=True, save_format='h5', filepath='mpath')
```

image

### DIFF
--- a/src/huggingface_hub/keras_mixin.py
+++ b/src/huggingface_hub/keras_mixin.py
@@ -233,6 +233,10 @@ def save_pretrained_keras(
             metadata["tags"] = [task_name]
 
     _create_model_card(model, save_directory, plot_model, metadata)
+    if model_save_kwargs.get('save_format')=='h5':
+        filepath = model_save_kwargs.pop('filepath', None)
+        if filepath:
+            save_directory = os.path.join(save_directory, filepath)
     tf.keras.models.save_model(
         model, save_directory, include_optimizer=include_optimizer, **model_save_kwargs
     )


rushic24:
@merveenoyan

adrinjalali (commented):
Thanks for the PR. Please add a test for the changes :)

rushic24:
> Thanks for the PR. Please add a test for the changes :)

Sure, can you just enable the workflows for once, I need to check if the tests pass in the github. They do pass locally.

rushic24:
@adrinjalali Is that normal? They do pass here.
tests

src/huggingface_hub/keras_mixin.py
```diff
@@ -233,6 +233,8 @@ def save_pretrained_keras(
             metadata["tags"] = [task_name]
 
     _create_model_card(model, save_directory, plot_model, metadata)
+    if model_save_kwargs.get("save_format") == "h5":
```
nateraw, line 236: Any reason why this would be a `model_save_kwarg` instead of a regular kwarg?

src/huggingface_hub/keras_mixin.py
```diff
@@ -530,6 +532,8 @@ def _from_pretrained(
             )
         else:
             storage_folder = model_id
+        if model_kwargs.get("custom_objects"):
```
nateraw, line 535: Same here with regards to this being a `model_kwarg` and not a regular kwarg. Also some docstring would help here.

nateraw (commented):
Hey there, thanks for looking into this and your contribution!

I think this is a tricky issue, however. I'm not sure we want to use H5 format at all if at all possible. I believe it's suggested to use `SavedModel` everywhere. If there are issues with `SavedModel` that H5 solves, maybe we try to see if we can solve them `SavedModel` first. 

I think this will actually add significant complexity in terms of usability, and will break in a variety of use cases. Imho I would avoid adding this feature altogether, but if we truly deem this necessary it will need some really good tests, docstrings, and example snippets to clear up its intended usage.

rushic24:
> Hey there, thanks for looking into this and your contribution!
> 
> I think this is a tricky issue, however. I'm not sure we want to use H5 format at all if at all possible. I believe it's suggested to use `SavedModel` everywhere. If there are issues with `SavedModel` that H5 solves, maybe we try to see if we can solve them `SavedModel` first.
> 
> I think this will actually add significant complexity in terms of usability, and will break in a variety of use cases. Imho I would avoid adding this feature altogether, but if we truly deem this necessary it will need some really good tests, docstrings, and example snippets to clear up its intended usage.

Hey, 
Thanks for looking into it. This is needed because some keras models don't save without 'h5'. For e.g. If I want to save this transformer https://github.com/rushic24/keras-io/blob/master/examples/nlp/neural_machine_translation_with_transformer.py#L355.

merveenoyan:
@rushic24 Hello Rushi, 
As far as I know it's quite discouraged to serialize models in the manner you mentioned. The models can be pushed to different repositories and then built locally in the manner you mentioned. When we save models as `SavedModel`, we have everything we need to serve a model with anything inside TFX ecosystem and that's why we support this format.

merveenoyan:
@rushic24 we appreciate contributions, thanks a lot for the input ❤️

nateraw:
Yea thanks again @rushic24 - feel free to discuss further in the issues if the solution above doesn't work for you

PR #891, opened 2022-06-01
`metadata_update`: work on a copy of the upstream file, to not mess up the cache

julien-c:
otherwise if you're using the same file from another script or library you'll have incorect file.

Side effect: With this change, we can probably remove the flag `force_download=True`

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -1051,8 +1051,9 @@ def hf_hub_download(
                 "We have no connection or you passed local_files_only, so"
                 " force_download is not an accepted option."
             )
-        commit_hash = revision
-        if not REGEX_COMMIT_HASH.match(revision):
+        if REGEX_COMMIT_HASH.match(revision):
+            commit_hash = revision
+        else:
             ref_path = os.path.join(storage_folder, "refs", revision)
             with open(ref_path) as f:
                 commit_hash = f.read()
--- a/src/huggingface_hub/repocard.py
+++ b/src/huggingface_hub/repocard.py
@@ -1,6 +1,8 @@
 import dataclasses
 import os
 import re
+import shutil
+import tempfile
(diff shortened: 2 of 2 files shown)


osanseviero (approved):
LGTM!

lvwerra (approved):
LGTM!

adrinjalali (changes requested):
Could you please write a test for this?

julien-c:
not sure what to test here TBH =)

adrinjalali:
> not sure what to test here TBH =)

- Download repo (which also caches it)
- change the metadata/work with it
- make sure the cached file hasn't changed

Isn't that what this PR enables?

julien-c:
Ok, makes sense. Want to take a stab at it @adrinjalali?

tests/test_repocard.py
```diff
@@ -240,6 +241,22 @@ def test_update_existing_result_with_overwrite(self):
...
+            f"{USER}/{self.REPO_NAME}",
+            filename=REPOCARD_NAME,
+            use_auth_token=self._token,
+        )
+
+        metadata_update(
+            f"{USER}/{self.REPO_NAME}", new_metadata, token=self._token, overwrite=True
+        )
+
+        self.assertNotEqual(metadata_load(path), new_metadata)
```
adrinjalali, line 258: can we also check the original file is not touched?

adrinjalali (commented):
Thanks for the tests @LysandreJik

tests/test_repocard.py
LysandreJik: Handled in [5fabdf8](https://github.com/huggingface/huggingface_hub/pull/891/commits/5fabdf86b27b75da6704d2c8dffb16c3814e2557)

adrinjalali (approved):
LGTM, thanks @LysandreJik

PR #893, opened 2022-06-02
CI enable codecov

adrinjalali:
This PR enables codecov reports on this repo.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -42,10 +42,10 @@ jobs:
       - run: |
           if [[ "${{ matrix.test_repository }}" == "Repository only" ]]
           then
-            pytest -Werror::FutureWarning --log-cli-level=INFO -sv ./tests/test_repository.py -k RepositoryTest
-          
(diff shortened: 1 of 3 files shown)


adrinjalali:
The files are being uploaded: 

https://app.codecov.io/gh/huggingface/huggingface_hub/pulls

This should be ready for a review/merge: @LysandreJik @osanseviero

LysandreJik (commented):
90,31%, that's quite nice, I didn't expect such a high number!

How can we do to have easy visibility on each PR? In the past we were working with comments, but I understand that's not your favored approach @adrinjalali.

adrinjalali:
A step is added to the CI list when it finishes, and it fails when it's not happy. We can start with that and then add the comments if we're not happy with how that works. WDYT?

LysandreJik:
Yes, that sounds good to me.

adrinjalali:
@osanseviero should we merge this one?

adrinjalali:
Let's see how it works and experiment.

PR #895, opened 2022-06-07
Display endpoint URL in whoami command

juliensimon:
This is a convenient way to check what hub endpoint you're logged on to (public hub, private hub, etc.):

```
% huggingface-cli whoami
https://private-hub.huggingface.cloud
juliensimon
```

The codebase has some informational messages with hard-coded references to huggingface.co. I've left these alone, maybe we should clean them up at some point.

### DIFF
--- a/src/huggingface_hub/commands/user.py
+++ b/src/huggingface_hub/commands/user.py
@@ -19,10 +19,10 @@ from typing import List, Union
 
 from huggingface_hub.commands import BaseHuggingfaceCLICommand
 from huggingface_hub.constants import (
+    ENDPOINT,
     REPO_TYPES,
     REPO_TYPES_URL_PREFIXES,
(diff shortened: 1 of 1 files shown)


LysandreJik:
I'm offering to only print it when it's different to the default one, as otherwise it may be a bit weird:
```
% huggingface-cli whoami
https://huggingface.co
juliensimon
```

My first reaction when seeing this is wondering why the devs left a random print statement in the code :)

juliensimon:
That's fine with me :)

LysandreJik:
I've opened a PR on your PR here to move the print statement as I was mentioning in my comment above: https://github.com/juliensimon/huggingface_hub/pull/1

Let me know if this works for you, I find it clearer this way but open to different opinions :)

LysandreJik (approved):
Thanks!

PR #900, opened 2022-06-10
FIX Avoid creating repository when it exists on remote

merveenoyan:
This PR partially fixes https://github.com/huggingface/huggingface_hub/issues/672. 

I observed that the following error pops up when I create a repository using huggingface API `create_repo` and then have a local repository cloning from the URL **using a token** ([here](https://github.com/huggingface/huggingface_hub/issues/672#issuecomment-1150972003))
`ValueError: No space_sdk provided. create_repo expects space_sdk to be one of ['gradio', 'streamlit', 'static'] when repo_type is 'space'`
So I noticed the following block was causing the problem:

```
if token is not None:
                whoami_info = self.client.whoami(token)
                user = whoami_info["name"]
                valid_organisations = [org["name"] for org in whoami_info["orgs"]]

                if namespace is not None:
                    repo_id = f"{namespace}/{repo_id}"
                repo_url += repo_id

                scheme = urlparse(repo_url).scheme
                repo_url = repo_url.replace(f"{scheme}://", f"{scheme}://user:{token}@")

                    self.client.create_repo(
                        repo_id=repo_id,
                        token=token,
                        repo_type=self.repo_type,
                        exist_ok=True,
                        private=self.private,
                    )
            else:
                if namespace is not None:
                    repo_url += f"{namespace}/"
                repo_url += repo_id
```

Only name being valid shouldn't be enough to create a repository, we have to check if the repository exists on remote so I added a small check on that (thanks @muellerzr for pointing out to particular HfApi function that does it).

With this, `space_sdk` error should be gone. I'll add test once @LysandreJik approves my logic in here.

Also this is separate but IDK if it's good to have an attribute `clone_from` and the `clone_from()` function. I can refactor couple of things I saw there that's not good to me as well.

### DIFF
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -5,7 +5,6 @@ import subprocess
 import tempfile
 import threading
 import time
-import warnings
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
@@ -648,10 +647,12 @@ class Repository:
                     try:
                         _ = HfApi().repo_info(f"{namespace}/{repo_id}")
                     except HTTPError:
-                        warnings.warn(
-                            "There's no repository in the given link in clone_from."
-                            " Creating a new repository.",
-                            UserWarning,
+                        self.client.create_repo(
+                            repo_id=repo_id,
+                            token=token,
+                            repo_type=self.repo_type,
+                            exist_ok=True,
+                            private=self.private,
                         )
 
             else:
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -648,13 +648,20 @@ class Repository:
                             f"{repo_id}", repo_type=self.repo_type, token=token
                         )
                     except HTTPError:
-                        self.client.create_repo(
-                            repo_id=repo_id,
-                            token=token,
-                            repo_type=self.repo_type,
-                            exist_ok=True,
-                            private=self.private,
-                        )
+                        if self.repo_type is "space":
+                            raise ValueError(
+                                "Creating a Space through passing Space link to"
+                                " clone_from is not allowed. Make sure the Space exists"
+                                " on Hugging Face Hub."
+                            )
+                        else:
+                            self.client.create_repo(
+                                repo_id=repo_id,
+                                token=token,
+                                repo_type=self.repo_type,
+                                exist_ok=True,
+                                private=self.private,
+                            )
 
             else:
                 if namespace is not None:
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -224,7 +224,7 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
     """
     try:
         with open(filename, "rb") as f:
-            content = f.read(10 * (1024 ** 2))  # Read a maximum of 10MB
+            content = f.read(10 * (1024**2))  # Read a maximum of 10MB
 
         # Code sample taken from the following stack overflow thread
         # https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python/7392391#7392391
@@ -684,7 +684,8 @@ class Repository:
                         env.update({"GIT_LFS_SKIP_SMUDGE": "1"})
 
                     run_subprocess(
-                        f"{'git clone' if self.skip_lfs_files else 'git lfs clone'} {repo_url} .".split(),
+                        f"{'git clone' if self.skip_lfs_files else 'git lfs clone'} {repo_url} ."
+                        .split(),
                         self.local_dir,
                         env=env,
                     )
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -224,7 +224,7 @@ def is_binary_file(filename: Union[str, Path]) -> bool:
     """
     try:
         with open(filename, "rb") as f:
-            content = f.read(10 * (1024**2))  # Read a maximum of 10MB
+            content = f.read(10 * (1024 ** 2))  # Read a maximum of 10MB
 
         # Code sample taken from the following stack overflow thread
         # https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python/7392391#7392391
@@ -684,8 +684,7 @@ class Repository:
                         env.update({"GIT_LFS_SKIP_SMUDGE": "1"})
 
                     run_subprocess(
-                        f"{'git clone' if self.skip_lfs_files else 'git lfs clone'} {repo_url} ."
-                        .split(),
+                        f"{'git clone' if self.skip_lfs_files else 'git lfs clone'} {repo_url} .".split(),
                         self.local_dir,
                         env=env,
                     )
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -645,7 +645,7 @@ class Repository:
 
                 if namespace == user or namespace in valid_organisations:
                     try:
-                        _ = HfApi().repository_info(f"{namespace}/{repo_id}")
+                        _ = HfApi().repo_info(f"{namespace}/{repo_id}")
                     except HTTPError:
                         self.client.create_repo(
                             repo_id=repo_id,
--- a/src/huggingface_hub/repository.py
+++ b/src/huggingface_hub/repository.py
@@ -642,7 +642,6 @@ class Repository:
 
                 scheme = urlparse(repo_url).scheme
                 repo_url = repo_url.replace(f"{scheme}://", f"{scheme}://user:{token}@")
-                breakpoint()
                 if namespace == user or namespace in valid_organisations:
                     try:
                         _ = HfApi().repo_info(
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/repository.py
```diff
@@ -643,13 +644,17 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
...
-                        repo_id=repo_id,
-                        token=token,
-                        repo_type=self.repo_type,
-                        exist_ok=True,
-                        private=self.private,
-                    )
+                    try:
+                        _ = HfApi().repo_info(f"{namespace}/{repo_id}")
+                    except HTTPError:
+                        self.client.create_repo(
```
adrinjalali, line 650: I'm very confused, why would `clone_from` create a repo remotely under any circumstances?

adrinjalali (commented):
Also tests? :D

src/huggingface_hub/repository.py
merveenoyan: It's because link given to `clone_from` might not be valid. Maybe it's better if I put a warning saying the remote repository doesn't exist instead. :)

merveenoyan: or do we want to not create at all and just raise a warning?

merveenoyan: I removed repo creation and added a warning.

merveenoyan: AH OK

merveenoyan: I just realized there are many tests that will fail and we'll break backwards compatibility if we change behavior of `clone_from`, I also thought that it was only used to clone existing repository when creating repository on local. pinging @LysandreJik about why the design is that way 😅

merveenoyan:
_Super weird thing_
I can do following:
```
api = HfApi()

temp_repo_url = api.create_repo(
            repo_id="merve/temp", token=token, 
            repo_type="space", space_sdk="gradio"
        )
repo = Repository("./dummy", clone_from=temp_repo_url,
        use_auth_token=token
        )
```

But when I run the test I've written (below) I still get the same error. 
```
def test_init_clone_from(self):
        temp_repo_url = self._api.create_repo(
            repo_id=f"{self.REPO_NAME}-temp", token=self._token, 
            repo_type="space", space_sdk="gradio"
        )
        repo = Repository(WORKING_REPO_DIR, clone_from=temp_repo_url,
        use_auth_token=self._token
        )
        self._api.delete_repo(
                repo_id=f"valid_org/{self.REPO_NAME}", token=self._token
            )
```

I have other stuff to do so will leave this for monday.

src/huggingface_hub/repository.py
merveenoyan: (ok apparently you left a comment but I didn't see it bcs I didn't refresh the page lol)

merveenoyan:
@osanseviero yup I'm sure (I uninstalled and installed again)
What I'm testing here is that it should bypass creating the repository remotely given there's a repository remotely already (which we check with `HfApi.repo_info()`)

src/huggingface_hub/repository.py
julien-c: (i agree that Repository clone_from shouldn't create a repo)

julien-c: (can't find at the moment the past discussion with a potential explicit `Repository.create() -> Repository`)

LysandreJik: Yes, I also think that `clone_from` shouldn't create the repository if it doesn't exist, it isn't really its place to do so. I wouldn't say removing it is a bugfix as when I implemented it I shared it as such, but we can deprecate it for sure.

merveenoyan:
@osanseviero it passed 😏

osanseviero:
This seems to be passing the tests now 😄

src/huggingface_hub/repository.py
merveenoyan: I say we merge this and then deprecate it.

adrinjalali (approved):
I'm happy to merge and deprecate in a separate PR.

src/huggingface_hub/repository.py
LysandreJik: SGTM!

src/huggingface_hub/repository.py
```diff
@@ -641,15 +642,20 @@ def clone_from(self, repo_url: str, use_auth_token: Union[bool, str, None] = Non
...
+                        _ = HfApi().repo_info(
+                            f"{repo_id}", repo_type=self.repo_type, token=token
+                        )
+                    except HTTPError:
+                        self.client.create_repo(
+                            repo_id=repo_id,
+                            token=token,
+                            repo_type=self.repo_type,
+                            exist_ok=True,
+                            private=self.private,
```
LysandreJik, line 656: Wouldn't this also need a `space_sdk` in case the repository in question is a space? i.e., if the repo is a space that doesn't exist, it'll try to create it but fail at doing so.

As we're deprecating, maybe that's intended and we keep that use-case as it is (i.e., not working)

LysandreJik (commented):
Looks good to me! Only left a comment regarding `space_sdk`.

src/huggingface_hub/repository.py
merveenoyan: I will raise a `ValueError` in case someone tries to create Space that doesn't exist because otherwise I have to add `space_sdk` as a class attribute to `Repository` (we're later deprecating the repository creation in `clone_from` anyway so there's no point to add that). I'll add two tests, one for model repo and one for Space repo to see if the error is raised.  We can merge tomorrow ^^

LysandreJik: Sounds great, thank you @merveenoyan!

merveenoyan:
Please do not review I know what the problem with tests is :D

merveenoyan:
I added three tests:
- `test_init_clone_from()` checks that when space exists it's clone well
- `test_clone_from_space()` makes sure the space is not created if it doesn't exist
- `test_clone_from_model()` makes sure that model is created if it doesn't exist.

merveenoyan:
One of the tests are not passing because moon-landing staging is down 😅

LysandreJik:
@merveenoyan it seems you're using an outdated version of `black`. Could you updated to the one required by `huggingface_hub` and run the code quality fix? Should be something like the following, from the root of your clone:

```
pip install -U -e .[quality]
make style
make quality
```

@SBrandeis, @XciD, would you have why we're getting errors 500 on the `test_init_clone_from` test? It seems like creating a space using `create_repo` doesn't work and throws an error 500. Reproducible code example:

```py
temp_repo_url = create_repo(
    repo_id=f"temp",
    token=token,
    repo_type="space",
    space_sdk="streamlit"
)
```

I tried with both `streamlit`, `gradio`. However, `static` seems to work just fine!

christophe-rannou:
@LysandreJik I just created a repo using the CLI with success. Maybe there's an issue with your token ?

LysandreJik:
@christophe-rannou the token is that of the CI, we're using the same token for all CI. Were you on the production endpoint or the staging endpoint?

christophe-rannou:
Production one, it's the staging one that is not working ?

XciD:
> Production one, it's the staging one that is not working ?

Seems related to CI, will investigate.

merveenoyan:
This one fails `HfApiEndpointsTest.test_create_update_and_delete_space_repo` with 500 like it was failing for my PR, see below where the error comes from (due to picking other SDKs for others)

```
for sdk in SPACES_SDK_TYPES:
            SPACE_REPO_NAME = space_repo_name(sdk)
            self._api.create_repo(
                repo_id=SPACE_REPO_NAME,
                token=self._token,
                repo_type=REPO_TYPE_SPACE,
                space_sdk=sdk,
            )
            res = self._api.update_repo_visibility(
                repo_id=SPACE_REPO_NAME,
                token=self._token,
                private=True,
                repo_type=REPO_TYPE_SPACE,
            )
            self.assertTrue(res["private"])
            res = self._api.update_repo_visibility(
                repo_id=SPACE_REPO_NAME,
                token=self._token,
                private=False,
                repo_type=REPO_TYPE_SPACE,
            )
```

LysandreJik:
This exact test fails on `main` since https://github.com/huggingface/huggingface_hub/pull/898 was merged, I would offer to skip that test until this is resolved, and merge this PR. Ok for you @merveenoyan ?

merveenoyan:
@LysandreJik sure!

osanseviero:
Was this supposed to be merged? As @LysandreJik mentioned, I think we should add the flag/decorator to skip the test so the tests are not failing (` @unittest.SkipTest` for example)

LysandreJik:
No problem @osanseviero, I opened a PR to fix the problem there: https://github.com/huggingface/huggingface_hub/pull/908

But now we have another problem with our CI because of the new datasets release, so waiting until that is fixed to merge it.

XciD:
About the spaces bug, we found it. 
Waiting for https://github.com/huggingface/moon-landing/pull/3252/commits/f2e8068348e225ba1f07c5d27a38843253e23dcb to be merged.

XciD:
> About the spaces bug, we found it. Waiting for [huggingface/moon-landing@f2e8068](https://github.com/huggingface/moon-landing/commit/f2e8068348e225ba1f07c5d27a38843253e23dcb) to be merged.

Fixed

PR #904, opened 2022-06-13
Test that TensorFlow is not imported on startup

lhoestq:
TF takes some time to be imported, and also uses some GPU memory.

I just added a test to make sure that in the future it's never imported by default when
```python
import huggingface_hub
```
is called

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,0 +1,7 @@
+import sys
+
+
+def test_import_huggingface_hub_doesnt_import_tensorfow():
+    import huggingface_hub  # noqa
+
+    assert "tensorflow" not in sys.modules
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,7 +1,26 @@
 import sys
+import unittest
+
+from huggingface_hub.file_download import is_tf_available
 
 
+def require_tf(test_case):
+    """
+    Decorator marking a test that requires TensorFlow, graphviz and pydot.
+
+    These tests are skipped when TensorFlow, graphviz and pydot are installed.
+
+    """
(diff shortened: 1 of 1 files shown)


adrinjalali:
Since we do lazy loading now in `huggingface_hub`, [almost] nothing should be imported on `import huggingface_hub` anyway. But I like the test.

lhoestq:
Yup I noticed that :) Looking forward to have a new release with lazy loading !

This test is just to make sure we make sure we don't do it without noticing, as it was done in https://github.com/huggingface/huggingface_hub/pull/679

osanseviero (approved):
LGTM, thanks for the test!

tests/test_import.py
```diff
@@ -0,0 +1,26 @@
...
+    These tests are skipped when TensorFlow is not installed.
+
+    """
+    if not is_tf_available():
+        return unittest.skip("test requires Tensorflow")(test_case)
+    else:
+        return test_case
+
+
+@require_tf
```
osanseviero, line 20: Actually, will this test ever be executed? If it uses `require_tf`, i will never run. The only test that builds tensorflow is ` pytest --cov -Werror::FutureWarning -sv ./tests/test_keras*` at https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/python-tests.yml. I think we need to remove the decorator?

lhoestq:
I ended up adding `pytest ./tests/test_tf*`.

This is because running the keras tests at the same time would import TF, so the test would always fail.

Let me know if it's good for you

osanseviero:
Yes, this works for me!

PR #910, opened 2022-06-15
Remove deprecations

LysandreJik:
This PR removes deprecations that should be removed with v0.8.0.

### DIFF
--- a/src/huggingface_hub/README.md
+++ b/src/huggingface_hub/README.md
@@ -97,9 +97,9 @@ programmatic way of creating a repo, deleting it (`⚠️ caution`), pushing a
 single file to a repo or listing models from the Hub, you'll find helpers in
 `hf_api.py`. Some example functionality available with the `HfApi` class:
 
-* `login()`
+* set_access_token()
+* unset_access_token()
 * `whoami()`
-* `logout()`
 * `create_repo()`
 * `list_repo_files()`
 * `list_repo_objects()`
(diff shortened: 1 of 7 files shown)


src/huggingface_hub/README.md
```diff
@@ -97,9 +97,9 @@ programmatic way of creating a repo, deleting it (`⚠️ caution`), pushing a
 single file to a repo or listing models from the Hub, you'll find helpers in
 `hf_api.py`. Some example functionality available with the `HfApi` class:
 
-* `login()`
+* set_access_token()
+* unset_access_token()
```
julien-c, line 101:
```suggestion
* `set_access_token()`
* `unset_access_token()`
```

julien-c (commented):
looks good to me!

src/huggingface_hub/commands/user.py
```diff
@@ -366,35 +335,9 @@ def login_token_event(t):
 
     token_finish_button.on_click(login_token_event)
 
-    def login_password_event(t):
```
osanseviero, line 369: I love when code is deleted 🔥

tests/test_hf_api.py
```diff
@@ -109,37 +106,36 @@ class HfApiCommonTest(unittest.TestCase):
 
 class HfApiLoginTest(HfApiCommonTest):
     def setUp(self) -> None:
-        erase_from_credential_store(USER)
+        erase_from_credential_store(USERNAME_PLACEHOLDER)
```
osanseviero, line 109: Why is this change needed?

osanseviero (approved):
Very nice PR 🔥 

From what I recall and quick check all downstream libraries should have pinned versions so it should not break anything, but let's keep an eye for things that could potentially break

tests/test_hf_api.py
LysandreJik: Because beforehand we were saving the username/password in the credential cache, and that's not the case anymore!

Now we're saving `hf_user:token` instead, with `USERNAME_PLACEHOLDER` being `hf_user`.

PR #912, opened 2022-06-16
Invert deprecation for create_repo

LysandreJik:
Unfortunately, this breaking change was not handled correctly in `transformers` and would have broken all existing `push_to_hub` methods. This reverts the deprecation for this method only, and sets it until 0.10.0


sgugger (approved):
LGTM!

PR #913, opened 2022-06-16
Constant was accidentally removed during deprecation transition

lyle-nel:
This constant was available before the deprecation notice, so it should be available during the warning transition period.
Note that the sentence-transformers library makes use of this constant [link to code](https://github.com/UKPLab/sentence-transformers/blob/3433c001e5a9e0991ea09e85f8861f38eb0d6583/sentence_transformers/util.py#L407)

### DIFF
--- a/src/huggingface_hub/snapshot_download.py
+++ b/src/huggingface_hub/snapshot_download.py
@@ -12,3 +12,4 @@ warnings.warn(
 )
 
 from ._snapshot_download import *  # noqa
+from .constants import REPO_ID_SEPARATOR  # noqa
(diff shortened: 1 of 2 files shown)


lyle-nel:
@LysandreJik I can't account for why the coverage checks would fail. The test clearly covers the bug that we are fixing here. Any thoughts?

osanseviero (approved):
This PR Looks good! Thanks for the fix! :fire: 

Note that when importing from snapshot_download you get the following warning:

```
/home/osanseviero/miniconda3/lib/python3.9/site-packages/huggingface_hub/snapshot_download.py:6: FutureWarning: snapshot_download.py has been made private and will no longer be available from version 0.11. Please use `from huggingface_hub import snapshot_download` to import the only public function in this module. 
```

so by version 0.11 this variable will be removed. `sentence-transformers` should likely also pin the version of `huggingface_hub` it's using.

PR #914, opened 2022-06-17
[Keras Mixin] - Flattening out nested configurations for better table parsing.

### LINKED ISSUE #911: Hyperparameters should be parsed recursively for Keras mixin
In this [model card](https://huggingface.co/keras-io/SimSiam) I realized that the learning rate scheduler break the table parsing (I didn't observe this case for other learning rates 😅).
The ideal table looks like the [one in this model card](https://huggingface.co/keras-io/ProbabalisticBayesianModel-Wine).
Would you like to take a look? 👀  @ariG23498

ariG23498:
Fixes #911

CC: @merveenoyan


merveenoyan:
Thank you so much for owning this! Is it okay if you could post a before + after model cards in both the non-nested case and nested case?

ariG23498:
@merveenoyan @nateraw I have opened this draft PR up for review!

Should we first merge #917?

merveenoyan:
@ariG23498 kindly pinging @nateraw

src/huggingface_hub/keras_mixin.py
```diff
@@ -26,10 +27,27 @@
     import tensorflow as tf
 
 
+def flatten_dict(dictionary, parent_key="", sep="."):
```
nateraw, line 30: Parameter descriptions would be nice here. Also, if we aren't using `parent_key` or `sep`, should we drop those params?

nateraw: also is there a risk here that something within this dict might not be serializable? Just curious.

nateraw (commented):
Seems ok to me! I would add docstrings as we go here, though...don't like to have functions without docstrings piling up here. 😄 Once you do that, it LGTM

Thanks for the contribution!

src/huggingface_hub/keras_mixin.py
ariG23498: > Parameter descriptions would be nice here. Also, if we aren't using parent_key or sep, should we drop those params?

I think if we decide on using `.` as the seperator, we can drop the `sep` parameter. We cannot drop the `parent_key` as this forms the basis of recursing over the parent keys and joining it with child keys in the nested dictionary.

ariG23498: > also is there a risk here that something within this dict might not be serializable? Just curious.

IMO we should not have anything un-serializable as this was packed inside the `model.optimizer` as the configuration. But I might be wrong here as well.

ariG23498:
@merveenoyan @nateraw I have taken care of the suggested changes.

The card with the current changes look like this:
image

nateraw (approved):
LGTM

PR #922, opened 2022-06-22
🐛 Fix PR creation for a repo the user does not own

SBrandeis:
There is a bug preventing anyone to create a Pull Request with `upload_file`. This bug has two parts:

- `fetch_upload_modes` (pre-upload call) was missing the `?create_pr=1` query parameter when requesting the pre-upload edpoint, resulting in a 403 HTTP error
-  `post_lfs_batch_info` would specify `body.ref.name` when `revision` is passed, resulting in an HTTPError 403 when trying to open a PR

This pull request addresses those two points

cc: @lewtun @abhishekkrthakur 

cc @coyotte508 : not sure `?create_pr=1` should be required for the pre-upload endpoint, wdyt?


src/huggingface_hub/_commit_api.py
```diff
@@ -173,7 +173,6 @@ def upload_lfs_files(
     repo_type: str,
     repo_id: str,
     token: str,
-    revision: str,
```
osanseviero, line 176: What was the use of this?

If we remove it, we would break backwards compatibility. I would add a deprecation warning that it will be removed in 0.11

SBrandeis: This particular method is private so I think we can change it however we want (am I correct?)

osanseviero: I didn't realize this was internal, then it SG!

osanseviero (approved):

Wauplin (approved):

PR #940, opened 2022-07-06
Add additional repo card utils from `modelcards` repo

nateraw:
This PR aims to add repo card utilities built out in [nateraw/modelcards](https://github.com/nateraw/modelcards) to this repo. Note that there are some lingering [issues](https://github.com/nateraw/modelcards/issues) we may want to resolve and update both there and here before merging.

Some todos:

- [x] Update existing repo card utils to use new dataclasses + ensure backwards compatibility
- [x] Add default model card and include it in package here
- [x] Do not merge until default model card is updated to latest spec we defined offline
- [x] Add tests from `modelcards` to this repo
  - [x] Add additional tests for private repos
  - [x] (maybe?) Add additional tests that support alternative base endpoint (i.e. staging env)
- [x] Check on min/max versions of Jinja2 and update accordingly in `setup.py`
- [x] Check all docstrings work as intended (original repo tested with doctests, I don't think we do that here?)
- [x] Add section to docs that explains usage

Latest walkthrough of this PR in Colab: [Open In Colab](https://colab.research.google.com/gist/nateraw/c5ba361def8fa7c949af851ae2acc7e8/modelcard-in-huggingface-hub-demo.ipynb)


nateraw:
This PR is ready for a first round of reviews. The docstrings may be out of date a bit (which is why I'll leave this as draft), but the tests are all in. The example colab in my message above shows the usage. 

I don't want to add anything in docs for now in case there are big changes from review I need to make.

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,288 @@
...
+    def save(self, filepath: Union[Path, str]):
+        r"""Save a RepoCard to a file.
+
+        Args:
+            filepath (`Union[Path, str]`): Filepath to the markdown file to save.
+
+        Example:
+            >>> from huggingface_hub import RepoCard
+            >>> card = RepoCard("---\nlanguage: en\n---\n# This is a test repo card")
+            >>> card.save("/tmp/test.md")
```
LysandreJik, line 86:
```suggestion
        Example:
            ```python
            >>> from huggingface_hub import RepoCard
            >>> card = RepoCard("---\nlanguage: en\n---\n# This is a test repo card")
            >>> card.save("/tmp/test.md")
            ```
```

LysandreJik: (please apply the same to all code examples)

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,288 @@
...
+    ):
+        """Push a RepoCard to a Hugging Face Hub repo.
+
+        Args:
+            repo_id (`str`):
+                The repo ID of the Hugging Face Hub repo to push to. Example: "nateraw/food".
+            token (`str`, *optional*):
+                Authentication token, obtained with `huggingface_hub.HfApi.login` method. Will default to
+                the stored token.
+            repo_type (`str`, *optional*):
```
LysandreJik, line 169:
```suggestion
            repo_type (`str`, *optional*, defaults to `"model"`):
```

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+        metrics: Optional[Union[str, List[str]]] = None,
+        eval_results: Optional[List[EvalResult]] = None,
+        model_name: Optional[str] = None,
+        **kwargs,
+    ):
+        """Model Card Metadata that is used by Hugging Face Hub when included at the top of your README.md
+
+        Args:
+            language (`Union[str, List[str]]`, *optional*):
+                Language of model's training data or metadata. Example: `'en'`. Defaults to `None`.
```
LysandreJik, line 95: Maybe specify these are `ISO 639-1` language codes

tests/test_modelcard.py
```diff
@@ -0,0 +1,199 @@
+import logging
```
LysandreJik, line 1: Please align with other tests in the library by putting all test in a class that inherits from `unittest.TestCase`, and using `self.assertXXX` instead of `assert` statements.

setup.py
```diff
@@ -19,6 +19,7 @@ def get_version() -> str:
     "typing-extensions>=3.7.4.3",  # to be able to import TypeAlias
     "importlib_metadata;python_version<'3.8'",
     "packaging>=20.9",
+    "Jinja2",
```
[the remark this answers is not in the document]
adrinjalali: I'd say jinja templates are what we should use for modelcards:
- Users can define their own template and pass it to these tools, and it's a well documented format which can easily be extended.
- we're adding html and js code in the modelcard, and HTML is not a [regular language](https://en.wikipedia.org/wiki/Regular_language) and therefore [cannot be parsed with a regex](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454)

But I completely agree that jinja is probably not the best dependency for hfh, and that's partly why I've advocated for modelcards to be its own project/repo rather than being inside hfh.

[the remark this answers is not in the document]
osanseviero: Friendly ping on this to reach a final consensus on this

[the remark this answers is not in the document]
julien-c: let's add it as optional dependency i.e. only install it when doing it `pip install -e .[cards]` or equivalent?

src/huggingface_hub/repocard_data.py
julien-c: technically the backend validation says this:

> It must be an ISO 639-1, 639-2 or 639-3 code (two/three letters), or a special value like "code", "multilingual"

julien-c (commented):
Played with it a little bit, I really like it! 

This looks to be in good shape.

Aside from small comments, I wonder why you didn't keep the notion of `ModelCard` inheriting from `RepoCard` (which applies to all repo types) which i quite liked in https://github.com/nateraw/modelcards ?
And same for `CardData`, IMO we should have a repo-type-agnostic implem of `CardData` and then children dataclasses that define the relevant fields

My other question was, do we really want to bundle the (or a?) model card template in the library, or should we host it on the Hub and load the template from there too. 
I think there might be advantages and drawbacks to both approaches.

setup.py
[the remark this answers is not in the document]
LysandreJik: Yes, if adding jinja2 to the `extras_require` is an option, then I'd be more than happy to have this merged in. This extra requirement can be required by the downstream libraries that install `huggingface_hub` in order to manage their model cards, so it's a good compromise for me.

tests/test_modelcard.py
nateraw: :( It looks so much cleaner this way though, imo. But okay 😛

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+    # The arguments passed during `Metric.compute()`.
+    # Example for `bleu`: max_order: 4
+    metric_args: Optional[Dict[str, Any]] = None
+
+    # If true, indicates that evaluation was generated by Hugging Face (vs. self-reported).
+    verified: Optional[bool] = None
+
+
+@dataclass
+class CardData:
```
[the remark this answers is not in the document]
nateraw: I did give that a go here the other day: https://github.com/nateraw/modelcards/pull/57

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+    def __init__(
+        self,
+        language: Optional[Union[str, List[str]]] = None,
```
osanseviero, line 81: WDYT of forcing users to pass all the parameters by keyword. It will make it easier to keep this up to date and there will be less issues with wrong order of params. E.g. a la https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L1428-L1438

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+        self.metrics = metrics
+        self.eval_results = eval_results
+        self.model_name = model_name
+        self.__dict__.update(kwargs)
+
+        if self.eval_results:
+            if type(self.eval_results) == EvalResult:
+                self.eval_results = [self.eval_results]
+            if self.model_name is None:
+                raise ValueError("`eval_results` requires `model_name` to be set.")
```
osanseviero, line 149:
```suggestion
                raise ValueError("Passing `eval_results` requires `model_name` to be set.")
```

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def __repr__(self):
+        return self.to_yaml()
+
+
+def model_index_to_eval_results(model_index: List[Dict[str, Any]]):
+    """Takes in a model index and returns a list of `huggingface_hub.EvalResult` objects.
+
+    A detailed spec of the model index can be found here:
+    https://github.com/huggingface/hub-docs/blob/main/modelcard.md
```
osanseviero, line 181: Maybe https://github.com/huggingface/hub-docs/blob/main/modelcard.md?plain=1 is a better link

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+
+    Args:
+        model_index (`List[Dict[str, Any]]`):
+        A model index data structure, likely coming from a README.md file on the
+        Hugging Face Hub.
+
+    Returns:
+        - model_name (`str`):
```
osanseviero, line 189: I'm not sure this format works with the doc builder

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+    def to_yaml(self):
+        """Dumps CardData to a YAML block for inclusion in a README.md file."""
+        return yaml.dump(self.to_dict(), sort_keys=False).strip()
+
+    def __repr__(self):
+        return self.to_yaml()
+
+
+def model_index_to_eval_results(model_index: List[Dict[str, Any]]):
+    """Takes in a model index and returns a list of `huggingface_hub.EvalResult` objects.
```
osanseviero, line 178: This is actually returning a tuple of a string and a list iiuc

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+
+    Example:
+        >>> from huggingface_hub.repocard_data import model_index_to_eval_results
+        >>> # Define a minimal model index
+        >>> model_index = [
+        ...     {
+        ...         "name": "my-cool-model",
+        ...         "results": [
+        ...         {
+        ...             "task": {
```
osanseviero, line 204: Nit: the formatting is a bit broken across this. Some indents are missing

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+        ...             {
+        ...                 "type": "accuracy",
+        ...                 "value": 0.9
+        ...             }
+        ...             ]
+        ...         }
+        ...         ]
+        ...     }
+        ... ]
+        >>> model_name, eval_results = model_index_to_eval_results(model_index)
```
osanseviero, line 221: Do we really need to return the `model_name`? I would have expected this to return the `eval_results` only.

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,369 @@
...
+
+    # Metrics are reported on a unique task-and-dataset basis.
+    # Here, we make a map of those pairs and the associated EvalResults.
+    task_and_ds_types_map = dict()
+    for eval_result in eval_results:
+        task_and_ds_pair = (eval_result.task_type, eval_result.dataset_type)
+        if task_and_ds_pair in task_and_ds_types_map:
+            task_and_ds_types_map[task_and_ds_pair].append(eval_result)
+        else:
+            task_and_ds_types_map[task_and_ds_pair] = [eval_result]
```
osanseviero, line 329: Alternatively

```suggestion
    task_and_ds_types_map = defaultdict(list)
    for eval_result in eval_results:
        task_and_ds_pair = (eval_result.task_type, eval_result.dataset_type)
        ask_and_ds_types_map[task_and_ds_pair].append(eval_result)
```

src/huggingface_hub/repocard_types.py
```diff
@@ -1,105 +0,0 @@
-from dataclasses import dataclass, field
```
osanseviero, line 1: Bye bye code :wave:

osanseviero (commented):
Thanks for this PR! It's very cool! :fire: I left some questions and high-level suggestions, but this is in a great state already!

* First of all, thanks for sharing the colab! It makes it super easy and nice to play with the API very quickly. I think a lot of it can be converted to some nice docs :mechanical_arm: 
* I agree with @julien-c on having different classes for each type of repo card, with `ModelCard`, `SpacesCard`, etc.
* I think having the main template bundled in the library is nice, so I would keep as is. If people want to provide new templated in the future we could have some additional functionality to use templates from the Hub, which might motivate people to create their own templates, but I think for now focusing in the current one is better!
* I didn't review the tests yet as they are probably going to change a bit :) 

Thanks again! :rocket:

src/huggingface_hub/repocard_data.py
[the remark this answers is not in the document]
nateraw: I believe separating them all out makes it harder to read, but "better" structurally. I think its a little too complicated for what's really going on, but if you folks prefer going that direction, I have no problem updating to do that here 😄

src/huggingface_hub/repocard_data.py
nateraw: all the docstrings are a bit screwed up atm.

src/huggingface_hub/repocard_data.py
nateraw: Model name is required in the eval results.

src/huggingface_hub/repocard_data.py
nateraw: hmm will give this one a go, should work.

src/huggingface_hub/repocard_types.py
nateraw: hehe

tests/test_modelcard.py
LysandreJik: Cleanliness is a subjective point of view while conventions are not :wink:

setup.py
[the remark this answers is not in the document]
nateraw: Opted to just put it in `testing` requirements. Since Jinja is only required by `from_template` and its only one dep, I figure its a bit overkill to add a whole new extras category.

src/huggingface_hub/repocard.py
nateraw: is it ok if I add a new line before the closing backticks? this way doctests still run as expected (even though they aren't being used in our CI)

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,288 @@
...
+            card_path = hf_hub_download(
+                repo_id_or_path,
+                REPOCARD_NAME,
+                repo_type=repo_type,
+                use_auth_token=token,
+            )
+
+        return cls(Path(card_path).read_text())
+
+    def validate(self, repo_type="model"):
```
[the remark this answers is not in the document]
nateraw: I actually find it fairly useful. Also, by leaving it public we can document it in reference docs so users understand what's going on if something fails.

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,288 @@
...
+        """Validates card against Hugging Face Hub's model card validation logic.
+        Using this function requires access to the internet, so it is only called
+        internally by `huggingface_hub.ModelCard.push_to_hub`.
```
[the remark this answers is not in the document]
nateraw: Had this discussion w/ Adrin in original PR in `modelcards`, we opted to have it not return anything. It errors out if theres an issue, so that's how we're communicating it vs. returning `True`/`False`.

nateraw:
Ok folks, I've:

- Updated tests to `unittest.TestCase` instead of `pytest`
- Refactored to have `Repocard` be parent of `ModelCard` and `DatasetCard`
- Refactored to have `CardData` be parent of `ModelCardData` and `DatasetCardData`
- Updated to have Jinja2 not be installed by default, only as extra
- Added docs (both a guide + reference documentation)

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
+# Creating and Sharing Model Cards
+
+The `huggingface_hub` library provides a Python interface to create, share, and update Model Cards.
+Visit [the dedicated documentation page](https://huggingface.co/docs/hub/models-cards)
+for a deeper view of what Model Cards on the Hub are, and how they work under the hood.
+
+## Loading a Model Card from the Hub
+
+To load an existing card from the hub, you can use the [`ModelCard.load`] function. Here, we'll load the card from [`nateraw/vit-base-beans`](https://huggingface.co/nateraw/vit-base-beans).
```
osanseviero, line 9:
```suggestion
To load an existing card from the Hub, you can use the [`ModelCard.load`] function. Here, we'll load the card from [`nateraw/vit-base-beans`](https://huggingface.co/nateraw/vit-base-beans).
```

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
...
+# My Model Card
+"""
+
+card = ModelCard(text)
+card.data.to_dict() == {'language': 'en', 'license': 'mit'}  # True
+```
+
+Another way you might want to do this is with f-strings. In the following example, we:
+
+- Use [`ModelCardData.to_yaml`] to convert metadata we defined to YAML so we can use it to insert the YAML block in the model card
```
osanseviero, line 44:
```suggestion
- Use [`ModelCardData.to_yaml`] to convert metadata we defined to YAML so we can use it to insert the YAML block in the model card.
```

osanseviero, line 44: `ModelCardData` has not been mentioned so far, so maybe adding one sentence about it before could be useful

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
...
+---
+
+# Model Card for MyCoolModel
+
+This model does this and that.
+
+This model was created by [@nateraw](https://hf.co/nateraw).
+```
+
+As you update the card data, you can validate the card is still valid against the Hub by calling [`ModelCard.validate`]. This ensures that the card passes any validation rules set up on the Hugging Face Hub.
```
osanseviero, line 155: Maybe it's worth adding some links to jinja documentation for people that want to create their templates.

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
...
+    more_resources="https://github.com/huggingface/huggingface_hub",
+    **card_data.to_dict()  # Pass along any card data vals that might fill out part of the template
+)
+card.save('my_model_card_2.md')
+print(card)
+```
+
+## Sharing Model Cards
+
+If you're authenticated with the Hugging Face Hub (either by using `huggingface-cli login` or `huggingface_hub.notebook_login()`), you can push cards to hub by simply calling [`ModelCard.push_to_hub`]. Let's take a look at how to do that...
```
osanseviero, line 177:
```suggestion
If you're authenticated with the Hugging Face Hub (either by using `huggingface-cli login` or `huggingface_hub.notebook_login()`), you can push cards to the Hub by simply calling [`ModelCard.push_to_hub`]. Let's take a look at how to do that...
```

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+You can check out the resulting card [here](https://huggingface.co/nateraw/hf-hub-modelcards-pr-test/blob/main/README.md).
+
+If you instead wanted to push a card as a pull request, you can just say `create_pr=True` when calling `push_to_hub`:
+
+```python
+card.push_to_hub(repo_id, create_pr=True)
+```
+
+A resulting PR created from this command can be seen [here](https://huggingface.co/nateraw/hf-hub-modelcards-pr-test/discussions/3).
```
osanseviero, line 217: Very cool!

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+### Including Evaluation Results
+
+To include evaluation results in the metadata `model-index`, you can pass an [`EvalResult`] or a list of `EvalResult` with your associated evaluation results. Under the hood it'll create the `model-index` when you call `card.data.to_dict()`.
```
osanseviero, line 221: Maybe link to https://huggingface.co/docs/hub/models-cards#evaluation-results

setup.py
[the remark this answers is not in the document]
osanseviero: Should this be removed from `install_requires` ?

src/huggingface_hub/file_download.py
```diff
@@ -151,6 +159,14 @@ def get_fastcore_version():
     return _fastcore_version
 
 
+def is_jinja_available():
+    return _jinja_available
+
+
+def get_jinja_version():
```
osanseviero, line 166: I don't think this is used anywhere. Should we remove it?

tests/test_repocard.py
```diff
@@ -360,3 +388,346 @@ def test_update_new_result_new_dataset(self):
...
+        with self.caplog.at_level(logging.WARNING):
+            card = ModelCard.load(sample_path)
+        self.assertIn(
+            "Invalid model-index. Not loading eval results into CardData.",
+            self.caplog.text,
+        )
+        self.assertIsNone(card.data.eval_results)
+
+    def test_load_model_card_from_file(self):
+
```
osanseviero, line 599:
```suggestion
```

tests/test_repocard_data.py
```diff
@@ -0,0 +1,197 @@
...
+                                "value": 0.9,
+                            }
+                        ],
+                    },
+                ],
+            }
+        ]
+        model_name, eval_results = model_index_to_eval_results(model_index)
+
+        self.assertEqual(len(eval_results), 3)
```
osanseviero, line 102: hmmm I Might be missing something but aren't there only 2 eval results above?

osanseviero: Ah we have two metrics in the first result. Shouldn't there be 2 metrics in the same result or do we split into multiple results?

osanseviero (approved):
Very cool PR! This is very very nice, and the documentation for it is very helpful!

src/huggingface_hub/repocard_data.py
Wauplin: 👍 for this

tests/test_repocard.py
```diff
@@ -119,7 +134,20 @@
 repo_name = partial(repo_name, prefix="dummy-hf-hub")
 
 
-class RepocardTest(unittest.TestCase):
+def require_jinja(test_case):
+    """
+    Decorator marking a test that requires Jinja2.
+
+    These tests are skipped when Jinja2 is not installed.
+
```
Wauplin, line 142:
```suggestion
```

Wauplin (commented):
Just finished to review this PR. Thanks A LOT @nateraw for all the impressive implementation, documentation and testing. I have left a bunch of small comments but it's overall very good work 👍 🔥 🔥

src/huggingface_hub/file_download.py
osanseviero: You're right! Yes, let's do this but in a separate PR as it came before.

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,288 @@
...
+                repo_type=repo_type,
+                commit_message=commit_message,
+                commit_description=commit_description,
+                create_pr=create_pr,
+                revision=revision,
+            )
+        return url
+
+    @classmethod
+    def from_template(
```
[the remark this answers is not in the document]
osanseviero: No strong opinion to be honest. `from_data` makes sense to me. WDYT @nateraw ?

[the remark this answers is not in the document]
julien-c: kinda like `from_data` too

setup.py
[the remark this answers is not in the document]
nateraw: Whoops! you're right.

src/huggingface_hub/templates/datasetcard_template.md
```diff
@@ -0,0 +1,126 @@
...
+
+- **Homepage:**
+- **Repository:**
+- **Paper:**
+- **Leaderboard:**
+- **Point of Contact:**
+
+### Dataset Summary
+
+[More Information Needed]
```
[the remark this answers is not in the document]
nateraw: Yes we should definitely make it more interactive. Perhaps I'll leave that to separate, smaller PR since this one has gotten so big

tests/test_repocard_data.py
nateraw: There are 3 eval results:
- 2 for the cats-vs-dogs dataset and 
- one for the beans dataset. 

The `EvalResult` object is a flattened representation, so each instance represents one evaluation result coming from one dataset.

nateraw: So, the above would be something like this:

```python
from huggingface_hub import EvalResult

eval_results = [
    EvalResult(dataset_type='cats_vs_dogs', metric_type='acc', metric_value=0.85, ...),
    EvalResult(dataset_type='cats_vs_dogs', metric_type='f1', metric_value=0.9, ...),
    EvalResult(dataset_type='beans', metric_type='acc', metric_value=0.9, ...)
]
```

src/huggingface_hub/repocard.py
```diff
@@ -131,78 +571,75 @@ def metadata_eval_result(
...
+        ...                             'name': 'Accuracy',
+        ...                             'verified': False
+        ...                         }
+        ...                     ]
+        ...                 }
+        ...             ]
+        ...         }
+        ...     ]
+        ... }
+        True
```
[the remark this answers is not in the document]
nateraw: this is a doctest, so `results == ...` is just saying the results I got on the previous line equal the dict you're seeing. The `True` is the output. We arent running doctests in CI, but they are really useful to make sure the example snippets actually work.

src/huggingface_hub/repocard_data.py
nateraw: yea probably a good call! I only intended users to use kwargs anyways, so this makes sense to me.

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,470 @@
...
+        else:
+            raise ImportError(
+                "Using RepoCard.from_template requires Jinja2 to be installed. Please"
+                " install it with `pip install Jinja2`."
+            )
+
+        template_path = template_path or cls.default_template_path
+        content = jinja2.Template(Path(template_path).read_text()).render(
+            card_data=card_data.to_yaml(), **template_kwargs
+        )
```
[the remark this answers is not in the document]
nateraw: Sure, I dig this...I think I actually had it written similar to this in one of the earlier versions of `modelcards`.

Anything to think about wrt whether this could break something/cause some unexpected behavior?

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,470 @@
...
+
+class ModelCard(RepoCard):
+    card_data_class = ModelCardData
+    default_template_path = TEMPLATE_MODELCARD_PATH
+    repo_type = "model"
+
+    @classmethod
+    def from_template(
+        cls,
+        card_data: CardData,
```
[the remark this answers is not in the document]
nateraw: v nice catch, thanks

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,470 @@
...
+
+            ```
+        """
+        return super().from_template(card_data, template_path, **template_kwargs)
+
+
+def _detect_line_ending(content: str) -> Literal["\r", "\n", "\r\n", None]:
+    """
+    same implem as in Hub server, keep it in sync
+    """
```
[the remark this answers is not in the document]
nateraw: @julien-c do you have a good description for this? 😅 I didn't really fully understand when you issued the pr for it, so maybe you'd be best to explain

src/huggingface_hub/repocard.py
```diff
@@ -265,149 +690,91 @@ def metadata_update(
...
+        >>> from huggingface_hub import metadata_update
+        >>> metadata = {'model-index': [{'name': 'RoBERTa fine-tuned on ReactionGIF',
+        ...             'results': [{'dataset': {'name': 'ReactionGIF',
+        ...                                      'type': 'julien-c/reactiongif'},
+        ...                           'metrics': [{'name': 'Recall',
+        ...                                        'type': 'recall',
+        ...                                        'value': 0.7762102282047272}],
+        ...                          'task': {'name': 'Text Classification',
+        ...                                   'type': 'text-classification'}}]}]}
+        >>> url = metadata_update("julien-c/reactiongif-roberta", metadata)
```
nateraw, line 705: TODO - use internal testing repo instead. This one keeps getting blown up whenever I run doctests 😂

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
nateraw: I'm not a big fan of `from_data`, as I think it's a bit more misleading. We aren't really creating the card 100% from data - a lot of the content is going to come from the template. And keeping the name with "template" in it at least makes sense when you get an error about not having Jinja2 (imo). 

I could be convinced to change it, though, if anybody is very opinionated about it!

src/huggingface_hub/repocard.py
```diff
@@ -1,30 +1,470 @@
...
+        headers = {"Accept": "text/plain"}
+
+        try:
+            r = requests.post(
+                "https://huggingface.co/api/validate-yaml", body, headers=headers
+            )
+            r.raise_for_status()
+        except requests.exceptions.HTTPError as exc:
+            if r.status_code == 400:
+                raise RuntimeError(r.text)
```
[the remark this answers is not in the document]
nateraw: SGTM, updated.

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
Wauplin: Ok thanks for the explanation. Never used it but it seems cool 👍

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
Wauplin: As long as we create a new dictionary with `.copy()` and that we give priority to user-defined kwargs I don't see what could go wrong.

A problem can happen if a user defines its own jinja template, use an optional variable in it that has the same name as an attribute of `RepoCardData` -without having the same value/same meaning- and that we overwrite it from card_data even though the user did not pass it. Said like this it seems a quite specific use-case and the user always have the possibility to update its template to use a different var name.

Always difficult to look for a blind spot though 😫

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
Wauplin: Then a description is definitely needed 😁

tests/test_repocard_data.py
osanseviero: Makes sense, thanks for explaining!

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
julien-c: hmm yeah it's just a helper to determine the line endings for a file

This is useful when updating a model card, e.g. changing just one metadata item, on a model that was created in Windows. If you don't honor the existing line ending you get a diff on all lines (and the resulting file has mixed line ending)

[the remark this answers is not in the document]
julien-c: not a huge deal considering most users have \n line endings, but we encountered this issue when doing mass-modification of model cards on the Hub (affected a few % of the model cards)

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
Wauplin: Seems no one is so opinionated but we still don't have a consensus though 😁. 
Maybe `RepoCard.render_with_data` (or `RepoCard.render_from_data(...)`) is more convincing ? It's less the name that I would give to a classmethod but we have both the templating and the data aspects.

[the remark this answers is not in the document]
Wauplin: And in any case, if no naming reach a perfect consensus I suggest with keep `RepoCard.from_template` just because it's good enough and so that we can close this (huge) PR :)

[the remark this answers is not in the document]
adrinjalali: As an alternative design, the pattern we chose in `skops` is for the user to create an instance of the object, then add contents for the sections through instance methods. From the users' perspective, the code looks like this, and the implementation is here: https://github.com/skops-dev/skops/blob/main/skops/card/_model_card.py

``` python
# this constructor would take a path to a template in a future iteration, for now supporting only
# the default template.
# Instead of calling the metadata `CardData`, we call it `metadata` which is easier
# to understand from the users' perspective.
model_card = card.Card(model=sklearn_estimator, metadata=...)

# The user can also access and change/add metadata by accessing the metadata object
model_card.metadata.license = "mit"

limitations = "This model is not ready to be used in production."
model_description = (
    "This is a HistGradientBoostingClassifier model trained on breast cancer dataset."
    " It's trained with Halving Grid Search Cross Validation, with parameter grids on"
    " max_leaf_nodes and max_depth."
)
model_card_authors = "skops_user"
get_started_code = (
    "import pickle\nwith open(pkl_filename, 'rb') as file:\n    clf = pickle.load(file)"
)
citation_bibtex = "bibtex\n@inproceedings{...,year={2020}}"

# they can add info in a single call, or chain calls to `add`, since all `add*` methods
# return self.
model_card.add(
    citation_bibtex=citation_bibtex,
    get_started_code=get_started_code,
    model_card_authors=model_card_authors,
    limitations=limitations,
    model_description=model_description,
)

# adding metrics
eval_descr = (
    "The model is evaluated on test data using accuracy and F1-score with macro"
    " average."
)
model_card.add(eval_method=eval_descr)

...
model_card.add_metrics(**{"accuracy": accuracy, "f1 score": f1})

# or save and add plots
...
model_card.add_plot(**{"Confusion matrix": "confusion_matrix.png"})

# or add tabular information about hyper parameter search space and resuts
...
model_card.add_table(
    # they decide if the table should be in a `<details>` tag.
    folded=True,
    **{
        "Hyperparameter search results": cv_results,
        "Classification report": clf_report,
    },
)

# And finally, this is how the model is _rendered_ into a file.
model_card.save(Path(local_repo) / "README.md")

# However, they could also render the model into a string:
rendered = model_card.render()
```

docs/source/how-to-model-cards.mdx
nateraw: Done!

docs/source/how-to-model-cards.mdx
nateraw: Done!

src/huggingface_hub/repocard.py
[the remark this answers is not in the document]
nateraw: I prefer that we leave it as `from_template` for now - I am opinionated enough to push for that. I think it makes it very clear that we're using a 'template', which implies the need for the extra `Jinja2` dep. I promise I'm not just being lazy and not wanting to change the name 😅 .

As for the usage described by @adrinjalali - I prefer the way we have implemented here, as it results in less code I think, and effectively does the same thing.

nateraw:
Ok folks, I'm ready to call this good to go. Any small updates to docs/etc can be done in future small PRs, as long as there's nothing major lingering that anybody is super opinionated about.

Feel free to give this a final look over.

---

@Wauplin, as a sanity check, can you please double check I didn't mess anything up when I rebased? one of my most [recent commits](https://github.com/huggingface/huggingface_hub/pull/940/commits/8ea02af2feb619c730d364c472054c5fab178ca9) ended up showing me codecov twice, causing tests to fail for some reason, but it resolved itself when I pushed to this branch again.

Should be fine, just would like another pair of eyes on it. Thank you!!

meg-huggingface (commented):
Great work!!

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+The [`CardData`] object is the parent class of [`ModelCardData`] and [`DatasetCardData`].
+
+[[autodoc]] huggingface_hub.repocard_data.CardData
+
+## Model Cards
+
+[[autodoc]] ModelCard
+
+[[autodoc]] ModelCardData
+
```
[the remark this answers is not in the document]
julien-c: i'm not too bent on the name but on HF side i like that symmetry:

`A X Card is materialized by a README.md file inside a X repo` where X can be {model, dataset, Space}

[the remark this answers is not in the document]
julien-c: we can always document that it's "also usually known as a Data Card", ofc

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,317 @@
...
+
+```python
+card_data = ModelCardData(language='en', license='mit', library_name='keras')
+card = ModelCard.from_template(
+    card_data,
+    model_id='my-cool-model',
+    model_description="this model does this and that",
+    developers="Nate Raw",
+    more_resources="https://github.com/huggingface/huggingface_hub",
+    **card_data.to_dict()  # Pass along any card data vals that might fill out part of the template
```
Wauplin, line 175:
```suggestion
```

No need for this anymore as card_data values are passed along by default

docs/source/how-to-model-cards.mdx
```diff
@@ -0,0 +1,317 @@
...
```
Wauplin, line 205:
```suggestion
```

same here

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+Visit the [dedicated documentation page](https://huggingface.co/docs/hub/models-cards) for a deeper view of what 
+Model Cards on the Hub are, and how they work under the hood. You can also check out our [Model Cards guide](../how-to-model-cards) to 
+get a feel for how you would use these utilities in your own projects.
+
+## Repo Card
+
+The `RepoCard` object is the parent class of [`ModelCard`] and [`DatasetCard`].
+
+[[autodoc]] huggingface_hub.repocard.RepoCard
+
```
Wauplin, line 13:
```suggestion
[[autodoc]] huggingface_hub.repocard.RepoCard
    - __init__
    - all
```

Repocard init does a lot of things and is really well documented so I would add it explicitely to the generated doc.

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+The [`CardData`] object is the parent class of [`ModelCardData`] and [`DatasetCardData`].
```
Wauplin, line 14:
```suggestion
## Card Data

The [`CardData`] object is the parent class of [`ModelCardData`] and [`DatasetCardData`].
```

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+[[autodoc]] huggingface_hub.repocard_data.CardData
+
+## Model Cards
+
+[[autodoc]] ModelCard
+
+[[autodoc]] ModelCardData
```
Wauplin, line 22:
```suggestion
### ModelCard

[[autodoc]] ModelCard

### ModelCardData

[[autodoc]] ModelCardData
```

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+## Dataset Cards
+
+[[autodoc]] DatasetCard
+
+[[autodoc]] DatasetCardData
+
```
Wauplin, line 29:
```suggestion
## Dataset Cards

Dataset cards are also known as Data Cards in the ML Community.

### DatasetCard

[[autodoc]] DatasetCard

### DatasetCardData

[[autodoc]] DatasetCardData

```

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,538 @@
...
+            The name of the dataset configuration used in `load_dataset()`.
+            Example: fr in `load_dataset("common_voice", "fr")`. See the `datasets` docs for more info:
+            https://hf.co/docs/datasets/package_reference/loading_methods#datasets.load_dataset.name
+        dataset_split (`str`, *optional*):
+            The split used in `load_dataset()`. Example: "test".
+        dataset_revision (`str`, *optional*):
+            The revision (AKA Git Sha) of the dataset used in `load_dataset()`.
+            Example: 5503434ddd753f426f4b38109466949a1217c2bb
+        dataset_args (`Dict[str, Any]`, *optional*):
+            The arguments passed during `Metric.compute()`. Example for `bleu`: max_order: 4.
```
Wauplin, line 44:
```suggestion
        dataset_args (`Dict[str, Any]`, *optional*):
            The arguments passed during `Metric.compute()`. Example for `bleu`: `{"max_order": 4}`
```

docs/source/package_reference/cards.mdx
```diff
@@ -0,0 +1,40 @@
...
+[[autodoc]] EvalResult
+
+[[autodoc]] huggingface_hub.repocard_data.model_index_to_eval_results
+
+[[autodoc]] huggingface_hub.repocard_data.eval_results_to_model_index
+
+[[autodoc]] huggingface_hub.repocard.metadata_eval_result
+
+[[autodoc]] huggingface_hub.repocard.metadata_update
```
Wauplin, line 40:
```suggestion
## Utilities

### EvalResult

[[autodoc]] EvalResult

### model_index_to_eval_results

[[autodoc]] huggingface_hub.repocard_data.model_index_to_eval_results

### eval_results_to_model_index

[[autodoc]] huggingface_hub.repocard_data.eval_results_to_model_index

### metadata_eval_result

[[autodoc]] huggingface_hub.repocard.metadata_eval_result

### metadata_update

[[autodoc]] huggingface_hub.repocard.metadata_update
```

docs/source/package_reference/cards.mdx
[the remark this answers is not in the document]
Wauplin: I added a line in @nateraw's documentation for that

src/huggingface_hub/repocard_data.py
```diff
@@ -0,0 +1,538 @@
...
+            The arguments passed during `Metric.compute()`. Example for `bleu`: `{"max_order": 4}`
+        metric_name: (`str`, *optional*):
+            A pretty name for the metric. Example: "Test WER".
+        metric_config: (`str`, *optional*):
+            The name of the metric configuration used in `load_metric()`.
+            Example: bleurt-large-512 in `load_metric("bleurt", "bleurt-large-512")`.
+            See the `datasets` docs for more info: https://huggingface.co/docs/datasets/v2.1.0/en/loading#load-configurations
+        metric_args: (`Dict[str, Any]`, *optional*):
+            The arguments passed during `Metric.compute()`. Example for `bleu`: max_order: 4
+        verified: (`bool`, *optional*):
```
Wauplin, line 53:
```suggestion
        metric_name (`str`, *optional*):
            A pretty name for the metric. Example: "Test WER".
        metric_config (`str`, *optional*):
            The name of the metric configuration used in `load_metric()`.
            Example: bleurt-large-512 in `load_metric("bleurt", "bleurt-large-512")`.
            See the `datasets` docs for more info: https://huggingface.co/docs/datasets/v2.1.0/en/loading#load-configurations
        metric_args (`Dict[str, Any]`, *optional*):
            The arguments passed during `Metric.compute()`. Example for `bleu`: max_order: 4
        verified (`bool`, *optional*):
```

Wauplin (approved):

Wauplin:
@nateraw I double-checked everything and I think we are good to go.
Thanks again for the huge work done in this PR both in the implementation and in documenting everything !

PR #941, opened 2022-07-06
:bug: Fix multilinguality filter value

nateraw:
There were some updates to dataset metadata on the Hub that caused the `multilinguality` filter to break and thus broke the CI. This PR fixes that issue.

CC @muellerzr


muellerzr (approved):
Thanks for the fix!

julien-c:
hmm, can you point to the dataset repo update that changed this?

I don't think this is related to a server change, but to a dataset repo metadata (the server doesn't store metadata schema, it just aggregates metadata from all repos)

nateraw:
@julien-c I have no clue what caused this to break 😅 - I just figured out a fix and moved on.

PR #951, opened 2022-07-21
✨ Add `files_metadata` option to repo_info

SBrandeis:
Nice to have in #943

cc @mariosasko @lhoestq 
cc @huggingface/moon-landing-back 

This PR adds a new `files_metadata` flag to `HfApi.repo_info`.
When set to `True`, this flag fetches additional metadata in`{Model|Dataset|Space}Info.siblings`:

```python
>>> from huggingface_hub import model_info
>>> model = model_info("bert-base-uncased", files_metadata=True)
>>> model.siblings
# [
#     RepoFile(rfilename='.gitattributes', size='437', blob_id='4d564b466df31bf165495398c0d8ef51ddd2ee99', lfs='None'),
#     RepoFile(rfilename='LICENSE', size='11356', blob_id='f49a4e16e68b128803cc2dcea614603632b04eac', lfs='None'),
#     [...]
#     RepoFile(rfilename='flax_model.msgpack', size='438064459', blob_id='5dfe9adac0748dcecbb14ab068f6a9406d7348da', lfs='{'sha256': 'ea201fabe466ef7182f1f687fb5be4b62a73d3a78883f11264ff7f682cdb54bf', 'size': 438064459}'),
#     RepoFile(rfilename='pytorch_model.bin', size='440473133', blob_id='ba5d19791be1dd7992e33bd61f20207b0f7f50a5', lfs='{'sha256': '097417381d6c7230bd9e3557456d726de6e83245ec8b24f529f60198a67b203a', 'size': 440473133}'),
# ]
```


LysandreJik (approved):
Great, thanks @SBrandeis! Only left a comment.

src/huggingface_hub/hf_api.py
```diff
@@ -166,10 +166,26 @@ class RepoFile:
...
+            with Git LFS. It's `None` otherwise.
     """
 
     def __init__(self, rfilename: str, **kwargs):
         self.rfilename = rfilename  # filename relative to the repo root
+
+        # Optional file metadata
+        self.size: Optional[int] = kwargs.pop("size", None)
+        self.blob_id: Optional[str] = kwargs.pop("blobId", None)
+        self.lfs: Optional[dict] = kwargs.pop("lfs", None)
```
LysandreJik, line 187: This means that these will always be present on this object even if we didn't ask for optional metadata. I think it would result in leaner objects if we just removed this and had the two lines below 

```
        for k, v in kwargs.items():
            setattr(self, k, v)
```

take care of the attribution. This way no `self.blob_id == None` on all objects when not asking for this info. 

WDYT?

SBrandeis: Honestly, I am not a fan of dynamic objects like this
While I understand it can be practical as a maintainer, I'm not convinced by the user experience for several reasons:
- It does not play well with type checkers
- Documentation basically says "there can be more stuff in here, but I can't tell you what'"
- On a personal note, I would prefer a dict-like interface (`__getitem__` vs `setattr`). IMO, `"key" in repo_file` is nicer (more idiomatic?) than `hasattr(repo_file, "key")`

TL;DR: I'd rather have `blob_id` always set but sometimes `None`, than having it set only from times to times

That being said, I can make the change if maintainers think the proposed `setattr` loop is better

LysandreJik: I understand where you're coming from, and I don't have a super strong feeling about this so if you find it more simple this way, happy to keep it like this.

SBrandeis: What do others think about this ? @osanseviero @adrinjalali @julien-c

adrinjalali: I also struggle both as a user and as a developer to understand what an object does when its `__init__` accepts everything and the possible attributes are not documented.

In a case like this, I would find an API which is very explicit on what the attributes can be much more understandable.

If we really need to have some optional extra parameters, we can have them in a separate attribute which is a dict and includes "other" stuff.

For example, the docstring of this class doesn't even mention that everything can be passed to its constructor and that they're stored.

Also, from what I see our `*Info` classes can be all dataclasses, can't they?

lhoestq: Small pref to have these attributes all the time, but set to None by default.

> I also struggle both as a user and as a developer to understand what an object does when its __init__ accepts everything and the possible attributes are not documented.

I think it's fine if those if we add them to `__init__` with a default None value. It should make it clearer that those attributes exist all the time, and have a default None value.

> Also, from what I see our *Info classes can be all dataclasses, can't they?

I think dataclasses can't have **kwargs if I'm not mistaken

src/huggingface_hub/hf_api.py
```diff
@@ -166,10 +166,26 @@ class RepoFile:
             file name, relative to the repo root. This is the only attribute
             that's guaranteed to be here, but under certain conditions there can
             certain other stuff.
+        size (`int`, *optional*):
+            The file's size, in bytes. This attribute is present when `files_metadata` argument
+            of [`repo_info`] is set to `True`. It's `None` otherwise.
+        blob_id (`str`, *optional*):
+            The file's git OID. This attribute is present when `files_metadata` argument
+            of [`repo_info`] is set to `True`. It's `None` otherwise.
+        lfs (`dict`, *optional*):
```
osanseviero, line 175: Should we add typing for the dictionary as it's uniform? `Dict[str,str]`

src/huggingface_hub/hf_api.py
osanseviero: Note that we already have the pattern of having attributes set to `None` by default in `huggingface_hub`, so this approach would be consistent with the rest

E.g.
```
list_models(limit=2)
```
The above will give `config:None` for example, while the following will set the config to the right dictionary.

```
list_models(limit=2, fetch_config=True)
```

src/huggingface_hub/hf_api.py
```diff
@@ -259,8 +275,8 @@ class DatasetInfo:
             date of last commit to repo
         tags (`Listr[str]`, *optional*):
             List of tags.
-        siblings (`List[Dict]`, *optional*):
-            list of files that constitute the Space
+        siblings (`List[RepoFile]`, *optional*):
+            list of [`huggingface_hub.hf_api.RepoFile`] objects that constitute the dataset
```
osanseviero, line 279:
```suggestion
            list of [`huggingface_hub.hf_api.RepoFile`] objects that constitute the dataset.
```

src/huggingface_hub/hf_api.py
```diff
@@ -193,8 +209,8 @@ class ModelInfo:
             List of tags.
         pipeline_tag (`str`, *optional*):
             Pipeline tag to identify the correct widget.
-        siblings (`List[Dict]`, *optional*):
-            list of files that constitute the Space
+        siblings (`List[RepoFile]`, *optional*):
+            list of ([`huggingface_hub.hf_api.RepoFile`]) objects that constitute the model
```
osanseviero, line 213:
```suggestion
            list of ([`huggingface_hub.hf_api.RepoFile`]) objects that constitute the model.
```

src/huggingface_hub/hf_api.py
```diff
@@ -166,10 +166,26 @@ class RepoFile:
...
             certain other stuff.
+        size (`int`, *optional*):
+            The file's size, in bytes. This attribute is present when `files_metadata` argument
+            of [`repo_info`] is set to `True`. It's `None` otherwise.
+        blob_id (`str`, *optional*):
+            The file's git OID. This attribute is present when `files_metadata` argument
+            of [`repo_info`] is set to `True`. It's `None` otherwise.
+        lfs (`dict`, *optional*):
+            The file's LFS metadata (has two keys: `sha256` and `size`). This attribute is present when
+            `files_metadata` argument of [`repo_info`] is set to `True` and the file is stored
```
osanseviero, line 177:
```suggestion
            `files_metadata` argument of [`repo_info`] is set to `True` and the file is stored.
```

tests/test_hf_api.py
```diff
@@ -928,6 +929,31 @@ def test_model_info_with_security(self):
             {"containsInfected": False},
         )
 
+    @with_production_testing
+    def test_model_info_with_file_metadata(self):
```
osanseviero, line 933: I would also add another test that shows that the metadata is not included in the default call

osanseviero (approved):
Thanks a lot for this PR! Very cool!

src/huggingface_hub/hf_api.py
julien-c: or even better to type the actual attributes sha256 and size

src/huggingface_hub/hf_api.py
julien-c: For consistency with e.g. `ModelInfo`, I'm would add them to `__init__` with a default None value.

The reason for the `setattr` catch-all in all those classes is to maximize forward compatibility: if we add something to the server (which we sometimes do), past versions of the library support it out of the box

Also related, note that in the past a few versions of `huggingface_hub` required a non-optional attribute that was deprecated (the infamous `key`, cc @lhoestq) and that . So it's important that everything is kept optional

src/huggingface_hub/hf_api.py
julien-c: arf yeah, i guess this docstring documents the class, not its init. So the change is correct

src/huggingface_hub/hf_api.py
```diff
@@ -1352,10 +1398,14 @@ def repo_info(
...
+            files_metadata (`bool`, *optional*):
+                Whether or not to retrieve metadata for files in the repository
+                (size, LFS metadata, etc). Defaults to `False`.
 
         Returns:
-            `Union[SpaceInfo, DatasetInfo, ModelInfo]`: The repository
-            information.
+            `Union[SpaceInfo, DatasetInfo, ModelInfo]`: The repository information, as a
+            [`huggingface_hub.hf_api.DatasetInfo`], [`huggingface_hub.hf_api.ModelInfo`]
+            or [`huggingface_hub.hf_api.SpaceInfo`] object.
```
julien-c, line 1408: not sure we need to expose it here

tests/test_hf_api.py
julien-c: (this is testing the server more than the client though)

julien-c (approved):

src/huggingface_hub/hf_api.py
SBrandeis: Why not ?

julien-c: `repo_info` is just a helper exposing a common subset of params accross repo types.

But no strong opinion, we can keep it here if you think it's going to be useful

lhoestq (approved):

LysandreJik (approved):

tests/test_hf_api.py
```diff
@@ -928,6 +929,31 @@ def test_model_info_with_security(self):
...
+            [isinstance(file.size, int) for file in files], [True] * len(files)
+        )
+        self.assertTrue(any([file.lfs is not None for file in files]))
+        self.assertListEqual(
+            [
+                file.lfs is None or isinstance(file.lfs, dict) and "sha256" in file.lfs
+                for file in files
+            ],
+            [True] * len(files),
+        )
```
Wauplin, line 955: Quick advice here (maybe more a personal opinion), I always try to favor code readability especially in the tests. Here I would rather loop over the file and assert stuff on their content instead of building artificial lists `[True] * len(files)` that slow down understanding.


```suggestion
        at_least_one_lfs = False
        for file in files:
            self.assertTrue(isinstance(file.blob_id, str))
            self.assertTrue(isinstance(file.size, int))
            if file.lfs is not None:
                at_least_one_lfs = True
                self.assertTrue(isinstance(file.lfs, dict) and "sha256" in file.lfs)
        self.assertTrue(at_least_one_lfs)
```

tests/test_hf_api.py
```diff
@@ -1083,6 +1109,30 @@ def test_dataset_info(self):
...
+
```
Wauplin, line 1135: Other suggestion, I would also refactor the siblings check in a separate helper function and use it in both model and dataset tests. Next time new metadata is added, only this helper would need an update.

EDIT: I realise I am sometimes too much into refactoring. Having self-contained and readable tests is nice as well :)

```suggestion
        self._check_siblings_metadata(files)

    def _check_siblings_metadata(self, files: List[RepoFile])
        """Check requested metadata has been received from the server."""
        at_least_one_lfs = False
        for file in files:
            self.assertTrue(isinstance(file.blob_id, str))
            self.assertTrue(isinstance(file.size, int))
            if file.lfs is not None:
                at_least_one_lfs = True
                self.assertTrue(isinstance(file.lfs, dict) and "sha256" in file.lfs)
        self.assertTrue(at_least_one_lfs)

```

Wauplin (approved):
@SBrandeis I approved with optional suggestions. It is more about sharing a hint about how I would I done things 🙂

src/huggingface_hub/hf_api.py
```diff
@@ -156,20 +156,47 @@ def repo_type_and_id_from_hf_id(
...
+    size: int
+    sha256: str
+
+
 class RepoFile:
     """
-    Data structure that represents a public file inside a repo, accessible from
-    huggingface.co
+        Data structure that represents a public file inside a repo, accessible from
+        huggingface.co
```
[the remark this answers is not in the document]
SBrandeis: I don't think it matters in the end here - cc @mishig25 
Also, this can be reverted if you want :)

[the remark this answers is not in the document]
Wauplin:
```suggestion
    Data structure that represents a public file inside a repo, accessible from
    huggingface.co

File: tests/test_hf_api.py [hunk 11]
```diff
@@ -45,6 +46,7 @@
     MetricInfo,
     ModelInfo,
     ModelSearchArguments,
+    RepoFile,
```
Wauplin, line 49:
```suggestion
    RepoFile,
```

PR #956, opened 2022-07-25
[#940] Better inheritance for repo card utils

nateraw:
Working out of a separate branch off of #940 since this will be a pretty big update and I want to keep the example colab working in that PR.

This PR aims to refactor so we have a base `RepoCard` object with `ModelCard` and `DatasetCard` inheriting from it. Likewise, `CardData` will have `ModelCardData` and `DatasetCardData` inheriting from it. 

Updated colab for this branch:
[Open In Colab](https://colab.research.google.com/gist/nateraw/e5ee295a3a5847e375130757a186e96d/modelcard-in-huggingface-hub-inheritance-demo.ipynb)


nateraw:
This isn't final, it's just the core of the code to move from no inheritance to using inheritance. everything else (docs, docstrings, more tests, etc) will be updated in #940. 

Curious if anybody has any objections/comments on how I've restructured here. If not, will be merging and moving forward with this.

src/huggingface_hub/repocard.py
```diff
@@ -65,18 +76,7 @@ def __init__(self, content: str):
...
-                model_name, eval_results = model_index_to_eval_results(model_index)
-                data_dict["model_name"] = model_name
-                data_dict["eval_results"] = eval_results
-            except KeyError:
-                logger.warning(
-                    "Invalid model-index. Not loading eval results into CardData."
-                )
-
-        self.data = CardData(**data_dict)
+        self.data = self.card_data_class(**data_dict)
```
osanseviero, line 79: Nice!

src/huggingface_hub/repocard.py
```diff
@@ -289,12 +286,25 @@ def from_template(
...
         return cls(content)
 
 
+class ModelCard(RepoCard):
+    card_data_class = ModelCardData
+    default_template_path = TEMPLATE_MODELCARD_PATH
+    repo_type = "model"
+
+
+class DatasetCard(RepoCard):
```
osanseviero, line 302: Cool!

src/huggingface_hub/repocard_data.py
```diff
@@ -76,6 +81,38 @@ class EvalResult:
...
+    def to_dict(self):
+        """Converts CardData to a dict.
+
+        Returns:
+            `dict`: CardData represented as a dictionary ready to be dumped to a YAML
+            block for inclusion in a README.md file.
+        """
+
+        data_dict = copy.deepcopy(self.__dict__)
+        self._to_dict(data_dict)
```
osanseviero, line 96: As discussed via DM, I'm not a fan of hacking `_to_dict` to make this work. Simply implementing `to_dict` for each might be cleaner, but I have a feeling others might prefer the `_to_dict` approach (the current one). WDYT @LysandreJik?

osanseviero (approved):
The overall approach looks good to me!

src/huggingface_hub/templates/datasetcard_template.md
```diff
@@ -0,0 +1,126 @@
...
+  - [Personal and Sensitive Information](#personal-and-sensitive-information)
+- [Considerations for Using the Data](#considerations-for-using-the-data)
+  - [Social Impact of Dataset](#social-impact-of-dataset)
+  - [Discussion of Biases](#discussion-of-biases)
+  - [Other Known Limitations](#other-known-limitations)
+- [Additional Information](#additional-information)
+  - [Dataset Curators](#dataset-curators)
+  - [Licensing Information](#licensing-information)
+  - [Citation Information](#citation-information)
+  - [Contributions](#contributions)
```
julien-c, line 30: I would remove this (same for model card) given that we display a TOC in the UX (for both datasets and models now) so I don't think those are needed (cc @osanseviero)

## dataset card UI:

image

## model card UI:

image

julien-c: (also cc @gary149 )

julien-c: (can be in a more relevant PR, of course)

src/huggingface_hub/repocard.py
```diff
@@ -29,12 +31,21 @@
...
+    Path(__file__).parent / "templates" / "datasetcard_template.md"
+)
+
 REGEX_YAML_BLOCK = re.compile(r"---[\n\r]+([\S\s]*?)[\n\r]+---[\n\r]")
 
 logger = get_logger(__name__)
 
 
-class ModelCard:
+class RepoCard:
```
julien-c, line 43: thoughts on making this abstract i.e. you have to actually instantiate a ModelCard or a DatasetCard? (not sure if it's even possible in Python though)

src/huggingface_hub/repocard_data.py
```diff
@@ -76,6 +81,38 @@ class EvalResult:
 
 @dataclass
 class CardData:
+    def __init__(self, **kwargs):
```
julien-c, line 84: same here, should we make this abstract

julien-c (approved):

src/huggingface_hub/templates/datasetcard_template.md
osanseviero: Yes, there was a huge discussion around this in #model-cards that they want the ToC both in UX and in the model card content. The logic behind it was that many institutions print model cards and having them in the content is important

julien-c: any way you have a link to the convo handy? 

Usually what happens on website is that you can specify that the table of contents gets printed with some CSS rules like `@media print { ... }`

ie. the webpage looks different when printed vs. on the Web

WDYT?

osanseviero: That works for me!

osanseviero: (this is unrelated to this PR so feel free to go ahead and merge)

src/huggingface_hub/repocard.py
nateraw: hmm will think on this. perhaps would be nice to warn if `RepoCard` is instantiated by itself? idk

nbd either way I assume

src/huggingface_hub/repocard_data.py
nateraw: Yea...this one bothers me a bit more. Mostly because in `modelcards` this was the name of the object you'd use, so it could cause some confusion.

PR #965, opened 2022-08-04
feat: 🎸 add HfApi.update_repo_settings to set a repo as gated

severo:
It also allows to disable/enable the discussions.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1645,6 +1645,93 @@ class HfApi:
         )
         _raise_for_status(r)
 
+    def update_repo_settings(
+        self,
+        repo_id: str = None,
+        *,
+        private: Optional[bool] = None,
+        gated: Optional[bool] = None,
+        discussions_disabled: Optional[bool] = None,
+        token: Optional[str] = None,
+        organization: Optional[str] = None,
+        repo_type: Optional[str] = None,
+        name: str = None,
+    ) -> Dict[str, bool]:
+        """Update the settings of a repository.
+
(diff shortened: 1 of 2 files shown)


julien-c:
not sure if there's a need to expose this programmatically to end users? 
WDYT @osanseviero @Wauplin?


If not, maybe keep it in your downstream code or in a helper

severo:
OK

osanseviero:
Probably there is not much interest in programmatically gating repos. Maybe we can revisit this PR if we get requests on this

severo:
You're right, it's much simpler to keep it in my downstream code.

cakiki:
I currently find myself needing to gate access to 498 datasets and feel it would be useful to expose that in HfApi, or maybe in the web interface on organization level (i.e. every dataset I publish in an org has to be gated with the following user agreement).

severo:
Here is the code I'm using, in case it helps:

https://github.com/huggingface/datasets-server/blob/main/e2e/tests/fixtures/hub.py#L24-L92

cakiki:
I sadly cannot access that file; any chance you could include it in a gist? :smiley_cat:

severo:
Oh, sorry, here it is: https://gist.github.com/severo/0e9a1d8fb91f5ecb9c123ac48fbe6ea4

btw, we plan to opensource datasets-server in the next months.

julien-c:
well... i stand corrected then! there is user demand for this!

Should we merge/add this feature then? WDYT?

julien-c:
(though by "user", i mean @cakiki-level power user 😅)

PR #1048, opened 2022-09-14
👽️ Add size parameter for lfsFiles when committing on the hub

coyotte508:


coyotte508:
@Wauplin sorry I don't understand by the coverage decreased so much :exploding_head:

Wauplin (approved):
Looks good to me, thanks for taking care of it :)

About code coverage, the check is updated each time 1 of the test is finished. So for the first minutes it's really low since only a few tests or finished. The coverage is reliable only once all checks have passed which is the case now !

Pierrci:
Thanks @coyotte508 @Wauplin!

PR #1052, opened 2022-09-14
Further clarifying Model Card sections

meg-huggingface:


nateraw (changes requested):
At the top of this file, can you please add the suffix `-optional` to the TOC reference links for citation and model examination? Lines 16 and 19.

Wauplin (approved):

PR #1076, opened 2022-09-22
Added metadata option to fastai utils

merveenoyan:
This PR adds metadata option to fast.ai utils.

The model card looks like [this](https://huggingface.co/merve/fastai-test) when pushed with following code:
```python
push_to_hub_fastai(
            learner=dummy_model,
            repo_id="merve/fastai-test",
            tags=["image-classification"]
        )
```

### DIFF
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -2,11 +2,11 @@
 
 Before you start, you will need to setup your environment by installing the appropriate packages.
 
-🤗 `huggingface_hub` is tested on **Python 3.7+**.
+`huggingface_hub` is tested on **Python 3.7+**.
 
 ## Install with pip
 
-It is highly recommended to install 🤗 `huggingface_hub` in a [virtual environment](https://docs.python.org/3/library/venv.html).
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/fastai_utils.py
```diff
@@ -226,10 +227,14 @@ def _create_model_card(repo_dir: Path):
             Directory where model card is created.
     """
     readme_path = repo_dir / "README.md"
-
+    metadata["library_name"] = "fastai"
+    model_card = "---\n"
+    model_card += yaml.dump(metadata, default_flow_style=False)
+    model_card += "---\n\n"
+    model_card += README_TEMPLATE
```
osanseviero, line 234: Maybe this could use the `modelcards` utils now that we have them. WDYT?

Here is a guide on it https://huggingface.co/docs/huggingface_hub/main/en/how-to-model-cards#from-text

merveenoyan: I thought of this as a short term fix but it makes sense to use modelcards. I'll modify the PR.

PR #1191, opened 2022-11-15
Parent Model --> Finetuned from model

meg-huggingface:
This change matches the wording in the template to the corresponding metadata variable.


Wauplin (approved):
LGTM, thanks for making the change @meg-huggingface . I'll merge it and add it to the 0.11 release of this week.

PR #1256, opened 2022-12-08
[Minor doc tweak] *currently...

julien-c:
(as it might change in the future)  cc @huggingface/infra-team


Wauplin (approved):
True :grin:

PR #1257, opened 2022-12-08
`datasetcard_template`: I think linking to a GH user does not make se…

julien-c:
…nse anymore now that dataset repos are fully on the Hub


lhoestq (approved):
Indeed, thanks

osanseviero (approved):
LGTM

PR #1265, opened 2022-12-13
Remove obsolete comment

severo:
The pagination is now active on the Hub (10.000 items)


Wauplin (approved):
Thanks for the PR @severo ! Good catch indeed :)

PR #1269, opened 2022-12-15
fix: authorization header should never be forwarded

rtrompier:
For the private hub, we don't use the same host url.
But we don't need to forward the Authorization header in any case, because S3 url already contains credentials information in a query params.


XciD (approved):

julien-c (commented):
sounds reasonable to me

Wauplin (approved):
Looks good. 
Checks are failing but it's unrelated (InferenceAPI)

PR #1345, opened 2023-02-16
Remove Authorization header when following redirects

XciD:
I thought that https://github.com/huggingface/huggingface_hub/blob/21a1e370f54e0af188d3ec8cbfd638b96917ab9e/src/huggingface_hub/file_download.py#L1137 was enough

Seems like we have multiple path for the same end


Wauplin (approved):

julien-c:
nice catch

PR #1347, opened 2023-02-17
Add repo_type to repo_info docstring

albertvillanova:
Add repo_type to repo_info docstring.


Wauplin (approved):
Thanks for the catch @albertvillanova !

PR #1405, opened 2023-03-22
🧹 Use `HUGGINGFACE_HEADER_X_LINKED_ETAG` const

julien-c:


Wauplin (approved):
Thanks for the catch!

Issue #1384, opened 2023-03-09
notebook_login() does not update UI on Databricks


fwetdb:
### Describe the bug

When using the `notebook_login` function on Databricks, the output is not adjusted [when calling `clear_output`](https://github1s.com/huggingface/huggingface_hub/blob/HEAD/src/huggingface_hub/_login.py#L217-L218) because this is not supported on Databricks and there is little chance that it will be supported in the future.

There is an easy fix. The content of the VBox can be adjusted (cleared explicitly) and the subsequent print statements can be appended to the VBox content.

I work for Databricks and I will provide a PR if you signal your willingness to take a look.

What do you think?

### Reproduction

from huggingface_hub import notebook_login

notebook_login()

### System info

```shell
Copy-and-paste the text below in your GitHub issue.

- huggingface_hub version: 0.12.1
- Platform: Linux-5.15.0-1030-aws-x86_64-with-glibc2.35
- Python version: 3.10.6
- Running in iPython ?: Yes
- iPython shell: DatabricksShell
- Running in notebook ?: Yes
- Running in Google Colab ?: No
- Token path ?: /root/.cache/huggingface/token
- Has saved token ?: False
- Configured git credential helpers: cache
- FastAI: N/A
- Tensorflow: 2.11.0
- Torch: 1.13.1+cpu
- Jinja2: 2.11.3
- Graphviz: N/A
- Pydot: N/A
- Pillow: 9.2.0
- hf_transfer: N/A
- ENDPOINT: https://huggingface.co
- HUGGINGFACE_HUB_CACHE: /root/.cache/huggingface/hub
- HUGGINGFACE_ASSETS_CACHE: /root/.cache/huggingface/assets
- HF_HUB_OFFLINE: False
- HF_TOKEN_PATH: /root/.cache/huggingface/token
- HF_HUB_DISABLE_PROGRESS_BARS: None
- HF_HUB_DISABLE_SYMLINKS_WARNING: False
- HF_HUB_DISABLE_IMPLICIT_TOKEN: False
- HF_HUB_ENABLE_HF_TRANSFER: False
```

Wauplin:
Hi @fwetdb, thanks for reporting the issue and proposing your help. The suggested idea makes sense to me. Would you do it only in Databrick notebooks or would that be applied ik any notebook? (Note that I'm not against it if it doesn't change the workflow for the user :) )

fwetdb:
[...]

PR #1414, opened 2023-03-29
Fix: notebook_login() does not update UI on Databricks

fwetdb:
Fixes [1384](https://github.com/huggingface/huggingface_hub/issues/1384)

The PR adjusts notebook_login() to use the ipywidgets Box output to print the feedback.
This enables the logic to work in all notebooks that are compatible with ipywidgets - including Databricks where `clear_output` does not work.

Examples for the code flow:
Screenshot 2023-03-29 at 12 08 35

Screenshot 2023-03-29 at 12 08 48

Screenshot 2023-03-29 at 12 09 00

Screenshot 2023-03-29 at 12 16 49

### DIFF
--- a/src/huggingface_hub/_login.py
+++ b/src/huggingface_hub/_login.py
@@ -206,10 +206,6 @@ def notebook_login() -> None:
     )
     display(login_token_widget)
 
-    def add_string_to_widget_output(*strings: str) -> None:
-        content_str = " ".join(strings)
-        login_token_widget.children = login_token_widget.children + (widgets.Label(content_str),)
-
     # On click events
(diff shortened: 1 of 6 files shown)


julien-c:
LGTM too and quite happy that we improve our support of Databricks / Databricks notebooks 👍

fwetdb:
Thank you @Wauplin and everyone else involved!
The adjustments look good to me and I just double-checked them on Databricks.
So, good to merge from my side.

Wauplin (approved):
Good, thanks @fwetdb for double-checking this!
I'm merging the PR then :tada:

PR #1433, opened 2023-04-11
Hffs misc improvements

mariosasko:
Improvements:
* pass the `ls` kwargs to `glob` in `expand_path`
* register `hf` as a `fsspec` protocol via the `entry_points` (explained [here](https://filesystem-spec.readthedocs.io/en/latest/developer.html#implementing-a-backend)).`fsspec.register_implementation("hf", huggingface_hub.HfFileSystem)` before using the HF filesystem with `fsspec` integrations is no longer required after this change

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,10 @@ setup(
     package_dir={"": "src"},
     packages=find_packages("src"),
     extras_require=extras,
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for the PR, looks good :+1:

PR #1482, opened 2023-05-22
Avoid `ImportError` when importing `WebhooksServer` and Gradio is not installed

mariosasko:
Avoids the ImportError when doing `from huggingface_hub import *` without having Gradio installed.

(The error is now raised when initializing a `WebhooksServer` object.)


Wauplin (approved):
Makes sense, thanks for the PR. Feel free to merge it when the CI is green :)

Out of curiosity, are you using this feature? What's the use case?

mariosasko:
@Wauplin No, but I use `from huggingface_hub import *` for debugging 🙂.

PR #1487, opened 2023-05-28
Update _hf_folder.py

SimonKitSangChu:
Allow permission error, especially for HPC environment


Wauplin (approved):
Yep, makes sense even if it's a corner case. Thanks for opening a PR! :) 

I made a small change to ignore any Exception, just in case. Worth case scenario, the user has to log in again.

Wauplin (approved):
Yep, makes sense even if it's a corner case. Thanks for opening a PR! :) 

I made a small change to ignore any Exception, just in case. Worth case scenario, the user has to log in again.

PR #1530, opened 2023-06-27
Set timeout to 10s

ydshieh:
The goal is to verify the fix on infra side works. We can either keep it 10s if everything is fine, or change back to 60s if we prefer.


Wauplin (commented):
Thanks for opening the PR @ydshieh 

Could you update the docstring of _request_wrapper. Instead of 60, write `10` as a default value. Thanks!

```py
timeout (`float`, *optional*, defaults to `60`):
```

Wauplin (approved):

PR #1577, opened 2023-08-03
[docs] Fix table

stevhliu:
Fixes [table](https://huggingface.co/docs/huggingface_hub/main/en/guides/inference#supported-tasks) that displays the supported tasks for the `InferenceClient` which is a bit disjointed right now.


Wauplin (approved):
Thanks @stevhliu !

PR #1649, opened 2023-09-08
Correct inference guide

martinbrose:
Just a few corrections in the inference guide:

- Inference Endpoints have inconsistent links within the guide. I believe it should always point to the docs: https://huggingface.co/docs/inference-endpoints/index. But maybe it is intentionally pointing to the main landing page for endpoints?
- The link for the Zero Shot Classification task is pointing to the description of the  Zero Shot Image Classification task.


Wauplin (approved):
Makes sense, thanks for spotting this :)

PR #1677, opened 2023-09-18
Correct typo in upload guide

martinbrose:
Just a quick fix for the upload guide.

It erroneously refers to "**two** types of operations supported by [`create_commit`]" and then lists **three** in the list below.


Wauplin (approved):
Good catch! Thanks :)

PR #1679, opened 2023-09-18
Longer progress bar description before trim, drop D-word 

poedator:
Addresses issue #1674:

Basically allows longer part of the filename being visible during model loading, at the expense of the word "Downloading".
See more details about rationale and possible alternatives at #1674


Wauplin (approved):
A 3-line PR that makes life easier. Thanks for opening it! :)

PR #1686, opened 2023-09-21
Add German concepts guide

martinbrose:
References https://github.com/huggingface/huggingface_hub/issues/1645

**Key features**

- Translates `concepts/git_vs_http` to German


Wauplin (approved):
Thanks for translating this concept guide @martinbrose! Looks good to me. I took the liberty to remove the _"package reference"_ section from the DE docs. It's perfectly fine to not have the same amount of content in EN and DE (as docstrings are written in EN anyway). I think it's best like this rather than trying to hack a way to link to the other docs.

PR #1717, opened 2023-10-08
Add `token` parameter to `HfApi`'s `snapshot_download` and `hf_hub_download`

mariosasko:
Adds a `token` parameter to `HfApi`'s `snapshot_download`/`hf_hub_download` to be consistent with the rest of the methods.

(It would come in handy in https://github.com/huggingface/datasets/pull/6269 🙂)


Wauplin (approved):
Makes sense!

PR #1722, opened 2023-10-12
Fix UnboundLocalError when using commit context manager

hahunavth:
References #1721 

### Changes Made:
The error has been fixed by checking the number of files to be staged. If there are fewer than 5 files, a notification is created without causing an error.


Wauplin (approved):
Thanks for opening the PR @hahunavth !

PR #1731, opened 2023-10-12
Properly configures `every` in `test_sync_and_squash_history`

jamesbraza:
I discovered `test_sync_and_squash_history` was improperly configured:
- `every` was set at 1-sec, but its comment said 0.1-sec
- A `time.sleep` below was set for 0.5-sec

This led to a race condition.  This solves one of the CI issues hit in https://github.com/huggingface/huggingface_hub/pull/1727.


Wauplin (approved):
Oops yes, true. Thanks for fixing this @jamesbraza! :hugs:

PR #1775, opened 2023-10-24
Minor doc fixes

pcuenca:
Some small fixes I found while playing with the collections and repo APIs.


Wauplin (approved):
Thanks for noticing and opening a PR @pcuenca ! Looks better now :)

PR #1815, opened 2023-11-10
Faster `HfFileSystem.glob`

lhoestq:
Following https://github.com/huggingface/huggingface_hub/pull/1809

I added the `expand_info`argument to `_ls_tree` instead of `detail` (it always returns dicts anyway) so that glob can still pass `detail=True` but with `expand_info=False`.

I also replaced the `detail` argument in `info()` with this one, since `info()` always return a dict anyway (which is consistent with `detail=True`).

(also let me know if you prefer naming the new argument `expand` instead of `expand_info` for consistency with the Hub API)


src/huggingface_hub/hf_file_system.py
```diff
@@ -368,6 +373,11 @@ def _ls_tree(
                 out.append(cache_path_info)
         return out
 
+    def glob(self, path, maxdepth=None, **kwargs):
+        # Set expand_info=False by default to get a x10 speed boost
+        kwargs = {"expand_info": False, **kwargs}
```
mariosasko, line 378:
```suggestion
        kwargs = {"expand_info": detail or False, **kwargs}
```
This value should be `True` if `detail=True` to be consistent with the spec. 

We need more tricks to optimize the glob logic in `datasets`  anyway.

mariosasko (commented):
`expand_info` is more descriptive, so I slightly prefer it over `expand`

src/huggingface_hub/hf_file_system.py
lhoestq: Good catch ! I fixed it with https://github.com/huggingface/huggingface_hub/pull/1815/commits/1cfb2f19cb2d9f64f21e4ed0e7dd7c8b4fa8519e

mariosasko (approved):
Thanks, LGTM!

Wauplin (approved):
Looks good to me! So this PR is essentially to not fetch details by default in `glob` right? All the rest is only renaming variables?

lhoestq:
Yes exactly :) I think `datasets` will use `glob(..., detail=True, expand_info=False)`

PR #1850, opened 2023-11-21
Fix common path in `_ ls_tree`

lhoestq:
The common path can be one of the directories to refresh.

This was causing a bug where the common path was the parent directory instead, and therefore `glob` would wrongly return content from the parent directory depending on the cache content

now this code returns the correct files

```python
import fsspec
from huggingface_hub import HfFileSystem

fs: HfFileSystem
path = "hf://datasets/severo/bug-hfh-lfs/data/*"

fs = fsspec.get_fs_token_paths(path)[0]
print(fs.glob(path, detail=True))

print("=" * 100)

fs = fsspec.get_fs_token_paths(path)[0]
print(fs.glob(path, detail=True))
```

this bug was introduced in https://github.com/huggingface/huggingface_hub/pull/1809

### DIFF
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -4,6 +4,7 @@ import tempfile
 from collections import deque
 from dataclasses import dataclass
 from datetime import datetime
+from itertools import chain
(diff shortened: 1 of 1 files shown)


mariosasko (approved):
Thanks, LGTM!

PR #1859, opened 2023-11-23
Update modelcard_template.md

EziOzoani:
changes to differentiate between the different uses of hardware
This is Part 1 of the updates to address [this issue](https://github.com/huggingface/hub-docs/issues/1121)


Wauplin (approved):
Thanks for updating @EziOzoani !

As mentioned by @osanseviero in https://github.com/huggingface/hub-docs/pull/1127 this introduces a breaking change but I don't think it's too problematic (I prefer to have a clean template for users).

PR #1860, opened 2023-11-24
Fix IndexError when empty string for credential.helper

SID262000:
This PR addresses #1858

When `credential.helper=` we need to just strip off any whitespaces and return the empty string, as compared to the other scenario `credential.helper=osykeychain` or similar values where we return the first value only considering the fact that `credential.helper` is multivalued.

### DIFF
--- a/src/huggingface_hub/utils/_git_credential.py
+++ b/src/huggingface_hub/utils/_git_credential.py
@@ -39,10 +39,11 @@ def list_credential_helpers(folder: Optional[str] = None) -> List[str]:
         #       Example: `credential.https://huggingface.co.helper=store`
         #       See: https://github.com/huggingface/huggingface_hub/pull/1138#discussion_r1013324508
         for line in output.split("\n"):
-            return sorted(  # Sort for nice printing
+            l_cred_helper = sorted(  # Sort for nice printing
                 # Might have some duplicates
                 re.findall(r"^\s*credential\.helper\s*=\s*(\w*)\s*$", line)
             )
+        return l_cred_helper
     except subprocess.CalledProcessError as exc:
         raise EnvironmentError(exc.stderr)
 
--- a/src/huggingface_hub/utils/_git_credential.py
+++ b/src/huggingface_hub/utils/_git_credential.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Contains utilities to manage Git credentials."""
+import re
 import subprocess
 from typing import List, Optional
 
@@ -38,16 +39,10 @@ def list_credential_helpers(folder: Optional[str] = None) -> List[str]:
         #       Example: `credential.https://huggingface.co.helper=store`
         #       See: https://github.com/huggingface/huggingface_hub/pull/1138#discussion_r1013324508
         for line in output.split("\n"):
-            if "credential.helper" in line:
-                if line.split("=")[-1]:
-                    value = line.split("=")[-1].split()[0]
-                else:
-                    value = line.split("=")[-1].strip()
-        return sorted(  # Sort for nice printing
-            {  # Might have some duplicates
-                value
-            }
-        )
+            return sorted(  # Sort for nice printing
+                # Might have some duplicates
+                re.findall(r"^\s*credential\.helper\s*=\s*(\w*)\s*$", line)
+            )
     except subprocess.CalledProcessError as exc:
         raise EnvironmentError(exc.stderr)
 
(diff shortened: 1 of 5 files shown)


src/huggingface_hub/utils/_git_credential.py
```diff
@@ -37,9 +37,15 @@ def list_credential_helpers(folder: Optional[str] = None) -> List[str]:
...
+        for line in output.split("\n"):
+            if "credential.helper" in line:
+                if line.split("=")[-1]:
+                    value = line.split("=")[-1].split()[0]
+                else:
+                    value = line.split("=")[-1].strip()
         return sorted(  # Sort for nice printing
             {  # Might have some duplicates
-                line.split("=")[-1].split()[0] for line in output.split("\n") if "credential.helper" in line
+                value
```
Wauplin, line 48: We must return a set of all git credential helpers configured on the machine. So here we would need to return a value for each line with `"credential.helper="` in it.

src/huggingface_hub/utils/_git_credential.py
```diff
@@ -37,9 +37,15 @@ def list_credential_helpers(folder: Optional[str] = None) -> List[str]:
         # NOTE: If user has set an helper for a custom URL, it will not we caught here.
         #       Example: `credential.https://huggingface.co.helper=store`
         #       See: https://github.com/huggingface/huggingface_hub/pull/1138#discussion_r1013324508
+        for line in output.split("\n"):
+            if "credential.helper" in line:
+                if line.split("=")[-1]:
+                    value = line.split("=")[-1].split()[0]
+                else:
+                    value = line.split("=")[-1].strip()
```
Wauplin, line 45: Here I think the parsing will break if line = ` credential.helper=  ` since `line.split("=")[-1]` will evaluate to True but `line.split("=")[-1].split()[0]` will fail.

Would you like to try to fix this once and for all using a regex? I started something [here](https://regex101.com/r/dSmWyB/1) that in theory should work for all the cases we want to handle (still has to be tested for real).

image

Wauplin (commented):
Thanks for opening a PR @SID262000! I reviewed and left some comments. Two main things:
- _maybe_ a regex would be easier here to handle all tricky possibilities
- we should still return multiple values => it is possible to configure multiple git credential helpers on a machine

src/huggingface_hub/utils/_git_credential.py
SID262000: I see, thanks for pointing it out.

src/huggingface_hub/utils/_git_credential.py
SID262000: Sure, I'll re-evaluate the parsing.

Regarding the usage of Regex,
It sounds more robust approach, many thanks for putting down the sample regex and sharing it with me. I'll test it for other potential use cases and incorporate the same in my ongoing PR

SID262000:
I modified the above Regex slightly from `"^\s*credential\.helper\s*=\s*(\w+)\s*"` to `"^\s*credential\.helper\s*=\s*(\w*)\s*$"` to capture empty strings as well.
Now the above change captures all the values for `credential.helper=` and fixes it all once and for all using Regex.

src/huggingface_hub/utils/_git_credential.py
```diff
@@ -37,11 +38,12 @@ def list_credential_helpers(folder: Optional[str] = None) -> List[str]:
...
-            {  # Might have some duplicates
-                line.split("=")[-1].split()[0] for line in output.split("\n") if "credential.helper" in line
-            }
-        )
+        for line in output.split("\n"):
+            l_cred_helper = sorted(  # Sort for nice printing
+                # Might have some duplicates
+                re.findall(r"^\s*credential\.helper\s*=\s*(\w*)\s*$", line)
+            )
+        return l_cred_helper
```
Wauplin, line 46: Using `re.findall` is good. However, you don't need to run it on each line one by one. You can run on the whole output at once and use `re.MULTILINE` as flag. Here is how it looks like when you apply it only once:

```suggestion
        return sorted(  # Sort for nice printing
        set(  # Might have some duplicates
            re.findall(
                r"^\s*credential\.helper\s*=\s*(\w+)\s*$",
                output,
                flags=re.MULTILINE | re.IGNORECASE,
            )
        )
    )
```

SID262000: Thanks @Wauplin for the feedback. I've incorporated the above changes as indicated.

PR #1864, opened 2023-11-26
fix some typo

pkking:


Wauplin (approved):
Thanks for noticing it and making a PR to fix this @pkking! Looks nicer now :)

PR #1866, opened 2023-11-27
Update references to hub-docs

mishig25:
Follow up to https://github.com/huggingface/huggingface.js/pull/296

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -258,7 +258,7 @@ class BlobSecurityInfo(TypedDict, total=False):
 class TransformersInfo(TypedDict, total=False):
     auto_model: str
     custom_class: Optional[str]
-    # possible `pipeline_tag` values: https://github.com/huggingface/hub-docs/blob/f2003d2fca9d4c971629e858e314e0a5c05abf9d/js/src/lib/interfaces/Types.ts#L79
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -258,7 +258,7 @@ class BlobSecurityInfo(TypedDict, total=False):
 class TransformersInfo(TypedDict, total=False):
     auto_model: str
     custom_class: Optional[str]
-    # possible `pipeline_tag` values: https://github.com/huggingface/hub-docs/blob/f2003d2fca9d4c971629e858e314e0a5c05abf9d/js/src/lib/interfaces/Types.ts#L79
+    # possible `pipeline_tag` values: https://github.com/huggingface/huggingface.js/blob/3ee32554b8620644a6287e786b2a83bf5caf559c/packages/tasks/src/pipelines.ts#L72
```
julien-c, line 261: update link to main?

mishig25: I followed the previous format, which was referencing a commit. Presumably, because a line was being rererences `#L79`. 

i can do `https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/pipelines.ts#L72`. However, there is a risk of `pipelines.ts` being updated and `#L72` pointing to something else. For simplicity, I can just do  `https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/pipelines.ts` without the line hash. Wdyt ?

julien-c: ah yes ok, that's fine to keep a hash then 👍

Wauplin (approved):
Thanks for the update @mishig25! Agree with keeping hash + line number.

Issue #1872, opened 2023-11-29
`TypeError`: huggingface_hub.hf_file_system.HfFileSystem.find() got multiple values for keyword argument 'maxdepth'


wasertech:
### Describe the bug

I noticed this bug a few times now but could not find anything on it and since it's a bit cryptic I will explain for posterity; if you mismatch the following libs....

```
❯ pip show huggingface-hub
Name: huggingface-hub
Version: 0.20.0.dev0
...
❯ pip show datasets
Name: datasets
Version: 2.15.0
...
```

...and try to load a dataset, it will raise the following `TypeError`:
> huggingface_hub.hf_file_system.HfFileSystem.find() got multiple values for keyword argument 'maxdepth'

### Reproduction

```python
ds = load_dataset("wasertech/AGI", split="train") # can be any set
```

### Logs

```shell
Downloading readme: 100%|███████████████████████████████████████████████████████████████████████████████| 23.6k/23.6k [00:00<00:00, 16.3MB/s]
Traceback (most recent call last):
...
File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/assistant/nlp/chains/schema/vectorstore.py", line 82, in get_guidebook_docs
    guidebook = load_dataset("wasertech/AGI", split="train")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/datasets/load.py", line 2128, in load_dataset
    builder_instance = load_dataset_builder(
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/datasets/load.py", line 1814, in load_dataset_builder
    dataset_module = dataset_module_factory(
                     ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/datasets/load.py", line 1511, in dataset_module_factory
    raise e1 from None
  File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/datasets/load.py", line 1495, in dataset_module_factory
    ).get_module()
      ^^^^^^^^^^^^
  File "/home/waser/Projets/Assistant/venv/lib/python3.11/site-packages/datasets/load.py", line 1047, in get_module
    data_files = DataFilesDict.from_patterns(
[...]

PR #1875, opened 2023-11-29
Remove `maxdepth` param from `HfFileSystem.glob`

mariosasko:
`maxdepth` was added to the default `fsspec`'s `glob` in version `2023.9.0`, so this PR removes `maxdepth` from the `HfFileSystem.glob`'s signature to avoid an error in the older `fsspec` versions.

Fix https://github.com/huggingface/huggingface_hub/issues/1872


lhoestq (approved):
good catch !

Wauplin (approved):
Thanks @mariosasko :+1:

PR #1957, opened 2024-01-04
[HfFileSystem] Fix `rm` on branch

lhoestq:
This code returns EntryNotFoundError even if the file exists

```python
fs.rm("datasets/username/datasetname@branch/data.txt")
```

and would only work on `main`.


Wauplin (commented):
Thanks for reporting and fixing this @lhoestq! Do you mind adding a regression test for it in `test_hf_file_system.py`? Thanks in advance!

mariosasko (approved):
LGTM, thanks!

Wauplin (approved):
Thanks!

PR #1958, opened 2024-01-04
Update the documentation of add_collection_item

FremyCompany:
Based on feedback from @justalphie


Wauplin (approved):
Thanks for opening this PR @FremyCompany! Well documented!

PR #1971, opened 2024-01-12
Finished migration from `setup.cfg` to `pyproject.toml`

jamesbraza:
Most of the configs in `setup.cfg` were all unused/stale as of https://github.com/huggingface/huggingface_hub/pull/1323 almost a year ago -- replaced with `ruff`. So this PR:
- Migrates `pytest` config to `pyproject.toml`
- Removes mostly unused `setup.cfg`
- Removes old `flake8: noqa` comment in favor of `ruff: noqa`


Wauplin (approved):
Thanks a lot for cleaning this up @jamesbraza! Changes looks good to me :) 
We kept stale config for isort/flake for contributors to slowly moved out from them in favor of ruff. But should be good by now :smile:

PR #1983, opened 2024-01-16
fix: actual error won't be raised while `force_download=True`

scruel:


Wauplin (approved):
Looks good! Thanks for iterating on it :hugs:

Wauplin (approved):
Thanks for making the changes @scruel! Looks good to me now :tada:

PR #1987, opened 2024-01-17
Newer `pre-commit`

jamesbraza:
- Ran `pre-commit autoupdate` for newer versions
- Updated `pre-commit` config to ignore `tests/cassettes/` from YAML checks
- Ran `pre-commit` on all files


Wauplin (approved):
Thanks for the cleaning PR @jamesbraza! Will update my pre-commit hook then :)

PR #1990, opened 2024-01-18
Removing unused `SETUP_CFG` path variable

jamesbraza:
https://github.com/huggingface/huggingface_hub/pull/1971 removed `setup.cfg`, this PR just removes an unused `Path` variable for it

### DIFF
--- a/utils/check_static_imports.py
+++ b/utils/check_static_imports.py
@@ -26,7 +26,6 @@ from huggingface_hub import _SUBMOD_ATTRS
 
 
 INIT_FILE_PATH = Path(__file__).parents[1] / "src" / "huggingface_hub" / "__init__.py"
-SETUP_CFG_PATH = Path(__file__).parents[1] / "setup.cfg"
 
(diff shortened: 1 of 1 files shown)


Wauplin (approved):
Thanks @jamesbraza! Failed test is unrelated indeed. I'll have to investigate it. It's not a flaky test, it's also failing in other PRs. I suspect a version update in keras/tensorflow. Anyway, not related so let's merge this!

jamesbraza:
Yeah looking at the `pip freeze` of the failed CI job:

```none
...
tb-nightly==2.16.0a20240118
tensorboard-data-server==0.7.2
tensorflow==2.16.0
tensorflow-io-gcs-filesystem==0.35.0
...
tf-estimator-nightly==2.14.0.dev2023080308
tf_keras-nightly==2.16.0.dev2024011811
...
```

For some reason it's installing the nightly version of 2.16, which doesn't make sense to me because that's not what `setup.py` calls out for the `tensorflow` extra.  The only place `tf-nightly` is mentioned is in `utils/_runtime.py`.

Anyways, feel free to tag me on a PR for the fix, if you can figure it out

PR #1991, opened 2024-01-18
`make style` using `pre-commit`

jamesbraza:
This PR makes a few changes for greater repo tooling synergy

- Moves CI to parity with local machines by reusing `make style`
- Expands `pre-commit` to match `make style`, then moves `make style` to invoke `pre-commit`
- Adds missing `pre-commit` to dev dependencies

Benefits:

- Brings `huggingface_hub` tooling closer to standard within Python community
- `pre-commit` will skip over unnecessary files, unlike old `make style`
- `pre-commit` config is no longer "lagging", it's sync'd with repo via CI

### DIFF
--- a/.github/workflows/python-quality.yml
+++ b/.github/workflows/python-quality.yml
@@ -27,11 +27,7 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install .[dev]
-      - run: ruff check tests src contrib # linter
-      - run: ruff format --check tests src contrib # formatter
-      - run: python utils/check_contrib_list.py
-      - run: python utils/check_static_imports.py
-      - run: python utils/generate_async_inference_client.py
+      - run: make style
 
       # Run type checking at least on huggingface_hub root file to check all modules
       # that can be lazy-loaded actually exist.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,7 @@
+---
+default_language_version:
+    python: python3
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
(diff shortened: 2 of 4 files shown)


jamesbraza:
To be clear, if the hugging face team isn't interested in this, I am not attached to it. I just thought there's some opportunities for greater synergy within the repo

Wauplin (commented):
Thanks for opening this PR (and discussion) @jamesbraza . The precommit config is indeed a bit outdated (main reason being that I don't use it myself). 

> Moves CI to parity with local machines by reusing make style

For the record, the "why?" we don't have parity in CI is because having them separated should make it easier to check which step failed. The steps are the same but separated in github action steps instead of 1 single command. Anyway, I'm not against harmonizing this (no big deal to have to open to see the error).

However, the CI shouldn't be aligned with `make style` but rather with `make quality`. The former modifies the files inplace while the later is a checker that will fail in case of wrong unexpected format (+ make quality runs mypy).

> Expands pre-commit to match make style, then moves make style to invoke pre-commit

I've made some tests from your branch and at least for me the `python -m pre_commit run --all-files` command takes ~2s to complete while `make style` takes ~0.20s on `main`. I don't know what's causing this decrease in speed but it's quite degrading the developer experience.

Would it be possible to tell `pre-commit` to run `make style` (and not the other way around) so that precommit hook/make style/CI are aligned but without changing the `make style` usage that is very fast and well configured already.

> Adds missing pre-commit to dev dependencies

I'm fine with this if the above can be fixed :)

jamesbraza:
Thanks for the comments.

I think the 1.8-sec overhead of `pre-commit` may come from its `--all-files` checking for `.gitignore`'d files, evaluating file type filtration, etc. In other words, `pre-commit` adds logic "on top" of the vanilla tools, whereas the current `make style` just directly invokes the tools.

> Would it be possible to tell `pre-commit` to run `make style` (and not the other way around)? So that precommit hook/make style/CI are aligned but without changing the `make style` usage that is very fast and well configured already.

I think with this route (redirecting to `make style`), it loses benefits of `pre-commit` such as file filtration based on Git, version management (e.g. `pre-commit autoupdate`), centralized tool configuration, and auto-integration with `git commit`, which is a core part of this PR. For example, to add `toml-sort`, it would be another tool in the `pre-commit` chain.

Anyways, I don't want to push sizable tooling changes that aren't desired by the core developers, so ima just close this out

PR #1995, opened 2024-01-19
[PyTorchModelHubMixin] Improve docs

NielsRogge:
This PR improves the documentation of PyTorchModelHubMixin, and also discusses some potential improvements.

To do:

- [x] add safetensors serialization by default
- [x] add `max_shard_size` support
- [ ] basically copy whatever is needed from https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2201

### DIFF
--- a/src/huggingface_hub/hub_mixin.py
+++ b/src/huggingface_hub/hub_mixin.py
@@ -1,16 +1,25 @@
+import collections
 import json
 import os
+import re
 from pathlib import Path
-from typing import Dict, List, Optional, Type, TypeVar, Union
+from typing import Dict, List, Optional, Tuple, Type, TypeVar, Union
+
+
+if is_safetensors_available():
+    from safetensors.torch import save_file as safe_save_file
+    from safetensors.torch import storage_ptr, storage_size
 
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hub_mixin.py
```diff
@@ -289,37 +289,48 @@ def push_to_hub(
...
 
-
     >>> class MyModel(nn.Module, PyTorchModelHubMixin):
-    ...     def __init__(self):
+    ...     def __init__(self, config):
     ...         super().__init__()
-    ...         self.param = nn.Parameter(torch.rand(3, 4))
-    ...         self.linear = nn.Linear(4, 5)
+    ...         self.param = nn.Parameter(torch.rand(num_channels, hidden_size))
+    ...         self.linear = nn.Linear(hidden_size, num_classes)
```
NielsRogge, line 312: cc @Wauplin it would be great if we could do things like `config.num_channels`, `config.hidden_size` (i.e. access all config keys as attributes). For that, I assume, we would need to make sure models do super()__init__(config)

PR #2000, opened 2024-01-25
Disable tqdm progress bar if no TTY attached

mssalvatore:
When dockerized applications write to STDOUT/STDERR, the applications can block due to logging back pressure (see
https://docs.docker.com/config/containers/logging/configure/#configure-the-delivery-mode-of-log-messages-from-container-to-log-driver6

HuggingFace's TGI container is one such example (see https://github.com/huggingface/text-generation-inference/issues/1186).

Setting tqdm's `disable=None` will disable the progress bar if no tty is attached and help to resolve TGI's issue #1186.

References:
    https://github.com/huggingface/text-generation-inference/issues/1186#issuecomment-1906809624
    https://github.com/huggingface/text-generation-inference/issues/1186#issuecomment-1908278377

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -505,7 +505,9 @@ def http_get(
         total=total,
         initial=resume_size,
         desc=displayed_filename,
-        disable=bool(logger.getEffectiveLevel() == logging.NOTSET),
+        disable=True if (logger.getEffectiveLevel() == logging.NOTSET) else None,
+        # ^ set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
+        # see https://github.com/huggingface/huggingface_hub/pull/2000
     ) as progress:
         if hf_transfer and total is not None and total > 5 * DOWNLOAD_CHUNK_SIZE:
             supports_callback = "callback" in inspect.signature(hf_transfer.download).parameters
--- a/src/huggingface_hub/lfs.py
+++ b/src/huggingface_hub/lfs.py
@@ -409,7 +409,10 @@ def _upload_parts_hf_transfer(
     desc = operation.path_in_repo
     if len(desc) > 40:
         desc = f"(…){desc[-40:]}"
-    disable = bool(logger.getEffectiveLevel() == logging.NOTSET)
+
+    # set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
+    # see https://github.com/huggingface/huggingface_hub/pull/2000
+    disable = True if (logger.getEffectiveLevel() == logging.NOTSET) else None
 
     with tqdm(unit="B", unit_scale=True, total=total, initial=0, desc=desc, disable=disable) as progress:
         try:


Wauplin:
Hi @mssalvatore, thanks for investigating this issue! Just to be sure, the solution for TGI would be to disable progress bars when downloading? For what I understand the PR simply change that if `logger.getEffectiveLevel() != logging.NOTSET` then the `disable` value is set to `None` instead of `False`. Is that the expected fix?

For your info, disabling progress bars can be done by setting [`HF_HUB_DISABLE_PROGRESS_BARS=1`](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfhubdisableprogressbars) as an environment variable. Maybe that could be a temporary fix for you before figuring out at which level we want to fix this (I'm still not so sure if this is a `huggingface_hub` bug).

pinging @OlivierDehaene @Narsil who maintain TGI

mssalvatore:
@Wauplin, yes, that's the intended fix. It preserves the current behavior which is that the progress bar is disabled if `logger.getEffectiveLevel() != logging.NOTSET`. When the `disable` value is set to `None`, the bar is disabled if no TTY is attached. I'll give `HF_HUB_DISABLE_PROGRESS_BARS` a try and see if it resolves the issue.

Regarding whether or not `huggingface_hub` should accept this patch: is there a reason you would ever want the progress bar to be enabled if no TTY were attached?

Wauplin:
TIL from [tqdm's docs](https://tqdm.github.io/docs/tqdm/):

image

So yes, definitely agree with this PR! I wasn't aware of the nuance between `None` and `False` is this particular case.

src/huggingface_hub/file_download.py
```diff
@@ -505,7 +505,7 @@ def http_get(
         total=total,
         initial=resume_size,
         desc=displayed_filename,
-        disable=bool(logger.getEffectiveLevel() == logging.NOTSET),
+        disable=True if (logger.getEffectiveLevel() == logging.NOTSET) else None,
```
Wauplin, line 508:
```suggestion
        disable=True if (logger.getEffectiveLevel() == logging.NOTSET) else None,
        # ^ set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
        # see https://github.com/huggingface/huggingface_hub/pull/2000
```

^ could you add a comment that `disable=None` has an implicit meaning in this context?

Wauplin (commented):
Thanks @mssalvatore for finding out about this and opening this PR! The change looks good to me now (didn't know about this default behavior before). Could you also update the tqdm in [lfs.py](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/lfs.py#L412)  as well (same fix + same comment). Thanks in advance!

src/huggingface_hub/file_download.py
mssalvatore: Sure

mssalvatore: Does the project have a limit on line length?

mssalvatore: I pushed the fix for `lfs.py`. My question is whether or not it would be preferable to make this change in `utils/tqdm.py` instead. The question is whether you'd rather force `True or None` to be the behavior. This has the benefit of ensuring this bug doesn't creep in somewhere else, but the drawback of changing the interface of `tqdm` throughout `huggingface_hub`.

```diff
diff --git a/src/huggingface_hub/utils/tqdm.py b/src/huggingface_hub/utils/tqdm.py
index cae9adb..c3dddad 100644
--- a/src/huggingface_hub/utils/tqdm.py
+++ b/src/huggingface_hub/utils/tqdm.py
@@ -129,6 +129,10 @@ class tqdm(old_tqdm):
     def __init__(self, *args, **kwargs):
         if are_progress_bars_disabled():
             kwargs["disable"] = True
+
+        if kwargs["disable"] is False:
+            kwargs["disable"] = None
+
         super().__init__(*args, **kwargs)
 
     def __delattr__(self, attr: str) -> None:
```

Wauplin: I would prefer not to do that, just to keep the `disable=False` behavior possible (can't find a useful use case but you never know). Thanks for suggesting though, we might reevaluate in the future.

Wauplin (approved):
Thanks for making the changes @mssalvatore ! Failed tests in CI are unrelated so let's get it merge :)

lllyasviel:
This just wasted me 3 hours to debug what caused by progress bar disappeared. Whoever merge this must be completely nonsense. Please do consider revert this as soon as possible!

Wauplin:
> Whoever merge this must be completely nonsense. Please do consider revert this as soon as possible!

Hi, merger here :wave:  What was your setup exactly and what solution would you suggest to address the issue described in the PR description?

juhyeon-cha:
> > Whoever merge this must be completely nonsense. Please do consider revert this as soon as possible!
> 
> Hi, merger here 👋 What was your setup exactly and what solution would you suggest to address the issue described in the PR description?

Hugging Face already has a feature to disable tqdm by name. Is there a need to separately disable based on TTY status?

PR #2003, opened 2024-01-26
Fix grammar

Anthonyg5005:
Line 105 had a grammar mistake in file ‘src/huggingface_hub/_login.py’


Wauplin (approved):
Thanks for noticing and opening a PR for it @Anthonyg5005!

PR #2013, opened 2024-02-04
Fix doc

jordane95:


Wauplin (approved):
Good catch! Thanks for opening the PR @jordane95 :heart:

PR #2028, opened 2024-02-15
Deprecate `ModelFilter/DatasetFilter`

druvdub:
Related to #1676 

Moved all attributes from ModelFilter class to list_models arguments and added a filter query builder to build a query. Updated list_models to take both filter as an Iterable and also update the filter list based on other arguments in the method

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1380,6 +1380,12 @@ class HfApi:
         *,
         filter: Union[ModelFilter, str, Iterable[str], None] = None,
         author: Optional[str] = None,
+        library: Optional[Union[str, List[str]]] = None,
+        language: Optional[Union[str, List[str]]] = None,
+        model_name: Optional[str] = None,
+        task: Optional[Union[str, List[str]]] = None,
+        trained_dataset: Optional[Union[str, List[str]]] = None,
+        tags: Optional[Union[str, List[str]]] = None,
         search: Optional[str] = None,
         emissions_thresholds: Optional[Tuple[float, float]] = None,
         sort: Union[Literal["last_modified"], str, None] = None,
@@ -1389,12 +1395,6 @@ class HfApi:
         cardData: bool = False,
         fetch_config: bool = False,
         token: Optional[Union[bool, str]] = None,
-        library: Optional[Union[str, List[str]]] = None,
-        language: Optional[Union[str, List[str]]] = None,
-        model_name: Optional[str] = None,
-        task: Optional[Union[str, List[str]]] = None,
-        trained_dataset: Optional[Union[str, List[str]]] = None,
-        tags: Optional[Union[str, List[str]]] = None,
         pipeline_tag: Optional[str] = None,
     ) -> Iterable[ModelInfo]:
         """
@@ -1407,6 +1407,24 @@ class HfApi:
             author (`str`, *optional*):
                 A string which identify the author (user or organization) of the
                 returned models
+            library (`str` or `List`, *optional*):
+                A string or list of strings of foundational libraries models were
+                originally trained from, such as pytorch, tensorflow, or allennlp.
+            language (`str` or `List`, *optional*):
+                A string or list of strings of languages, both by name and country
+                code, such as "en" or "English"
+            model_name (`str`, *optional*):
+                A string that contain complete or partial names for models on the
+                Hub, such as "bert" or "bert-base-cased"
+            task (`str` or `List`, *optional*):
+                A string or list of strings of tasks models were designed for, such
+                as: "fill-mask" or "automatic-speech-recognition"
+            trained_dataset (`str` or `List`, *optional*):
+                A string tag or a list of string tags of the trained dataset for a
+                model on the Hub.
+            tags (`str` or `List`, *optional*):
+                A string tag or a list of tags to filter models on the Hub by, such
+                as `text-generation` or `spacy`.
             search (`str`, *optional*):
                 A string that will be contained in the returned model ids.
             emissions_thresholds (`Tuple`, *optional*):
@@ -1437,24 +1455,6 @@ class HfApi:
                 If `None` or `True` and machine is logged in (through `huggingface-cli login`
                 or [`~huggingface_hub.login`]), token will be retrieved from the cache.
                 If `False`, token is not sent in the request header.
-            library (`str` or `List`, *optional*):
-                A string or list of strings of foundational libraries models were
-                originally trained from, such as pytorch, tensorflow, or allennlp.
-            language (`str` or `List`, *optional*):
-                A string or list of strings of languages, both by name and country
-                code, such as "en" or "English"
-            model_name (`str`, *optional*):
-                A string that contain complete or partial names for models on the
-                Hub, such as "bert" or "bert-base-cased"
-            task (`str` or `List`, *optional*):
-                A string or list of strings of tasks models were designed for, such
-                as: "fill-mask" or "automatic-speech-recognition"
-            trained_dataset (`str` or `List`, *optional*):
-                A string tag or a list of string tags of the trained dataset for a
-                model on the Hub.
-            tags (`str` or `List`, *optional*):
-                A string tag or a list of tags to filter models on the Hub by, such
-                as `text-generation` or `spacy`.
             pipeline_tag (`str`, *optional*):
                 A string pipeline tag to filter models on the Hub by, such as `summarization`
 
@@ -1500,32 +1500,44 @@ class HfApi:
             raise ValueError("`emissions_thresholds` were passed without setting `cardData=True`.")
 
         path = f"{self.endpoint}/api/models"
+        model_str = ""
         headers = self._build_hf_headers(token=token)
         params = {}
+        filter_list = []
 
         if filter is not None:
             if isinstance(filter, ModelFilter):
                 params = self._unpack_model_filter(filter)
             else:
-                params = self._build_model_filter(
-                    author=author,
-                    library=library,
-                    language=language,
-                    model_name=model_name,
-                    task=task,
-                    trained_dataset=trained_dataset,
-                    tags=tags,
-                )
-
-                if isinstance(filter, str):
-                    params.update({"filter": params.get("filter", set()).add(filter)})
-                else:
-                    params.update({"filter": params.get("filter", set()).update(filter)})
+                params.update({"filter": filter})
 
             params.update({"full": True})
-        if author is not None:
+
+        # Build the filter list
+        if author:
+            model_str = f"{author}/"
             params.update({"author": author})
-        if search is not None:
+        if model_name:
+            model_str += model_name
+        if library:
+            filter_list.extend([library] if isinstance(library, str) else library)
+        if task:
+            filter_list.extend([task] if isinstance(task, str) else task)
+        if trained_dataset:
+            if not isinstance(trained_dataset, (list, tuple)):
+                trained_dataset = [trained_dataset]
+            for dataset in trained_dataset:
+                if not dataset.startswith("dataset:"):
+                    dataset = f"dataset:{dataset}"
+                filter_list.append(dataset)
+        if language:
+            filter_list.extend([language] if isinstance(language, str) else language)
+        if tags:
+            filter_list.extend([tags] if isinstance(tags, str) else tags)
+        if model_str:
+            params.update({"search": model_str})
+
+        if search:
             params.update({"search": search})
         if sort is not None:
             params.update({"sort": "lastModified" if sort == "last_modified" else sort})
@@ -1544,6 +1556,11 @@ class HfApi:
             params.update({"cardData": True})
         if pipeline_tag:
             params.update({"pipeline_tag": pipeline_tag})
+
+        filter_value = params.get("filter", [])
+        if filter_value:
+            filter_list.extend([filter_value] if isinstance(filter_value, str) else list(filter_value))
+        params.update({"filter": filter_list})
 
         # `items` is a generator
         items = paginate(path, params=params, headers=headers)
@@ -1603,64 +1620,6 @@ class HfApi:
         elif isinstance(model_filter.language, str):
             filter_list.append(model_filter.language)
         query_dict["filter"] = tuple(filter_list)
-        return query_dict
-
-    def _build_model_filter(
-        self,
-        author: Optional[str] = None,
-        library: Optional[Union[str, List[str]]] = None,
-        language: Optional[Union[str, List[str]]] = None,
-        model_name: Optional[str] = None,
-        task: Optional[Union[str, List[str]]] = None,
-        trained_dataset: Optional[Union[str, List[str]]] = None,
-        tags: Optional[Union[str, List[str]]] = None,
-    ) -> Dict[str, Any]:
-        """
-        Builds a filter query for the list_models methods.
-        """
-        model_str = ""
-        filter_list: List[str] = []
-
-        # Handling author
-        if author:
-            model_str = f"{author}/"
-
-        # Handling model_name
-        if model_name:
-            model_str += model_name
-
-        # Handling library
-        if library:
-            filter_list.extend([library] if isinstance(library, str) else library)
-
-        # Handling tasks
-        if task:
-            filter_list.extend([task] if isinstance(task, str) else task)
-
-        # Handling trained_dataset
-        if trained_dataset:
-            if not isinstance(trained_dataset, (list, tuple)):
-                trained_dataset = [trained_dataset]
-            for dataset in trained_dataset:
-                if not dataset.startswith("dataset:"):
-                    dataset = f"dataset:{dataset}"
-                filter_list.append(dataset)
-
-        # Handling language
-        if language:
-            filter_list.extend([language] if isinstance(language, str) else language)
-
-        # Handling tags
-        if tags:
-            filter_list.extend([tags] if isinstance(tags, str) else tags)
-
-        query_dict: Dict[str, Any] = {}
-
-        if model_str:
-            query_dict["search"] = model_str
-        if filter_list:
-            query_dict["filter"] = set(filter_list)
-
         return query_dict
 
     @validate_hf_hub_args
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1538,8 +1538,13 @@ class HfApi:
                 params = self._unpack_model_filter(filter)
             else:
                 params = self._build_filter_query(
-                    author=author, library=library, language=language, model_name=model_name,
-                    task=task, trained_dataset=trained_dataset, tags=tags
+                    author=author,
+                    library=library,
+                    language=language,
+                    model_name=model_name,
+                    task=task,
+                    trained_dataset=trained_dataset,
+                    tags=tags,
                 )
 
                 if isinstance(filter, str):
@@ -1637,12 +1642,15 @@ class HfApi:
         return query_dict
 
     def _build_filter_query(
-            self,
-            author: Optional[str] = None, library: Optional[Union[str, List[str]]] = None,
-            language: Optional[Union[str, List[str]]] = None, model_name: Optional[str] = None,
-            task: Optional[Union[str, List[str]]] = None, trained_dataset: Optional[Union[str, List[str]]] = None,
-            tags: Optional[Union[str, List[str]]] = None
-    ) -> Dict[str, Union[str, Tuple[str, ...]]]:
+        self,
+        author: Optional[str] = None,
+        library: Optional[Union[str, List[str]]] = None,
+        language: Optional[Union[str, List[str]]] = None,
+        model_name: Optional[str] = None,
+        task: Optional[Union[str, List[str]]] = None,
+        trained_dataset: Optional[Union[str, List[str]]] = None,
+        tags: Optional[Union[str, List[str]]] = None,
+    ) -> Dict[str, Any]:
         """
         Builds a filter query for the list_models methods.
         """
@@ -1659,15 +1667,12 @@ class HfApi:
 
         # Handling library
         if library:
-            filter_list.extend(
-                [library] if isinstance(library, str) else library
-            )
+            filter_list.extend([library] if isinstance(
+                library, str) else library)
 
         # Handling tasks
         if task:
-            filter_list.extend(
-                [task] if isinstance(task, str) else task
-            )
+            filter_list.extend([task] if isinstance(task, str) else task)
 
         # Handling trained_dataset
         if trained_dataset:
@@ -1680,17 +1685,14 @@ class HfApi:
 
         # Handling language
         if language:
-            filter_list.extend(
-                [language] if isinstance(language, str) else language
-            )
+            filter_list.extend([language] if isinstance(
+                language, str) else language)
 
         # Handling tags
         if tags:
-            filter_list.extend(
-                [tags] if isinstance(tags, str) else tags
-            )
+            filter_list.extend([tags] if isinstance(tags, str) else tags)
 
-        query_dict: Dict[str, Union[str, Tuple[str, ...]]] = {}
+        query_dict: Dict[str, Any] = {}
 
         if model_str:
             query_dict["search"] = model_str
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1628,12 +1628,6 @@ class HfApi:
         *,
         filter: Union[DatasetFilter, str, Iterable[str], None] = None,
         author: Optional[str] = None,
-        search: Optional[str] = None,
-        sort: Union[Literal["last_modified"], str, None] = None,
-        direction: Optional[Literal[-1]] = None,
-        limit: Optional[int] = None,
-        full: Optional[bool] = None,
-        token: Optional[str] = None,
         benchmark: Optional[Union[str, List[str]]] = None,
         dataset_name: Optional[str] = None,
         language_creators: Optional[Union[str, List[str]]] = None,
@@ -1642,6 +1636,12 @@ class HfApi:
         size_categories: Optional[Union[str, List[str]]] = None,
         task_categories: Optional[Union[str, List[str]]] = None,
         task_ids: Optional[Union[str, List[str]]] = None,
+        search: Optional[str] = None,
+        sort: Optional[Union[Literal["last_modified"], str]] = None,
+        direction: Optional[Literal[-1]] = None,
+        limit: Optional[int] = None,
+        full: Optional[bool] = None,
+        token: Optional[str] = None,
     ) -> Iterable[DatasetInfo]:
         """
         List datasets hosted on the Huggingface Hub, given some filters.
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -1469,23 +1526,37 @@ def list_models(
...
         path = f"{self.endpoint}/api/models"
         headers = self._build_hf_headers(token=token)
         params = {}
+
         if filter is not None:
             if isinstance(filter, ModelFilter):
                 params = self._unpack_model_filter(filter)
             else:
-                params.update({"filter": filter})
+                params = self._build_filter_query(
```
druvdub, line 1540: Waiting to confirm on how to handle different scenarios for filter

Wauplin:
Hi @druvdub! Thanks for the PR :heart:  I've tried to answer all of your questions in https://github.com/huggingface/huggingface_hub/issues/1676#issuecomment-1945594952. Please let me know if some parts are still unclear :)

Regarding this PR, would it be possible to fix the styling? The PR change diff is very big at the moment due to some whitespaces/newlines added or removed. This makes the PR quite difficult to review. You can fix this with the following steps:
1. install the dev dependencies:
```
pip install -e ".[dev]"
```
2. run the `make style` command (which runs ruff linter and formatter + a few scripts) 
```
make style
``` 
3. (optional) once done, you should be able to run `run quality` without any error (checks styling + type annotations):
```
make quality
```

I believe the current changes were due to a "autoformat" or "format on save" feature in your IDE. Make sure to always run `make style` before committing changes :) Thanks in advance!

druvdub:
> Hi @druvdub! Thanks for the PR ❤️ I've tried to answer all of your questions in [#1676 (comment)](https://github.com/huggingface/huggingface_hub/issues/1676#issuecomment-1945594952). Please let me know if some parts are still unclear :)
> 
> Regarding this PR, would it be possible to fix the styling? The PR change diff is very big at the moment due to some whitespaces/newlines added or removed. This makes the PR quite difficult to review. You can fix this with the following steps:
> 
> 1. install the dev dependencies:
> 
> ```
> pip install -e ".[dev]"
> ```
> 
> 2. run the `make style` command (which runs ruff linter and formatter + a few scripts)
> 
> ```
> make style
> ```
> 
> 3. (optional) once done, you should be able to run `run quality` without any error (checks styling + type annotations):
> 
> ```
> make quality
> ```
> 
> I believe the current changes were due to a "autoformat" or "format on save" feature in your IDE. Make sure to always run `make style` before committing changes :) Thanks in advance!

Yup no worries, for some reason the makefile did not work for some reason, So I had to now manually run the ruff checks

src/huggingface_hub/hf_api.py
```diff
@@ -340,7 +340,8 @@ class CommitInfo(str):
     pr_num: Optional[str] = field(init=False)
 
     # legacy url for `str` compatibility (ex: url to uploaded file, url to uploaded folder, url to PR, etc.)
-    _url: str = field(repr=False, default=None)  # type: ignore  # defaults to `commit_url`
+    # type: ignore  # defaults to `commit_url`
+    _url: str = field(repr=False, default=None)
```
druvdub, line 344: FYI: `mypy src` raises an error here. Not sure if I should address it

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -221,3 +236,10 @@ class ModelFilter:
     task: Optional[Union[str, List[str]]] = None
     trained_dataset: Optional[Union[str, List[str]]] = None
     tags: Optional[Union[str, List[str]]] = None
+
+    def __post_init__(self):
+        warnings.warn(
+            "ModelFilter is being deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_models` method.",
+            FutureWarning,
```
druvdub, line 243: Updated the category from `DeprecatedWarning` to `FutureWarning` because 
> DeprecatedWarning: Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).
i.e. you won't get a warning in interepreter if you use the class as it is, unless we update the filters for warning.

src/huggingface_hub/hf_api.py
```diff
@@ -1654,7 +1794,22 @@ def list_datasets(
...
+                    multilinguality=multilinguality,
+                    size_categories=size_categories,
+                    task_categories=task_categories,
+                    task_ids=task_ids,
+                )
+
+                if isinstance(filter, str):
+                    params.update({"filter": params.get("filter", set()).add(filter)})
+                else:
+                    params.update({"filter": params.get("filter", set()).update(filter)})
```
druvdub, line 1812: Used set to avoid duplication in cases where a user passes both keyword argument and a filter
e.g.
```python

filter="language:en"
language="language:en"
```
can revert it to a tuple if needed

druvdub:
@Wauplin I think this should be all in terms of functionality changes. Should I update READMEs and completely remove `ModelFilter/DatasetFilter`.

And finally in terms of test changes should I just update the existing tests to take arguments instead of `ModelFilter/DatasetFilter`

Wauplin:
Thanks for all the modifications and taking care of the special cases @druvdub! I'll have a look at it on tomorrow if that's fine with you :)

druvdub:
> Thanks for all the modifications and taking care of the special cases @druvdub! I'll have a look at it on tomorrow if that's fine with you :) 

Yea sounds good to me :)

src/huggingface_hub/hf_api.py
Wauplin, line 344:
```suggestion
    _url: str = field(repr=False, default=None)  # type: ignore  # defaults to `commit_url`
```
This should be on a single-line to avoid mypy complaining

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -73,6 +74,10 @@ class DatasetFilter:
     compatible with the REST API. For all parameters capitalization does not
     matter.
 
+    <Tip warning={true}>
+        The `DatasetFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the [`list_datasets`].
+    </Tip>
```
Wauplin, line 79:
```suggestion
    <Tip warning={true}>
        The `DatasetFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_datasets`].
    </Tip>
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -151,6 +156,13 @@ class DatasetFilter:
     task_categories: Optional[Union[str, List[str]]] = None
     task_ids: Optional[Union[str, List[str]]] = None
 
+    def __post_init__(self):
+        warnings.warn(
+            "DatasetFilter is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_datasets` method.",
+            category=FutureWarning,
+            stacklevel=2,
```
Wauplin, line 163:
```suggestion
```
no need for that (all our deprecation warning use default stacklevel)

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -159,6 +171,10 @@ class ModelFilter:
     compatible with the REST API. For all parameters capitalization does not
     matter.
 
+    <Tip warning={true}>
+        The `ModelFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the [`list_models`].
+    </Tip>
```
Wauplin, line 176:
```suggestion
    <Tip warning={true}>
        The `ModelFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_models`].
    </Tip>
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -221,3 +238,10 @@ class ModelFilter:
     task: Optional[Union[str, List[str]]] = None
     trained_dataset: Optional[Union[str, List[str]]] = None
     tags: Optional[Union[str, List[str]]] = None
+
+    def __post_init__(self):
+        warnings.warn(
+            "ModelFilter is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_models` method.",
+            FutureWarning,
+            stacklevel=2,
```
Wauplin, line 246:
```suggestion
```
(same as above)

src/huggingface_hub/utils/endpoint_helpers.py
Wauplin: Yes that's perfect! This is also what we use in [`./utils/_deprecate.py`](https://github.com/huggingface/huggingface_hub/blob/133bf03ebf22e33f558a3a4041e4ec4cca6e9e00/src/huggingface_hub/utils/_deprecation.py) (no need to use these utils here). Nice idea to use the `__post_init__` method to deprecate dataclasses :)

src/huggingface_hub/hf_api.py
```diff
@@ -1389,6 +1390,13 @@ def list_models(
         cardData: bool = False,
         fetch_config: bool = False,
         token: Optional[Union[bool, str]] = None,
+        library: Optional[Union[str, List[str]]] = None,
+        language: Optional[Union[str, List[str]]] = None,
+        model_name: Optional[str] = None,
+        task: Optional[Union[str, List[str]]] = None,
+        trained_dataset: Optional[Union[str, List[str]]] = None,
+        tags: Optional[Union[str, List[str]]] = None,
+        pipeline_tag: Optional[str] = None,
```
Wauplin, line 1399: I would move those arguments up, between `author` and `search` => this way we start with all possible filters and then have the sort/direction/limit/full/etc. options.
(same below in docstring)

src/huggingface_hub/hf_api.py
```diff
@@ -1474,11 +1503,26 @@ def list_models(
         path = f"{self.endpoint}/api/models"
         headers = self._build_hf_headers(token=token)
         params = {}
+
         if filter is not None:
             if isinstance(filter, ModelFilter):
                 params = self._unpack_model_filter(filter)
             else:
-                params.update({"filter": filter})
+                params = self._build_model_filter(
```
Wauplin, line 1511: I think it's best to have all the `_build_model_filter` inside `list_models`. Once `ModelFilter` are completely removed (and therefore `_unpack_model_filter`), we will have everything in one place which will be more consistent that the current code. Same way as you handled `pipeline_tag` below

src/huggingface_hub/hf_api.py
```diff
@@ -1572,6 +1676,14 @@ def list_datasets(
...
         full: Optional[bool] = None,
         token: Optional[str] = None,
+        benchmark: Optional[Union[str, List[str]]] = None,
+        dataset_name: Optional[str] = None,
+        language_creators: Optional[Union[str, List[str]]] = None,
+        language: Optional[Union[str, List[str]]] = None,
+        multilinguality: Optional[Union[str, List[str]]] = None,
+        size_categories: Optional[Union[str, List[str]]] = None,
+        task_categories: Optional[Union[str, List[str]]] = None,
+        task_ids: Optional[Union[str, List[str]]] = None,
```
Wauplin, line 1686: Same comment as for `list_models` (would prefer to have them between `author` and `search`).

src/huggingface_hub/hf_api.py
```diff
@@ -1654,7 +1794,22 @@ def list_datasets(
             if isinstance(filter, DatasetFilter):
                 params = self._unpack_dataset_filter(filter)
             else:
-                params.update({"filter": filter})
+                params = self._build_dataset_filter(
```
Wauplin, line 1797: Same comment as for `list_models` (would prefer to have everything in `list_datasets`).

src/huggingface_hub/hf_api.py
Wauplin: _I think_ it's fine to keep it simple without the `set` overhead. I don't think we should optimize for users passing `list_datasets(..., filter="language:en", language="language:en")` (which is very unlikely to happen). In any case, if we the parameter is duplicated in the URL, the search query should still work so no need to worry about it.

---
(nit) to handle both raw strings and lists in such a case I would do something like this:
```py
                if isinstance(filter, str):
                    filter = [filter]
                params.update({"filter": params.get("filter", set()).update(filter)})
```
which is clearer to read IMO. Drawback is that the `filter` value is modified but that's not really an issue in this precise case.

src/huggingface_hub/hf_api.py
```diff
@@ -3612,7 +3818,8 @@ def create_commit(
             revision=unquoted_revision,  # first-class methods take unquoted revision
             create_pr=create_pr,
             num_threads=num_threads,
-            free_memory=False,  # do not remove `CommitOperationAdd.path_or_fileobj` on LFS files for "normal" users
+            # do not remove `CommitOperationAdd.path_or_fileobj` on LFS files for "normal" users
+            free_memory=False,
```
Wauplin, line 3822:
```suggestion
            free_memory=False,  # do not remove `CommitOperationAdd.path_or_fileobj` on LFS files for "normal" users
```
(something might be off with your `make style` command :thinking: )

src/huggingface_hub/hf_api.py
```diff
@@ -8314,7 +8521,8 @@ def _prepare_upload_folder_additions(
     # List files from folder
     relpath_to_abspath = {
         path.relative_to(folder_path).as_posix(): path
-        for path in sorted(folder_path.glob("**/*"))  # sorted to be deterministic
+        # sorted to be deterministic
+        for path in sorted(folder_path.glob("**/*"))
```
Wauplin, line 8525:
```suggestion
        for path in sorted(folder_path.glob("**/*")) # sorted to be deterministic
```

src/huggingface_hub/hf_api.py
```diff
@@ -8323,7 +8531,8 @@ def _prepare_upload_folder_additions(
     prefix = f"{path_in_repo.strip('/')}/" if path_in_repo else ""
     return [
         CommitOperationAdd(
-            path_or_fileobj=relpath_to_abspath[relpath],  # absolute path on disk
+            # absolute path on disk
+            path_or_fileobj=relpath_to_abspath[relpath],
```
Wauplin, line 8535:
```suggestion
            path_or_fileobj=relpath_to_abspath[relpath],  # absolute path on disk
```

Wauplin (commented):
Thanks for the PR @druvdub! Logic-wise it looks good to me! :fire: I left a couple of comments regarding the code + API but overall very nice one :) Once addressed, we will have to review some tests, especially if ModelFilter / DatasetFilter are used (since deprecation warning make the CI fail, on purpose). I can help on this part if needed (we have a `expect_deprecation` decorator for tests that are meant to break).

src/huggingface_hub/hf_api.py
druvdub: @Wauplin haven't changed anything in the makefile, so it automatically formats these comments over the code, I'll revert those lines

druvdub:
> Thanks for the PR @druvdub! Logic-wise it looks good to me! 🔥 I left a couple of comments regarding the code + API but overall very nice one :) Once addressed, we will have to review some tests, especially if ModelFilter / DatasetFilter are used (since deprecation warning make the CI fail, on purpose). I can help on this part if needed (we have a `expect_deprecation` decorator for tests that are meant to break).

Hey @Wauplin I have addressed all the comments so far, and hopefully I can look at the tests now.
In the meantime, is this expected? 
image

druvdub:
> Thanks for the PR @druvdub! Logic-wise it looks good to me! 🔥 I left a couple of comments regarding the code + API but overall very nice one :) Once addressed, we will have to review some tests, especially if ModelFilter / DatasetFilter are used (since deprecation warning make the CI fail, on purpose). I can help on this part if needed (we have a `expect_deprecation` decorator for tests that are meant to break).

Might need to add another decorator for classes if we plan to use `@expect_deprecation` or potentially modify the tests to handle warnings with `warning.catch_warnings`. I'll look into it

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -73,6 +74,10 @@ class DatasetFilter:
     compatible with the REST API. For all parameters capitalization does not
     matter.
 
+    <Tip warning={true}>
+        The `DatasetFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_datasets`].
+    </Tip>
+
```
Wauplin, line 80: (otherwise tips are not rendered properly - see https://github.com/huggingface/huggingface_hub/pull/2028#issuecomment-1948712119)

```suggestion
    <Tip warning={true}>

        The `DatasetFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_datasets`].

    </Tip>

```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -159,6 +170,10 @@ class ModelFilter:
     compatible with the REST API. For all parameters capitalization does not
     matter.
 
+    <Tip warning={true}>
+        The `ModelFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_models`].
+    </Tip>
```
Wauplin, line 175: (otherwise tips are not rendered properly - see https://github.com/huggingface/huggingface_hub/pull/2028#issuecomment-1948712119)

```suggestion
    <Tip warning={true}>

        The `ModelFilter` class is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to [`list_models`].

    </Tip>
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -221,3 +237,9 @@ class ModelFilter:
     task: Optional[Union[str, List[str]]] = None
     trained_dataset: Optional[Union[str, List[str]]] = None
     tags: Optional[Union[str, List[str]]] = None
+
+    def __post_init__(self):
+        warnings.warn(
+            "ModelFilter is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_models` method.",
```
Wauplin, line 243: To make it work with `@expect_deprecation` in the tests

```suggestion
            "'ModelFilter' is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_models` method.",
```

src/huggingface_hub/utils/endpoint_helpers.py
```diff
@@ -151,6 +156,12 @@ class DatasetFilter:
     task_categories: Optional[Union[str, List[str]]] = None
     task_ids: Optional[Union[str, List[str]]] = None
 
+    def __post_init__(self):
+        warnings.warn(
+            "DatasetFilter is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_datasets` method.",
```
Wauplin, line 161: To make it work with `@expect_deprecation` in the tests

```suggestion
            "'DatasetFilter' is deprecated and will be removed in huggingface_hub>=0.24. Please pass the filter parameters as keyword arguments directly to the `list_datasets` method.",
```

Wauplin (approved):
Thanks for iterating over this PR! Implementation looks good to me now :)

> In the meantime, is this expected?

Tips might not be formatted correctly if they contains special anchor (like urls). This is a known issue to solve in our parser but a workaround exists. A quick solution is to add newlines between the <Tip> tag and the text itself (see comments below) and it should fix the documentation page. Thanks for noticing!

> And similarly do we need to update the README.md files as well

AFAIK there is no mention of `ModelFilter`/`DatasetFilter` in the READMEs. No need to mention them "just" for the deprecation warning.

> Might need to add another decorator for classes if we plan to use @expect_deprecation or potentially modify the tests to handle warnings with warning.catch_warnings unless we deprecate the filter argument in list_models and list_datasets.

We can already use `@expect_deprecation` without needing to add a custom decorator. You just have to update the warning message in the code to add 'quotes'.

> what I am thinking about are the additional scenarios
where we pass the the filter keyword arguments and ModelFilter as well. Do we need them assuming ModelFilter will be removed in sometime

No need to test the cases where direct parameters + ModelFilter are passed at the same time. It's counter-intuitive + the user would have a direct deprecation warning when implementing it (+it will only be supported for a few version, as you said).

druvdub:
Sounds good, everything should be done now :)

tests/test_hf_api.py
```diff
@@ -1737,25 +1737,29 @@ def test_list_datasets_no_filter(self):
         self.assertGreater(len(datasets), 100)
         self.assertIsInstance(datasets[0], DatasetInfo)
 
+    @expect_deprecation
```
Wauplin, line 1740:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1737,25 +1737,29 @@ def test_list_datasets_no_filter(self):
...
     def test_filter_datasets_by_author_and_name(self):
         f = DatasetFilter(author="huggingface", dataset_name="DataMeasurementsFiles")
         datasets = list(self._api.list_datasets(filter=f))
         self.assertEqual(len(datasets), 1)
         self.assertTrue("huggingface" in datasets[0].author)
         self.assertTrue("DataMeasurementsFiles" in datasets[0].id)
 
+    @expect_deprecation
```
Wauplin, line 1748:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1737,25 +1737,29 @@ def test_list_datasets_no_filter(self):
...
     def test_filter_datasets_by_benchmark(self):
         f = DatasetFilter(benchmark="raft")
         datasets = list(self._api.list_datasets(filter=f))
         self.assertGreater(len(datasets), 0)
         self.assertTrue("benchmark:raft" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1755:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1737,25 +1737,29 @@ def test_list_datasets_no_filter(self):
...
     def test_filter_datasets_by_language_creator(self):
         f = DatasetFilter(language_creators="crowdsourced")
         datasets = list(self._api.list_datasets(filter=f))
         self.assertGreater(len(datasets), 0)
         self.assertTrue("language_creators:crowdsourced" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1762:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1766,21 +1770,25 @@ def test_filter_datasets_by_language_only(self):
         self.assertTrue("language:en" in datasets[0].tags)
         self.assertTrue("language:fr" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1773:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1766,21 +1770,25 @@ def test_filter_datasets_by_language_only(self):
...
     def test_filter_datasets_by_multilinguality(self):
         datasets = list(self._api.list_datasets(filter=DatasetFilter(multilinguality="multilingual")))
         self.assertGreater(len(datasets), 0)
         self.assertTrue("multilinguality:multilingual" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1779:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1766,21 +1770,25 @@ def test_filter_datasets_by_language_only(self):
...
     def test_filter_datasets_by_size_categories(self):
         datasets = list(self._api.list_datasets(filter=DatasetFilter(size_categories="100K<n<1M")))
         self.assertGreater(len(datasets), 0)
         self.assertTrue("size_categories:100K<n<1M" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1785:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1766,21 +1770,25 @@ def test_filter_datasets_by_language_only(self):
...
     def test_filter_datasets_by_task_categories(self):
         datasets = list(self._api.list_datasets(filter=DatasetFilter(task_categories="audio-classification")))
         self.assertGreater(len(datasets), 0)
         self.assertTrue("task_categories:audio-classification" in datasets[0].tags)
 
+    @expect_deprecation
```
Wauplin, line 1791:
```suggestion
    @expect_deprecation("DatasetFilter")
```

tests/test_hf_api.py
```diff
@@ -1858,21 +1866,25 @@ def test_list_metrics(self):
         self.assertIsInstance(metrics[0], MetricInfo)
         self.assertTrue(any(metric.description for metric in metrics))
 
+    @expect_deprecation
```
Wauplin, line 1869:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1858,21 +1866,25 @@ def test_list_metrics(self):
...
     def test_filter_models_by_author(self):
         models = list(self._api.list_models(filter=ModelFilter(author="muellerzr")))
         self.assertGreater(len(models), 0)
         self.assertTrue("muellerzr" in models[0].modelId)
 
+    @expect_deprecation
```
Wauplin, line 1875:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1858,21 +1866,25 @@ def test_list_metrics(self):
...
     def test_filter_models_by_author_and_name(self):
         # Test we can search by an author and a name, but the model is not found
         models = list(self._api.list_models(filter=ModelFilter("facebook", model_name="bart-base")))
         self.assertTrue("facebook/bart-base" in models[0].modelId)
 
+    @expect_deprecation
```
Wauplin, line 1881:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1858,21 +1866,25 @@ def test_list_metrics(self):
...
     def test_failing_filter_models_by_author_and_model_name(self):
         # Test we can search by an author and a name, but the model is not found
         models = list(self._api.list_models(filter=ModelFilter(author="muellerzr", model_name="testme")))
         self.assertEqual(len(models), 0)
 
+    @expect_deprecation
```
Wauplin, line 1887:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1884,6 +1896,7 @@ def test_filter_models_with_library(self):
         )
         self.assertGreater(len(models), 0)
 
+    @expect_deprecation
```
Wauplin, line 1899:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1892,11 +1905,13 @@ def test_filter_models_with_task(self):
         models = list(self._api.list_models(filter=ModelFilter(task="dummytask")))
         self.assertEqual(len(models), 0)
 
+    @expect_deprecation
```
Wauplin, line 1908:
```suggestion
    @expect_deprecation("ModelFilter")
```

tests/test_hf_api.py
```diff
@@ -1892,11 +1905,13 @@ def test_filter_models_with_task(self):
...
     def test_filter_models_by_language(self):
         res_fr = list(self._api.list_models(filter=ModelFilter(language="fr")))
         res_en = list(self._api.list_models(filter=ModelFilter(language="en")))
         self.assertGreater(len(res_en), len(res_fr))
 
+    @expect_deprecation
```
Wauplin, line 1914:
```suggestion
    @expect_deprecation("ModelFilter")
```

Wauplin (commented):
`@expect_deprecation` requires an argument (which method/class is deprecated)

druvdub:
@Wauplin does the code coverage report imply I need to add tests covering the keyword arguments for filter?

Wauplin (approved):
PR is ready to be merged! Good job @druvdub :fire: :fire: :fire:

PR #2044, opened 2024-02-24
List models tweaks

julien-c:
Tweaks following [#2028](https://github.com/huggingface/huggingface_hub/issues/2028)


Wauplin (approved):
Thanks @julien-c !

PR #2073, opened 2024-03-01
Fix commonpath in read-only filesystem

stevelaskaridis:
## Description

When downloading a model, commonpath is defined as the LCA (Lowest Common Ancestor) of huggingface's cachedir and target output directory. When this ends-up being a directory that is not read-only (e.g. /), an OSError (errno: 30) is thrown.

This PR catches that error and tries to check symlink creation in the destination directory.

## To reproduce

```python
model_path = snapshot_download(repo_id=model,
                               local_dir='/tmp',
                               cache_dir=None, # TRANSFORMERS_CACHE default (i.e. ~/.cache/huggingface/hub/)
                               token=os.environ.get('HF_TOKEN', None))
```

### Original behaviour

Tested on MacOS 14.1.2 and python 3.11.4, the output was:

```
OSError: [Errno 30] Read-only file system: '/tmpzeu59eox'
```

### New behaviour

The model is persisted as intended.

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -1,4 +1,5 @@
 import copy
+import errno
 import fnmatch
 import inspect
 import io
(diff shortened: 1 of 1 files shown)


stevelaskaridis:
I am not even sure why symlinks are originally tested on `commonpath`, but tried to stay as close as possible to the initial functionality.

Wauplin (approved):
Makes sense yes, thanks for reporting the issue and fixing it in a PR! 
For the record, we might not need this fix in a near future once we revampt `to_local_dir` logic in https://github.com/huggingface/huggingface_hub/issues/1738.

PR #2086, opened 2024-03-04
[PyTorchModelHubMixin] Fix saving model with shared tensors

NielsRogge:
This PR uses `save_model` instead of `save_file` in order to properly save shared tensors for the `PyTorchModelHubMixin`.

In the future, this may be replaced by save_file again when we support saving sharded checkpoints, which deduplicates shared tensors as in the Transformers library.


Wauplin (approved):
Let's get this merged.

PR #2092, opened 2024-03-06
rm unnecessary early makedirs in file_download.py::hf_hub_download()

poedator:
This helps to avoid early directory creation for model names that are not found in the HF hub.
For valid model names the directory gets created below. 
see https://github.com/huggingface/huggingface_hub/issues/2091 for discussion


Wauplin (approved):
Thanks for opening the PR @poedator! Looks good to me :) I'll merge once CI gets green. I also tried locally and it does the job :+1:

PR #2241, opened 2024-04-22
add warning tip that list_deployed_models only searches over cache

MoritzLaurer:
Adding a small warning tip based on this [internal discussion](https://huggingface.slack.com/archives/C016D661PAN/p1713796729692209?thread_ts=1713796161.795809&cid=C016D661PAN).

General thought: I'm wondering how useful the `list_deployed_models` method actually is for users if they don't know if the list is up to date and they need to check with InferenceClient.get_model_status anyways. Could we have a cached list that is updated every N minutes or something like that? No idea how complicated that would be to implemented in the backend. @Wauplin @Narsil


Wauplin (approved):
Thanks @MoritzLaurer!

(I also share concerns about the usability of this method TBH :confused:)

PR #2252, opened 2024-04-25
Standardise token Parameter description and Type Definition

lappemic:
This PR harmonizes all `token` parameter descriptions and type definitions as discussed in #2251.

#### Updated Type Definitions

All functions now uniformly use the following type definition for the `token` parameter:
```python
token: Union[bool, str, None] = None
```

The decision to use `Union[bool, str, None]` for the type definition allows for explicit disabling of authentication (`False`), use of a specific token (`str`), or defaulting to a saved token (`None`).

#### Updated Docstring Descriptions
The docstring for the `token` parameter has been standardized across all relevant functions to:
```
token (Union[bool, str, None], optional):
    A valid user access token (string). Defaults to the locally saved
    token, which is the recommended method for authentication (see
    https://huggingface.co/docs/huggingface_hub/quick-start#authentication).
    To disable authentication, pass `False`.
```

The standardized description clarifies when and how the default token is used, how to explicitly disable authentication, and where to find more information on obtaining tokens.

I am looking forward to your feedback.

### DIFF
--- a/src/huggingface_hub/_multi_commits.py
+++ b/src/huggingface_hub/_multi_commits.py
@@ -273,8 +273,8 @@ def multi_commit_create_pull_request(
     commit_message: str,
     commit_description: Optional[str],
     strategy: MultiCommitStrategy,
-    token: Optional[str],
     repo_type: Optional[str],
+    token: Union[str, bool, None] = None,
 ) -> DiscussionWithDetails:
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thank you for this harmonization @lappemic!

PR #2280, opened 2024-05-13
fix/issue 2090 : Add a `repo_files` command, with recursive deletion.

OlivierKessler01:
## Implement a new `repo-files` command to manipulate files (deletion...)

Fixes [https://github.com/huggingface/huggingface_hub/issues/2235](https://github.com/huggingface/huggingface_hub/issues/2235)

#### Crux
I've encoutered an issue when implementing the folder matching as per this issue specification : 
```bash
huggingface-cli repo-files <repo_id> delete folder/
```

To implement the pattern lexer I reused `huggingface_hub.hf_api._prepare_folder_deletions()`, which is already used in the `upload` command to lexe/tokenize allow and disallow lists. 

The function is a wrapper around `fnmatch.fnmatch()` (via `huggingface_hub.utils._paths.filter_repo_objects()`). It matches files against unix shell-style file matching patterns.

Unix shell-style automatons respond to this case in an rather unconvenient way :

```
>>> fnmatch.fnmatch("nested/file.bin", "nested/")
False
```
#### Solution implemented

It adds "*" at the end of the patterns that end with "/". Given git untracks any folder that is empty, deleting all files in the folder effectively deletes the folder. Did it in the new `huggingface_hub.hf_api.delete_files_r()` function, because updating `huggingface_hub.utils._paths.filter_repo_objects()` at a lower-level would alter the behavior of various functions like the `upload` command and the commit scheduler in a non-BC compatible way.

What do  you think about it ? 
```bash
huggingface-cli upload --delete="nested/"
#the line above still doesn't work, but it's consistent (logic unchanged, interface unchanged). 

huggingface-cli repo-files delete nested/ #interpreted as `huggingface-cli repo-files delete nested/*`
#the line above works as per https://github.com/huggingface/huggingface_hub/issues/2235 specs.
```

All this has been documented in the function doc.

#### What could have been done instead 

Remove support for this
`huggingface-cli repo-files <repo_id> delete folder/`

Natively support this through already existing logic, and let Git untrack the folder.
`huggingface-cli repo-files <repo_id> delete folder/*`

### Tests added (duration that might disrupt DevX and project output) 
```bash
20.46s call     tests/test_hf_api.py::CommitApiTest::test_delete_files_r
```
### Remarks

Feel free to tell me if you'd like something different here :)


src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
             parent_commit=parent_commit,
         )
 
+    @validate_hf_hub_args
+    def delete_files_r(
```
Wauplin, line 4776:
```suggestion
    def delete_files(
```

I'd rename it `delete_files` instead (the `_r` is not very explicit)

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+        self,
+        patterns: List[str],
+        repo_id: str,
```
Wauplin, line 4779:
```suggestion
        repo_id: str,
        delete_patterns: List[str],
```

I'd place `repo_id` first (as in the command line) and rename to `delete_patterns` to be more explicit.

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+        repo_type: Optional[str] = None,
+        revision: Optional[str] = None,
+        commit_message: Optional[str] = None,
+        commit_description: Optional[str] = None,
+        create_pr: Optional[bool] = None,
+        parent_commit: Optional[str] = None,
+    ) -> CommitInfo:
+        """
+        Recursively deletes files in a given repository based on Unix shell-style
+        file matching patterns.
```
Wauplin, line 4791:
```suggestion
        Delete files from a repository on the Hub.
```

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+
+        Git based under the hood (server-side), meaning if you delete all files
+        from a directory, git untracks it. It works this way recursively
+        through the sub folders. Meaning a whole directory structure is untracked
+        by deleting their respectives files at all levels.
+
```
Wauplin, line 4797:
```suggestion
        If a folder path is provided, the entire folder is deleted as well as all files it contained.

```

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+        Also if one or many of the patterns from the `patterns` argument to
+        this function have a trailing `/` (as in `folder/`),
+        this function will interpret it as `/*` (as in `folder/*`) to comply with Unix
+        patterns (Unix shell standards don't match `folder/a` to `folder/` but
+        match it to `folder/*`) when deleting all files in the folders(s).
+        By the previous corollary, the folder(s) will disappear because all
+        their files are deleted.
```
Wauplin, line 4804:
```suggestion
```

I think we don't need this (see below)

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+
+        Args:
+            path_in_repo (`str`):
+                Relative folder path in the repo, for example: `"checkpoints/1fec34a"`.
+            repo_id (`str`):
+                The repository from which the folder will be deleted, for example:
+                `"username/custom_transformers"`
```
Wauplin, line 4811:
```suggestion
            repo_id (`str`):
                The repository from which the folder will be deleted, for example:
                `"username/custom_transformers"`
            delete_patterns (`List[str]`):
                List of files or folders to delete. Each string can either be a file path, a folder path or a Unix shell-style wildcard.
                E.g. `["file.txt", "folder/", "data/*.parquet"]`
```

`path_in_repo` is not a correct arg (`delete_patterns` instead) + should switch with repo_id since the change in args above.

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+            token (Union[bool, str, None], optional):
+                A valid user access token (string). Defaults to the locally saved
+                token, which is the recommended method for authentication (see
+                https://huggingface.co/docs/huggingface_hub/quick-start#authentication).
+                To disable authentication, pass `False`.
+                to the stored token.
+            repo_type (`str`, *optional*):
+                Set to `"dataset"` or `"space"` if the folder is in a dataset or
+                space, `None` or `"model"` if in a model. Default is `None`.
```
Wauplin, line 4820:
```suggestion
            repo_type (`str`, *optional*):
                Type of the repo to delete files from. Can be `"model"`, `"dataset"` or `"space"`. Defaults to `"model"`.
```

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+            revision (`str`, *optional*):
+                The git revision to commit from. Defaults to the head of the `"main"` branch.
+            commit_message (`str`, *optional*):
+                The summary / title / first line of the generated commit. Defaults to
+                `f"Delete folder {path_in_repo} with huggingface_hub"`.
```
Wauplin, line 4825:
```suggestion
                The summary (first line) of the generated commit. Defaults to
                `f"Delete files using huggingface_hub"`.
```

src/huggingface_hub/hf_api.py
```diff
@@ -8994,6 +9079,7 @@ def _parse_revision_from_pr_url(pr_url: str) -> str:
 upload_folder = api.upload_folder
 delete_file = api.delete_file
 delete_folder = api.delete_folder
+delete_files_r = api.delete_files_r
```
Wauplin, line 9082:
```suggestion
delete_files = api.delete_files
```

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+                `revision` is set and is not a branch name (example: a commit oid), an
+                `RevisionNotFoundError` is returned by the server.
+            parent_commit (`str`, *optional*):
+                The OID / SHA of the parent commit, as a hexadecimal string. Shorthands (7 first characters) are also supported.
+                If specified and `create_pr` is `False`, the commit will fail if `revision` does not point to `parent_commit`.
+                If specified and `create_pr` is `True`, the pull request will be created from `parent_commit`.
+                Specifying `parent_commit` ensures the repo has not changed before committing the changes, and can be
+                especially useful if the repo is updated / committed to concurrently.
+        """
+        patterns = [p + "*" if p[-1] == "/" else p for p in patterns]
```
Wauplin, line 4841: Are we sure this is needed? When creating a commit, a `CommitOperationDelete` operation with `"folder/"` (no `*`) should already delete the entire folder for us. So I'm surprised if `"*"` is needed. If that's the case, it should be fixed in `_prepare_folder_deletions` directly.

src/huggingface_hub/hf_api.py
```diff
@@ -4772,6 +4772,92 @@ def delete_file(
...
+
+        f_patterns = " ".join(patterns)
+        return self.create_commit(
+            repo_id=repo_id,
+            repo_type=repo_type,
+            token=token,
+            operations=operations,
+            revision=revision,
+            commit_message=(
+                commit_message if commit_message is not None else f"Delete files {f_patterns} with huggingface_hub"
```
Wauplin, line 4854: (nit)

I would add a

```py
if commit_message is None:
    commit_message = f"Delete files {f_patterns} with huggingface_hub"
```

before the `return self.create_commit` calls. Can make investigation/debugging easier in the future.

src/huggingface_hub/hf_api.py
```diff
@@ -8771,7 +8857,7 @@ def _build_hf_headers(
             headers=self.headers,
         )
 
-    def _prepare_upload_folder_deletions(
+    def _prepare_folder_deletions(
```
Wauplin, line 8860: Nice to reuse this!

src/huggingface_hub/hf_api.py
```diff
@@ -8794,7 +8880,6 @@ def _prepare_upload_folder_deletions(
 
         # List remote files
         filenames = self.list_repo_files(repo_id=repo_id, revision=revision, repo_type=repo_type, token=token)
-
```
Wauplin, line 8797: better to keep this line for readability

tests/test_command_repo_files.py
```diff
@@ -0,0 +1,149 @@
+import unittest
+from argparse import ArgumentParser
+from unittest.mock import Mock, patch
+
+from huggingface_hub.commands.repo_files import DeleteSubCommand, RepoFilesCommand
+
+from .testing_utils import DUMMY_MODEL_ID
+
+
+class TestRepoFilesCommand(unittest.TestCase):
```
Wauplin, line 10: Can you move this to `tests/test_cli.py` please? Keep it as a new test class as you did but not in a new module.

tests/test_command_repo_files.py
```diff
@@ -0,0 +1,149 @@
...
+    def setUp(self) -> None:
+        """
+        Set up CLI as in `src/huggingface_hub/commands/huggingface_cli.py`.
+        """
+        self.parser = ArgumentParser("huggingface-cli", usage="huggingface-cli <command> [<args>]")
+        commands_parser = self.parser.add_subparsers()
+        RepoFilesCommand.register_subcommand(commands_parser)
+
+    @patch("huggingface_hub.commands.repo_files.HfApi.delete_files_r")
+    def test_delete(self, delete_files_r_mock: Mock) -> None:
```
Wauplin, line 20:
```suggestion
    @patch("huggingface_hub.commands.repo_files.HfApi.delete_files")
    def test_delete(self, delete_files_mock: Mock) -> None:
```

(+ update variables below)

tests/test_hf_api.py
```diff
@@ -467,6 +467,83 @@ def test_get_full_repo_name(self):
         repo_name_with_no_org = self._api.get_full_repo_name("model", organization="org")
         self.assertEqual(repo_name_with_no_org, "org/model")
 
+    @use_tmp_repo()
+    def test_delete_files_r(self, repo_url: RepoUrl) -> None:
```
Wauplin, line 471:
```suggestion
    def test_delete_files(self, repo_url: RepoUrl) -> None:
```

tests/test_hf_api.py
```diff
@@ -467,6 +467,83 @@ def test_get_full_repo_name(self):
...
+        repo_id = repo_url.repo_id
+        fixtures = [
```
Wauplin, line 473: In this case, I'd rather have a separate test per example. This way we ensure the tests are not dependent on eachother by using a different tmp repo each time.

Here a potential problem is that between each subTest we are deleting files to have a clean state. But since `delete_files` is actually the method we want to test, it might error out/have a weird behavior without us noticing.

Wauplin: Actually, what I would do to test this:
1. have a separate TestCase class specific for `delete_files`
2. in `setUp`, create a repo + upload some files to it (always the same for each test).
3. for each test you want to try, write a separate test that calls `delete_files` with different args. Then you call `self._api.list_repo_files` and you check that only the files you wanted still exists. Testing `hf_hub_download` for that is a too heavy, it's preferable to list files without downloading.

(4. it is also important to check cases where deleting files will fail (say if files don't exist for example). These corner cases should be clearly identified, tested and documented.)

src/huggingface_hub/commands/repo_files.py
```diff
@@ -0,0 +1,118 @@
...
+
+class DeleteSubCommand:
+    def __init__(self, args):
+        self.args = args
+        self.repo_id: str = args.repo_id
+        self.token: str = args.token
+        self.repo_type: Optional[str] = args.repo_type
+        self.revision: Optional[str] = args.revision
+        self.api: HfApi = HfApi(token=args.token, library_name="huggingface-cli")
+        self.patterns: List[str] = args.patterns
```
Wauplin, line 56: Also add `--create-pr` / `--commit-message` / `--commit-description` as optional arguments.

Wauplin (commented):
Hi @OlivierKessler01, thanks a lot for this extensive work! It's a very good first version and I've reviewed it top to bottom. You'll find my comments below. Please let me know if you have any questions :) Thanks again!

src/huggingface_hub/hf_api.py
OlivierKessler01: Yep it is needed, as I documented it in the docstring and in the PR description : 

Here is what you get if you comment this line : 
image

Fixing it in `_prepare_folder_deletions` will alter the `upload` command behavior related to deletion (see PR description). Should we still do it ?

tests/test_hf_api.py
OlivierKessler01: Just implemented it as you'd like it :) see for yourself. 

Also added a new function to initialize a local temp folder. 
 I could also have done it that way ((pseudo code not-tested) : 
```python
def use_tmp_local_folder(self, test_fn: T) -> Callable[[T], T]:
    def _inner_use_local_folder(test_fn: T) -> T:
        @wraps(test_fn)
        def _inner(*args, **kwargs):
            if len(args) > 0 and isinstance(args[0], object):
                self:unittest.TestCase = args[0]
                if self.__class__ != unittest.TestCase:
                    raise TypeError(
                        f"Method {test_fn.__name__} can only be called from an instance of the 'unittest.TestCase' class"
                    )
            else:
                raise TypeError("This decorator can only be used on non-static methods of a 'unittest.TestCase' class.")

            self.tmp_dir = tempfile.mkdtemp()
            self.tmp_file = os.path.join(self.tmp_dir, "temp")
            self.tmp_file_content = "Content of the file"
            with open(self.tmp_file, "w+") as f:
                f.write(self.tmp_file_content)
            os.makedirs(os.path.join(self.tmp_dir, "nested"))
            self.nested_tmp_file = os.path.join(self.tmp_dir, "nested", "file.bin")
            with open(self.nested_tmp_file, "wb+") as f:
                f.truncate(1024 * 1024)

            self.addCleanup(rmtree_with_retry, self.tmp_dir)

            return test_fn(*args, **kwargs)

        return _inner

    return _inner_use_tmp_local_folder

```

src/huggingface_hub/hf_api.py
Wauplin: Indeed. I had a look into it and I think it should be done in `filter_repo_objects` directly. I just pushed https://github.com/huggingface/huggingface_hub/pull/2280/commits/19d6b199473e5f11ec563fa90510a6c731df6363 to your PR, hope you don't mind.

tests/test_hf_api.py
Wauplin: Hi @OlivierKessler01, I just pushed https://github.com/huggingface/huggingface_hub/pull/2280/commits/e953469ec57633eb530a1be50b369912142b563c to refacto a bit the tests. While the tests you wrote were valid, I've updated them for readability and simplicity. In particular:
- no need to create a local folder since what we want is to create a repo with files in it. This can be done directly with `create_commit`
- it's better to name tests with more explicit names than `test_one`, `test_two`, etc. This way it gives information for people reading those tests in the future.
- I've added more tests about nested folders and patterns.
- It's better to only test what the unit test is supposed to test. For instance, no need to check the files present in the repo after the first commit. If the upload commit fails (for a breaking change for instance), it will be covered by another test in the test suite. Usually you test 1 feature at a time in the unit test and assume all other methods are working correctly.  
- I feel that the `_check` method was hidding too many things to the reader. Sometimes it's better to explicitly test what you want to test . For example, testing only the fact that a file has been deleted rather than testing all files that should have remained.

Wauplin (commented):
Hi @OlivierKessler01, many thanks for the changes! I've pushed a few commits (see above) to finalize it and now everything looks good to me.

Last thing to update when implementing a new feature: adding a section in the docs! In this case, it should be here: https://github.com/huggingface/huggingface_hub/blob/main/docs/source/en/guides/cli.md. Which correspond [to these docs](https://huggingface.co/docs/huggingface_hub/guides/cli). I think a section between `huggingface-cli upload` and `huggingface-cli scan-cache` would be good. You can have:
```
## huggingface-cli repo-files

(... some introduction content ...)

### Delete files

(... docs of your PR with 2-3 examples ...)

## huggingface-cli scan-cache

(...)
```

src/huggingface_hub/commands/repo_files.py
OlivierKessler01: Done !

OlivierKessler01:
Hi @Wauplin thank you for the contribution. I just added the documentation block, have a look :)

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
 https://huggingface.co/Wauplin/my-cool-model/tree/main
 ```
 
+## huggingface-cli repo-files
+
+If you want to update or delete files from a Hugging Face repository, use the `huggingface-cli repo-files` command.
```
Wauplin, line 434:
```suggestion
If you want to delete files from a Hugging Face repository, use the `huggingface-cli repo-files` command. 
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
 https://huggingface.co/Wauplin/my-cool-model/tree/main
 ```
 
+## huggingface-cli repo-files
+
+If you want to update or delete files from a Hugging Face repository, use the `huggingface-cli repo-files` command. 
+
+### Delete files
+
+The `huggingface-cli repo-files "<repo_id>" delete` sub-command allows you to delete files from a repository. Here are some usage examples.
```
Wauplin, line 438:
```suggestion
The `huggingface-cli repo-files <repo_id> delete` sub-command allows you to delete files from a repository. Here are some usage examples.
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
...
+
+Delete a folder :
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete folder/  
+Files correctly deleted from repo. Commit: https://huggingface.co/Wauplin/my-cool-mo...
+```
+
+Delete multiple files explicitely:
```
Wauplin, line 446:
```suggestion
Delete multiple files: 
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
...
+
+Delete a folder :
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete folder/  
+Files correctly deleted from repo. Commit: https://huggingface.co/Wauplin/my-cool-mo...
+```
+
+Delete multiple files explicitely: 
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete fold2/file1.txt fold3/file2.py
```
Wauplin, line 448:
```suggestion
>>> huggingface-cli repo-files Wauplin/my-cool-model delete file.txt folder/pytorch_model.bin
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
...
+
+Delete multiple files explicitely: 
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete fold2/file1.txt fold3/file2.py 
+Files correctly deleted from repo. Commit: https://huggingface.co/Wauplin/my-cool-mo...
+```
+
+Use Unix-style wildcards to delete sets of files: 
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete folder2/*.txt
```
Wauplin, line 454:
```suggestion
>>> huggingface-cli repo-files Wauplin/my-cool-model delete *.txt folder/*.bin 
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
...
+
+Use Unix-style wildcards to delete sets of files: 
+```bash
+>>> huggingface-cli repo-files Wauplin/my-cool-model delete folder2/*.txt 
+Files correctly deleted from repo. Commit: https://huggingface.co/Wauplin/my-cool-mo...
+```
+
+### Specify a token
+
+To access private or gated repositories, you must use a token. By default, the token saved locally (using `huggingface-cli login`) will be used. If you want to authenticate explicitly, use the `--token` option:
```
Wauplin, line 460:
```suggestion
To delete files from a repo you must be authenticated and authorized. By default, the token saved locally (using `huggingface-cli login`) will be used. If you want to authenticate explicitly, use the `--token` option:
```

docs/source/en/guides/cli.md
```diff
@@ -429,6 +429,40 @@ By default, the `huggingface-cli upload` command will be verbose. It will print
...
+
+```bash
+>>> huggingface-cli repo-files --token=hf_**** Wauplin/my-cool-model delete example.txt
```
Wauplin, line 463:
```suggestion
>>> huggingface-cli repo-files --token=hf_**** Wauplin/my-cool-model delete file.txt 
```

src/huggingface_hub/commands/repo_files.py
```diff
@@ -0,0 +1,128 @@
...
+        self.args = args
+        self.repo_id: str = args.repo_id
+        self.repo_type: Optional[str] = args.repo_type
+        self.revision: Optional[str] = args.revision
+        self.api: HfApi = HfApi(token=args.token, library_name="huggingface-cli")
+        self.patterns: List[str] = args.patterns
+        self.commit_message: Optional[str] = args.commit_message
+        self.commit_description: Optional[str] = args.commit_description
+        self.create_pr: bool = args.create_pr
+        self.token: str = args.token
```
Wauplin, line 59:
```suggestion
        self.token: Optional[str] = args.token
```

Wauplin (approved):
Yay! Thanks again @OlivierKessler01 for all the work on this PR :rocket: I just had a look to the docs and I think we are good to go. Let's wait for the CI to complete and I'll merge.

Wauplin:
And, merged!

PR #2291, opened 2024-05-29
Fix ModelHubMixin coders

gorold:
This PR implements the fixes as discussed in #2283 and the relevant test cases.
@Wauplin


Wauplin (commented):
Thanks for the clean implementation and tests @gorold! Logic looks good to me. I've left some comments below, mostly related to the fact that I'd prefer to have the "typing helpers" implemented and tested separately from the hub mixin module. Apart from that, I think we'll be good to go :+1:

src/huggingface_hub/hub_mixin.py
```diff
@@ -302,20 +317,39 @@ def _is_jsonable(cls, value: Any) -> bool:
...
+            expected_type = next(takewhile(lambda x: x is not type(None), get_args(expected_type)))
         for type_, (_, decoder) in cls._hub_mixin_coders.items():
-            if issubclass(expected_type, type_):
-                return decoder(value)
+            if inspect.isclass(expected_type) and issubclass(expected_type, type_):
+                try:
+                    decoded_value = decoder(value)
+                except Exception as e:
+                    if value is None:
+                        return None
```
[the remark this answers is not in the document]
gorold: Maybe we should check if it is optional and None and fails? We could have the following scenario, then should we return None or CustomType(None)?

```python
class CustomType:
    def __init__(self, val: Optional[int]):
        self.val = val


def encoder(arg):
    return arg.val


def decoder(arg):
    return CustomType(arg)


custom_type: Optional[CustomType]
```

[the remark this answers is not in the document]
gorold: This might be a pathological case... haha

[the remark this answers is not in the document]
Wauplin: In the case above I would return `None` - and sorry for the encoder. But yeah, this is clearly a corner-case and better not to optimize for it.

src/huggingface_hub/utils/_typing.py
```diff
@@ -48,3 +50,17 @@ def is_jsonable(obj: Any) -> bool:
...
+    if get_origin(type_) is Union or isinstance(type_, UnionType):
+        union_args = get_args(type_)
+        if len(union_args) == 2 and type(None) in union_args:
+            return True
+    return False
+
+
+def unwrap_simple_optional_type(optional_type: Type) -> Type:
+    """Unwraps a simple optional type, i.e. returns Type from Optional[Type]."""
+    return next(filterfalse(lambda x: x is type(None), get_args(optional_type)))
```
Wauplin, line 66: Current implementation is valid but I'd prefer a more readable version like this:

```suggestion
    for arg in get_args(optional_type):
        if arg is not type(None):
            return arg
    raise ValueError(f"'{optional_type}' is not an optional type")
```

This way, we don't have a `next` + `itertools` + `lambda` just to get a value from a tuple.

tests/test_utils_typing.py
```diff
@@ -47,3 +53,70 @@ def test_is_jsonable_failure(data):
...
+        (Union[CustomType, None], True),
+        (int, False),
+        (None, False),
+        (Union[int, float, None], False),
+        (Union[Union[int | float], None], False),
+        (Optional[int | float], False),
+    ],
+)
+def test_is_simple_optional_type(type_: Type, is_optional: bool):
+    assert is_simple_optional_type(type_) == is_optional
```
Wauplin, line 75:
```suggestion
    assert is_simple_optional_type(type_) is is_optional
```

(nit)

tests/test_utils_typing.py
```diff
@@ -47,3 +53,70 @@ def test_is_jsonable_failure(data):
...
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="requires python3.10 or higher")
```
Wauplin, line 78: I was about to put a comment for those but you already tested them. Thank you! :heart:

Wauplin (commented):
Thanks for the changes @gorold! It's looking good now. Just left a minor comment otherwise I think we are good to merge :)

src/huggingface_hub/utils/_typing.py
```diff
@@ -14,7 +14,9 @@
 # limitations under the License.
 """Handle typing imports based on system compatibility."""
 
-from typing import Any, Callable, Literal, TypeVar
+from itertools import filterfalse
+from types import UnionType
```
Wauplin, line 18: Looks like `UnionType` is not available on Python 3.8 :confused:

gorold: 😢 made the updates!

Wauplin (approved):
Thanks for the changes @gorold! That's very neat way of handling it :)

Wauplin (approved):

gorold:
This should fix it. 3.8 was still failing because the interpreter doesn't recognise the pipe operator with types, so we need to guard it to only branch to those tests in >=3.10.

There seem to be 2 test cases which fail, but they don't seem like related tests?

tests/test_offline_utils.py::test_offline_with_timeout
tests/test_utils_http.py::TestConfigureSession::test_get_session_in_forked_process

Wauplin:
Thanks for fixing the tests @gorold! The two remaining ones are indeed unrelated (I'm currently having a look at them in a separate PR). About the solution to test the pipe operator, could you add `from __future__ import annotations` at the beginning of the test module so that all tests are run on all Python versions? It would make the test file easier to read I think.

Wauplin:
Ah true, thanks for trying it. I prefer the `eval` solution better if you don't mind? At least tests are explicitly skipped instead of not defined. Last round of changes, I promise! :smile:

gorold:
No problem, made the changes.

Wauplin (approved):
All good! Thanks again for fixing this :)

PR #2371, opened 2024-07-03
`torch`  extra requires for `safetensors` dep

qgallouedec:
This change is needed, otherwise numpy won't be installed. See https://github.com/huggingface/safetensors/blob/079781fd0dc455ba0fe851e2b4507c33d0c0d407/bindings/python/pyproject.toml#L35


Wauplin (approved):
Thanks @qgallouedec !

PR #2396, opened 2024-07-17
fix: Handle single return value.

28Smiles:
See https://github.com/huggingface/huggingface.js/pull/803


Wauplin (approved):
Thanks @28Smiles!

When updating `_client.py`, the async client must be updated as well which I did in https://github.com/huggingface/huggingface_hub/commit/b5e0d76e09bfc82c1be14e787de9b0ba2da40e46 but pushed it to the main repo instead of yours :see_no_evil: Nevermind, let's merge this and everything will be back to normal :smile: Right on time for next release!

PR #2397, opened 2024-07-18
Reduce max shard size

osanseviero:
ClamAV scans files of up to 3.9GB. By lowering the shard size, we can ensure files are scanned


julien-c (commented):
hmm not sure this is enough reason to break the current behavior

(also ClamAV will soon not be the only tool we have)

osanseviero:
Closing as discussed :)

PR #2432, opened 2024-07-31
fix .huggingface to .cache/huggingface in doc

lizzzcai:
fix the doc


Wauplin (approved):
Thanks for noticing and opening a PR @lizzzcai! :hugs:

PR #2434, opened 2024-07-31
Correct "login" to "log in" when used as verb

DePasqualeOrg:
"login" is a noun, and "log in" is a verb. I've corrected instances of "login" where it is used as a verb.


Wauplin (approved):
You are right! Thanks for correcting it :)

PR #2490, opened 2024-08-27
fix secrets inference endpoints

LuisBlanche:
Fix to allow not passing a secrets at creation time (passing secrets = None to API returns 422)


Wauplin (approved):
Thanks for finding out and fixing it quickly @LuisBlanche :heart_hands:

PR #2500, opened 2024-09-01
Solved: Need to add the max-workers argument to the huggingface-cli command

devymex:
https://github.com/huggingface/huggingface_hub/issues/2499


Wauplin (approved):
Looks good! Thanks for the contribution @devymex, it makes a lot of sense! :hugs:

PR #2503, opened 2024-09-02
fix empty siblings

lhoestq:
`{model,dataset,space}_info.siblings` should be `[]` if the repo is empty no ? currently it's `None` even if the Hub API returns `"sibligns": []` in the JSON response

this causes unexpected though minor errors in the Viewer

e.g. it indirectly causes the histogram filtering UI of the [laion's viewer](https://huggingface.co/datasets/laion/relaion2B-multi-research-safe) to be disabled


Wauplin (approved):
Makes sense yes! Thanks for the PR :)

PR #2532, opened 2024-09-11
Implemented https://github.com/huggingface/huggingface_hub/issues/2516

yuxi-liu-wired:
Two updates:

1. Implemented https://github.com/huggingface/huggingface_hub/issues/2516
2. Added `extras["torch"]` to `[dev]` in `setup.py` since `safetensors` is required for testing mixin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ extras["tensorflow-testing"] = [
 extras["testing"] = (
     extras["cli"]
     + extras["inference"]
+    + extras["torch"]
     + [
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for the prompt and quick PR @yuxi-liu-wired! This looks good to me. I would just remove the dev dependency you've added (see below). Otherwise we should be good to merge :tada:

Wauplin (approved):
Thanks!

PR #2551, opened 2024-09-18
Fix HfApi.grant_access for datasets and spaces

### LINKED ISSUE #2550: Method HfApi.grant_access only works for models
The `HfApi.grant_access` method only works with models and does not work with datasets or spaces.

The reason is that this method ignores the parameter `repo_type`: https://github.com/huggingface/huggingface_hub/blob/f1a7ed49de66d5fcd75ddee0a2739daef23a9c47/src/huggingface_hub/hf_api.py#L8976

This was introduced in v0.20.0 by this PR:
- #1905

albertvillanova:
Fix `HfApi.grant_access` for datasets and spaces.

Fix #2550.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -8973,7 +8973,7 @@ class HfApi:
             repo_type = constants.REPO_TYPE_MODEL
 
         response = get_session().post(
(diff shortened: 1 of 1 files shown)


Wauplin (approved):
:facepalm: 

My bad, thanks for reporting and fixing it!

PR #2585, opened 2024-10-01
Enforce parameters as a List of Dictionaries in modelinfo.safetensors

adiaholic:
~This PR modifies the `model_info` method to ensure that the `SafeTensorsInfo` instance is initialized with a list of dictionaries for the `parameters` attribute, as intended by the class definition.~

This PR changes the type definition of `parameters` attribute in `SafeTensorsInfo` to be a `Dict` instead of `List[Dict[str, int]]`

The model_info response for repo_id = `allenai/Molmo-72B-0924` is of the form:

```
"safetensors": {
            "parameters": {
                "F32": 73308285952
            },
            "total": 73308285952
        },
```

which is ~not~ expected since `parameters` should be of type ~`List[Dict[str, int]]`~ `Dict[str, int]`

This PR **might** serve as an extension for #2190

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -358,7 +358,7 @@ class TransformersInfo(dict):
 
 @dataclass
 class SafeTensorsInfo(dict):
-    parameters: List[Dict[str, int]]
+    parameters: Dict[str, int]
     total: int
 
     def __post_init__(self):  # hack to make SafeTensorsInfo backward compatible


hanouticelina:
Hi @adiaholic, thanks for spotting the mismatch! the issue is actually in the type definition: The `parameters` field in `SafeTensorsInfo` should be `Dict[str, int]`, not `List[Dict[str, int]]`. 
We should update the type annotation instead as it is defined here : [huggingface/huggingface.js/packages/tasks/src/model-data.ts ](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/model-data.ts#L117).

adiaholic:
> Hi @adiaholic, thanks for spotting the mismatch! the issue is actually in the type definition: The `parameters` field in `SafeTensorsInfo` should be `Dict[str, int]`, not `List[Dict[str, int]]`. We should update the type annotation instead as it is defined here : [huggingface/huggingface.js/packages/tasks/src/model-data.ts ](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/model-data.ts#L117).

Thank you for reviewing this and providing clarification. I've made some updates to `SafeTensorsInfo`, and I hope they meet the requirements. Please feel free to let me know if any further adjustments are needed.

hanouticelina (approved):
All good! thanks again @adiaholic 🤗

PR #2601, opened 2024-10-11
[Inference Client] Factorize inference payload build

hanouticelina:
This PR  is a first attempt to factorize the payload build in multiple `InferenceClient` methods. In fact, there's some repetitive logic across several methods for handling inputs and parameters, so here we introduce a new (private) helper function to factorize this logic.

## Key changes

1. Adding a helper function that:
    - handles both raw content (images or audio) and string/dict inputs uniformly.
    - base64 encodes raw content when at least one parameter is present.
    - filters out None values from the parameters.
    - .. and returns an `_InferenceInputs` object containing the json payload and raw data if any. (I don't have a strong opinion on this, we can also return a Tuple instead).
2. A unit test to verify the correct behavior of the refactored logic.

⚠️ These changes only affect the internal/private functionality of the `InferenceClient` and `AsyncInferenceClient`.

### DIFF
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -2848,9 +2848,11 @@ class InferenceClient:
         ```
         """
 
-        parameters = {"candidate_labels": labels, "multi_label": multi_label}
-        if hypothesis_template is not None:
-            parameters["hypothesis_template"] = hypothesis_template
+        parameters = {
+            "candidate_labels": labels,
+            "multi_label": multi_label,
+            "hypothesis_template": hypothesis_template,
+        }
         payload = _prepare_payload(text, parameters=parameters)
         response = self.post(
             **payload,
@@ -2911,7 +2913,7 @@ class InferenceClient:
             raise ValueError("You must specify at least 2 classes to compare.")
 
         inputs = {"image": _b64_encode(image), "candidateLabels": ",".join(labels)}
-        parameters = {"hypothesis_template": hypothesis_template} if hypothesis_template is not None else None
+        parameters = {"hypothesis_template": hypothesis_template}
         payload = _prepare_payload(inputs, parameters=parameters)
         response = self.post(
             **payload,
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -2924,6 +2924,10 @@ class InferenceClient:
         inputs: Union[str, Dict[str, Any], ContentT],
         parameters: Optional[Dict[str, Any]] = None,
     ) -> _InferenceInputs:
+        """
+        Prepare payload for an API request, handling various input types and parameters.
+        """
+
         def is_raw_content(inputs: Union[str, ContentT]) -> bool:
             return isinstance(inputs, (bytes, Path)) or (
                 isinstance(inputs, str) and inputs.startswith(("http://", "https://"))
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -37,6 +37,8 @@ import logging
 import re
 import time
 import warnings
+from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Literal, Optional, Union, overload
 
 from requests import HTTPError
@@ -107,6 +109,12 @@ logger = logging.getLogger(__name__)
 
 
 MODEL_KWARGS_NOT_USED_REGEX = re.compile(r"The following `model_kwargs` are not used by the model: \[(.*?)\]")
+
+
+@dataclass
+class _InferenceInputs:
+    json: Optional[Dict[str, Any]] = None
+    raw_data: Optional[ContentT] = None
 
 
 class InferenceClient:
@@ -364,18 +372,8 @@ class InferenceClient:
         ```
         """
         parameters = {"function_to_apply": function_to_apply, "top_k": top_k}
-        if all(parameter is None for parameter in parameters.values()):
-            # if no parameters are provided, send audio as raw data
-            data = audio
-            payload: Optional[Dict[str, Any]] = None
-        else:
-            # Or some parameters are provided -> send audio as base64 encoded string
-            data = None
-            payload = {"inputs": _b64_encode(audio)}
-            for key, value in parameters.items():
-                if value is not None:
-                    payload.setdefault("parameters", {})[key] = value
-        response = self.post(json=payload, data=data, model=model, task="audio-classification")
+        payload = self._prepare_payload(audio, parameters=parameters)
+        response = self.post(json=payload.json, data=payload.raw_data, model=model, task="audio-classification")
         return AudioClassificationOutputElement.parse_obj_as_list(response)
 
     def audio_to_audio(
@@ -988,7 +986,7 @@ class InferenceClient:
         [DocumentQuestionAnsweringOutputElement(answer='us-001', end=16, score=0.9999666213989258, start=16, words=None)]
         ```
         """
-        payload: Dict[str, Any] = {"question": question, "image": _b64_encode(image)}
+        inputs: Dict[str, Any] = {"question": question, "image": _b64_encode(image)}
         parameters = {
             "doc_stride": doc_stride,
             "handle_impossible_answer": handle_impossible_answer,
@@ -999,10 +997,8 @@ class InferenceClient:
             "top_k": top_k,
             "word_boxes": word_boxes,
         }
-        for key, value in parameters.items():
-            if value is not None:
-                payload.setdefault("parameters", {})[key] = value
-        response = self.post(json=payload, model=model, task="document-question-answering")
+        payload = self._prepare_payload(inputs, parameters=parameters)
+        response = self.post(json=payload.json, model=model, task="document-question-answering")
         return DocumentQuestionAnsweringOutputElement.parse_obj_as_list(response)
 
     def feature_extraction(
@@ -1060,17 +1056,14 @@ class InferenceClient:
         [ 0.28552425, -0.928395  , -1.2077185 , ...,  0.76810825, -2.1069427 ,  0.6236161 ]], dtype=float32)
         ```
         """
-        payload: Dict = {"inputs": text}
         parameters = {
             "normalize": normalize,
             "prompt_name": prompt_name,
             "truncate": truncate,
             "truncation_direction": truncation_direction,
         }
-        for key, value in parameters.items():
-            if value is not None:
-                payload.setdefault("parameters", {})[key] = value
-        response = self.post(json=payload, model=model, task="feature-extraction")
+        payload = self._prepare_payload(text, parameters=parameters)
+        response = self.post(json=payload.json, model=model, task="feature-extraction")
         np = _import_numpy()
         return np.array(_bytes_to_dict(response), dtype="float32")
 
@@ -1119,12 +1112,9 @@ class InferenceClient:
         ]
         ```
         """
-        payload: Dict = {"inputs": text}
         parameters = {"targets": targets, "top_k": top_k}
-        for key, value in parameters.items():
-            if value is not None:
-                payload.setdefault("parameters", {})[key] = value
-        response = self.post(json=payload, model=model, task="fill-mask")
+        payload = self._prepare_payload(text, parameters=parameters)
+        response = self.post(json=payload.json, model=model, task="fill-mask")
         return FillMaskOutputElement.parse_obj_as_list(response)
 
     def image_classification(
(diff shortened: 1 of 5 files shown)


src/huggingface_hub/inference/_generated/_async_client.py
```diff
@@ -3058,18 +2981,55 @@ async def zero_shot_image_classification(
...
+        inputs: Union[str, Dict[str, Any], ContentT],
+        parameters: Optional[Dict[str, Any]] = None,
+    ) -> _InferenceInputs:
+        """
+        Prepare payload for an API request, handling various input types and parameters.
+        """
+
+        def is_raw_content(inputs: Union[str, Dict[str, Any], ContentT]) -> bool:
+            return isinstance(inputs, (bytes, Path)) or (
+                isinstance(inputs, str) and inputs.startswith(("http://", "https://"))
```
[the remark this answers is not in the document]
Wauplin: What do you think of modifying the method signature to

```py
def _prepare_payload(
        inputs: Union[str, Dict[str, Any], ContentT],
        parameters: Optional[Dict[str, Any]],
        expect_binary: bool,
    )
```
?

For tasks that expect a binary input (`image_to_*`, `audio_to_*`), you pass `_prepare_payload(..., expect_binary=True)`.

This way you could have a logic like this:

```py
is_binary = isinstance(inputs, (bytes, Path))

if expect_binary and not is_binary and not isinstance(inputs, str):
    raise ValueError(...)  # should be a binary or at least a string (local path or url)

if expect_binary and not has_parameter:
    return _InferenceInputs(raw_data=inputs)

if not expect_binary and is_binary:
    raise ValueError(...)  # cannot be a binary

# else set as "inputs" in a json payload
...
```

src/huggingface_hub/inference/_client.py
```diff
@@ -109,6 +111,12 @@
 MODEL_KWARGS_NOT_USED_REGEX = re.compile(r"The following `model_kwargs` are not used by the model: \[(.*?)\]")
 
 
+@dataclass
+class _InferenceInputs:
+    json: Optional[Dict[str, Any]] = None
+    raw_data: Optional[ContentT] = None
```
Wauplin, line 117: Potentially slightly overkill. No need to be extremely precise on the return type of `_prepare_payload`. You can also add `-> Dict` as return type with output values like `{"payload": ...}` or `{"data": ...}` (it's never both at the same time).

And then call it like this:

```py
response = self.post(
    **_prepare_payload(inputs, parameters),
    model=model,
    task="document-question-answering",
)
```

src/huggingface_hub/inference/_client.py
```diff
@@ -1166,19 +1156,8 @@ def image_classification(
...
-        else:
-            data = None
-            payload = {"inputs": _b64_encode(image)}
-            for key, value in parameters.items():
-                if value is not None:
-                    payload.setdefault("parameters", {})[key] = value
-
-        response = self.post(json=payload, data=data, model=model, task="image-classification")
+        payload = self._prepare_payload(image, parameters=parameters)
+        response = self.post(json=payload.json, data=payload.raw_data, model=model, task="image-classification")
```
Wauplin, line 1160: Typically here (based on suggestions above), it would lead to something like this:

```suggestion
        response = self.post(
            **_prepare_payload(inputs, parameters, expect_binary=True),
            model=model,
            task="image-classification",
       )
```

(i.e. not a static method, `expect_binary=True` and using `**` on dict return value)

Wauplin (commented):
Thanks @hanouticelina! I'm sorry I realized to late that I commented on both `_client.py` and `_async_client.py` but all comments applied to both (since it's autogenerated). My main concern is about determining if a task expects binaries as input or not (see below).

src/huggingface_hub/inference/_generated/_async_client.py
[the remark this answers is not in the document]
hanouticelina: hum yes you're right! actually I did not update  `image_to_text` as we don't have any parameters or logic to either send a json payload or a raw data:
```python
response = self.post(data=image, model=model, task="image-to-text")
output = ImageToTextOutput.parse_obj(response)
return output[0] if isinstance(output, list) else output
```
but of course your point is totally valid.
Having a flag seems to cover all the cases. In the beginning I though about having a Input type enum and add a `input_type` arg to `_prepare_payload()` but it's simpler to just use a `expect_binary` flag. I don't have a better solution either for now 😕 
I will fix the suggestions and I will get back to this :)

hanouticelina:
thanks @Wauplin for the review! I addressed your suggestions and I ended up adding an `expect_binary` flag as it's the simplest way to handle the special case of image and audio (path, URL and binary) inputs :)

src/huggingface_hub/inference/_common.py
```diff
@@ -259,6 +256,50 @@ def _bytes_to_image(content: bytes) -> "Image":
...
+    """
+    if parameters is None:
+        parameters = {}
+    parameters = {k: v for k, v in parameters.items() if v is not None}
+    has_parameters = len(parameters) > 0
+
+    is_binary = isinstance(inputs, (bytes, Path))
+    # If expect_binary is True, inputs must be a binary object or a local path or a URL.
+    if expect_binary and not is_binary and not isinstance(inputs, str):
+        raise ValueError("Expected binary inputs or a local path or a URL.")
```
Wauplin, line 279:
```suggestion
        raise ValueError(f"Expected binary inputs or a local path or a URL. Got {inputs}.")
```

src/huggingface_hub/inference/_common.py
```diff
@@ -259,6 +256,50 @@ def _bytes_to_image(content: bytes) -> "Image":
...
+    # Send inputs as raw content when no parameters are provided
+    if expect_binary and not has_parameters:
+        return {"data": inputs}
+    # If expect_binary is False, inputs must not be a binary object.
+    if not expect_binary and is_binary:
+        raise ValueError("Unexpected binary inputs.")
```
Wauplin, line 285:
```suggestion
        raise ValueError(f"Unexpected binary inputs. Got {inputs}")
```

src/huggingface_hub/inference/_common.py
```diff
@@ -259,6 +256,50 @@ def _bytes_to_image(content: bytes) -> "Image":
...
+
+    json: Dict[str, Any] = {}
+    # If inputs is a dict, update the json payload with its content
+    if isinstance(inputs, dict):
+        json.update(inputs)
```
Wauplin, line 290: I'm surprised this is needed. I saw that `question_answering` uses it (and that it was already the case before this PR):

```py
        inputs: Dict[str, Any] = {"question": question, "context": context}
        payload = _prepare_payload(inputs, parameters=parameters)
```

But according [to the specs](https://github.com/huggingface/huggingface.js/blob/4b8aa177c4931e8d0f3e1958eecc640285ff5b23/packages/tasks/src/tasks/question-answering/spec/input.json#L8) and to what I would expect as well, the payload should be 

```js
{
    "inputs": {
        "question": ...,
        "context": ...,
    },
    parameters: { ... }
}
``` 

So I think there was a bug in the implementation before this PR and we should fix it in this one. Could you check how things are supposed to behave on a real example? If `inputs` is always needed -which I think is the case- that would simplify things.

Wauplin: **EDIT:** I checked in https://huggingface.co/deepset/roberta-base-squad2 inference widget and it's indeed the case:

image

Wauplin (commented):
Looking good! I have a question related to question answering and then we should be good to merge

src/huggingface_hub/inference/_common.py
hanouticelina: I tried sending this payload (without `inputs`) and it worked. I fixed it though to align with the specs.
```python
{
    "question": "What's my name?",
    "context": "My name is Clara and I live in Berkeley.",
    "parameters": {"top_k": 3},
}
```

src/huggingface_hub/inference/_common.py
```diff
@@ -259,6 +256,47 @@ def _bytes_to_image(content: bytes) -> "Image":
...
+    if not expect_binary and is_binary:
+        raise ValueError(f"Unexpected binary inputs. Got {inputs}")  # type: ignore
+
+    json: Dict[str, Any] = {}
+    # If inputs is a bytes-like object, encode it to base64
+    if isinstance(inputs, (bytes, Path)) or (isinstance(inputs, str) and inputs.startswith(("http://", "https://"))):
+        json["inputs"] = _b64_encode(inputs)
+    # If inputs is a string or a dict, send it as is
+    elif isinstance(inputs, (dict, str)):
+        json["inputs"] = inputs
```
Wauplin, line 293:
```suggestion
    # Otherwise (string, dict, list) send it as is
    else:
        json["inputs"] = inputs
```

Let's forward any remaining thing otherwise users might experience weird issues where the inputs is not set. Typically some tasks work both with a string or a list of string. We don't want to promote that usage but if we can continue to support it, that's for the best. Better to have an InferenceClient that is opinionated in its type annotations + parameters but not constrained if (pseudo-)invalid data is passed.

src/huggingface_hub/inference/_common.py
```diff
@@ -259,6 +256,47 @@ def _bytes_to_image(content: bytes) -> "Image":
...
+    if expect_binary and not has_parameters:
+        return {"data": inputs}
+    # If expect_binary is False, inputs must not be a binary object.
+    if not expect_binary and is_binary:
+        raise ValueError(f"Unexpected binary inputs. Got {inputs}")  # type: ignore
+
+    json: Dict[str, Any] = {}
+    # If inputs is a bytes-like object, encode it to base64
+    if isinstance(inputs, (bytes, Path)) or (isinstance(inputs, str) and inputs.startswith(("http://", "https://"))):
+        json["inputs"] = _b64_encode(inputs)
```
Wauplin, line 290:
```suggestion
    if expect_binary:
        json["inputs"] = _b64_encode(inputs)
```

Given the conditions above, we know inputs is binary or url/path if we reached this point with `expect_binary` been True. 

The problem with `if ... or (isinstance(inputs, str) and inputs.startswith(("http://", "https://"))):` is that if a user passes a URL for a non-binary task (like text classification) we don't want to fetch that URL.

tests/test_inference_client.py
```diff
@@ -1080,3 +1084,121 @@ def test_resolve_chat_completion_url(
...
+                "inputs": "encoded_data",
+                "parameters": {"param1": "value1"},
+            },
+            None,
+        ),
+        # Case 9: inputs is a URL string without parameters
+        (
+            "http://example.com",
+            None,
+            True,
```
Wauplin, line 1172: Would be good to test URL + `expect_binary = False` => shouldn't be encoded. (cf comment above).
Same for case 9 and 10.

Wauplin (commented):
Been a bit picky on corner cases here but I do think it's worth it :innocent:

src/huggingface_hub/inference/_common.py
hanouticelina: yes my bad! I thought I pushed this fix, I git stash these changes when I switched to something else and forgot it 🤦‍♀️

hanouticelina: fixed in [9ac79df](https://github.com/huggingface/huggingface_hub/pull/2601/commits/9ac79df56507586cfada4853afcd33a63dc872f7).

src/huggingface_hub/inference/_generated/_async_client.py
```diff
@@ -3058,13 +2982,11 @@ async def zero_shot_image_classification(
         if len(labels) < 2:
             raise ValueError("You must specify at least 2 classes to compare.")
 
-        payload = {
-            "inputs": {"image": _b64_encode(image), "candidateLabels": ",".join(labels)},
-        }
-        if hypothesis_template is not None:
-            payload.setdefault("parameters", {})["hypothesis_template"] = hypothesis_template
+        inputs = {"image": _b64_encode(image), "candidateLabels": ",".join(labels)}
+        parameters = {"hypothesis_template": hypothesis_template} if hypothesis_template is not None else None
```
Wauplin, line 2986:
```suggestion
        parameters = {"hypothesis_template": hypothesis_template}
```

(nit)

Wauplin (approved):
Thanks @hanouticelina! That should make the inference client more reliable on corner cases in the future. And reduce the amount of duplicate code :smile:

src/huggingface_hub/inference/_client.py
```diff
@@ -2986,13 +2910,11 @@ def zero_shot_image_classification(
         if len(labels) < 2:
             raise ValueError("You must specify at least 2 classes to compare.")
 
-        payload = {
-            "inputs": {"image": _b64_encode(image), "candidateLabels": ",".join(labels)},
-        }
-        if hypothesis_template is not None:
-            payload.setdefault("parameters", {})["hypothesis_template"] = hypothesis_template
+        inputs = {"image": _b64_encode(image), "candidateLabels": ",".join(labels)}
+        parameters = {"hypothesis_template": hypothesis_template} if hypothesis_template is not None else None
```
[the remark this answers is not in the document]
hanouticelina: thanks! found another similar check in another place, fixed it in [c8e7cc8](https://github.com/huggingface/huggingface_hub/pull/2601/commits/c8e7cc830a87b774caec959ac74f58078568f5da)

hanouticelina:
thanks @Wauplin! I think we're good to merge this one

PR #2616, opened 2024-10-18
Update hf filesystem "get all files" example

SwayStar123:
The example is wrong. With the example an error is received
```ValueError: Invalid pattern: '**' can only be an entire path component```


SwayStar123:
@hanouticelina Can you review this

hanouticelina (approved):
Hello @SwayStar123, thanks for the PR! the correct pattern for recursive matching using `HfFileSystem.glob()` is indeed `**/*.csv` :)

SwayStar123:
I cannot merge this, only someone with write access can @hanouticelina

PR #2621, opened 2024-10-21
Update how file's security metadata is retrieved following changes in the API response

hanouticelina:
Following the deployment of [huggingface-internal/moon-landing/pull/11077](https://github.com/huggingface-internal/moon-landing/pull/11077), from now on, the API response of `/{repo_type}s/{repo_id}/tree/{revision}/{path_in_repo}` returns the security scan metadata of a file in the field `securityFileStatus` instead of `security`. This PR reflects these changes and should fix the failing tests.


Wauplin (approved):
Thanks for adapting this @hanouticelina!
Looks good and should be backward and future-compatible :crossed_fingers: 

cc @coyotte508 @krampstudio for viz'

PR #2627, opened 2024-10-22
[`fix`] Add utf8 encoding to read_text to avoid Windows charmap crash

tomaarsen:
Hello!

## Pull Request overview
* Add utf8 encoding to `read_text` to avoid Windows charmap crash

## Details
If I call `push_to_hub` with a README that uses any non-"standard" character, as is common in my automatically generated model cards in Sentence Transformers, then it'll crash on Windows due to this error: https://stackoverflow.com/questions/14630288/unicodeencodeerror-charmap-codec-cant-encode-character-maps-to-undefined

This is always easily fixed by ensuring that the README is read in utf8.

cc @Wauplin @hlky

- Tom Aarsen


hlky:
LGTM!

PR #2620, opened 2024-10-21
bugfix huggingface-cli command execution in python3.8

PineApple777:
When I logged-in with huggingface-cli command in basic ubuntu 20.04 docker image which uses python 3.8, below error occurred.

- command
```
Huggingface-cli login
huggingface-cli login --token hf_xxxx
```
- error log
```
Traceback (most recent call last):
  File "/usr/local/bin/huggingface-cli", line 5, in <module>
    from huggingface_hub.commands.huggingface_cli import main
  File "/usr/local/lib/python3.8/dist-packages/huggingface_hub/commands/huggingface_cli.py", line 26, in <module>
    from huggingface_hub.commands.user import UserCommands
  File "/usr/local/lib/python3.8/dist-packages/huggingface_hub/commands/user.py", line 165, in <module>
    class AuthSwitchCommand(BaseUserCommand):
  File "/usr/local/lib/python3.8/dist-packages/huggingface_hub/commands/user.py", line 203, in AuthSwitchCommand
    def _select_token_name_tui(self, token_names: list[str]) -> Optional[str]:
TypeError: 'type' object is not subscriptable
```

I think that this issue occurred because python3.8 does not support generic types and this leads to bugs at below points. so, I modified the below code to use `List` type within a typing alias. Because huggingface-hub still officially supports Python 3.8, this bug should be resolved. (this issue occur in version 0.26.0)

PR #2628, opened 2024-10-23
Add user CLI unit tests

hanouticelina:
This PR adds unit tests for user CLI commands (login, logout, auth switch and auth list), helping us catch command-related bugs (like #2620) early in our CI pipeline. 
while the underlying authentication functionality is already tested in `tests/test_auth.py`, these tests focus on ensuring the CLI commands themselves work reliably across different environments.


Wauplin (approved):
Thanks for this extensive test suite!

PR #2633, opened 2024-10-24
[InferenceClient] Automatically handle outdated task parameters

hanouticelina:
This PR will complete some tasks from #2063. 
The PR adds the possibility to automatically check for outdated parameters of task methods in the inference client, aligning these task methods with their corresponding parameter specs. More specifically, the script will **check** and **update** parameter types and docstrings. The latter is the most possible to happen as parameter descriptions tend to change more frequently.
- ✅  detect and update outdated parameter's type and/or docstring in method signature.
- ✅ detect and update outdated parameter's type and/or docstring in method docstrings.

some changes in the inference client code are automatically generated after re-running `utils/generate_inference_types.py` and `utils/check_task_parameters.py` to be able to import some missing type aliases from `huggingface_hub/inference/_generated_types` module.

⚠️ the docstring parsing code is a little bit clunky as it's quite difficult to extract or update a parameter's docstring in a structured way using libcst. this implementation relies on heuristics heavily based on the "format" of the method docstring.


Wauplin (approved):
Thanks @hanouticelina, impressive work!

> ⚠️ the docstring parsing code is a little bit clunky as it's quite difficult to extract or update a parameter's docstring in a structured way using libcst. this implementation relies on heuristics heavily based on the "format" of the method docstring.

Let's hope it'll not break too often (I'm quite confident). Let's remember the utility is only meant to help us and is not publicly released so it's fine to have more clunky logic there.

PR #2636, opened 2024-10-27
Support python 3.13

hanouticelina:
this PR adds python 3.13 support. Following the attempt on adding python 3.13 previously in #2605, the webhooks tests were failing because of an import issue in gradio. This is fixed now in the latest gradio version (5.4.0) in [gradio-app/gradio/#9757](https://github.com/gradio-app/gradio/pull/9757).

### changes made
- run CI on Python 3.13.
- run torch@latest only on python 3.12, since there is no support of python 3.13 yet.
- pin new supported version in `setup.py`.
- update documentation.


Wauplin (approved):
Thanks @hanouticelina! Great you followed Gradio's update

PR #2653, opened 2024-11-04
[InferenceClient] Update `max_tokens` and `max_new_tokens` default value in docstring

### LINKED ISSUE #2652: `max_tokens`'s default in docstring of `InferenceClient::chat_completion`'s  is 20 but that of TGI is 100
Hi there!🤗

`InferenceClient::chat_completion`'s docstring states that `max_tokens`'s default value is `20` but it seems that of the TGI is `100`.

https://github.com/huggingface/huggingface_hub/blob/8a99deb4e1c2c70ed180d4429e9da48902cd39bf/src/huggingface_hub/inference/_client.py#L586-L587

https://github.com/huggingface/text-generation-inference/blob/6e3220529df5906ae586031873b7865e9923040b/router/src/lib.rs#L939

hanouticelina:
fixes #2652.

tiny PR to fix incorrect default value of `max_tokens` and `max_new_tokens` in `InferenceClient.text_generation()` and `InferenceClient.chat_completion()` docstrings, reflecting the default values in [text-generation-inference/lib.rs](https://github.com/huggingface/text-generation-inference/blob/main/router/src/lib.rs). 
The default value in `InferenceClient.chat_completion()` was incorrectly documented as 20 instead of 100. The `max_new_tokens` parameter default value was also added in `InferenceClient.text_generation()` docstring, given that it's an important generation parameter.

### DIFF
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -584,7 +584,7 @@ class InferenceClient:
                 Whether to return log probabilities of the output tokens or not. If true, returns the log
                 probabilities of each output token returned in the content of message.
             max_tokens (`int`, *optional*):
-                Maximum number of tokens allowed in the response. Defaults to 20.
+                Maximum number of tokens allowed in the response. Defaults to 100.
             n (`int`, *optional*):
                 UNUSED.
             presence_penalty (`float`, *optional*):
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for the update @hanouticelina! 
Most likely due to a difference between TGI and `transformers` at some point. Better to document the TGI default values :+1:

PR #2657, opened 2024-11-07
Don't write the ref file unless necessary

d8ahazard:
Updating snapshot_download so that it doesn't re-write the commit hash *every* single time. 

This has been incredibly problematic if we're keeping a repo saved to disk, and snapshot_download can't write to that disk.

This should at least partially resolve the issue.

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -258,16 +258,22 @@ def snapshot_download(
     # If the file exists, read the commit hash from it
     write_commit_hash = True
     if os.path.exists(ref_path):
-        with open(ref_path) as f:
-            existing_commit_hash = f.read()
-        if existing_commit_hash == commit_hash:
-            write_commit_hash = False
+        try:
+            with open(ref_path) as f:
+                existing_commit_hash = f.read()
+            if existing_commit_hash == commit_hash:
+                write_commit_hash = False
+        except Exception as e:
+            logger.warning(f"Error reading commit hash from {ref_path}: {e}")
 
-    # Write the commit hash to the ref file if it doesn't exist or is different
-    if revision != commit_hash and write_commit_hash:
-        os.makedirs(os.path.dirname(ref_path), exist_ok=True)
-        with open(ref_path, "w") as f:
-            f.write(commit_hash)
+    try:
+        # Write the commit hash to the ref file if it doesn't exist or is different
+        if revision != commit_hash and write_commit_hash:
+            os.makedirs(os.path.dirname(ref_path), exist_ok=True)
+            with open(ref_path, "w") as f:
+                f.write(commit_hash)
+    except Exception as e:
+        logger.warning(f"Error writing commit hash to {ref_path}: {e}")
 
     # we pass the commit_hash to hf_hub_download
     # so no network call happens if we already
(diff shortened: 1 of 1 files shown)


Wauplin:
What's the underlying issue you are facing? Is it a permission error? In that case I'd be more inclined to add a try/except instead of additional logic to read before writing (more error-prone)

d8ahazard:
If the model is on a read-only drive, this will break. My logic fixes that.
Ideally, you should probably have both. Only write when you need to, and
also wrap it all in a try catch so it doesn't break the whole program if
something fails.

d8ahazard:
> What's the underlying issue you are facing? Is it a permission error? In that case I'd be more inclined to add a try/except instead of additional logic to read before writing (more error-prone)

See my additional commit - I added exception handling for both parts of dealing with the ref file.

src/huggingface_hub/_snapshot_download.py
```diff
@@ -254,11 +254,26 @@ def snapshot_download(
...
+            logger.warning(f"Error reading commit hash from {ref_path}: {e}")
+
+    try:
+        # Write the commit hash to the ref file if it doesn't exist or is different
+        if revision != commit_hash and write_commit_hash:
+            os.makedirs(os.path.dirname(ref_path), exist_ok=True)
+            with open(ref_path, "w") as f:
+                f.write(commit_hash)
+    except Exception as e:
+        logger.warning(f"Error writing commit hash to {ref_path}: {e}")
```
Wauplin, line 276: Hi @d8ahazard, thanks for the modification. I'd be more inclined to have a more concise logic as it's simpler to maintain. This might cause some unnecessary warnings in the logs in case commit hash is already correct but I don't think it's too problematic as long as the error is caught and ignored.

Also as a nit, it's preferable to catch `OSError` instead of any `Exception`.

```suggestion
    if revision != commit_hash:
        ref_path = os.path.join(storage_folder, "refs", revision)
        try:
            os.makedirs(os.path.dirname(ref_path), exist_ok=True)
            with open(ref_path, "w") as f:
                f.write(commit_hash)
    except OSError as e:
        logger.warning(f"Ignored error while writing commit hash to {ref_path}: {e}.")
```

src/huggingface_hub/_snapshot_download.py
```diff
@@ -256,9 +256,12 @@ def snapshot_download(
...
         ref_path = os.path.join(storage_folder, "refs", revision)
-        os.makedirs(os.path.dirname(ref_path), exist_ok=True)
-        with open(ref_path, "w") as f:
-            f.write(commit_hash)
+        try:
+            os.makedirs(os.path.dirname(ref_path), exist_ok=True)
+            with open(ref_path, "w") as f:
+                f.write(commit_hash)
+    except OSError as e:
+        logger.warning(f"Ignored error while writing commit hash to {ref_path}: {e}.")
```
Wauplin, line 264:
```suggestion
        except OSError as e:
            logger.warning(f"Ignored error while writing commit hash to {ref_path}: {e}.")
```

Wauplin (approved):
Thanks @d8ahazard!

d8ahazard:
> Thanks @d8ahazard!

Thank you! You have no idea how many headaches this is going to solve for me. ;)

PR #2660, opened 2024-11-11
Fix the error caused by reading `.DS_Store` when running the `scan-cache` command.

johnmai-dev:
As title. 

Same as #2112

iTerm2 2024-11-11 17 42 11


Wauplin (approved):
Makes sense! Thanks for spotting and fixing this issue :hugs:

PR #2696, opened 2024-12-05
[Serialization] add argument to pass shared tensors names to drop when saving

hanouticelina:
Related to [transformers#35080](https://github.com/huggingface/transformers/issues/35080).

Currently, when saving a torch state dict with shared tensors, the key to keep is chosen alphabetically which might not be the desired behavior. This PR simply adds an optional `state_dict_keys_to_discard` argument to both `save_torch_state_dict()` and `save_torch_model()` that allows users to specify which keys should be discarded in priority when duplicates are found.
_Note_ : The logic behind choosing which key should be discarded in priority is framework/model-specific and should be handled by the user. For example, for `transformers.PreTrainedModel` models, this is specified in the `_tied_weights_keys` attribute, example usage:

```python
model = ... # A model inheriting from PreTrainedModel
save_torch_model(
    model,
    "path/to/folder",
    state_dict_keys_to_discard=model._tied_weights_keys,
)
```
or with `save_torch_state_dict()`
```python
model = ... # A model inheriting from PreTrainedModel
save_torch_state_dict(
    model.state_dict(),
    "path/to/folder",
    state_dict_keys_to_discard=model._tied_weights_keys,
)
```

### DIFF
--- a/src/huggingface_hub/serialization/_torch.py
+++ b/src/huggingface_hub/serialization/_torch.py
@@ -42,6 +42,7 @@ def save_torch_model(
     metadata: Optional[Dict[str, str]] = None,
     safe_serialization: bool = True,
     is_main_process: bool = True,
+    shared_tensors_to_discard: Optional[List[str]] = None,
 ):
     """
     Saves a given torch model to disk, handling sharding and shared tensors issues.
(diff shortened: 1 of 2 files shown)


SunMarc (commented):
Nice ! Thanks for adding this !

Wauplin (approved):
Thank you !

PR #2698, opened 2024-12-06
Enable tqdm progress in cloud environments

cbensimon:
Cloud environments usually enable [`TQDM_POSITION=-1`](https://github.com/tqdm/tqdm#:~:text=Consoles%20in%20general%3A%20require%20support%20for%20carriage%20return) to disable carriage return and get clean tqdm progress logging

The idea of this PR is to use this environment variable as a hint to force enabling tqdm progress despite the absence of TTY

This will at least re-enable download progress in Spaces

https://github.com/huggingface/huggingface_hub/pull/2000 follow-up

### DIFF
--- a/src/huggingface_hub/lfs.py
+++ b/src/huggingface_hub/lfs.py
@@ -39,6 +39,7 @@ from .utils import (
 )
 from .utils._lfs import SliceFileObj
 from .utils.sha import sha256, sha_fileobj
+from .utils.tqdm import is_tqdm_disabled
 
 
 if TYPE_CHECKING:
@@ -430,17 +431,13 @@ def _upload_parts_hf_transfer(
     if len(desc) > 40:
         desc = f"(…){desc[-40:]}"
 
-    # set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
-    # see https://github.com/huggingface/huggingface_hub/pull/2000
-    disable = True if (logger.getEffectiveLevel() == logging.NOTSET) else None
-
     with tqdm(
         unit="B",
         unit_scale=True,
         total=total,
         initial=0,
         desc=desc,
-        disable=disable,
+        disable=is_tqdm_disabled(logger.getEffectiveLevel()),
         name="huggingface_hub.lfs_upload",
     ) as progress:
         try:


Wauplin (commented):
Thanks for investigating this @cbensimon. Solution looks ok to me :+1: 
Could you update [this line](https://github.com/huggingface/huggingface_hub/blob/78b4e5ef12a8000894df86f306b387998b4b39a7/src/huggingface_hub/lfs.py#L435) as well please?

src/huggingface_hub/utils/tqdm.py
```diff
@@ -196,6 +198,16 @@ def are_progress_bars_disabled(name: Optional[str] = None) -> bool:
...
 
 
+def is_tqdm_disabled(log_level: int) -> Optional[bool]:
+    if log_level == logging.NOTSET:
+        return True
+    if os.getenv("TQDM_POSITION") == "-1":
+        return False
+    return None
+    # ^ set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
+    # see https://github.com/huggingface/huggingface_hub/pull/2000
```
hanouticelina, line 208: (nit)
```suggestion
def is_tqdm_disabled(log_level: int) -> Optional[bool]:
    """
    Determine if tqdm progress bars should be disabled based on logging level and environment settings.

    see https://github.com/huggingface/huggingface_hub/pull/2000 and https://github.com/huggingface/huggingface_hub/pull/2698.
    """
    if log_level == logging.NOTSET:
        return True
    if os.getenv("TQDM_POSITION") == "-1":
        return False
    return None
```

hanouticelina (commented):
TIL the nuance between disable=False and disable=None in tqdm. 

same comment as @Wauplin about updating [src/huggingface_hub/lfs.py#L435](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/lfs.py#L435). Otherwise looks good to me! thanks @cbensimon!

Wauplin (approved):
Thanks @cbensimon !

hanouticelina (approved):
thanks @cbensimon for the PR!

PR #2711, opened 2024-12-16
[InferenceClient] Keep `generate_kwargs` instead of `generation_parameters`

hanouticelina:
Following up on the failing test in transformers related to Hub spec compliance for pipelines (see the related test [here](https://github.com/huggingface/transformers/blob/66531a1ec3e2aafe7ffb23a9ca715cfb67b9fea0/tests/test_pipeline_mixin.py#L932)). The test failure occurs because there's a mismatch between `generate_kwargs` (current name in transformers) and `generation_parameters` (name in huggingface.js specs).
For now, we'll maintain the use of `generate_kwargs` since the renaming hasn't been implemented in transformers. This is purely a naming convention issue, and we'll update it to `generation_parameters` when we do the full rename in `transformers`.

Note that in huggingface.js, this field was originally named `generate` (not `generate_kwargs`). It was later renamed to `generation_parameters` (see this issue: https://github.com/huggingface/huggingface.js/issues/923). This rename was safe to do since InferenceAPI and Inference Endpoints do not support `generate` or `generate_kwargs` anyways.

### DIFF
--- a/src/huggingface_hub/inference/_generated/types/automatic_speech_recognition.py
+++ b/src/huggingface_hub/inference/_generated/types/automatic_speech_recognition.py
@@ -80,6 +80,9 @@ class AutomaticSpeechRecognitionParameters(BaseInferenceType):
     """Parametrization of the text generation process"""
     return_timestamps: Optional[bool] = None
     """Whether to output corresponding timestamps with the generated text"""
+    # Will be deprecated in the future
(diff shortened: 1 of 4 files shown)


hanouticelina:
since it's a naming convention only and doesn't have any effect on any feature, I will merge this :) I will also open a PR later in huggingface.js to revert the renaming.

PR #2719, opened 2024-12-24
Add http proxies for HfApi

TPEC:
Add http proxies for HfApi.__init__, and pass proxies from snapshot_download to HfApi.

* The default value of proxies is None, which means no proxy(keep the previous action).
* If proxies is set, all http requests to the endpoint will go through proxies.
* Pass proxies from snapshot_download() to HfApi.__init__.

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -151,6 +151,7 @@ def snapshot_download(
                 user_agent=user_agent,
                 endpoint=endpoint,
                 headers=headers,
+                proxies=proxies,
             )
             repo_info = api.repo_info(repo_id=repo_id, repo_type=repo_type, revision=revision, token=token)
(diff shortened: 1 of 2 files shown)


hanouticelina:
Hi @TPEC, thanks for your contribution! If you want to use a custom proxy, you can use [huggingface_hub.configure_http_backend()](https://huggingface.co/docs/huggingface_hub/v0.27.0/en/package_reference/utilities#huggingface_hub.configure_http_backend) that allows you to configure how HTTP calls are made. If you need to use proxies in your environment, you can configure them globally like this:
```python
from huggingface_hub import configure_http_backend
import requests

def custom_backend():
    session = requests.Session()
    session.proxies = your_proxies
    return session

configure_http_backend(custom_backend)
```
I'll be closing this PR since proxy configuration is already handled through `configure_http_backend()`.

PR #2721, opened 2024-12-26
[BUG] Fix 404 NOT FOUND issue caused by endpoint tail slash

Mingqi2:
The issue was triggered when user added a **tail slash '/'** in `HF_ENDPOINT`, like `$set HF_ENDPOINT=https://hf-mirror.com/`. 

Then it's interesting (stupid?) to visit a link like - [https://hf-mirror.com//api/models/sentence-transformers/all-MiniLM-L6-v2/revision/main](https://hf-mirror.com/api/models/sentence-transformers/all-MiniLM-L6-v2/revision/main) by huggingface.

Okay then you got 404 and don't know why. Bad experience and you should take care of them in the code with urljoin and carefully strip.

### DIFF
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -2,7 +2,7 @@ import os
 import re
 import typing
 from typing import Literal, Optional, Tuple
-form urllib.parse import urljoin
+from urllib.parse import urljoin
 
 # Possible values for env variables
 
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -63,7 +63,9 @@ _staging_mode = _is_true(os.environ.get("HUGGINGFACE_CO_STAGING"))
 
 _HF_DEFAULT_ENDPOINT = "https://huggingface.co"
 _HF_DEFAULT_STAGING_ENDPOINT = "https://hub-ci.huggingface.co"
-ENDPOINT = os.getenv("HF_ENDPOINT") or (_HF_DEFAULT_STAGING_ENDPOINT if _staging_mode else _HF_DEFAULT_ENDPOINT)
+ENDPOINT = os.getenv("HF_ENDPOINT", "").rstrip("/") or (
+    _HF_DEFAULT_STAGING_ENDPOINT if _staging_mode else _HF_DEFAULT_ENDPOINT
+)
 
 HUGGINGFACE_CO_URL_TEMPLATE = urljoin(ENDPOINT, "/{repo_id}/resolve/{revision}/{filename}")
 HUGGINGFACE_HEADER_X_REPO_COMMIT = "X-Repo-Commit"
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -2,7 +2,7 @@ import os
 import re
 import typing
 from typing import Literal, Optional, Tuple
-
+form urllib.parse import urljoin
 
 # Possible values for env variables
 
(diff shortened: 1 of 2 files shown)


Mingqi2:
@Wauplin could you please review this PR? thanks.

src/huggingface_hub/constants.py
```diff
@@ -65,7 +65,7 @@ def _as_int(value: Optional[str]) -> Optional[int]:
 _HF_DEFAULT_STAGING_ENDPOINT = "https://hub-ci.huggingface.co"
 ENDPOINT = os.getenv("HF_ENDPOINT") or (_HF_DEFAULT_STAGING_ENDPOINT if _staging_mode else _HF_DEFAULT_ENDPOINT)
```
hanouticelina, line 66:
```suggestion
ENDPOINT = os.getenv("HF_ENDPOINT", "").rstrip("/") or (
    _HF_DEFAULT_STAGING_ENDPOINT if _staging_mode else _HF_DEFAULT_ENDPOINT
)
```
I'd rather normalize `ENDPOINT` here as it is used throughout the codebase.

src/huggingface_hub/constants.py
```diff
@@ -65,7 +65,7 @@ def _as_int(value: Optional[str]) -> Optional[int]:
...
 
-HUGGINGFACE_CO_URL_TEMPLATE = ENDPOINT + "/{repo_id}/resolve/{revision}/{filename}"
+HUGGINGFACE_CO_URL_TEMPLATE = urljoin(ENDPOINT, "/{repo_id}/resolve/{revision}/{filename}")
```
hanouticelina, line 68: `urljoin()` is not necessary here if `ENDPOINT` is already normalized. see my comment here.

src/huggingface_hub/utils/_http.py
```diff
@@ -341,6 +341,7 @@ def fix_hf_endpoint_in_url(url: str, endpoint: Optional[str]) -> str:
     endpoint = endpoint or constants.ENDPOINT
```
hanouticelina, line 341:
```suggestion
    endpoint = endpoint.rstrip("/") if endpoint else constants.ENDPOINT
```

src/huggingface_hub/utils/_http.py
```diff
@@ -341,6 +341,7 @@ def fix_hf_endpoint_in_url(url: str, endpoint: Optional[str]) -> str:
...
     # check if a proxy has been set => if yes, update the returned URL to use the proxy
     if endpoint not in (None, constants._HF_DEFAULT_ENDPOINT, constants._HF_DEFAULT_STAGING_ENDPOINT):
+        endpoint.rstrip('/')
```
hanouticelina, line 344:
```suggestion
    if endpoint not in (constants._HF_DEFAULT_ENDPOINT, constants._HF_DEFAULT_STAGING_ENDPOINT):
```
Unrelated to this PR but we could simplify this condition since `endpoint` cannot be `None` at this point!

hanouticelina (commented):
Hi @Mingqi2, one of the maintainers of `huggingface_hub` here 👋  
thanks a lot for your contribution! indeed, some servers might block double slashes in URLs as they can potentially lead to security vulnerabilities (see this stackoverflow [comment](https://stackoverflow.com/a/40683096) for more context). This is not the case for https://huggingface.co but since we also allow custom endpoints, we can handle this case properly with a minimal endpoint normalization!

src/huggingface_hub/constants.py
Mingqi2: Yes but I prefer to keep this way when we connect url parts. It can avoid potential issues caused by formats.

Mingqi2:
@hanouticelina thanks for your warm reply! I have commited most of your changes.

src/huggingface_hub/constants.py
```diff
@@ -2,7 +2,7 @@
 import re
 import typing
 from typing import Literal, Optional, Tuple
-
+form urllib.parse import urljoin
```
hanouticelina, line 5:
```suggestion
from urllib.parse import urljoin
```

src/huggingface_hub/constants.py
hanouticelina: Not a strong opinion on that, i'm okay to keep `urljoin()`

hanouticelina (commented):
just noticed a typo in the imports, once it's fixed, we will be good to merge! 🎉

Mingqi2:
> just noticed a typo in the imports, once it's fixed, we will be good to merge! 🎉

Good catch. It's committed thanks! Let's merge 🎉

hanouticelina (approved):
All good! thanks @Mingqi2 for your contribution! 🤗 
failing tests are unrelated and will be fixed when #2728 will be merged.

Issue #2720, opened 2024-12-25
Inference Client chat completion parameter logit_bias not working


joetaylor94:
### Describe the bug

According to [the docs](https://huggingface.co/docs/huggingface_hub/en/package_reference/inference_client#huggingface_hub.InferenceClient.chat_completion), the `logit_bias` parameter for the `chat_completion` function expects a "JSON object that maps tokens (specified by their token ID in the tokenizer) to an associated bias value from -100 to 100". The type annotation, however says that it should be an Optional[List[float]].

Indeed, if I try to pass in a dictionary, e.g.

`completion = client.chat_completion(
    model="meta-llama/Llama-3.3-70B-Instruct", 
	  messages=messages,
	  max_tokens=100,
    logit_bias={100: 4}
)` 

I get an `HTTPError: 422 Client Error: Unprocessable Entity for url` error. I can pass in a list of floats, but have no idea how this is supposed to encode logit biases without a mapping.

### Reproduction

```
from huggingface_hub import InferenceClient

client = InferenceClient(api_key="hf_xxx")

messages = [
	{
		"role": "user",
		"content": "The capital of France is"
	}
]

completion = client.chat_completion(
    model="meta-llama/Llama-3.3-70B-Instruct", 
	  messages=messages,
	  max_tokens=20,
    logit_bias={100: 4}
)

print(completion.choices[0].message)
```

### System info

```shell
N/A
```

juanwisz:
I think this is not correctly implemented on TGI

hanouticelina:
Hi @joetaylor94, as mentioned by @juanwisz, `logit_bias` is not implemented in TGI and it is included in the documentation as part of matching the same HTTP interface as the `OpenAI` client. Our documentation is automatically generated from TGI's, I will see if there is a way to add an "UNUSED" flag in our documentation as well.

PR #2724, opened 2024-12-30
[InferenceClient] flag `chat_completion()`'s `logit_bias` as UNUSED

hanouticelina:
Related to #2720.
logit_bias is not implemented in TGI (see the comment [here](https://github.com/huggingface/text-generation-inference/blob/23bc38b10d06f8cc271d086c26270976faf67cc2/router/src/lib.rs#L820)), this PR updates the parameter documentation to flag it as unused.


juanwisz:
Could we consider actually implementing logit_bias? It's a very valuable feature for controlling model outputs, I've opened  in TGI to discuss implementation details: https://github.com/huggingface/text-generation-inference/issues/2869

hanouticelina:
> Could we consider actually implementing logit_bias? It's a very valuable feature for controlling model outputs, I've opened in TGI to discuss implementation details: [huggingface/text-generation-inference#2869](https://github.com/huggingface/text-generation-inference/issues/2869)

@juanwisz This PR only aligns the client documentation with the current TGI implementation. Once/If `logit_bias` is implemented on TGI side, we'll update the documentation accordingly 🤗

LysandreJik:
Is it possible to remove that doc attribute altogether? If not, I would at least specify why it's unused and how to have users request for this argument to be used

hanouticelina:
> Is it possible to remove that doc attribute altogether? If not, I would at least specify why it's unused and how to have users request for this argument to be used

@LysandreJik `logit_bias` and `n` (i.e number of chat completion choices to generate for each input message) which is already flagged as unused, are implemented in the OpenAI API, and just like we expose these parameters in TGI to maintain OpenAI compatibility, we want to keep them in the client for the same reason —making `InferenceClient` a proper drop-in replacements for OpenAI client. I've updated their documentation to indicate they are currently not implemented in TGI while explaining why they are still included.

LysandreJik (approved):
Ok great! Thank you Célina !

Issue #2727, opened 2025-01-03
Calling `typing.get_type_hints` on a `ModelHubMixin` breaks


aliberts:
### Describe the bug

Calling `typing.get_type_hints` on a class inheriting from `ModelHubMixin` gets you a NameError when that function tries to resolve its type because `"DataclassInstance"` (which is what `ModelHubMixin` uses in its annotations) doesn't seem to be available in the scope when that happens.

I'll see if I can fix this and open a PR.

### Reproduction

```python
import typing
from huggingface_hub import ModelHubMixin

class MyModel(ModelHubMixin): ...

my_model = MyModel()
typing.get_type_hints(my_model)
```

### Logs

```shell
Traceback (most recent call last):
  File "/Users/simon/projects/draccus/demo2.py", line 7, in <module>
    typing.get_type_hints(my_model)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/hub_310/lib/python3.10/typing.py", line 1871, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/hub_310/lib/python3.10/typing.py", line 329, in _eval_type
    ev_args = tuple(_eval_type(a, globalns, localns, recursive_guard) for a in t.__args__)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/hub_310/lib/python3.10/typing.py", line 329, in <genexpr>
    ev_args = tuple(_eval_type(a, globalns, localns, recursive_guard) for a in t.__args__)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/hub_310/lib/python3.10/typing.py", line 327, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/hub_310/lib/python3.10/typing.py", line 694, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'DataclassInstance' is not defined
```

### System info

```shell
- huggingface_hub version: 0.27.0
- Platform: macOS-15.2-arm64-arm-64bit
- Python version: 3.10.16
- Running in iPython ?: No
- Running in notebook ?: No
- Running in Google Colab ?: No
- Running in Google Colab Enterprise ?: No
- Token path ?: /Users/simon/.cache/huggingface/token
- Has saved token ?: True
- Who am I ?: aliberts
[...]

PR #2729, opened 2025-01-03
Fix `typing.get_type_hints` call on a `ModelHubMixin`

aliberts:
## Description

Proposed fix for #2727

`typing.get_type_hints` tries to resolve the type of an object at runtime. This means that the `global` and `local` namespaces this function uses won't have access to whatever has been defined only for static type checkers (typically using `typing.TYPE_CHECKING`).

In this case, `DataclassInstance` is only defined in such a way.
This proposes to add a default fallback definition of `DataclassInstance` for runtime type checking.

### DIFF
--- a/src/huggingface_hub/hub_mixin.py
+++ b/src/huggingface_hub/hub_mixin.py
@@ -38,7 +38,7 @@ logger = logging.get_logger(__name__)
 
 # Type alias for dataclass instances, copied from https://github.com/python/typeshed/blob/9f28171658b9ca6c32a7cb93fbb99fc92b17858b/stdlib/_typeshed/__init__.pyi#L349
 class DataclassInstance(Protocol):
-    __dataclass_fields__: ClassVar[Dict[str, Field[Any]]]
+    __dataclass_fields__: ClassVar[Dict[str, Field]]
 
 
 # Generic variable that is either ModelHubMixin or a subclass thereof
--- a/src/huggingface_hub/hub_mixin.py
+++ b/src/huggingface_hub/hub_mixin.py
@@ -26,7 +26,7 @@ from .utils import (
 
 # Type alias for dataclass instances
 class DataclassInstance(Protocol):
-    __dataclass_fields__: ClassVar[dict[str, Field[Any]]]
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hub_mixin.py
```diff
@@ -25,7 +25,15 @@
...
-    from _typeshed import DataclassInstance
+    from _typeshed import DataclassInstance  # type: ignore
+else:
+
+    class DataclassInstance(Protocol):  # type: ignore
+        __dataclass_fields__: ClassVar[Dict[str, Field]]
+
+
+Dataclass = TypeVar("Dataclass", bound=DataclassInstance)
+DataclassType = Type[Dataclass]
```
hanouticelina, line 36:
```suggestion
# Type alias for dataclass instances
class DataclassInstance(Protocol):
    __dataclass_fields__: ClassVar[dict[str, Field[Any]]]
```
not a strong opinion on that and feel free to ignore, but maybe we can just define `DataclassInstance` directly here and remove the if/else and `_typeshed` dependency.

hanouticelina (approved):
thanks @aliberts for the fix! all good to me :)

src/huggingface_hub/hub_mixin.py
aliberts: Agreed, fixed: https://github.com/huggingface/huggingface_hub/pull/2729/commits/b1b9f5fde98d52501f20af53da1dfa77abbabcad

hanouticelina:
Added a test in [a5c49e4](https://github.com/huggingface/huggingface_hub/pull/2729/commits/a5c49e45f7c861bfc1519467ef9adb124550c525). thanks again @aliberts for the PR, we're good to merge once the CI is 🟢 I will release a patch with this fix shortly!

aliberts:
@hanouticelina not sure why but that definition doesn't work with python 3.8 as it's saying that dataclass.Field is not subscriptable. I've replaced `Field[Any]` with `Field` (https://github.com/huggingface/huggingface_hub/pull/2729/commits/93b53a7fdb05cfdc3d86c4ae568c6c08353cd4e6), should be good this time.

PR #2738, opened 2025-01-06
[HfApi] Remove unused parameter from method's docstring

### LINKED ISSUE #2731: `get_safetensors_metadata()` has no actual `filename` parameter despite showing one in the documentation
The docstring (and auto-generated web docs) for `get_safetensors_metadata()` claim there is a `filename` parameter, but there is none in the actual [function signature](https://github.com/huggingface/huggingface_hub/blob/3618a38f80ae3073b95ea0fba26f363d2ec4ece8/src/huggingface_hub/hf_api.py#L5400). Example code:

```
>>> from huggingface_hub import get_safetensors_metadata
>>> x = get_safetensors_metadata("dslim/bert-base-NER")  # This works fine
>>> x = get_safetensors_metadata("dslim/bert-base-NER", filename="model.safetensors")  # This fails
  File "<stdin>", line 1, in <module>
TypeError: HfApi.get_safetensors_metadata() got an unexpected keyword argument 'filename'
>>> import huggingface_hub
>>> huggingface_hub.__version__  # I'm using a slightly old version of HF Hub but the git blame says the function signature is entirely unchanged for the last 9 months
'0.24.6'
```

Ideally, an error like this should be caught by a pre-commit hook.

hanouticelina:
closes #2731.
`HfApi.get_safetensors_metadata()` does not take `filename` as a parameter but it still appears in the method's docstring. This PR fixes the docstring accordingly.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -5425,8 +5425,6 @@ class HfApi:
         Args:
             repo_id (`str`):
                 A user or an organization name and a repo name separated by a `/`.
-            filename (`str`):
-                The name of the file in the repo.
             repo_type (`str`, *optional*):
                 Set to `"dataset"` or `"space"` if the file is in a dataset or space, `None` or `"model"` if in a
                 model. Default is `None`.


hanouticelina:
bypassing a review approval as the PR only updates a method's docstring 🤗

PR #2812, opened 2025-01-30
[InferenceClient] Better handling of task parameters 

hanouticelina:
This PR:
- Fixes some discrepancies in `text-to-image` parameters (particularly for Together AI).
- Adds a new `extra_parameters` argument to `text_to_image`, `text_to_speech` and `text_to_video` to be able to pass a provider's unique parameters for these tasks.
- Improves the test suite of inference providers.


src/huggingface_hub/inference/_providers/fal_ai.py
```diff
@@ -134,7 +134,8 @@ def __init__(self):
 
     def _prepare_payload(self, inputs: Any, parameters: Dict[str, Any]) -> Dict[str, Any]:
         parameters = {k: v for k, v in parameters.items() if v is not None}
-        if "image_size" not in parameters and "width" in parameters and "height" in parameters:
+        parameters["image_size"] = parameters.pop("target_size", None)
```
hanouticelina, line 137: imo we should keep either target_size (as an enum) or (width, height) in the arguments of text_to_image. we currently support both 😕

Wauplin (approved):
Haven't tested anything locally but it looks good to me. Thanks for handling these discrepancies!

src/huggingface_hub/inference/_providers/fal_ai.py
hanouticelina: yes, i mean as a dict

src/huggingface_hub/inference/_client.py
```diff
@@ -2407,7 +2405,7 @@ def text_to_image(
         scheduler: Optional[str] = None,
         target_size: Optional[TextToImageTargetSize] = None,
         seed: Optional[int] = None,
-        **kwargs,
```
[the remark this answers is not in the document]
hanouticelina: i'll mention this in the next release notes! but it should be fine, if users were previously using `text_to_image` with the HF Inference API, this shouldn't be an issue since all API parameters were exposed as explicit method arguments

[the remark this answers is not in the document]
Wauplin: > this shouldn't be an issue since all API parameters were exposed as explicit method arguments

yes exactly

src/huggingface_hub/inference/_client.py
```diff
@@ -2443,6 +2441,9 @@ def text_to_image(
                 The size in pixel of the output image
             seed (`int`, *optional*):
                 Seed for the random number generator.
+            extra_parameters (`Dict[str, Any]`, *optional*):
```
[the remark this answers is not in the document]
hanouticelina: yep, added in [305c720](https://github.com/huggingface/huggingface_hub/pull/2812/commits/305c7201bc1e96cf5d7320fffecbee4f6dfdb8f8)

src/huggingface_hub/inference/_providers/fal_ai.py
hanouticelina: splitted the nested dict [d927057](https://github.com/huggingface/huggingface_hub/pull/2812/commits/d9270579f97856e372be058a27c5a8c33e664deb) and opened a PR in hf.js https://github.com/huggingface/huggingface.js/pull/1158

src/huggingface_hub/inference/_providers/fal_ai.py
```diff
@@ -134,7 +134,7 @@ def __init__(self):
 
     def _prepare_payload(self, inputs: Any, parameters: Dict[str, Any]) -> Dict[str, Any]:
         parameters = {k: v for k, v in parameters.items() if v is not None}
-        if "image_size" not in parameters and "width" in parameters and "height" in parameters:
+        if "width" in parameters and "height" in parameters:
```
Wauplin, line 137: what if only one if passed btw?

Wauplin (approved):
Looks good to me!

src/huggingface_hub/inference/_providers/fal_ai.py
hanouticelina: we should be able to send only one if specified, the other one would be set to the default value. I'll fix that

hanouticelina: no actually for fal-ai, you either send both, or neither.

Wauplin: ok thanks for checking :+1:

Wauplin (approved):
approving, though I'm not sure to understand what's happening in `utils/generate_inference_types.py`. I trust it we result is correct :)

hanouticelina:
should be good now, let's merge!

julien-c (commented):
(nit) i think OpenAI calls this param `extra_body` (https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#extra-parameters or from the Python openai sdk), but i think our name is probably fine too 🤷

Wauplin:
Then I'm down to go with `extra_body` before making a release for consistency

julien-c:
up to you (i'm not sure it's exactly the same use case, but my brain is fried)

hanouticelina:
It's actually the same use case, you can send an extra parameter using openai client, but i'm not sure if it's `extra_body` or `extra_query` (they expose both). Let me double check and open a PR to use the same naming on our side.

julien-c:
extra_query goes into the URL query params no? (just guessing)

hanouticelina:
yep, just wanted to double check in their docstrings, since it's not really documented. 
Indeed, `extra_body` -> adds parameters to the JSON body and `extra_query` -> adds parameters as query string to the URL.
Let's rename `extra_parameters` to `extra_body` then!

PR #2826, opened 2025-02-03
[InferenceClient] Fix `automatic-speech-recognition` output parsing and add `extra_body` as an argument

hanouticelina:
should fix https://github.com/huggingface/huggingface_hub/issues/2706#issuecomment-2631115722 and it should work correctly with fal as well, since the timestamp chunks is returned in `timestamp` instead of `timestamps` (see [here](https://fal.ai/models/fal-ai/whisper/api#type-WhisperChunk)).
cc @kiansierra

### DIFF
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -463,6 +463,7 @@ class InferenceClient:
         audio: ContentT,
         *,
         model: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


kiansierra:
Yes, now it's working as expected.

One thing that I couldn't get with calling the client post, was send the other parameters, but I guess that might depend on the server side code too

https://huggingface.co/openai/whisper-large-v3-turbo

PR #2836, opened 2025-02-05
[InferenceClient] Add dynamic inference providers mapping

hanouticelina:
Companion PR to https://github.com/huggingface/huggingface.js/pull/1173. This is to give a first idea on how to implement the dynamic inference providers mapping before moving on with the TODOs.

TODOs: (a bit similar to the hf.js PR)
- handle prod/staging status.
- handle if model supports both `text-generation` and `conversational`.
- add and update tests.
- ✅ The `HfApi().model_info` call is cached.


src/huggingface_hub/inference/_providers/fal_ai.py
```diff
@@ -12,31 +12,6 @@
...
-        "Warlord-K/Sana-1024": "fal-ai/sana",
-    },
-    "text-to-speech": {
-        "m-a-p/YuE-s1-7B-anneal-en-cot": "fal-ai/yue",
-    },
-    "text-to-video": {
-        "genmo/mochi-1-preview": "fal-ai/mochi-v1",
-        "tencent/HunyuanVideo": "fal-ai/hunyuan-video",
-    },
-}
```
julien-c, line 38: i'm wondering if, for Developer experience, we should still support an (empty by default) mapping that Partners can use locally before implementing the server-side mapping

julien-c: (and in the future let's add a link to some doc on how to implement the server-side mapping here)

src/huggingface_hub/hf_api.py
```diff
@@ -132,56 +132,47 @@
...
+    "resourceGroup",
     "sha",
     "siblings",
-    "spaces",
     "tags",
-    "transformersInfo",
     "trendingScore",
-    "widgetData",
     "usedStorage",
-    "resourceGroup",
```
Wauplin, line 162: wow, a lot of them have been removed ?

hanouticelina: No i might messed up the reordering 🙈 let me fix that

Wauplin:
(updated the PR to align with hf.js client + did some refacto. Still some tests to fix I suppose ^^)

Wauplin:
I've completed the PR. I still need to check a few things but it should be nearly done now. I took the opportunity to -once again :see_no_evil:- refactor how the providers are defined. We now use a base class that is inherited in all providers. Steps to `prepare_request` are split in atomic methods that can be tested individually (to prepare api_token, mapped_model, base_url, full url, payload, body, etc.). Goal is to reuse as much things as possible and therefore making the introduction of a new provider effortless. The case of LLMs is particularly interesting to optimize (see `sambanova.py` implementation). 

I've also refactored the providers' tests to test only the parts that are overwritten in subclasses. The problem with current tests were that a lot was duplicated and it required the provider to be live on production (to get mapping). Since implementation is now centralized, we can also centralize the tests (still needs to be done). And testing atomic methods prevent from having to test end-to-end with production mapping.

src/huggingface_hub/inference/_providers/hf_inference.py
```diff
@@ -1,131 +1,66 @@
...
+            provider="hf-inference",
+            base_url=constants.INFERENCE_PROXY_TEMPLATE.format(provider="hf-inference"),
+            task=task,
         )
 
-    def map_model(self, model: Optional[str]) -> str:
-        return model if model is not None else get_recommended_model(self.task)
+    def _prepare_api_key(self, api_key: Optional[str]) -> str:
+        # special case: for HF Inference we allow not providing an API key
+        return get_token()  # type: ignore[return-value]
```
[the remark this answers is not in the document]
hanouticelina: just noticed that `test_prepare_request` seems to be failing because of that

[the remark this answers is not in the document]
Wauplin: Definitely an oversight from me. Fixed it in https://github.com/huggingface/huggingface_hub/pull/2836/commits/d041ceb3772acb97f7a8910c6378a5ad49b7bdfa

Wauplin (approved):
Current state looks good to me :) @hanouticelina can you have a last look at it?

hanouticelina:
All good on my side! some tests are failing but seem unrelated but i'm not sure if they are flaky or not, first time seeing connection errors from the model download cdn https://cdn-lfs-dev.hf.co

PR #2863, opened 2025-02-14
Add Hyperbolic provider

hanouticelina:
Added a _base_ provider helper class for `text-generation` and `conversational` tasks, this will make it easier to add LLMs-only providers in the future (suggested by @Wauplin in private).


Wauplin (approved):
All good! Thanks for iterating :)

Issue #2767, opened 2025-01-21
Use `endpoint.wait()` when updating an endpoint


datavistics:
**Is your feature request related to a problem? Please describe.**
Id like to be able to use `endpoint.wait()` when updating an endpoint.

**Describe the solution you'd like**
Id like to be able to use `endpoint.wait()` when updating an endpoint.

**Describe alternatives you've considered**
You could delete and re-create the endpoint. But its bad for everyone. 

**Additional context**
None

hanouticelina:
closed as completed by #2867

PR #2867, opened 2025-02-15
bug fix in inference_endpoint wait function for proper waiting on update

Ajinkya-25:
#2767  
### Problem Statement:
Previously, the wait() function in the InferenceEndpoint class waited for the endpoint to become available but did not properly handle the case where the endpoint was updating. This could lead to premature exits before the update was fully completed.
### Changes Implemented:
Added a check for InferenceEndpointStatus.UPDATING and InferenceEndpointStatus.UPDATE_FAILED

Before considering the endpoint as ready, the function now ensures that the endpoint is not in the updating state.
This prevents returning an endpoint that is still in the process of updating.

### Potential Issues Fixed:
Previously, wait() could return immediately after an update request, even if the endpoint was still being updated.
The function is now more robust and ensures that users do not interact with an endpoint that is still updating.


src/huggingface_hub/_inference_endpoints.py
```diff
@@ -210,7 +210,10 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
             if self.url is not None:
                 # Means the URL is provisioned => check if the endpoint is reachable
                 response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
-                if response.status_code == 200:
+                if response.status_code == 200 and self.status not in [
+                    InferenceEndpointStatus.UPDATING,
+                    InferenceEndpointStatus.UPDATE_FAILED,
+                ]:
```
hanouticelina, line 216: this could allow some intermediate states (like INITIALIZING) to pass through if they return a 200 status code (not sure if it's the case)

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -210,7 +210,10 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
             if self.url is not None:
                 # Means the URL is provisioned => check if the endpoint is reachable
                 response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
```
hanouticelina, line 212: Let's do something like that instead that requires a RUNNING state, which implicitly means not updating/initializing/failed.
```suggestion
            if self.status == InferenceEndpointStatus.FAILED:
                raise InferenceEndpointError(
                    f"Inference Endpoint {self.name} failed to deploy. Please check the logs for more information."
                )
            if self.status == InferenceEndpointStatus.UPDATE_FAILED:
                raise InferenceEndpointError(
                    f"Inference Endpoint {self.name} failed to update. Please check the logs for more information."
                )
            if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
                # Verify the endpoint is actually reachable
                response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
                if response.status_code == 200:
                    logger.info("Inference Endpoint is ready to be used.")
                    return self
```

hanouticelina (commented):
Hi @Ajinkya-25, thanks a lot for your contribution, I left comments to make the approach more robust :)
also, could you add a test in [huggingface_hub/tests/test_inference_endpoints.py](https://github.com/huggingface/huggingface_hub/blob/main/tests/test_inference_endpoints.py) similar to [this one](https://github.com/huggingface/huggingface_hub/blob/f629886b76e6134154270431a67397aef11408e9/tests/test_inference_endpoints.py#L191) please?

src/huggingface_hub/_inference_endpoints.py
Ajinkya-25: Thanks For Reviewing And Suggesting The Changes. I have implemented the suggested changes as they provide a more robust solution to the bug.

tests/test_inference_endpoints.py
```diff
@@ -109,6 +109,38 @@
...
+        "image": {"huggingface": {}},
+        "secret": {"token": "my-token"},
+    },
+    "status": {
+        "createdAt": "2023-10-26T12:41:53.263078506Z",
+        "createdBy": {"id": "6273f303f6d63a28483fde12", "name": "Wauplin"},
+        "updatedAt": "2023-10-26T12:41:53.263079138Z",
+        "updatedBy": {"id": "6273f303f6d63a28483fde12", "name": "Wauplin"},
+        "private": None,
+        "state": "updating",
```
Wauplin, line 138: AFAIU what we want to test is that we are indeed waiting for the endpoint to update before returning, even if the endpoint has a `url` ready in its payload (since an endpoint that is updating is still accessible with previous version). So adding a line like this should do the trick:

```suggestion
        "state": "updating",
        "url": "https://vksrvs8pc1xnifhq.us-east-1.aws.endpoints.huggingface.cloud",
```

tests/test_inference_endpoints.py
```diff
@@ -244,6 +276,24 @@ def test_wait_failed(mock_get: Mock):
...
+        InferenceEndpoint.from_raw(MOCK_UPDATE, namespace="foo"),
+        InferenceEndpoint.from_raw(MOCK_UPDATE, namespace="foo"),
+        InferenceEndpoint.from_raw(MOCK_UPDATE, namespace="foo"),
+        InferenceEndpoint.from_raw(MOCK_UPDATE, namespace="foo"),
+    ]
+    with pytest.raises(InferenceEndpointTimeoutError):
+        endpoint.wait(timeout=0.1, refresh_every=0.05)
+
+    assert endpoint.status == "updating"
+    assert len(mock_get.call_args_list) >= 2
```
Wauplin, line 294: This test is not wrong but I would rather test that if we have "update" ,"update, "update, "running" then the endpoint will indeed wait until the endpoint is running before returning. Prior to this PR, the `.wait` call would have returned immediately because `MOCK_UPDATE` has a `url` field. 

```suggestion
        InferenceEndpoint.from_raw(MOCK_RUNNING, namespace="foo"),
    ]
    endpoint.wait(refresh_every=0.05)
    assert endpoint.status == "running"
```

Wauplin (commented):
Good and clean PR, thanks @Ajinkya-25 ! I've left two comments in the test section as I think we should focus the test on _"if an endpoint is updating, we should not return until it is "running" again, even if an URL is already provisioned for it"_.

tests/test_inference_endpoints.py
Ajinkya-25: Thanks For Suggestions. I Think There Are Not Sufficient Values For Effect As It Gives callback Error Every Time.
Screenshot 2025-02-19 165715
Also We Need To Fake Status Code as 200.

tests/test_inference_endpoints.py
```diff
@@ -244,6 +278,27 @@ def test_wait_failed(mock_get: Mock):
...
+@patch("huggingface_hub._inference_endpoints.get_session")
+def test_wait_update(mock_get_session, mock_get_inference_endpoint):
+    """Test that wait() returns when the endpoint transitions to running."""
+    endpoint = InferenceEndpoint.from_raw(MOCK_INITIALIZING, namespace="foo")
+    # Create an iterator that yields three MOCK_UPDATE responses,and then infinitely yields MOCK_RUNNING responses.
+    responses = chain(
+        [InferenceEndpoint.from_raw(MOCK_UPDATE, namespace="foo")] * 3,
+        repeat(InferenceEndpoint.from_raw(MOCK_RUNNING, namespace="foo")),
+    )
+    mock_get_inference_endpoint.side_effect = lambda *args, **kwargs: next(responses)
```
Wauplin, line 291: Though this is perfectly valid, I prefer the way you did it previously with an explicit list (it makes it more readable IMO). Is the problem that you need to repeat the last side effect (RUNNING) an unknown number of times?

Ajinkya-25: I first used a list with three MOCK_UPDATE responses, but it caused a StopIteration error. After some research, I realized it was because the iterator ran out of values. Switching to chain() and repeat() fixed it by making sure the test keeps returning MOCK_RUNNING after the updates, preventing the error. Still learning testing, but this worked well! 🚀. Also this test case needs to set status_code to 200 for testing and as i am writing test case first time it was hard to set it as 200                           can some function be added to testing files to set all this parameters(maybe a new feature). Thanks for the support.

Wauplin (approved):
LGTM but let's wait for @hanouticelina 's final review before merging :)

hanouticelina (approved):
Looks good! thanks a lot @Ajinkya-25 for your contribution, let's merge 🤗

PR #2879, opened 2025-02-19
Fix revision bug in _upload_large_folder.py

yuantuo666:
When using following code to upload folders:

```python

api = HfApi()
api.upload_large_folder(
    repo_id="amphion/Emilia-Dataset",
    repo_type="dataset",
    folder_path="/mnt/yodas",
    revision="refs/pr/10"
)
```

We got errors like "Failed to get upload mode: 404 Client Error: Not Found for url: https://huggingface.co/api/datasets/amphion/Emilia-Dataset/preupload/refs/pr/10 (Request ID: xxx)"

But the revision is already created manually and accessible via https://huggingface.co/datasets/amphion/Emilia-Dataset/tree/refs%2Fpr%2F10.

Here we note that the revision is URL-encoded in the link above (`refs%2Fpr%2F10`), but not encoded in the error message (`refs/pr/10`). By adding a `urllib.parse.quote` function solve this problem.

Besides, I've check other function calls in the same function, which handles this correctly. Probably, the person writing this code only tested on revision that did not contain "/" symbol.

### DIFF
--- a/src/huggingface_hub/_upload_large_folder.py
+++ b/src/huggingface_hub/_upload_large_folder.py
@@ -25,6 +25,7 @@ from datetime import datetime
 from pathlib import Path
 from threading import Lock
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from urllib.parse import quote
 
 from . import constants
 from ._commit_api import CommitOperationAdd, UploadInfo, _fetch_upload_modes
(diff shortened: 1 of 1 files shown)


yuantuo666:
Hi @Wauplin, 
Please kindly review this on this PR, thanks!

yuantuo666:
Thanks! It seems this PR is ready to be merged, please let me know if anything else need to be updated.

hanouticelina (approved):
thank you @yuantuo666 for fixing this! we will release a patch that includes this fix shortly!

hanouticelina:
@yuantuo666 we've just released a [patch](https://github.com/huggingface/huggingface_hub/releases/tag/v0.29.1) with this fix, you just have to upgrade the `huggingface_hub` version:
```bash
pip install huggingface_hub==0.29.1
```

PR #2896, opened 2025-03-02
Fix: Restore sys.stdout in notebook_login after error

LEEMINJOO:
## Issue
In `notebook_login`, when using `capture_output`, `sys.stdout` is redirected to `StringIO`, 
but if an error occurs, it is not restored, causing issues in `ipykernel`.

## Fix
- Ensure `sys.stdout` is restored using a `finally` block.


Wauplin (approved):
Nice finding, thanks for opening a PR for it!

PR #2902, opened 2025-03-04
Remove Cohere from tests and fix quality

hanouticelina:
Since there is no model mapping registered in https://huggingface.co/api/partners/cohere/models, let's not test Cohere for now and wait until we have the model mapping. wdyt @Wauplin ?


Wauplin (approved):
Sorry, I really thought CI was green :grimacing: :thinking:

PR #2912, opened 2025-03-06
Handle file-like and IO objects in `_curlify()`

hanouticelina:
This PR fixes a bug I ran into with `_curlify`: when the body request contains a `BytesIO` object, `_curlify()` attempts to get the length of this object, and obviously it fails because it doesn't have a __len__ method and we don't handle the case when the body contains file-like objects.


Wauplin (approved):
Makes sense!

PR #2937, opened 2025-03-18
[Inference Providers] update polling interval for fal-ai

hanouticelina:
Related to this discussion : https://github.com/huggingface/huggingface.js/pull/1292#discussion_r2000759123

Let's use the same interval polling for our clients. 0.5s should be smaller enough to fetch updates frequently enough and larger enough to avoid querying the API too often (hopefully).


Wauplin (approved):
Thanks!

PR #2939, opened 2025-03-19
Allow users with (at least) write permissions to run the style bot workflow

hanouticelina:
users with `admin` > `maintain` > `push` permissions are allowed to run the style bot.


sayakpaul (approved):
Fantastic!

Wauplin (approved):
Conceptual approval (as long as #security team is fine with it) :)

PR #2944, opened 2025-03-20
[DON'T MERGE] test doc builder

gante:
https://github.com/huggingface/doc-builder/pull/547/

We can see the runner being used:
Screenshot 2025-03-20 at 16 11 22


gante:
(closing: transformers found a solution that does not imply using groups, i.e. base GHA images are used)

PR #2949, opened 2025-03-21
[Inference Providers] fold OpenAI support into `provider` parameter

hanouticelina:
Quick PR to support OpenAI through the `provider` parameters, similar to what is done in the JS client: https://github.com/huggingface/huggingface.js/pull/1205 + fixing the snippets in the `new_provider.md` doc.


Wauplin (approved):
nice!

PR #2957, opened 2025-03-26
[Internal] Fix check parameters script

hanouticelina:
This PR addresses the issue highlighted by @Wauplin in #2956 where core parameters defined in the task dataclass were incorrectly flagged as "missing" in method docstrings. The problem occurred because we weren't filtering out core parameters from the task dataclass params. This fix ensures core parameters are properly filtered from the dataclass parameters as well (copy pasted from #2956) and fixes an unrelated issue when a parameter is only missing in the docstring and not in the method's signature.

There is some possible improvement that can be done to the script but this is not necessary (everything should be working fine now 🤞) and not prio for now, I will come back to this when I have some spare time!


Wauplin (approved):
Thanks for properly fixing this!

PR #2982, opened 2025-04-07
[Inference Providers] Fix `text-generation` when using an external provider

hanouticelina:
This PR fixes `text_generation` when used with an external provider. The response was not handled in the corresponding subclasses which made the task unusable with these providers (spotted this while debugging the HF integration in langchain). 
Also, note that the output schema of text generation is based on TGI [/generation](https://huggingface.github.io/text-generation-inference/#/Text%20Generation%20Inference/generate) and it's not incompatible with the OpenAI completion API schema which is followed by most of the providers. 
Changing the output schema of `text_generation` to make it OpenAI-compatible is a breaking change and I believe it's not worth it now, especially that the usage is low (?) for this task.


Wauplin (approved):
Thanks!

PR #2986, opened 2025-04-08
PoC: `InferenceClient` is also a `MCPClient`

julien-c:
## Required reading

https://modelcontextprotocol.io/quickstart/client

TL;DR: MCP is a standard API to expose sets of Tools that can be hooked to LLMs

## Summary of how to use this

```py
client = MCPClient(
    provider="together",
    model="Qwen/Qwen2.5-72B-Instruct",
)
await client.add_mcp_server(
    "node", ["--disable-warning=ExperimentalWarning", f"{os.path.expanduser('~')}/Desktop/hf-mcp/index.ts"]
    # in the future we'll (very likely) be able to just pass a remote URL,
    # but for now, we need the MCP server to run locally
)

response = await client.process_query(
    """
    find an app that generates 3D models from text,
    and also get the best paper about transformers
    """
)

```

## Open questions

- Should we implement this in `(Async)InferenceClient` directly? Or on a distinct class, like here?

## Where to find the MCP Server used here as an example

Note that you can replace it with any MCP Server, from this doc for instance: https://modelcontextprotocol.io/examples

https://gist.github.com/julien-c/0500ba922e1b38f2dc30447fb81f7dc6

## Script output

<details>
<summary>Generation from LLM with tools</summary>

### 3D Model Generation from Text

Here are some of the best apps that can generate 3D models from text:

1. **Shap-E**
   - **Link**: [huggingface.co/spaces/hysts/Shap-E](https://huggingface.co/spaces/hysts/Shap-E)
   - **Description**: This app supports both text-to-3D and image-to-3D generation.

2. **Hunyuan3D-1**
   - **Link**: [huggingface.co/spaces/tencent/Hunyuan3D-1](https://huggingface.co/spaces/tencent/Hunyuan3D-1)
   - **Description**: Another robust tool for generating 3D models from text and images.

3. **LGM**
   - **Link**: [huggingface.co/spaces/ashawkey/LGM](https://huggingface.co/spaces/ashawkey/LGM)
   - **Description**: Generates 3D content from images or text.

4. **3D-Adapter**
   - **Link**: [huggingface.co/spaces/Lakonik/3D-Adapter](https://huggingface.co/spaces/Lakonik/3D-Adapter)
   - **Description**: Generates 3D models from text descriptions.

5. **Fictiverse-Voxel_XL_Lora**
   - **Link**: [huggingface.co/spaces/mygyasir/Fictiverse-Voxel_XL_Lora](https://huggingface.co/spaces/mygyasir/Fictiverse-Voxel_XL_Lora)
   - **Description**: Generates 3D models from text prompts.

### Best Paper on Transformers

One of the most influential and highly cited papers on transformers is:

- **Title**: "Attention Is All You Need"
- **Authors**: Ashish Vaswani, Noam Shazeer, Niki Parmar, Jakob Uszkoreit, Llion Jones, Aidan N. Gomez, Łukasz Kaiser, Illia Polosukhin
- **Link**: [huggingface.co/papers/1706.03762](https://huggingface.co/papers/1706.03762)
- **Description**: This paper introduced the Transformer architecture, which has become a cornerstone in natural language processing and many other areas of deep learning.

If you are looking for more recent advancements, here are a few other notable papers:

- **Title**: "RoFormer: Enhanced Transformer with Rotary Position Embedding"
- **Link**: [huggingface.co/papers/2104.09864](https://huggingface.co/papers/2104.09864)
- **Description**: This paper introduces RoFormer, which improves the Transformer by using rotary position embeddings.

- **Title**: "Performer: Generalized Attention with Gaussian Kernels for Sequence Modeling"
- **Link**: [huggingface.co/papers/2009.14794](https://huggingface.co/papers/2009.14794)
- **Description**: This paper presents Performer, a more efficient and scalable version of the Transformer.

- **Title**: "Longformer: The Long-Document Transformer"
- **Link**: [huggingface.co/papers/2004.05150](https://huggingface.co/papers/2004.05150)
- **Description**: This paper introduces Longformer, which extends the Transformer to handle very long documents.

These resources should provide you with a solid foundation for both generating 3D models from text and understanding the latest advancements in transformer models.
</details>

### DIFF
--- a/mcp_client.py
+++ b/mcp_client.py
@@ -1,0 +1,125 @@
+import asyncio
+import json
+import os
+from contextlib import AsyncExitStack
+from typing import List, Optional
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+
+from huggingface_hub import AsyncInferenceClient, ChatCompletionInputTool, ChatCompletionOutput
+from huggingface_hub.inference._providers import PROVIDER_T
+
+
+class MCPClient:
+    def __init__(
+        self,
+        *,
+        provider: PROVIDER_T,
+        model: str,
+        api_key: Optional[str] = None,
+    ):
+        self.client = AsyncInferenceClient(
+            provider=provider,
+            api_key=api_key,
+        )
+        self.model = model
+        # Initialize MCP session and client objects
+        self.session: Optional[ClientSession] = None
+        self.exit_stack = AsyncExitStack()
+        self.available_tools: List[ChatCompletionInputTool] = []
+
+    async def add_mcp_server(self, command: str, args: List[str]):
+        """Connect to an MCP server
+
+        Args:
+            todo
+        """
+        server_params = StdioServerParameters(command=command, args=args, env={"HF_TOKEN": os.environ["HF_TOKEN"]})
+
+        stdio_transport = await self.exit_stack.enter_async_context(stdio_client(server_params))
+        self.stdio, self.write = stdio_transport
+        self.session = await self.exit_stack.enter_async_context(ClientSession(self.stdio, self.write))
+
+        await self.session.initialize()
+
+        # List available tools
+        response = await self.session.list_tools()
+        tools = response.tools
+        print("\nConnected to server with tools:", [tool.name for tool in tools])
+        self.available_tools += [
+            {
+                "type": "function",
+                "function": {
+                    "name": tool.name,
+                    "description": tool.description,
+                    "parameters": tool.inputSchema,
+                },
+            }
+            for tool in tools
+        ]
+
+    async def process_query(self, query: str) -> ChatCompletionOutput:
+        """Process a query using Claude and available tools"""
+        messages = [{"role": "user", "content": query}]
+
+        response = await self.client.chat.completions.create(
+            model=self.model,
+            messages=messages,
+            tools=self.available_tools,
+            tool_choice="auto",
+        )
+
+        # Process response and handle tool calls
+        tool_calls = response.choices[0].message.tool_calls
+        if tool_calls:
+            for tool_call in tool_calls:
+                function_name = tool_call.function.name
+                function_args = json.loads(tool_call.function.arguments)
+
+                # Execute tool call
+                result = await self.session.call_tool(function_name, function_args)
+                messages.append(
+                    {
+                        "tool_call_id": tool_call.id,
+                        "role": "tool",
+                        "name": function_name,
+                        "content": result.content[0].text,
+                    }
+                )
+
+        function_enriched_response = await self.client.chat.completions.create(
+            model=self.model,
+            messages=messages,
+        )
+
+        return function_enriched_response
+
+    async def cleanup(self):
+        """Clean up resources"""
+        await self.exit_stack.aclose()
+
+
+async def main():
+    client = MCPClient(
+        provider="together",
+        model="Qwen/Qwen2.5-72B-Instruct",
+        api_key=os.environ["HF_TOKEN"],
+    )
+    try:
+        await client.add_mcp_server(
+            "node", ["--disable-warning=ExperimentalWarning", f"{os.path.expanduser('~')}/Desktop/hf-mcp/index.ts"]
+        )
+        response = await client.process_query(
+            """
+            find an app that generates 3D models from text,
+            and also get the best paper about transformers
+            """
+        )
+        print("\n" + response.choices[0].message.content)
+    finally:
+        await client.cleanup()
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,11 @@ extras["tensorflow-testing"] = [
 
 extras["hf_xet"] = ["hf_xet>=0.1.4"]
 
+extras["mcp"] = [
+    "mcp>=1.6.0",
+    "aiohttp",  # for AsyncInferenceClient
+]
+
 extras["testing"] = (
     extras["cli"]
     + extras["inference"]


mcp_client.py
```diff
@@ -0,0 +1,125 @@
...
+            provider=provider,
+            api_key=api_key,
+        )
+        self.model = model
+        # Initialize MCP session and client objects
+        self.session: Optional[ClientSession] = None
+        self.exit_stack = AsyncExitStack()
+        self.available_tools: List[ChatCompletionInputTool] = []
+
+    async def add_mcp_server(self, command: str, args: List[str]):
```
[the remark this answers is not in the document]
julien-c: yes we can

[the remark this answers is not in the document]
julien-c: i mean we would need to store a map of `sessions`, but there's nothing preventing us from doing it, conceptually

mcp_client.py
```diff
@@ -0,0 +1,125 @@
...
+                    "name": tool.name,
+                    "description": tool.description,
+                    "parameters": tool.inputSchema,
+                },
+            }
+            for tool in tools
+        ]
+
+    async def process_query(self, query: str) -> ChatCompletionOutput:
+        """Process a query using Claude and available tools"""
```
mishig25, line 63:
```suggestion
        """Process a query using `self.model` and available tools"""
```
?

mcp_client.py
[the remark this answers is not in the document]
mishig25: let me know if the following question is out of scope.
Question: how would I connect to multiple MCP servers? Would it look like option 1 or option 2?

### Option 1
```
client1 = MCPClient()
client2 = MCPClient()
```

or 

### Option 2
```
client = MCPClient()
client.add_mcp_server(server1)
client.add_mcp_server(server2)
```

?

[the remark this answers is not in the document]
mishig25: another design question: `class MCPClient(AsyncInferenceClient)` vs `class AsyncInferenceClient(...args, mcp_clients: MCPClient[])`

mcp_client.py
julien-c: haha yes

mcp_client.py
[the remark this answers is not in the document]
julien-c: Option 2!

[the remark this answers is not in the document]
mishig25: okay, in that case, naming `add_mcp_server` make sense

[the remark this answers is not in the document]
julien-c: trying to one-shot the multiple servers support with Sonnet 3.7. Prompt:

```
change the "session" property to become "sessions", a dict of `ClientSession` indexed by tool name (which is a string)
```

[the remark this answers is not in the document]
julien-c: #2987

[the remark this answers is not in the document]
mishig25: if we go with `class MCPClient(AsyncInferenceClient)`, which is the current approach, why not just add methods `add_mcp_server` & `process_query` directly to `AsyncInferenceClient` AND not create `class MCPClient` in the first place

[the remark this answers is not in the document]
Wauplin: After sleeping on it, I do think that it's preferable to have `InferenceClient`/`AsyncInferenceClient` as the base classes to make inference and `MCPClient` a separate wrapper around them. My reasoning is that:
- it distinguishes clearly the role of each client (one is for raw request, the other for a specific protocol on top of it). Also makes the naming explicit + we can document it separately in a `MCPClient` doc page
- MCP protocol might evolve in the future to a 2.0 version. It'd be easier to implement a `MCPClient2` if that's the case
- I expect `MCPClient` to take more and more parameters in the future. For instance for secret management. Or for http-based communication instead of the current local stdio-transport thingy. It will not be a problem to extend a tiny client like `MCPClient` but would make `AsyncInferenceClient` even more bloated that currently

Happy to hear your thoughts on that otherwise :)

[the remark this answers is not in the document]
julien-c: yes i agree, will revert 2c7329c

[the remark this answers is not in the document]
Pierrci: Sorry to chime in unannounced, but from a very removed external user standpoint, I find this all very confusing - I just don't think what you coded should be called `MCPClient` 😄

When I came to this PR I was fully expecting `MCPClient` to be passed as parameter to `InferenceClient`, though I hear @Wauplin above, so why not a wrapper. But the end result is really more of an `InferenceClientWithEmbeddedMCP` to me, not an `MCPClient`.

That being said, it's just about semantics, but I'm kind of a semantics extremist, sorry about that (and feel free to completely disregard this message, as is very likely XD)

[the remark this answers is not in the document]
julien-c: > I was fully expecting `MCPClient` to be passed as parameter to `InferenceClient`

What do you mean as parameter? Do you have an example signature?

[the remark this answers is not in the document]
Pierrci: second option of https://github.com/huggingface/huggingface_hub/pull/2986#discussion_r2033308610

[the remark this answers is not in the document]
julien-c: ah yes, sure, we can probably add this I guess

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+            provider=provider,
+            api_key=api_key,
+        )
+        self.model = model
+        # Initialize MCP sessions as a dictionary of ClientSession objects
+        self.sessions: Dict[ToolName, ClientSession] = {}
+        self.exit_stack = AsyncExitStack()
+        self.available_tools: List[ChatCompletionInputTool] = []
+
+    async def add_mcp_server(self, command: str, args: List[str], env: Dict[str, str]):
```
grll, line 36: you would need to lighten a bit the requirements on your args if you want to make it work with SSE or the intent is just to support STDIO ? I see the rest seems to focus on stdio so maybe it's by design

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+        for tool in tools:
+            self.sessions[tool.name] = session
+
+        self.available_tools += [
+            {
+                "type": "function",
+                "function": {
+                    "name": tool.name,
+                    "description": tool.description,
+                    "parameters": tool.inputSchema,
```
grll, line 65: just a note that I have seen some MCP servers with jsonref in their description which sometimes confuses the model. In mcpadapt I had to resolve the jsonref before passing it to the model, might be minor for now

grll: confused or sometime plain unsupported by the model sdk like google genai...

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+
+from huggingface_hub import AsyncInferenceClient, ChatCompletionInputTool, ChatCompletionOutput
+from huggingface_hub.inference._providers import PROVIDER_T
+
+
+# Type alias for tool names
+ToolName: TypeAlias = str
+
+
+class MCPClient:
```
grll, line 18: could potentially also support 2 syntaxes like in the MCPClient for smolagents:
1. like you have with try and finally
2. one as a context manager, where you directly pass the the MCP client settings + the MCP servers with the tools you want to have this would allow to do something like "with MCPClient(...) as client:" and then just run client.chat.completion with the tools

grll (commented):
few comments out of my head, SSE supports would not be so hard to add and could really  be a nice addition

mcp_client.py
julien-c: for now, just Stdio, but in the future `Streaming HTTP` from what i've understood should be the way to go?

mcp_client.py
julien-c: Does the spec mention anything about whether jsonref is allowed or not?

julien-c:
Thanks for the review @grll!

mcp_client.py
grll: yes this is the new spec, but it is backward compatible and at the level you are working with in this PR I wouldnt expect much to change, probably the internals of the client will change but the client interface would remain the same. Which means if today you do something like add_mcp_server(StdioParameters | dict) dict being the arguments of the sse_client from the python sdk you could already support all the SSE servers + potentially future Streaming HTTP server with minor adjustments at most

julien-c:
@grll indeed, thanks. I think i'll merge this PR almost as-is, but we'd welcome further PRs to add SSE/HTTP and other stuff!

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+        await session.initialize()
+
+        # List available tools
+        response = await session.list_tools()
+        tools = response.tools
+        print("\nConnected to server with tools:", [tool.name for tool in tools])
+
+        # Map tool names to their server for later lookup
+        for tool in tools:
+            self.sessions[tool.name] = session
```
grll, line 57: you might have some funny situation happening here if multiple mcp servers have similarly named tools, almost missed that one!

grll: (nothing prevent two mcp servers to have conflicting tool name at the moment)

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+            tools=self.available_tools,
+            tool_choice="auto",
+        )
+
+        # Process response and handle tool calls
+        tool_calls = response.choices[0].message.tool_calls
+        if tool_calls is None or len(tool_calls) == 0:
+            return response
+
+        for tool_call in tool_calls:
```
Wauplin, line 87: Since we are in an async context, we should run the tools in coroutines in parallel 

(can be done in later PR)

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+        # List available tools
+        response = await session.list_tools()
+        tools = response.tools
+        print("\nConnected to server with tools:", [tool.name for tool in tools])
+
+        # Map tool names to their server for later lookup
+        for tool in tools:
+            self.sessions[tool.name] = session
+
+        self.available_tools += [
```
Wauplin, line 59: I would make `available_tools` a mapping of Tools instead like this:

```py
# in __init__
    (...)
    self.sessions: Dict[ToolName, ClientSession] = {}
    self.available_tools: Dict[ToolName, Tool] = {}
```

then to add tools:

```py
# in add_mcp_server
    (...)
    self.available_tools[tool.name] =  tool
```

and finally format them to `List[ChatCompletionInputTool]` only when you use them:

```py
# in `process_query`:
    tools= [
        {
            "type": "function",
            "function": {
                "name": tool.name,
                "description": tool.description,
                "parameters": tool.inputSchema,
            },
        }
        for tool in self.available_tools.items()
    ]
```

Wauplin: It feels more idiomatic to me to format only when needed. Also prevent having multiple times the same tool in the list sent to the LLM.

mcp_client.py
Wauplin: As a first solution, I would just raise a warning to the user to let them know we are overwriting a tool from another MCP server. We can always do more fancy work in a follow-up PR.

```suggestion
            if tool.name in self.sessions:
                warning.warns(f"Overwriting existing tool '{tool.name}'. This can lead to unexpected behavior. It is recommended to make sure MCP servers have different tool names.")
            self.sessions[tool.name] = session
```

 Having several tools with same names would be a problem anyway since we can't distinguish which tool the LLM chose.

Wauplin: in a follow-up PR we can define a namespace per mcp client to avoid conflicts (low prio)

julien-c: yes IMO low-prio, i haven't seen any issue in JS yet

mcp_client.py
```diff
@@ -0,0 +1,152 @@
...
+# Type alias for tool names
+ToolName: TypeAlias = str
+
+
+class MCPClient:
+    def __init__(
+        self,
+        *,
+        provider: PROVIDER_T,
+        model: str,
```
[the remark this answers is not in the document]
julien-c: ok

setup.py
```diff
@@ -57,6 +57,11 @@ def get_version() -> str:
 
 extras["hf_xet"] = ["hf_xet>=0.1.4"]
 
+extras["mcp"] = [
+    "mcp>=1.6.0",
+    "aiohttp",  # for AsyncInferenceClient
+]
```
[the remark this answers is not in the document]
julien-c: makes sense

mcp_client.py
Wauplin: I ended-up doing a 

```py
        for tool in response.tools:
            if tool.name in self.sessions:
                logger.warning(f"Tool '{tool.name}' already defined by another server. Skipping.")
                continue

            # Map tool names to their server for later lookup
            self.sessions[tool.name] = session
```

i.e. => warn and skip tool if duplicate naming (and post-pone proper solution with namespaces)

julien-c: sgtm

mcp_client.py
Wauplin: In the end I kept the syntax from JS client

Wauplin:
I just pushed changes to clean this PR. See PR description for list of changes https://github.com/huggingface/huggingface_hub/pull/2986#issue-2979490302. It should now be very close to JS SDK (no SSE / Streaming server support though it can be easily added later).

**Example:**

```py
import os

from huggingface_hub import ChatCompletionInputMessage, ChatCompletionStreamOutput, MCPClient

async def main():
    async with MCPClient(
        provider="nebius",
        model="Qwen/Qwen2.5-72B-Instruct",
    ) as client:
        await client.add_mcp_server(
            command="python",
            args=["hf_mcp.py"],  # https://gist.github.com/Wauplin/b6e5f1a39db843eedfa00738e4998589
            env={"HF_TOKEN": os.environ["HF_TOKEN"]},
        )

        messages = [
            {
                "role": "user",
                "content": "what are the most recently updated models from nvidia? Explain your reasoning.",
            }
        ]

        async for chunk in client.process_single_turn_with_tools(messages):
            # Log messages
            if isinstance(chunk, ChatCompletionStreamOutput):
                delta = chunk.choices[0].delta
                if delta.content:
                    print(delta.content, end="")

            # Or tool calls
            elif isinstance(chunk, ChatCompletionInputMessage):
                print(
                    f"\nCalled tool '{chunk.name}'. Result: '{chunk.content if len(chunk.content) < 100 else chunk.content[:100] + '...'}'"
                )


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

**Producing:**

```
/home/wauplin/projects/huggingface_hub/src/huggingface_hub/utils/_experimental.py:60: UserWarning: 'MCPClient.__init__' is experimental and might be subject to breaking changes in the future without prior notice. You can disable this warning by setting `HF_HUB_DISABLE_EXPERIMENTAL_WARNING=1` as environment variable.
  warnings.warn(
[05/13/25 20:15:17] INFO     Processing request of type ListToolsRequest                                                                            server.py:545
To find the most recently updated models from NVIDIA, we can use the `list_models` function from the Hugging Face Hub API. Specifically, we will filter by the author (`nvidia`) and sort the results by the `last_modified` date in descending order. This will give us the models that were most recently updated.

Here is the reasoning:
1. **Author Filter**: We filter by the author `nvidia` to ensure we only get models from NVIDIA.
2. **Sorting by Last Modified**: We sort the models by the `last_modified` date in descending order to get the most recently updated models at the top of the list.

Let's make the function call to get the most recently updated models from NVIDIA:

[05/13/25 20:15:22] INFO     Processing request of type CallToolRequest                                                                             server.py:545

Called tool 'list_models'. Result: '[ModelInfo(id='nvidia/Llama-3.3-70B-Instruct-FP8', author=None, sha=None, created_at=datetime.dateti...'
```

This should now be ready for review :)

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+        self.sessions: Dict[ToolName, "ClientSession"] = {}
+        self.exit_stack = AsyncExitStack()
+        self.available_tools: List[ChatCompletionInputTool] = []
+
+        # Initialize the AsyncInferenceClient
+        self.client = AsyncInferenceClient(model=model, provider=provider, api_key=api_key)
+
+    async def __aenter__(self):
+        """Enter the context manager"""
+        await self.client.__aenter__()
```
hanouticelina, line 54: just for the sake of ressource leakage, maybe we should enter `self.exist_stack` either here or we can do:
```python
async with elf.exit_stack:
      # call self.exit_stack.enter_async_context(..)
```
```suggestion
        await self.client.__aenter__()
        await self.exit_stack.__aenter__()
```

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+    ):
+        """Connect to an MCP server
+
+        Args:
+            command (str):
+                The command to run the MCP server.
+            args (List[str], optional):
+                Arguments for the command.
+            env (Dict[str, str], optional):
+                Environment variables for the command. Default to empty environment.
```
hanouticelina, line 78: (nit) based on mcp [docs](https://github.com/modelcontextprotocol/python-sdk/blob/5d33861cad4bc8e0f24f572ba0bac38f2f68b4f4/src/mcp/client/stdio/__init__.py#L70).
```suggestion
                Environment variables for the command. Default is to inherit the parent environment.
```

hanouticelina (commented):
🔥 
(tested the PR, it works as expected!)

src/huggingface_hub/inference/_mcp/utils.py
julien-c, line 1: cc @evalstate

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+        self,
+        *,
+        model: str,
+        provider: Optional[PROVIDER_OR_POLICY_T] = None,
+        api_key: Optional[str] = None,
+    ):
+        warnings.warn(
+            "'MCPClient' is experimental and might be subject to breaking changes in the future without prior notice.",
+            UserWarning,
+        )
```
julien-c, line 42: personally think we could omit this, but no strong opinion

julien-c (commented):
lgtm, feel free to merge and release whenever you want!

julien-c:
Pinging @albertvillanova for a quick review too if you have time!

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+class MCPClient:
+    def __init__(
+        self,
+        *,
+        model: str,
+        provider: Optional[PROVIDER_OR_POLICY_T] = None,
+        api_key: Optional[str] = None,
+    ):
+        warnings.warn(
+            "'MCPClient' is experimental and might be subject to breaking changes in the future without prior notice.",
```
evalstate, line 40: No strong opinion either, but the default warning formatter leaves a dangling `warnings.warn(` :( Move it to a docstring?

albertvillanova:
I can have a look at it this weekend.

src/huggingface_hub/inference/_mcp/mcp_client.py
albertvillanova, line 54: What is the point of entering the client context manager here?

It is only used within the `process_single_turn_with_tools` method, and the context manager is properly entered and exit there using a `with` block?
```suggestion
```

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+        self.client = AsyncInferenceClient(model=model, provider=provider, api_key=api_key)
+
+    async def __aenter__(self):
+        """Enter the context manager"""
+        await self.client.__aenter__()
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        """Exit the context manager"""
+        await self.client.__aexit__(exc_type, exc_val, exc_tb)
```
albertvillanova, line 59: See comment above.
```suggestion
```

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Args:
+            query: The user query to process
+            opts: Optional parameters including:
+                - exit_loop_tools: List of tools that should exit the generator when called
+                - exit_if_first_chunk_no_tool: Exit if no tool is present in the first chunks
+
+        Yields:
+            ChatCompletionStreamOutput chunks or ChatCompletionInputMessage objects
+        """
```
albertvillanova, line 152: This docstring needs to be updated: `self.model` should be replaced by `self.client` and the args should be updated as well.

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+        # Prepare tools list based on options
+        tools = self.available_tools
+        if exit_loop_tools is not None:
+            tools = [*exit_loop_tools, *self.available_tools]
+
+        # Create the streaming request
+        async with self.client:
```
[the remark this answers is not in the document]
albertvillanova: See my comment above: I proposed keeping the `with` block here and removing the `self.client` from the `__aenter__` above.
The reason is that `self.client` stores an HTTP response object each time it make a request: so better cleaning this resource after the request/response has been handled (here, in this part of the code), instead of keeping all responses if multiple calls to `process_single_turn_with_tools` are made.

albertvillanova (approved):
Thanks a lot for this addition! It is great that we finally have a Hugging Face MCP client: this will definitely make experimentation and integration much smoother.

I know this has already been discussed and there's no strong consensus, but I'd still like to share my opinion here for the record: as we've talked about privately, I feel that this is more than a simple MCP client: in my humble view, it already constitutes a single-step agent due to the use of an LLM. A multi-step agent could then be built by looping over this single step. That said, I realize this is mostly a matter of terminology in a space that's evolving really quickly, so it's not a big deal at all, just a remark worth surfacing.

On a more concrete note, I'd love to suggest that we consider supporting use cases that don't involve an LLM. For example, I'm thinking of scenarios where a user wants to build a code agent rather than a tool-calling agent, similar to what we do in `smolagents`, where tools are passed as Python functions and the model generates code that uses them. 

With this in mind, do you think it would make sense to also provide a sync version of the client? That could make it much easier to support tools exposed as synchronous Python functions and to integrate into existing sync workflows.

Curious to hear your thoughts. And thanks again for pushing this forward!

julien-c:
> I feel that this is more than a simple MCP client: in my humble view, it already constitutes a single-step agent due to the use of an LLM

I disagree: the official example of MCP Client does have a LLM: https://modelcontextprotocol.io/quickstart/client

Without a LLM, a bare MCP client would just be a standard JSON-over-RPC call ; IMO MCP is LLM-centric.

> already constitutes a single-step agent

For me Agent === multi-step, otherwise it's just LLM with tool calling.

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -0,0 +1,236 @@
...
+        await session.initialize()
+
+        # List available tools
+        response = await session.list_tools()
+        logger.debug("Connected to server with tools:", [tool.name for tool in response.tools])
+
+        for tool in response.tools:
+            if tool.name in self.sessions:
+                logger.warning(f"Tool '{tool.name}' already defined by another server. Skipping.")
+                continue
```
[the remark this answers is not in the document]
Wauplin: Yes agree :+1: (been discussed above but for now post-poned for later PR as you mentioned)

src/huggingface_hub/inference/_mcp/mcp_client.py
Wauplin: Good catch! I initially intended to delete the `async with self.client:` below but forgot about it. The issue with including `async with self.client:` within `process_single_turn_with_tools` is that concurrent usage of `process_single_turn_with_tools` may result in unexpected behaviors if one instance terminates the sessions for all others. By relocating the logic to `__aenter__`, the responsibility shifts to the end user, who can handle the client lifecycle themselves.

Wauplin: updated in https://github.com/huggingface/huggingface_hub/pull/2986/commits/b273cba153b8febfd01e5ef590459079ef34b26b

src/huggingface_hub/inference/_mcp/mcp_client.py
Wauplin: https://github.com/huggingface/huggingface_hub/pull/2986#discussion_r2098148909

src/huggingface_hub/inference/_mcp/mcp_client.py
Wauplin: moved it the docstring

src/huggingface_hub/inference/_mcp/mcp_client.py
Wauplin: moved it to the docstring in the end

Wauplin:
I addressed all the comments above and fixed the resources lifecycle (user decides when to open/close connections). 

Re: should the MCP client include an LLM, I do think that yes and I would argue that this is what an HF-based MCP client is for (i.e. the goal is to have minimal code to glue MCP servers to Inference Providers). Regarding exposing the MCP tools as python functions for a code agent, that can be explored separately I think (either in a separate class or with this class but with a separate method). I'm still unfamiliar with code agents and what they would expect from MCP Client.


**TL;DR:** are we finally ready to merge this? :smile:

Wauplin (approved):
I like self-approvals.

hanouticelina (approved):
🔥 🔥 🔥

PR #2987, opened 2025-04-08
mcp_client: Support multiple servers

julien-c:
Untested and heavily one-shooted by LLM

cc @mishig25

### DIFF
--- a/mcp_client.py
+++ b/mcp_client.py
@@ -2,13 +2,17 @@ import asyncio
 import json
 import os
 from contextlib import AsyncExitStack
-from typing import List, Optional
+from typing import Dict, List, Optional, TypeAlias
 
 from mcp import ClientSession, StdioServerParameters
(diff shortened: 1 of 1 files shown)


mishig25 (approved):
tested and the code works

lets merge it into mcp branch, and we can handle nits afterwrads

mcp_client.py
```diff
@@ -38,15 +42,20 @@ async def add_mcp_server(self, command: str, args: List[str]):
...
 
         # List available tools
-        response = await self.session.list_tools()
+        response = await session.list_tools()
         tools = response.tools
         print("\nConnected to server with tools:", [tool.name for tool in tools])
+
+        # Map tool names to their server for later lookup
+        for tool in tools:
+            self.sessions[tool.name] = session
```
mishig25, line 57: could be a collision if different servers have tools with same names

PR #2993, opened 2025-04-10
Removing workaround for deprecated code

bpronan:
Hub has been updated to include the schema and domain in the returned refresh route ([private link](https://github.com/huggingface-internal/moon-landing/pull/13072)). 

Removing the workaround to check for the deprecated case.


hanouticelina (approved):
looks good! thanks @bpronan

PR #3005, opened 2025-04-15
[Inference Providers] Support for LoRAs

hanouticelina:
This work mirrors the PR in the js client (see [huggingface.js#1347](https://github.com/huggingface/huggingface.js/pull/1347)).

This PR introduces the support of LoRAs for fal-ai, this should not be merged until the server-side [PR](https://github.com/huggingface-internal/moon-landing/pull/13246) is merged (private).

## Changes
- update `_prepare_mapped_model` to return a `ProviderMappingInfo` object. This object returns the provider id, the hf model id (optional) and the lora weights path (optional) that is expected to be sent in the payload. 
- update `_prepare_payload_as_dict` and `_prepare_payload_as_bytes` signatures to accept a `ProviderMappingInfo` object as an argument. 

the logic is implemented within the `TaskProviderHelper` base class, to avoid duplicating similar logic when we support LoRAs for the other providers. Note that LoRAs can also be used with other tasks like `text-to-video` as instance.

### DIFF
--- a/src/huggingface_hub/inference/_providers/_common.py
+++ b/src/huggingface_hub/inference/_providers/_common.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from functools import lru_cache
 from typing import Any, Dict, Optional, Union
 
@@ -34,6 +35,13 @@ def filter_none(d: Dict[str, Any]) -> Dict[str, Any]:
     return {k: v for k, v in d.items() if v is not None}
 
 
+@dataclass
+class ProviderMappingInfo:
+    provider_id: str
+    adapter_weights_path: Optional[str] = None
+    hf_model_id: Optional[str] = None
+
+
 class TaskProviderHelper:
     """Base class for task-specific provider helpers."""
 
@@ -61,28 +69,30 @@ class TaskProviderHelper:
         api_key = self._prepare_api_key(api_key)
 
         # mapped model from HF model ID
-        mapped_model = self._prepare_mapped_model(model)
+        provider_mapping_info = self._prepare_mapped_model(model)
 
         # default HF headers + user headers (to customize in subclasses)
         headers = self._prepare_headers(headers, api_key)
 
         # routed URL if HF token, or direct URL (to customize in '_prepare_route' in subclasses)
-        url = self._prepare_url(api_key, mapped_model)
+        url = self._prepare_url(api_key, provider_mapping_info.provider_id)
 
         # prepare payload (to customize in subclasses)
-        payload = self._prepare_payload_as_dict(inputs, parameters, mapped_model=mapped_model)
+        payload = self._prepare_payload_as_dict(inputs, parameters, provider_mapping_info=provider_mapping_info)
         if payload is not None:
             payload = recursive_merge(payload, extra_payload or {})
 
         # body data (to customize in subclasses)
-        data = self._prepare_payload_as_bytes(inputs, parameters, mapped_model, extra_payload)
+        data = self._prepare_payload_as_bytes(inputs, parameters, provider_mapping_info, extra_payload)
 
         # check if both payload and data are set and return
         if payload is not None and data is not None:
             raise ValueError("Both payload and data cannot be set in the same request.")
         if payload is None and data is None:
             raise ValueError("Either payload or data must be set in the request.")
-        return RequestParameters(url=url, task=self.task, model=mapped_model, json=payload, data=data, headers=headers)
+        return RequestParameters(
+            url=url, task=self.task, model=provider_mapping_info.provider_id, json=payload, data=data, headers=headers
+        )
 
     def get_response(
         self,
@@ -107,7 +117,7 @@ class TaskProviderHelper:
             )
         return api_key
 
-    def _prepare_mapped_model(self, model: Optional[str]) -> str:
+    def _prepare_mapped_model(self, model: Optional[str]) -> ProviderMappingInfo:
         """Return the mapped model ID to use for the request.
 
         Usually not overwritten in subclasses."""
@@ -116,7 +126,9 @@ class TaskProviderHelper:
 
         # hardcoded mapping for local testing
         if HARDCODED_MODEL_ID_MAPPING.get(self.provider, {}).get(model):
-            return HARDCODED_MODEL_ID_MAPPING[self.provider][model]
+            return ProviderMappingInfo(
+                provider_id=HARDCODED_MODEL_ID_MAPPING[self.provider][model],
+            )
 
         provider_mapping = _fetch_inference_provider_mapping(model).get(self.provider)
         if provider_mapping is None:
@@ -132,7 +144,15 @@ class TaskProviderHelper:
             logger.warning(
                 f"Model {model} is in staging mode for provider {self.provider}. Meant for test purposes only."
             )
-        return provider_mapping.provider_id
+        if provider_mapping.adapter == "lora":
+            adapter_weights_path = _fetch_lora_weights_path(model)
+            return ProviderMappingInfo(
+                adapter_weights_path=adapter_weights_path,
+                provider_id=provider_mapping.provider_id,
+                hf_model_id=model,
+            )
+
+        return ProviderMappingInfo(provider_id=provider_mapping.provider_id)
 
     def _prepare_headers(self, headers: Dict, api_key: str) -> Dict:
         """Return the headers to use for the request.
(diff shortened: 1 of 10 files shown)


src/huggingface_hub/inference/_providers/_common.py
```diff
@@ -116,7 +126,9 @@ def _prepare_mapped_model(self, model: Optional[str]) -> str:
 
         # hardcoded mapping for local testing
         if HARDCODED_MODEL_ID_MAPPING.get(self.provider, {}).get(model):
-            return HARDCODED_MODEL_ID_MAPPING[self.provider][model]
+            return ProviderMappingInfo(
+                provider_id=HARDCODED_MODEL_ID_MAPPING[self.provider][model],
```
Wauplin, line 130: Let switch from `HARDCODED_MODEL_ID_MAPPING` to `HARDCODED_MODEL_INFERENCE_MAPPING` which would become a `Dict[str, Dict[str, ProviderMappingInfo]]`

Wauplin (commented):
I haven't tested locally yet but left some high-level comments related to naming + structure to match as much as possible the JS counterpart.

src/huggingface_hub/inference/_providers/_common.py
```diff
@@ -34,6 +35,13 @@ def filter_none(d: Dict[str, Any]) -> Dict[str, Any]:
     return {k: v for k, v in d.items() if v is not None}
 
 
+@dataclass
+class ProviderMappingInfo:
+    provider_id: str
+    adapter_weights_path: Optional[str] = None
+    hf_model_id: Optional[str] = None
```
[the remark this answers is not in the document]
hanouticelina: we can keep `task`, `hf_model_id` and `status` optional no? we don't need these in our python implementation but we can definitely include them in the dataclass

src/huggingface_hub/inference/_providers/_common.py
```diff
@@ -233,6 +267,17 @@ def _fetch_inference_provider_mapping(model: str) -> Dict:
...
 
+@lru_cache(maxsize=None)
+def _fetch_lora_weights_path(model: str) -> str:
+    from huggingface_hub.hf_api import HfApi
+
+    repo_files = HfApi().list_repo_files(model)
+    safetensors_files = [f for f in repo_files if f.endswith(".safetensors")]
+    if len(safetensors_files) != 1:
+        raise ValueError(f"Expected exactly one safetensors file in repo {model}, got {len(safetensors_files)}.")
+    return safetensors_files[0]
```
Wauplin, line 278: Wondering if this check could be done server-side instead cc @SBrandeis 

My reasoning is that here a repo can be tagged as `adapter: "lora"` in the provider's mapping but it's only when we try to use it client-side that we get the error because adapters weights file cannot be found. Ideally it would be nice if:
1. the check can be done server-side when mapping is created
2. the "model info" API can return both `adapter: "lora"` and `adapter_weights_path: "..."` => would avoid having every user requesting to list files from the Hub to check something that could be checked once

Wauplin: If we can make those 2 suggested changes it will simplify logic both in JS and Python.
If not really possible then I'm fine with current solution (but still a bit confused that the error happens at runtime)

src/huggingface_hub/inference/_providers/fal_ai.py
```diff
@@ -61,14 +64,29 @@ class FalAITextToImageTask(FalAITask):
...
+                "width": payload.pop("width"),
+                "height": payload.pop("height"),
             }
-        return {"prompt": inputs, **parameters}
+        if provider_mapping_info.adapter_weights_path is not None:
+            lora_path = (
+                f"{ENDPOINT}/"
+                f"{provider_mapping_info.hf_model_id}/resolve/main/"
+                f"{provider_mapping_info.adapter_weights_path}"
+            )
```
Wauplin, line 84:
```suggestion
            lora_path = constants.HUGGINGFACE_CO_URL_TEMPLATE.format(
                repo_id=provider_mapping_info.hf_model_id,
                revision="main",
                filename=provider_mapping_info.adapter_weights_path,
            )
```

(nit) we have a constants for that

Wauplin (commented):
Looking forward to this!

src/huggingface_hub/inference/_providers/_common.py
```diff
@@ -132,7 +145,22 @@ def _prepare_mapped_model(self, model: Optional[str]) -> str:
...
+                task=provider_mapping.task,
+                status=provider_mapping.status,
+            )
+
+        return ProviderMappingInfo(
+            provider_id=provider_mapping.provider_id,
+            hf_model_id=model,
+            task=provider_mapping.task,
+            status=provider_mapping.status,
+        )
```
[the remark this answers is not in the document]
hanouticelina: nice catch!

> I was a bit confused at first about the difference between `ProviderMappingInfo` (reconstructed here) and `InferenceProviderMapping` (returned by server). Ideally we would like to merge both but not sure it's doable.

yes actually when you suggested to have a dataclass matching the js one, i thought maybe we can just use `InferenceProviderMapping` but that would require to make `hf_model_id` optional (so not really matching the js interface).

src/huggingface_hub/inference/_providers/_common.py
SBrandeis: Yes, let's expose `adapterWeightsPath` from the API 👍

hanouticelina: I updated the PR to use the precomputed `adapterWeightsPath` [f55e628](https://github.com/huggingface/huggingface_hub/pull/3005/commits/f55e62872bbc6440b2e2a07d560971672346bfaf)

src/huggingface_hub/inference/_providers/_common.py
```diff
@@ -233,6 +248,17 @@ def _fetch_inference_provider_mapping(model: str) -> Dict:
...
+@lru_cache(maxsize=None)
+def _fetch_lora_weights_path(model: str) -> str:
+    from huggingface_hub.hf_api import HfApi
+
+    repo_files = HfApi().list_repo_files(model)
+    safetensors_files = [f for f in repo_files if f.endswith(".safetensors")]
+    if len(safetensors_files) != 1:
+        raise ValueError(f"Expected exactly one safetensors file in repo {model}, got {len(safetensors_files)}.")
+    return safetensors_files[0]
+
```
SBrandeis, line 260: Not needed anymore if I understand correctly :)
```suggestion
```

SBrandeis (approved):
lgtm

hanouticelina:
✅ I tested the PR with :
```python


client = InferenceClient(
    provider="fal-ai",
)
image = client.text_to_image(
    prompt="a boy and a girl looking out of a window with a cat perched on the window sill",
    model="openfree/flux-chatgpt-ghibli-lora",
)

```
output image :

Wauplin (approved):
All good! Pre-approving but still curious about this hardcoded `model_name`

src/huggingface_hub/inference/_providers/fal_ai.py
```diff
@@ -61,14 +65,29 @@ class FalAITextToImageTask(FalAITask):
...
-        return {"prompt": inputs, **parameters}
+        if provider_mapping_info.adapter_weights_path is not None:
+            lora_path = constants.HUGGINGFACE_CO_URL_TEMPLATE.format(
+                repo_id=provider_mapping_info.hf_model_id,
+                revision="main",
+                filename=provider_mapping_info.adapter_weights_path,
+            )
+            payload["loras"] = [{"path": lora_path, "scale": 1}]
+            if provider_mapping_info.provider_id == "fal-ai/lora":
+                payload["model_name"] = "stabilityai/stable-diffusion-xl-base-1.0"
```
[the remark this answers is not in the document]
hanouticelina: this is a special case when using loras with stable diffusion : https://fal.ai/models/fal-ai/lora/api. For flux loras we call  https://fal.ai/models/fal-ai/flux-lora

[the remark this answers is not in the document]
Wauplin: Thanks both for the explanation/links. I'd add a small comment about it for future ourselves

```suggestion
            if provider_mapping_info.provider_id == "fal-ai/lora":
                # little hack: fal requires the base model for stable-diffusion-based loras but not for flux-based
                # See payloads in https://fal.ai/models/fal-ai/lora/api vs https://fal.ai/models/fal-ai/flux-lora/api
                payload["model_name"] = "stabilityai/stable-diffusion-xl-base-1.0"
```

Wauplin (approved):

hanouticelina:
i'm merging, CI is red for unrelated reasons (hub-ci seems to be down)

PR #3011, opened 2025-04-16
PoC: `provider="auto"`

julien-c:


src/huggingface_hub/inference/_client.py
```diff
@@ -332,12 +332,13 @@ def audio_classification(
         ]
         ```
         """
-        provider_helper = get_provider_helper(self.provider, task="audio-classification")
+        model_id = model or self.model
+        provider_helper = get_provider_helper(self.provider, task="audio-classification", model=model_id)  # type: ignore
```
Wauplin, line 336: Why the `# type: ignore` in all calls?

Is it because `model or self.model` can be None? If that's the case I would rather fix this in `get_provider_helper` directly => e.g. accept `model: Optional[str]` as input and have a `if (model or self.model) is None and provider == "auto": raise ValueError(...)` check afterwards

Wauplin: (in general I feel that `# type: ignore` is ok when we internally use something a bit hacky but in the case here it feels more like a normal behavior rather than a corner case)

src/huggingface_hub/inference/_providers/__init__.py
```diff
@@ -35,6 +35,7 @@
     "replicate",
     "sambanova",
     "together",
+    "auto",
```
Wauplin, line 38: I would define a `PROVIDER_OR_POLICY_T = Union[PROVIDER_T, Literal["auto"]` instead. This way we keep `PROVIDER_T` as the proper type describing providers

src/huggingface_hub/inference/_providers/__init__.py
```diff
@@ -118,21 +119,30 @@
...
+        raise ValueError(
+            f"Provider '{provider}' not supported. Available providers: {['auto'] + list(PROVIDERS.keys())}, "
+            "'auto' will automatically select the first provider available for the model, sorted "
+            "by the user's order in https://hf.co/settings/inference-providers."
+        )
+
+    if provider == "auto":
+        provider_mapping = _fetch_inference_provider_mapping(model)
+        provider = next(iter(provider_mapping))
+
```
Wauplin, line 145:
```suggestion
    if provider == "auto":
        provider_mapping = _fetch_inference_provider_mapping(model)
        provider = next(iter(provider_mapping))

    if provider not in PROVIDERS:
        raise ValueError(
            f"Provider '{provider}' not supported. Available values: 'auto' or any provider from {list(PROVIDERS.keys())}."
            "Passing 'auto' (default value) will automatically select the first provider available for the model, sorted "
            "by the user's order in https://hf.co/settings/inference-providers."
        )
```

better this way IMO to decorrelate the "policy" from the "is this provider supported" check. Kinda corner case but it's possible that a provider is listed on the Hub but not yet supported by the Python client.

src/huggingface_hub/inference/_client.py
```diff
@@ -228,7 +228,7 @@ def __init__(
                 )
 
         # Configure provider
-        self.provider = provider if provider is not None else "hf-inference"
+        self.provider = provider if provider is not None else "auto"
```
Wauplin, line 231: wondering if we should default to `None` here (meaning `self.provider` is of type `Union[Literal["auto"], PROVIDER_T, None]` and then in `get_provider_helper(self.provider, task=..., model=...)` we default to `"auto"`.

My reasoning is that in case of passing a URL as `model` (e.g. for local TGI or Inference Endpoint), the provider should default to `"hf-inference"` and not `"auto"` (which would fail downstream). If we delegate all the logic to `get_provider_helper`, we should be quite safe in term of corner cases.

Wauplin (commented):
Made a high-level review. Looks good if it's really possible to retrieve provider ordering based on `/api/models/<model_id>` call.

Also in internal typing, I would make sure to distinguish between providers (cerebras, cohere, etc.) and policies (auto, maybe "cheapest", "fastest", etc.) in the future.

src/huggingface_hub/inference/_providers/__init__.py
hanouticelina: definitely! addressed in [1f852d6](https://github.com/huggingface/huggingface_hub/pull/3011/commits/1f852d60bf9e719841d578b6f37697f09a69b45a)

src/huggingface_hub/inference/_client.py
hanouticelina: addressed in [1f852d6](https://github.com/huggingface/huggingface_hub/pull/3011/commits/1f852d60bf9e719841d578b6f37697f09a69b45a)

src/huggingface_hub/inference/_client.py
hanouticelina: ah yes, that's indeed much better. addressed in [7e7ab1f](https://github.com/huggingface/huggingface_hub/pull/3011/commits/7e7ab1f10d757ac42563a4263504184a64c15b65)

tests/test_inference_providers.py
```diff
@@ -925,3 +918,28 @@ def test_recursive_merge(dict1: Dict, dict2: Dict, expected: Dict):
...
+            "provider-b": mocker.Mock(),
+        },
+    )
+    helper = get_provider_helper(provider="auto", task="test-task", model="test-model")
+
+    # The helper should be the one from provider-a
+    assert helper is mock_provider_a_helper
+
+    PROVIDERS.clear()
+    PROVIDERS.update(original_providers)
```
Wauplin, line 945:
```suggestion
    PROVIDERS.pop("provider-a", None)
    PROVIDERS.pop("provider-b", None)
```

(nit) (just wary of concurrency issues if we clear + update since tests are done in parallel)

src/huggingface_hub/inference/_providers/__init__.py
```diff
@@ -118,24 +121,43 @@
...
+    if provider is None:
+        warnings.warn(
+            "Defaulting to 'auto' which will select the first provider available for the model, sorted by the user's order in https://hf.co/settings/inference-providers.",
+            UserWarning,
+        )
+        provider = "auto"
+
+    if provider == "auto":
+        if model is None:
+            raise ValueError("Specifying a model is required when provider is 'auto'")
```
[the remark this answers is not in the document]
julien-c: not 100% sure we really need to do auto-model selection to be honest, inference is very model-centric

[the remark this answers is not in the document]
julien-c: (i agree with all your other comments on this PR though)

src/huggingface_hub/inference/_providers/__init__.py
```diff
@@ -118,24 +121,43 @@
...
-        raise ValueError(f"Provider '{provider}' not supported. Available providers: {list(PROVIDERS.keys())}")
-    if task not in PROVIDERS[provider]:
         raise ValueError(
-            f"Task '{task}' not supported for provider '{provider}'. "
-            f"Available tasks: {list(PROVIDERS[provider].keys())}"
+            f"Provider '{provider}' not supported. Available values: 'auto' or any provider from {list(PROVIDERS.keys())}."
+            "Passing 'auto' (default value) will automatically select the first provider available for the model, sorted "
+            "by the user's order in https://hf.co/settings/inference-providers."
+        )
+    provider_tasks = PROVIDERS[provider]  # type: ignore
```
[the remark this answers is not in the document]
hanouticelina: I think you still need the `# type: ignore` for `provider_tasks = PROVIDERS.get(provider)`. I'm getting the same "a Argument 1 to "get" of "dict" has incompatible type error" from mypy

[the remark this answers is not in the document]
Wauplin: ok then... :shrug:

Wauplin (approved):

hanouticelina:
failing tests are unrelated (staging environment is currently mostly down) so let's merge!

PR #3013, opened 2025-04-18
add route payload to deploy Inference Endpoints

Vaibhavs10:
ref: https://huggingface.slack.com/archives/C016D661PAN/p1744900723918749?thread_ts=1742997838.592509&cid=C016D661PAN

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -7612,9 +7612,9 @@ class HfApi:
             type ([`InferenceEndpointType]`, *optional*):
                 The type of the Inference Endpoint, which can be `"protected"` (default), `"public"` or `"private"`.
             domain (`str`, *optional*):
-                The custom domain for the Inference Endpoint (e.g. `"api-inference.endpoints.huggingface.tech"`).
+                The custom domain for the Inference Endpoint (e.g. `"my-new-domain.endpoints.huggingface.tech"`).
             path (`str`, *optional*):
-                The custom path for the Inference Endpoint (e.g. `"/models/google-bert/bert-base-uncased"`).
+                The custom path for the Inference Endpoint, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
             cache_http_responses (`bool`, *optional*):
                 Whether to cache HTTP responses from the Inference Endpoint. Defaults to `False`.
             tags (`List[str]`, *optional*):
@@ -7937,9 +7937,9 @@ class HfApi:
                 Secret values to inject in the container environment.
 
             domain (`str`, *optional*):
-                The custom domain for the Inference Endpoint (e.g. `"api-inference.endpoints.huggingface.tech"`).
+                The custom domain for the Inference Endpoint (e.g. `"my-new-domain.endpoints.huggingface.tech"`).
             path (`str`, *optional*):
-                The custom path for the Inference Endpoint (e.g. `"/models/google-bert/bert-base-uncased"`).
+                The custom path for the Inference Endpoint, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
 
             cache_http_responses (`bool`, *optional*):
                 Whether to cache HTTP responses from the Inference Endpoint.
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -7612,9 +7612,9 @@ class HfApi:
             type ([`InferenceEndpointType]`, *optional*):
                 The type of the Inference Endpoint, which can be `"protected"` (default), `"public"` or `"private"`.
             domain (`str`, *optional*):
-                The custom domain for the Inference Endpoint (e.g. `"my-new-domain.cool-website.woof"`).
+                The custom domain for the Inference Endpoint deployment, if setup the inference endpoint will be available at this domain (e.g. `"my-new-domain.cool-website.woof"`).
             path (`str`, *optional*):
-                The custom path for the Inference Endpoint, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
+                The custom path to model artifacts, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
             cache_http_responses (`bool`, *optional*):
                 Whether to cache HTTP responses from the Inference Endpoint. Defaults to `False`.
             tags (`List[str]`, *optional*):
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -7906,6 +7909,12 @@ def update_inference_endpoint(
                 Inference Endpoint running on the `text-generation-inference` (TGI) framework (see examples).
             secrets (`Dict[str, str]`, *optional*):
                 Secret values to inject in the container environment.
+
+            domain (`str`, *optional*):
+                The custom domain for the Inference Endpoint (e.g. `"api-inference.endpoints.huggingface.tech"`).
```
oOraph, line 7940: The custom domain must belong to the user and have been created using this api call beforehand:
https://api.endpoints.huggingface.cloud/#post-/v2/network/domain/-namespace-

src/huggingface_hub/hf_api.py
```diff
@@ -7906,6 +7909,12 @@ def update_inference_endpoint(
...
+            path (`str`, *optional*):
+                The custom path for the Inference Endpoint (e.g. `"/models/google-bert/bert-base-uncased"`).
```
oOraph, line 7942: path is a prefix match and should end with a / (auto corrected but if they want to be able to reach /whatever/path
on the backend the path should be /whatever)

src/huggingface_hub/hf_api.py
oOraph: it can either be:
- one we reserve on the fly in our dns zone custom-whatever.us-east-1.aws.endpoints.huggingface.cloud (if not already taken) 
- or one the customer brings to us (then they should cname it to custom-domains.endpoints.huggingface.cloud)

Vaibhavs10: since it's optional, maybe we don't need to elaborate much? wdyt?

src/huggingface_hub/hf_api.py
Vaibhavs10: yay! perfect! - I updated the doc string

src/huggingface_hub/hf_api.py
oOraph: yeah sure, need to document better the feature anyway but you're right this should not be done here in the docstring

oOraph (approved):

src/huggingface_hub/hf_api.py
```diff
@@ -7607,6 +7611,14 @@ def create_inference_endpoint(
                 Secret values to inject in the container environment.
             type ([`InferenceEndpointType]`, *optional*):
                 The type of the Inference Endpoint, which can be `"protected"` (default), `"public"` or `"private"`.
+            domain (`str`, *optional*):
+                The custom domain for the Inference Endpoint (e.g. `"my-new-domain.cool-website.woof"`).
+            path (`str`, *optional*):
+                The custom path for the Inference Endpoint, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
```
Wauplin, line 7617: What does it means exactly? Is it that the endpoint should be served under `endpoints-(...)-url.co/models/google-bert/bert-base-uncased` or is it that the URL `endpoints-(...)-url.co/` points to the `/models/google-bert/bert-base-uncased` path of the docker image? (worth some clarification in the docstring IMO)

src/huggingface_hub/hf_api.py
```diff
@@ -7607,6 +7611,14 @@ def create_inference_endpoint(
...
+            cache_http_responses (`bool`, *optional*):
```
Wauplin, line 7618: (link this [private thread](https://huggingface.slack.com/archives/C016D661PAN/p1744979768406779?thread_ts=1742997838.592509&cid=C016D661PAN) to avoid forgetting about it - if this attribute is HF-only, let's remove it)

src/huggingface_hub/hf_api.py
```diff
@@ -7607,6 +7611,14 @@ def create_inference_endpoint(
...
+                Whether to cache HTTP responses from the Inference Endpoint. Defaults to `False`.
+            tags (`List[str]`, *optional*):
+                A list of tags to associate with the Inference Endpoint.
```
Wauplin, line 7621: Out of curiosity, how do we play with tags once set? (e.g. can we list endpoints based on tags? or is it a UI change?)

Wauplin (commented):
nice addition :)

src/huggingface_hub/hf_api.py
Vaibhavs10: IIRC @oOraph mentioned this is applicable for anyone with a custom domain registered as well, so I'd consider this to be generic.

Wauplin: Yes, worth having the parameter supported in `huggingface_hub`! Just wanted to be sure what it meant exactly (_The custom path for the Inference Endpoint_ is not explicit enough IMO)

Vaibhavs10: updating now

Vaibhavs10: updated here: https://github.com/huggingface/huggingface_hub/pull/3013/commits/54572d5dce4bf3a2061cd83706b7b444e54938ad

julien-c: i think it's when you want to expose the resulting endpoint's inference route under a custom path but you can check with @oOraph and @XciD

Wauplin (approved):

hanouticelina (approved):

Wauplin (approved):

Wauplin:
CI failure is unrelated, merging it now (code quality is :heavy_check_mark: though)

src/huggingface_hub/hf_api.py
```diff
@@ -7906,6 +7935,17 @@ def update_inference_endpoint(
...
             secrets (`Dict[str, str]`, *optional*):
                 Secret values to inject in the container environment.
+
+            domain (`str`, *optional*):
+                The custom domain for the Inference Endpoint deployment, if setup the inference endpoint will be available at this domain (e.g. `"my-new-domain.cool-website.woof"`).
+            path (`str`, *optional*):
+                The custom path to the deployed model, should start with a `/` (e.g. `"/models/google-bert/bert-base-uncased"`).
+
+            cache_http_responses (`bool`, *optional*):
+                Whether to cache HTTP responses from the Inference Endpoint.
```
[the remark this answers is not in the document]
Wauplin: yes I think so in the end. It's also publicly documented here: https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-. This method is already not much used so if it can make HF's people life easier, let's go for it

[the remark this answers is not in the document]
tomaarsen: Perfect, convenient, thanks!

PR #3044, opened 2025-05-02
[inference] Necessary breaking change: nest task-specific route inside of model route

julien-c:
The alternative is just to break this feature...

### DIFF
--- a/src/huggingface_hub/inference/_providers/hf_inference.py
+++ b/src/huggingface_hub/inference/_providers/hf_inference.py
@@ -42,7 +42,7 @@ class HFInferenceTask(TaskProviderHelper):
             return mapped_model
         return (
             # Feature-extraction and sentence-similarity are the only cases where we handle models with several tasks.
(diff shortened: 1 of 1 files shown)


julien-c:
> the goal here is to point `feature-extraction` and `sentence-similarity` tasks to the same model route so that the same TEI deployment can answer both kinds of calls

not necessarily TEI, but yeah, the same container would be able to answer both calls/tasks, unlike now where we need two (redundant) running containers (unless we'd make complex URL/route rewriting)

Can you confirm this @oOraph?

hanouticelina:
I added unit tests and tested the PR 👍 all good, let's merge this (the inference tests are passing).
```python


client = InferenceClient(provider="hf-inference")

similarities = client.sentence_similarity(
    model="intfloat/e5-base-v2",
    sentence="Hello, how are you?",
    other_sentences=["Hello, how are you?", "I am fine, thank you!"],
)

print(similarities)
```

PR #3045, opened 2025-05-02
Add the 'env' parameter to creating/updating Inference Endpoints

tomaarsen:
Hello!

## Pull Request overview
* Add the 'env' parameter to creating/updating Inference Endpoints

## Details
See discussion: https://huggingface.slack.com/archives/C016D661PAN/p1746191822733239?thread_ts=1746187906.460849&cid=C016D661PAN

Here's the API Spec: https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-

In short: I can pass `secrets`, but I cannot pass `env` environment variables. It's not extremely critical as using `secrets` instead works fine for my purposes as well, but `env` would be more fitting, and it makes sense to expose whatever the API Spec supports in the `huggingface_hub` functions.

- Tom Aarsen

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -7681,14 +7681,14 @@ class HfApi:
             ...     type="protected",
             ...     instance_size="x1",
             ...     instance_type="nvidia-a10g",
+            ...     env={
+            ...           "MAX_BATCH_PREFILL_TOKENS": "2048",
+            ...           "MAX_INPUT_LENGTH": "1024",
+            ...           "MAX_TOTAL_TOKENS": "1512",
+            ...           "MODEL_ID": "/repository"
+            ...         },
             ...     custom_image={
             ...         "health_route": "/health",
-            ...         "env": {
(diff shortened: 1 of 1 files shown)


oOraph (approved):

julien-c (commented):
let wait for a maintainer review but LGTM

hanouticelina (approved):
Looks good! thank you @tomaarsen.
based on the schema [here](https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-), it seems that we're not supposed to send `env` via the `image` field. If that's the case, could you update the [example](https://github.com/huggingface/huggingface_hub/blob/557576de090963bdc641954f56c705aa71b9a072/src/huggingface_hub/hf_api.py#L7670) in the docstring accordingly please?

Wauplin (approved):
(Too bad that naming is not consistent between Inference Endpoints `env`/`secrets` and Spaces `variables`/`secrets` but I guess that now it's done we can't do much :confused: . Not something to fix client-side anyway)

hanouticelina:
I took the liberty of updating the docstring, thanks again @tomaarsen for the PR! let's merge

PR #3050, opened 2025-05-05
Fix CACHE_DIR value in cache documentation

albertcthomas:
I think it was wrong to say that the CACHE_DIR was usually your user's home directory. I fixed it by putting the default value.


hanouticelina (approved):
thanks for spotting and fixing this!

PR #3062, opened 2025-05-08
Dont set scale to zero as default when creating an Endpoint

tomaarsen:
Hello!

## Pull Request overview
* Dont set scale to zero as default when creating an Endpoint

## Details
As discussed [internally](https://huggingface.slack.com/archives/C03E4DQ9LAJ/p1746625809773699?thread_ts=1746617597.432019&cid=C03E4DQ9LAJ), this PR sets the `scale_to_zero_timeout` to `None` by default. It's not entirely clear to me if the `min_replica`/`max_replica` is also intended to be set to `None` from the discussion though - we can make those changes still.

Also, when someone specifies the `scale_to_zero_timeout`, we can help them out by overriding the `min_replica` to 0, but we can also do nothing as users will get an error anyways that the `minReplica` should not be 1.

I've also added an example of how to create an endpoint that scales to zero, and I quickly tested that running that example with the new `scale_to_zero_timeout` & `min_replica` defaults indeed creates an endpoint that doesn't scale to zero.

cc @Wauplin @Vaibhavs10 @ErikKaum 

- Tom Aarsen


Vaibhavs10 (approved):

src/huggingface_hub/hf_api.py
```diff
@@ -7566,9 +7566,9 @@ def create_inference_endpoint(
         region: str,
         vendor: str,
         account_id: Optional[str] = None,
-        min_replica: int = 0,
+        min_replica: int = 1,
         max_replica: int = 1,
-        scale_to_zero_timeout: int = 15,
+        scale_to_zero_timeout: int | None = None,
```
hanouticelina, line 7571: the PEP 604 syntax won't work in python 3.8 (we still support python 3.8 in hfh).
```suggestion
        scale_to_zero_timeout: Optional[int] = None,
```

src/huggingface_hub/hf_api.py
```diff
@@ -7566,9 +7566,9 @@ def create_inference_endpoint(
         region: str,
         vendor: str,
         account_id: Optional[str] = None,
-        min_replica: int = 0,
+        min_replica: int = 1,
```
hanouticelina, line 7569: Can we set the default value to `None` here as well? if we do that, what would be the default value set server-side? cc @ErikKaum 

(if we can set the default balue to `None`, let's do the same for `max_replica`).

hanouticelina (commented):
Thanks @tomaarsen, looks good! I left two comments about the default values of `min_replica`/`max_replica` and another one about the typing syntax

src/huggingface_hub/hf_api.py
tomaarsen: Ah, I figured you used `from __future__ import annotations`! Will update

src/huggingface_hub/hf_api.py
hanouticelina: Okay then let's set the default value to 1, as you already did!

hanouticelina (approved):
Looks good, let's merge!

> Also, when someone specifies the scale_to_zero_timeout, we can help them out by overriding the min_replica to 0, but we can also do nothing as users will get an error anyways that the minReplica should not be 1.

I’d prefer not to add any pre-processing logic here or override `min_replica` based on the value of `scale_to_zero_timeout`. The error message is already quite clear as you mentioned:
```
Bad Request: 'scaleToZeroTimeout' can only be set when 'minReplica' is 0
```
I think it's better to let the user handle it explicitly.

PR #3063, opened 2025-05-08
Expand file lock scope to resolve concurrency issues during downloads

humengyu2012:
# Pull Request overview
Expand file lock scope to resolve concurrency issues during downloads

# Details
We want to use a network file system as the cache directory (HF_HOME). After testing various network file systems(such as NFS), we observed strange errors when multiple requests concurrently download the same file.

We found that some symlink operations are protected by locks, while others are not, which can lead to conflicts.
In particular, on file systems that lack symlink support, symlink operations degrade into copy and mv commands, which further exacerbate these conflicts.

Therefore, I created this PR to expand the scope of the flock in order to resolve these conflicts.


src/huggingface_hub/file_download.py
```diff
@@ -1130,19 +1130,21 @@ def _hf_hub_download_to_cache_dir(
...
+    # If file already exists, return it (except if force_download=True)
+    with WeakFileLock(lock_path):
+        if not force_download:
+            if os.path.exists(pointer_path):
+                return pointer_path
+
+            if os.path.exists(blob_path):
+                # we have the blob already, but not the pointer
+                _create_symlink(blob_path, pointer_path, new_blob=False)
+                return pointer_path
```
hanouticelina, line 1147: Two suggestions here:
- The pointer file is immutable once created, if it already exists, we can return safely without grabing the lock. 
- I think it's safer to double-check again that the `pointer_path` does not exist (line 1146 in my code suggestion) because another process can create the pointer in the interval between the outer if (` if not force_download and os.path.exists(blob_path)`) and the moment we acquired the lock, it's rare but stiil let's cover this race window.

```suggestion
 
    # pointer already exists -> immediate return
    if not force_download and os.path.exists(pointer_path):
        return pointer_path

    # Blob exists but pointer must be (safely) created -> take the lock
    if not force_download and os.path.exists(blob_path):
        with WeakFileLock(lock_path):
            if not os.path.exists(pointer_path):
                _create_symlink(blob_path, pointer_path, new_blob=False)
            return pointer_path
```

hanouticelina (commented):
Thanks @humengyu2012 for the PR! This indeed should fix the race condition on file systems that replace symlinks with non atomic copy -> rename 👍

src/huggingface_hub/file_download.py
hanouticelina: (this suggestion should fix the failing test)

humengyu2012:
Thanks for your review! I have made further changes. Could you take a look again? @hanouticelina

hanouticelina:
@bot /style

hanouticelina (approved):
thank you @humengyu2012 for the PR! let's merge

PR #3092, opened 2025-05-19
Surface 401 unauthorized errors more directly in `snapshot_download`

hanouticelina:
originally from @Vaibhavs10 in this slack internal [message](https://huggingface.slack.com/archives/C02V5EA0A95/p1747396386743699) who reported this user feedback: https://x.com/felix_red_panda/status/1923145537693405269

Currently, if a 401 error occurs during the initial API call in `snapshot_download` and is not specifically classified as a `RepositoryNotFoundError` or `GatedRepoError`, it can fall through to a generic `LocalEntryNotFoundError`. This requires users to inspect the full stack trace to identify the underlying 401 error.

This change modifies the error handling in `snapshot_download` to explicitly check for `HfHubHTTPError` with a status code of 401. If such an error is detected, it is now re-raised directly.

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -7,7 +7,13 @@ from tqdm.auto import tqdm as base_tqdm
 from tqdm.contrib.concurrent import thread_map
 
 from . import constants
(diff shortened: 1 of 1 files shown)


Wauplin (approved):
Makes sense indeed, thanks for the quick fix!

PR #3096, opened 2025-05-20
Improved progress reporting for Xet uploads

hoytak:
This PR adds detailed progress reporting for upload_files when hf_xet is used, showing both per-file progress and accurate total progress. Total progress speed, which includes both deduplication and data transfer, is also separated out into separate bars.

Requries xet-core / hf_xet at commit 4faec0b or later, which can be obtained using `pip install --pre "hf-xet==1.1.3dev0"`


hoytak:
Here is a video of the current version of it in action:


https://github.com/user-attachments/assets/1ef67fb1-537a-49e8-8b7b-a43dddb07eec

setup.py
```diff
@@ -19,6 +19,7 @@ def get_version() -> str:
     "requests",
     "tqdm>=4.42.1",
     "typing-extensions>=3.7.4.3",  # to be able to import TypeAlias
+    "rich>=14.0.0"
```
Wauplin, line 22: As mentioned in [slack](https://huggingface.slack.com/archives/C02V5EA0A95/p1747847427127569?thread_ts=1747771966.873299&cid=C02V5EA0A95) (private link) I'm quite reluctant to add [rich](https://github.com/Textualize/rich) as a required dependency :confused:

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
+from typing import List
+
+try:
+    from hf_xet import PyItemProgressUpdate, PyTotalProgressUpdate
+except ImportError as e:
+    raise ImportError(f"The current version of hf_xet does not yet support detailed upload progress; please upgrade ({e})")
```
Wauplin: Let's update the minimal hf_xet version in [`setup.py`](https://github.com/huggingface/huggingface_hub/blob/4cc31432f0875946250b3ccbfe2fcf56af2fca74/setup.py#L17) instead to ensure it's always supported

Wauplin:
@bot /style

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
...
+            w = self.description_width - 2
+            pad = "  "
+        else:
+            w = self.description_width
+            pad = ""
+            
+        if len(name) > w:
+            return pad + "..." + name[-(w - 3):]
+        else:
+            return pad + name.ljust(w, " ")
```
Wauplin: (nit) feels more pythonic and readable to me like this

```suggestion
        padding = "  " if indent else ""
        width = self.description_width - len(padding)        
    
        if len(name) > width:
            truncated = f"...{name[-(width - 3):]}"
            return f"{padding}{truncated}"
        return f"{padding}{name.ljust(width)}"
```

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
...
+        self.upload_bar.close()
+        for bar in self.current_bars: 
+            if bar:
+                bar.close()
+
+
+XetProgressTracker = ProgressReporter
+__all__ = [
+    "XetProgressTracker",
+]
```
Wauplin: I would rather rename `ProgressReporter` as `XetProgressTracker` directly and remove the `__all__` magic

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
...
+                bar = self.current_bars[bar_idx]
+                in_final_bar_mode = False
+
+            if bar is None:
+                self.current_bars[bar_idx] = bar = \
+                    tqdm(desc = self.format_desc(name, True), 
+                                position = 2 + bar_idx, # Set to the position past the initial bars.
+                                total = item.total_bytes, 
+                                initial = item.bytes_completed, 
+                                **self.tqdm_settings)
```
Wauplin: re-defining `bar` not needed here, right? (not used afterwards since the `elif`/`else`)

```suggestion
                self.current_bars[bar_idx] = tqdm(desc = self.format_desc(name, True), 
                                position = 2 + bar_idx, # Set to the position past the initial bars.
                                total = item.total_bytes, 
                                initial = item.bytes_completed, 
                                **self.tqdm_settings)
```

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
...
+
+class ProgressReporter:
+    def __init__(self, n_lines: int = 10, description_width : int = 40):
+        self.n_lines = n_lines
+        self.description_width = description_width
+
+        self.tqdm_settings = {"unit" : "B", 
+                              "unit_scale" : True, 
+                              "leave" : True, 
+                              "unit_divisor" : 1024,
```
Wauplin:
```suggestion
                              "unit_divisor" : 1000, 
```

I would prefer to use MB instead of MiB

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -0,0 +1,155 @@
...
+            del self.item_state[name]
+
+        # Now manually refresh each of the bars
+        for bar in self.current_bars: 
+            if bar:
+                bar.refresh()
+        
+        # Update overall bars
+        def postfix(speed): 
+            s = tqdm.format_sizeof(speed, divisor=1024) if speed is not None else "???"
```
Wauplin:
```suggestion
            s = tqdm.format_sizeof(speed) if speed is not None else "???"
```

same here (and `divisor=1000` is already [the default](https://tqdm.github.io/docs/tqdm/#format_sizeof))

src/huggingface_hub/_commit_api.py
```diff
@@ -569,38 +571,29 @@ def token_refresher() -> Tuple[str, int]:
...
     num_chunks = math.ceil(len(additions) / UPLOAD_BATCH_MAX_NUM_FILES)
     num_chunks_num_digits = int(math.log10(num_chunks)) + 1
-    for i, chunk in enumerate(chunk_iterable(additions, chunk_size=UPLOAD_BATCH_MAX_NUM_FILES)):
-        _chunk = [op for op in chunk]
 
-        bytes_ops = [op for op in _chunk if isinstance(op.path_or_fileobj, bytes)]
-        paths_ops = [op for op in _chunk if isinstance(op.path_or_fileobj, (str, Path))]
-        expected_size = sum(op.upload_info.size for op in bytes_ops + paths_ops)
+    if not are_progress_bars_disabled():
+        progress = XetProgressTracker(10)
```
Wauplin:
```suggestion
        progress = XetProgressTracker()
```

let's use the default value from the class directly (which is also `10`)

Wauplin (commented):
Thanks a lot for the work @hoytak ! I tested the PR both locally and in a colab to check how it behaves and needless to say that it's a great addition :fire: And thanks again for removing any extra dependency!

Regarding the implementation itself, I did not review in depth the hf_xet/python binding as it seems to "just work" + the comments make it clear what's going on.  I've left some comments to address (mostly cosmetic). Apart from that we should be good to merge.

---

Attaching some screenshots of my tests

image

image

hoytak:
> Thanks a lot for the work @hoytak ! I tested the PR both locally and in a colab to check how it behaves and needless to say that it's a great addition 🔥 And thanks again for removing any extra dependency!
> 

Great review!  Thank you.  

However, these screenshots are for the download path, while these changes are part of the upload path.   It should be similar, but just wanted to clarify.

Wauplin (approved):
Looks ready to be merged as soon as the CI is green! :tada:

Wauplin:
re-tested properly :facepalm: 

image

image

hoytak:
@Wauplin -- could I get another trigger for the workflows?  I didn't catch a python 3.9 issue locally.

hoytak:
@Wauplin Thank you!  Ready to merge if you are (I assume you have to do that?).

Wauplin:
Yes, ready to merge!

hanouticelina:
very nice! 🔥

PR #3103, opened 2025-05-22
[Internal] make `hf-xet` (again) a required dependency

hanouticelina:
Follow-up PR after #3079 in which we made `hf-xet` optional due to reports about the binaries being too big on Linux platform. Thanks to the Xet team, now the installed binaries of `hf-xet` are much much smaller! (eg. Linux went from ~96MB → ~14MB). Release notes: https://github.com/huggingface/xet-core/releases/tag/v1.1.2.

Given these improvements, we can now make `hf-xet` a required dependency again.

cc @Vaibhavs10 @jsulz

### DIFF
--- a/.github/conda/meta.yaml
+++ b/.github/conda/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - typing-extensions
     - packaging
     - pyyaml
+    - hf-xet
   run:
     - python
     - pip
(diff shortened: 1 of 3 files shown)


rajatarya:
(Sorry for being late on this - excited to see this happen, yay!)

and thanks!

Vaibhavs10 (commented):
thanks a lot @hanouticelina - you're the best! ❤️

PR #3106, opened 2025-05-23
Refactor metadata reading

julien-c:
## Summary
- factorize metadata reading logic in `_local_folder.py`

## Testing
- `make style`
- `make quality` *(fails: missing type stubs)*
- `make test` *(fails: test collection errors)*


Wauplin (commented):
While the code is correct and does what we want, I don't find it more readable (it de-duplicates the `try/except/paths.unlink()` logic at the cost of one level of abstraction more)

julien-c:
@copilot close the PR following maintainer feedback

PR #3165, opened 2025-06-17
fix(inference_endpoints): use GET `healthRoute` instead of GET / to check status

mfuntowicz:
This PR fixes a status handling issue with the newly introduced vLLM container on Inference Endpoint.

Currently, the library attempt to send a `GET /` to the endpoint URL and expect a `200 OK` to update the internal status to RUNNING.

While this works with TGI based container, vLLM do not expose such route and thus `GET /` leads to a `404 NOT FOUND` leaving the status in a on-going status pulling loop.

This PR proposes to move from `GET /` to `GET /health` which is exposed by both (and certainly more generally) other endpoint types.

### DIFF
--- a/src/huggingface_hub/_inference_endpoints.py
+++ b/src/huggingface_hub/_inference_endpoints.py
@@ -220,7 +220,7 @@ class InferenceEndpoint:
                 )
             if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
                 # Verify the endpoint is actually reachable
-                response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
+                response = get_session().get(f"{self.url}/health", headers=self._api._build_hf_headers(token=self._token))
                 if response.status_code == 200:
                     logger.info("Inference Endpoint is ready to be used.")
                     return self
--- a/src/huggingface_hub/_inference_endpoints.py
+++ b/src/huggingface_hub/_inference_endpoints.py
@@ -100,7 +100,7 @@ class InferenceEndpoint:
     namespace: str
     repository: str = field(init=False)
     status: InferenceEndpointStatus = field(init=False)
-    health_url: str = field(init=False)
+    health_route: str = field(init=False)
     url: Optional[str] = field(init=False)
 
     # Other fields
@@ -117,6 +117,7 @@ class InferenceEndpoint:
     # Internal fields
     _token: Union[str, bool, None] = field(repr=False, compare=False)
     _api: "HfApi" = field(repr=False, compare=False)
+    _health_url: str = field(repr=False, init=False, compare=False)
 
     @classmethod
     def from_raw(
@@ -221,7 +222,7 @@ class InferenceEndpoint:
                 )
             if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
                 # Verify the endpoint is actually reachable
-                response = get_session().get(self.health_url, headers=self._api._build_hf_headers(token=self._token))
+                response = get_session().get(self._health_url, headers=self._api._build_hf_headers(token=self._token))
                 if response.status_code == 200:
                     logger.info("Inference Endpoint is ready to be used.")
                     return self
@@ -242,6 +243,9 @@ class InferenceEndpoint:
         obj = self._api.get_inference_endpoint(name=self.name, namespace=self.namespace, token=self._token)  # type: ignore [arg-type]
         self.raw = obj.raw
         self._populate_from_raw()
+
+        # Update additional utility variables compute from the raw payload
+        self._health_url = f"{self.url.rstrip('/')}/{self.health_route.lstrip('/')}"
         return self
 
     def update(
@@ -401,7 +405,7 @@ class InferenceEndpoint:
         self.repository = self.raw["model"]["repository"]
         self.status = self.raw["status"]["state"]
         self.url = self.raw["status"].get("url")
-        self.health_url = self.raw["healthRoute"]
+        self.health_route = self.raw["healthRoute"]
 
         # Other fields
         self.framework = self.raw["model"]["framework"]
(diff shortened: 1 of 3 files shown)


Wauplin:
as discussed [on slack](https://huggingface.slack.com/archives/C07QRFB1JU8/p1750151998091269?thread_ts=1750110264.624129&cid=C07QRFB1JU8) (private), let's use the `healthRoute` provided by the API (I didn't know about this one^^)

image

mfuntowicz:
`healthRoute` is available at the root of the endpoint description payload

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -400,6 +401,7 @@ def _populate_from_raw(self) -> None:
         self.repository = self.raw["model"]["repository"]
         self.status = self.raw["status"]["state"]
         self.url = self.raw["status"].get("url")
+        self.health_url = self.raw["healthRoute"]
```
Wauplin:
```suggestion
        self.health_route = self.raw["healthRoute"]
```

for consistency, can you name it `health_route` in Python client ? (+ in other places)

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -220,7 +221,7 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
                 )
             if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
                 # Verify the endpoint is actually reachable
-                response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
+                response = get_session().get(self.health_url, headers=self._api._build_hf_headers(token=self._token))
```
Wauplin: Value returned by server is a route, not a URL

```suggestion
                health_url = f"{self.url.rstrip('/')}/{self.health_route.lstrip('/')}"
                response = get_session().get(health_url, headers=self._api._build_hf_headers(token=self._token))
```

tests/test_inference_endpoints.py
```diff
@@ -158,6 +162,7 @@ def test_from_raw_initialization():
     assert endpoint.revision == "11c5a3d5811f50298f278a704980280950aedb10"
     assert endpoint.task == "text-generation"
     assert endpoint.type == "protected"
+    assert endpoint.health_url == "/health"
```
Wauplin:
```suggestion
    assert endpoint.health_route == "/health"
```

src/huggingface_hub/_inference_endpoints.py
Wauplin: potentially `health_url` could be an additional class property (for easier reuse)

tests/test_inference_endpoints.py
Wauplin, line 1: Can you add a test around https://github.com/huggingface/huggingface_hub/blob/5a1526e309296c3a0c6cd8a05f46e55a6f2688f1/tests/test_inference_endpoints.py#L247 to check that `https://vksrvs8pc1xnifhq.us-east-1.aws.endpoints.huggingface.cloud/health` has been called (with `/health`) ?

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -219,8 +221,11 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
                     f"Inference Endpoint {self.name} failed to update. Please check the logs for more information."
                 )
             if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
+                # Update additional utility variables compute from the raw payload
+                self._health_url = f"{self.url.rstrip('/')}/{self.health_route.lstrip('/')}"
```
Wauplin:
```suggestion
```

let's remove the property for now and add it back if we need it in the future (sorry for the suggestion)

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -219,8 +221,11 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
...
+
                 # Verify the endpoint is actually reachable
-                response = get_session().get(self.url, headers=self._api._build_hf_headers(token=self._token))
+                response = get_session().get(self._health_url, headers=self._api._build_hf_headers(token=self._token))
```
Wauplin:
```suggestion
                _health_url = f"{self.url.rstrip('/')}/{self.health_route.lstrip('/')}"
                response = get_session().get(_health_url, headers=self._api._build_hf_headers(token=self._token))
```

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -116,6 +117,7 @@ class InferenceEndpoint:
     # Internal fields
     _token: Union[str, bool, None] = field(repr=False, compare=False)
     _api: "HfApi" = field(repr=False, compare=False)
+    _health_url: str = field(repr=False, init=False, compare=False)
```
Wauplin:
```suggestion
```

(cf above)

Wauplin (approved):
Let's revert _health_url entirely (sorry for the suggestion) and we should be good to go

tests/test_inference_endpoints.py
Wauplin: done in https://github.com/huggingface/huggingface_hub/pull/3165/commits/7bcfb6258e62caec06fd85475dd6ed00f021f7d5

Wauplin (approved):
I've added a test in https://github.com/huggingface/huggingface_hub/pull/3165/commits/7bcfb6258e62caec06fd85475dd6ed00f021f7d5. Time to merge :)

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -219,8 +220,11 @@ def wait(self, timeout: Optional[int] = None, refresh_every: int = 5) -> "Infere
                     f"Inference Endpoint {self.name} failed to update. Please check the logs for more information."
                 )
             if self.status == InferenceEndpointStatus.RUNNING and self.url is not None:
+                # Update additional utility variables compute from the raw payload
+
```
Wauplin:
```suggestion
```

hanouticelina (approved):
thank you!

Issue #3218, opened 2025-07-12
Failed to parse tools: key 'parameters' not found in mcp inference


nicoloddo:
### Describe the bug

When running a local agent with MCP inference tooling, an internal server error gets thrown saying "failed to parse tools". This is due to the default exit tools missing a key 'parameters' in their definition in the file **huggingface_hub/inference/_mcp/constants.py**.

The bug happens following directly the instruction in the Hugging Face MCP course, section lemonade-server at [link](https://huggingface.co/learn/mcp-course/en/unit2/lemonade-server), which makes it impossible to complete and gain experience with it.

### Reproduction

1. Install lemonade-server.ai
2. Install Qwen3-8B-GGUF in it
3. create agent.json with:
{
    "model": "Qwen3-8B-GGUF",
    "endpointUrl": "http://localhost:8000/api/"
}
4. tiny-agents run agent.json

### Logs

```shell
DEBUG:    Sending HTTP Request: POST http://127.0.0.1:58926/v1/chat/completions
DEBUG:    connect_tcp.started host='127.0.0.1' port=58926 local_address=None timeout=5.0 socket_options=None
DEBUG:    connect_tcp.complete return_value=<httpcore._backends.sync.SyncStream object at 0x00000238D7866800>
DEBUG:    send_request_headers.started request=<Request [b'POST']>
DEBUG:    send_request_headers.complete
DEBUG:    send_request_body.started request=<Request [b'POST']>
DEBUG:    send_request_body.complete
DEBUG:    receive_response_headers.started request=<Request [b'POST']>
DEBUG:    receive_response_headers.complete return_value=(b'HTTP/1.1', 500, b'Internal Server Error', [(b'Server', b'llama.cpp'), (b'Access-Control-Allow-Origin', b''), (b'Content-Type', b'application/json; charset=utf-8'), (b'Content-Length', b'546'), (b'Keep-Alive', b'timeout=5, max=100')])
[...]

PR #3214, opened 2025-07-11
Omit `parameters` in default tools of `tiny-agent`

hanouticelina:
Fixes #3213.
OpenAI validates every function -> parameters object as a full json schema, if the schema contains a "properties" field, at least one property has to be defined. according to OpenAI’s specs, omitting parameters is the correct way to express a function that takes no arguments:

Screenshot 2025-07-11 at 11 21 24

Issue #3213, opened 2025-07-11
Tool Schema not OpenAI compatible


debasisdwivedy:
### Describe the bug

Hi,

Currently tinyAgent is not OpenAI compatible. The tool exception fails with the error below. This is due to faulty constants named "TASK_COMPLETE_TOOL" and "ASK_QUESTION_TOOL" which are used for "EXIT_LOOP_TOOLS". The file present in  :  

**./.venv/lib/python3.13/site-packages/huggingface_hub/inference/_mcp/constants.py** 

The issue can be addressed in two possible ways:

1.  Change **constants.py** at [**./.venv/lib/python3.13/site-packages/huggingface_hub/inference/_mcp/constants.py**].

The properties field cannot be a blank dictionary. Either remove it or add something like below.

```
from __future__ import annotations

import sys
from pathlib import Path
from typing import List

from huggingface_hub import ChatCompletionInputTool


FILENAME_CONFIG = "agent.json"
FILENAME_PROMPT = "PROMPT.md"

DEFAULT_AGENT = {
    "model": "Qwen/Qwen2.5-72B-Instruct",
    "provider": "nebius",
    "servers": [
        {
            "type": "stdio",
            "config": {
                "command": "npx",
                "args": [
                    "-y",
                    "@modelcontextprotocol/server-filesystem",
                    str(Path.home() / ("Desktop" if sys.platform == "darwin" else "")),
                ],
            },
        },
        {
            "type": "stdio",
            "config": {
                "command": "npx",
                "args": ["@playwright/mcp@latest"],
            },
        },
    ],
}


DEFAULT_SYSTEM_PROMPT = """
You are an agent - please keep going until the user’s query is completely
resolved, before ending your turn and yielding back to the user. Only terminate
your turn when you are sure that the problem is solved, or if you need more
info from the user to solve the problem.
If you are not sure about anything pertaining to the user’s request, use your
tools to read files and gather the relevant information: do NOT guess or make
up an answer.
You MUST plan extensively before each function call, and reflect extensively
on the outcomes of the previous function calls. DO NOT do this entire process
by making function calls only, as this can impair your ability to solve the
problem and think insightfully.
""".strip()

MAX_NUM_TURNS = 10

TASK_COMPLETE_TOOL: ChatCompletionInputTool = ChatCompletionInputTool.parse_obj(  # type: ignore[assignment]
    {
        "type": "function",
        "function": {
            "name": "task_complete",
[...]

PR #3219, opened 2025-07-12
Fix mcp EXIT_LOOP_TOOLS tools not following the ChatCompletionInputFunctionDefinition model

nicoloddo:


nicoloddo:
I see this pr is related to: [https://github.com/huggingface/huggingface_hub/pull/3214/files](https://github.com/huggingface/huggingface_hub/pull/3214/files). 

Let's add a "trigger" dummy parameter to preserve compatibility with OpenAI schema, as suggested in [https://github.com/huggingface/huggingface_hub/issues/3213](https://github.com/huggingface/huggingface_hub/issues/3213)

nicoloddo:
This now solves both [https://github.com/huggingface/huggingface_hub/issues/3213](https://github.com/huggingface/huggingface_hub/issues/3213) and [https://github.com/huggingface/huggingface_hub/issues/3218](https://github.com/huggingface/huggingface_hub/issues/3218)

hanouticelina (approved):
I'm okay with adding a dummy parameters field here, seems like some frameworks do parse `parameters` even for tools that take no input, so indeed it's safer to include it explicitly. thank you @nicoloddo for the PR!

nicoloddo:
@hanouticelina no problem! About the ask question tool, I think the AI may start to call the tool instead of asking the question directly to the user, whom will not receive any output. 
I am not sure what is your approach for the exit loop: does the AI **need** to call a tool to exit? 
- If not, I would specify in the docstring that the tool should not really be called and that the agent should instead direct the question to the user, or even delete the tool altogether and change prompt to address exiting the loop;
- If yes, maybe I would pass the question through the tool and return it.

In any case, you might want to do something about that!

julien-c (commented):
makes sense imo as a way to maximize compatibility...

PR #3242, opened 2025-07-21
[Tiny Agents] Add tools to config

NielsRogge:
This PR enables tools inclusion/exclusion for a given agent. Users can specify the tools to include/exclude like so:

```diff
{
  "model": "meta-llama/Meta-Llama-3-8B-Instruct",
  "provider": "auto",
  "servers": [
    {
      "type": "stdio",
      "command": "npx",
      "args": ["@playwright/mcp@latest"],
+      "allowed_tools": {
+        "include": ["browser_click", "browser_close"]
      }
    }
  ]
}
```

Note: this feature was implemented using Claude Code.


examples/tool_filtering_demo/README.md
Wauplin, line 1: to be removed

examples/tool_filtering_demo/agent.json
Wauplin, line 1: to be moved to https://huggingface.co/datasets/tiny-agents/tiny-agents

tiny_agents.md
Wauplin, line 1: to be removed

src/huggingface_hub/inference/_mcp/types.py
```diff
@@ -10,24 +10,32 @@ class InputConfig(TypedDict, total=False):
...
+    exclude: NotRequired[List[str]]
+
+
 class StdioServerConfig(TypedDict):
     type: Literal["stdio"]
     command: str
     args: List[str]
     env: Dict[str, str]
     cwd: str
+    tools: NotRequired[ToolsConfig]
```
Wauplin: Can you rename it `allowed_tools` and accept only a list of strings as input? I can see pretty clearly the goal of allowing only a certain subset of tools but not necessarily excluding some. In the future if we really need it we can always add a new `forbidden_tools` property.

Note: this is taken from OpenAI's specs 
image

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -233,6 +248,50 @@ async def add_mcp_server(self, type: ServerType, **params: Any):
                 )
             )
 
+    def _filter_tools(
+        self, tools: List[Any], tools_config: Optional[Dict[str, Any]], all_tool_names: List[str]
+    ) -> List[Any]:
```
Wauplin: Haven't reviewed this logic myself but I feel that if we only have a `allowed_tools` list then a 1-line list comprehension should be enough

NielsRogge:
@claude address the comments

NielsRogge:
@claude address the comments

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -233,6 +248,28 @@ async def add_mcp_server(self, type: ServerType, **params: Any):
...
+            return tools
+
+        # Validate that specified tools exist
+        all_tool_names = [tool.name for tool in tools]
+        missing_tools = set(allowed_tools) - set(all_tool_names)
+        if missing_tools:
+            logger.warning(f"Tools specified in 'allowed_tools' not found on server: {list(missing_tools)}")
+
+        # Filter tools using list comprehension
+        return [tool for tool in tools if tool.name in allowed_tools]
```
Wauplin: This method is far too complex and verbose IMO. Would be better to remove it and replace by a simple `[tool for tool in tools if tool.name in allowed_tools]`.

tests/test_mcp_client.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
Wauplin: not a big fan of this test file (lots of mocks, quite verbose and hard to read). 

Would prefer to remove it and not add tests for the allowed_tools (MCP client entirely is not tested/experimental).

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -209,9 +215,18 @@ async def add_mcp_server(self, type: ServerType, **params: Any):
...
         response = await session.list_tools()
-        logger.debug("Connected to server with tools:", [tool.name for tool in response.tools])
+        all_tool_names = [tool.name for tool in response.tools]
+        logger.debug("Connected to server with tools:", all_tool_names)
+
+        # Filter tools based on allowed_tools configuration
+        filtered_tools = self._filter_tools(response.tools, allowed_tools)
+
+        if allowed_tools:
+            logger.info(
```
Wauplin: let put this message at a "debug" level as well (same as the one above)

Wauplin (commented):
Re-reviewed it. To be honest it's not very exciting to review auto-generated PRs like this one as it's often verbose and over-engineered. It feels that the whole PR could have been one-shotted if not AI-generated (or at least if reviewed before pushing). Anyway, now we are close to something good.

src/huggingface_hub/inference/_mcp/mcp_client.py
```diff
@@ -209,9 +215,18 @@ async def add_mcp_server(self, type: ServerType, **params: Any):
 
         # List available tools
         response = await session.list_tools()
-        logger.debug("Connected to server with tools:", [tool.name for tool in response.tools])
+        all_tool_names = [tool.name for tool in response.tools]
+        logger.debug("Connected to server with tools:", all_tool_names)
```
Wauplin:
```suggestion
        logger.debug("Connected to server with tools:", [tool.name for tool in response.tools])
```

PR #3248, opened 2025-07-22
Fix bad total size after resuming download (#3234)

DKingAlpha:
`python -m pytest tests/test_file_download.py`
```
65 passed in 80.20s (0:01:20)
```

`make quality`
```
  ❌ text_generation: errors found
    Parameter prompt is not documented
    Parameter details is not documented
    ......
```
not related, left untouched.


src/huggingface_hub/file_download.py
```diff
@@ -329,6 +326,15 @@ def _get_file_length_from_http_response(response: requests.Response) -> Optional
...
     """
 
+    # If HTTP response contains compressed body (e.g. gzip), the `Content-Length` header will
+    # contain the length of the compressed body, not the uncompressed file size.
+    # And at the start of transmission there's no way to know the uncompressed file size for gzip,
+    # thus we return None in that case.
+    content_encoding = response.headers.get("Content-Encoding")
+    if (content_encoding is not None) and ("identity" not in content_encoding.lower()):
+        # gzip/br/deflate/zstd etc
+        return None
```
hanouticelina: slight preference for better readability
```suggestion
    content_encoding = response.headers.get("Content-Encoding", "identity").lower()
    if content_encoding != "identity":
        # gzip/br/deflate/zstd etc
        return None
```

tests/test_file_download.py
```diff
@@ -1226,6 +1227,7 @@ def test_large_file_http_path_error(self):
         "hf_transfer not installed, so skipping large file download with hf_transfer check.",
     )
     @patch("huggingface_hub.file_download.constants.HF_HUB_ENABLE_HF_TRANSFER", True)
+    @patch("huggingface_hub.file_download.constants.HF_HUB_DISABLE_XET", False)
```
hanouticelina: this is useful only when running the tests locally, but we should set it to `True` here instead to disable xet path for this specific test and correctly fallback to `http_get`
```suggestion
    @patch("huggingface_hub.file_download.constants.HF_HUB_DISABLE_XET", True)
```

hanouticelina (commented):
Thanks @DKingAlpha, the fix looks good to me! i left two minor comments

DKingAlpha:
Thanks for your CR, that makes more sense.

hanouticelina (approved):
Looks good to me! thanks @DKingAlpha for the contribution 🤗

PR #3251, opened 2025-07-23
[Internal] Fix docstring param spacing check and `libcst` incompatibility with Python 3.13

hanouticelina:
this PR fixes two internal issues:

- `make quality` fails on Python 3.13 because `utils/check_inference_input_params.py` looks for the literal fragment
"            {param_name} (" in every docstring., not 100% sure but maybe it's related to CPython 3.13 stripping the common indentation of docstrings, see: https://docs.python.org/3/whatsnew/3.13.html#other-language-changes. 

- `libcst` (used to update python files) installation failure on (at least) Python 3.13: when setting up the repo locally with newer Python versions (e.g., 3.13), the installation of `libcst` fails.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,7 @@ extras["quality"] = [
     "ruff>=0.9.0",
     "mypy>=1.14.1,<1.15.0; python_version=='3.8'",
     "mypy==1.15.0; python_version>='3.9'",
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for fixing this!

PR #3270, opened 2025-07-28
[Jobs] Update uv image

lhoestq:
the new image does contain `git` which is required to install `uv` dependencies from source, e.g. from https://raw.githubusercontent.com/huggingface/trl/refs/heads/main/trl/scripts/sft.py


Wauplin (approved):
Looks good, thanks

PR #3279, opened 2025-07-31
Fix concurrent create_repo with exist_ok=True

lhoestq:
`create_repo(..., exist_ok=True)` may fail if called concurrently.

If the repo already exists it generally succeeds (server returns 409 and the repo url).
But when called concurrently the server returns 409 but with an error message:
```
{'error': 'A repository with this name already exists'}
```
Since it doesn't contain url , create_repo(..., exist_ok=True) raises an error
```
KeyError: 'url'
```

internal discussion to know whether this should be fixed client side or server side: https://huggingface.slack.com/archives/C02EMARJ65P/p1753959403168719?thread_ts=1753799202.135049&cid=C02EMARJ65P


lhoestq:
will be fixed server side

PR #3281, opened 2025-08-01
Add timeout info to Jobs guide docs

davanstrien:
As discussed in (internal) Slack thread https://huggingface.slack.com/archives/C07RD3RVD5W/p1753978425762419, it might be nice to have some more information on timeout so users know to set longer timeouts for training jobs, etc.

### DIFF
--- a/docs/source/en/guides/jobs.md
+++ b/docs/source/en/guides/jobs.md
@@ -202,7 +202,7 @@ That's it! You're now running code on Hugging Face's infrastructure.
 
 ## Configure Job Timeout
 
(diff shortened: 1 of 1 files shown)


davanstrien:
cc @apolinario

apolinario:
LGTM! Should we also include what the default timeout actually is?

julien-c (commented):
i think the default is 30 mins no?

davanstrien:
> i think the default is 30 mins no?

IUC yes, but don't know if this is likely to change. Maybe @azr @christophe-rannou can confirm?

PR #3303, opened 2025-08-13
[Jobs] Support commands in `hf jobs uv run`

lhoestq:
support `hf jobs uv run command` in addition to `hf jobs uv run script.py` as in the official `uv`

this is useful to run cli commands from python packages, e.g.

```bash
hf jobs uv run --with lighteval -s HF_TOKEN lighteval endpoint inference-providers "model_name=openai/gpt-oss-20b,provider=groq" "lighteval|gsm8k|0|0"
```

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -10286,10 +10286,10 @@ class HfApi:
 
         Args:
             script (`str`):
-                Path or URL of the UV script.
+                Path or URL of the UV script, or a command.
 
             script_args (`List[str]`, *optional*)
-                Arguments to pass to the script.
+                Arguments to pass to the script or command.
 
             dependencies (`List[str]`, *optional*)
                 Dependencies to use to run the UV script.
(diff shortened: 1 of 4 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -10346,8 +10367,8 @@ def run_uv_job(
         if namespace is None:
             namespace = self.whoami(token=token)["name"]
 
-        if script.startswith("http://") or script.startswith("https://"):
-            # Direct URL execution - no upload needed
+        if script.startswith("http://") or script.startswith("https://") or not script.endswith(".py"):
+            # Direct URL execution or command - no upload needed
```
hanouticelina: maybe more robust to check for the presence of a file here because the script could be any file type, like a `.sh` file, not just one ending in `.py`.
```suggestion
        is_url = script.startswith("http://") or script.startswith("https://")
        if is_url or not os.path.isfile(script):
            # Direct URL execution or command - no upload needed
```

hanouticelina (commented):
Thank you! I left one comment, otherwise everything else looks good to me

lhoestq:
let's merge this one ? (needs one approval)

hanouticelina (approved):
Thank you!

PR #3315, opened 2025-08-26
Fix infinite retry loop on HTTP 403 in upload_large_folder

nyuuzyou:
Fixes infinite retry loop during `upload_large_folder` uploads. In my case, the issue occurred with "Your storage patterns tripped our internal systems" HTTP 403 error.

### DIFF
--- a/src/huggingface_hub/_upload_large_folder.py
+++ b/src/huggingface_hub/_upload_large_folder.py
@@ -31,6 +31,7 @@ from . import constants
 from ._commit_api import CommitOperationAdd, UploadInfo, _fetch_upload_modes
 from ._local_folder import LocalUploadFileMetadata, LocalUploadFilePaths, get_local_upload_paths, read_upload_metadata
 from .constants import DEFAULT_REVISION, REPO_TYPES
+from .errors import HfHubHTTPError
 from .utils import DEFAULT_IGNORE_PATTERNS, filter_repo_objects, tqdm
 from .utils._cache_manager import _format_size
 from .utils._runtime import is_xet_available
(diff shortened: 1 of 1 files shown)


Wauplin:
Hi @nyuuzyou, thanks for looking into it. I understand the infinite loop is a problem however I don't like the idea of adding `should_ignore=True` to it. I'm fine to have a system to stop the process if an infinite loop is detected but simply ignoring the large files is bad practice as it's silencing the error to the end user.

The HTTP 403 issue you are facing usually happens if the repo content is too large, in which case the best workaround is to upgrade to a PRO plan for higher limits.

nyuuzyou:
The error description (storage one) is not hidden from user, but to be
honest I didn't test it on other permanent errors like auth problems.

In my case, this caused a permanent loop that just hashes files infinitely.
This change fixed it for me, but I'm happy to implement a more appropriate
fix or convert this PR to an issue for someone from HF to look into it
properly.


чт, 28 авг. 2025 г., 17:50 Lucain 'notifications at github.com'
***@***.***

Wauplin:
> This change fixed it for me

By "fixed it", do you mean the process has completed but files have not been uploaded to the Hub? Do you know which files have not been uploaded? (their size, the number, file type, etc.) Just to get an idea of the typology of files.

For context, this 403 error has been introduced recently server-side so yes, we want to do something about it but not sure yet what. Moving it to an issue would be good indeed so we can discuss it further (both internally but also with the community).

nyuuzyou:
Sorry for the long delay and my unclear comments earlier. Here's what the problem was:

I was using `hf upload-large-folder` for automated file uploads. When my account hit storage limits (which resulted in an permanent HTTP 403 "storage patterns tripped our internal systems"), upload-large-folder entered an infinite retry loop that kept hashing files endlessly for each retry attempt. With regular `hf upload`, this won't happen. It will fail as expected with an HTTPError when it reaches the 403. The problem is that the large folder upload function gets stuck in the retry logic.

The error occurs at the stage of creating a commit with files, so the files are actually successfully uploaded somewhere on the HF server before the 403 happens. This PR doesn't "fix" the server-side storage limitation. It just stops the infinite retry cycle so the process can end instead of running forever.  The 403 error is still reported to the user, but the tool doesn't get stuck in an endless loop of rehashing the same files.

I'll close this PR and create an issue instead, as the issue format will be more suitable as you mentioned.

PR #3320, opened 2025-08-29
[Internal] fix `ty`check quality 

hanouticelina:


setup.py
```diff
@@ -83,7 +83,7 @@ def get_version() -> str:
         "pytest-xdist",
         "pytest-vcr",  # to mock Inference
         "pytest-asyncio",  # for AsyncInferenceClient
-        "pytest-rerunfailures",  # to rerun flaky tests in CI
+        "pytest-rerunfailures<16.0",  # to rerun flaky tests in CI
```
hanouticelina, line 86: a major release of `pytest-rerunfailures` (https://pypi.org/project/pytest-rerunfailures/16.0/) appears to introduce a regression in how it interacts with xdist/execnet when serializing reports. pinning the version here

Issue #3364, opened 2025-09-15
`set rich_markup_mode=None` to disable rich in all cases https://typer.tiangolo.com/tutorial/commands/help/#rich-markdown-and-markup


hanouticelina:

hanouticelina:
completed by https://github.com/huggingface/huggingface_hub/pull/3365

PR #3365, opened 2025-09-15
Refactor CLI implementation using Typer

### LINKED ISSUE #3363: Refactor CLI implementation using Typer

hanouticelina:
Resolves #3363 and #3364.

This PR migrates the `hf` CLI from built-in `argparse` to [`Typer`](https://typer.tiangolo.com) using idiomatic, function based commands. this significantly reduces boilerplate, making the CLI easier to maintain, less error prone, and faster to extend with new `HfApi` features.

## What changed
- new dependencies: 
   - [`typer`](https://github.com/fastapi/typer), latest version is 0.17.4,  it should be stable enough and it's widely adopted.
   - [click](https://github.com/pallets/click), typer s foundation, pure Python, should be stable too. latest version is 8.2.2.
   - ⚠️ rich has been disabled in all cases in commit [(#3364) disable rich in all cases](https://github.com/huggingface/huggingface_hub/commit/93136efe5f4a3f7fd5c096579259ed77d9e189bc) to keep a consistent UI regardless of having `rich` installed or not.
 
- Main entrypoint: `cli/hf.py` now uses typer.Typer and mounts:
  - Top level single commands: `download`, `upload`, `upload-large-folder`, `env`, `version` (all defined in their respective files).
  - Groups: `auth`, `cache`, `repo` (with `tag` subgroup),` repo-files`, `jobs` (with `scheduled` and `uv` subgroups).
  - LFS helpers are still available but hidden in help: `lfs-enable-largefiles`,` lfs-multipart-upload`.
  - All argparse classes removed and replaced by Typer commands.

```bash                                                                                                                         
> hf --help
Usage: hf [OPTIONS] COMMAND [ARGS]...

  Hugging Face Hub CLI

Options:
  --install-completion  Install completion for the current shell.
  --show-completion     Show completion for the current shell, to copy it or
                        customize the installation.
  --help                Show this message and exit.

Commands:
  auth                 Manage authentication (login, logout, etc.).
  cache                Manage local cache directory.
  download             Download files from the Hub.
  env                  Print information about the environment.
  jobs                 Run and manage Jobs on the Hub.
  repo                 Manage repos on the Hub.
  repo-files           Manage files in a repo on the Hub.
  upload               Upload a file or a folder to the Hub.
  upload-large-folder  Upload a large folder to the Hub.
  version              Print information about the hf version.
```
- `tests/test_cli.py` has been updated accordingly. cf the documentation to write tests for typer cli: https://typer.tiangolo.com/tutorial/testing/#about-the-appcommand-decorator.

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -24,24 +24,24 @@ Once installed, you can check that the CLI is correctly setup:
 
 ```
 >>> hf --help
-usage: hf <command> [<args>]
+Usage: hf [OPTIONS] COMMAND [ARGS]...
 
-positional arguments:
-  {auth,cache,download,repo,repo-files,upload,upload-large-folder,env,version,lfs-enable-largefiles,lfs-multipart-upload}
-                        hf command helpers
-    auth                Manage authentication (login, logout, etc.).
-    cache               Manage local cache directory.
-    download            Download files from the Hub
-    repo                Manage repos on the Hub.
-    repo-files          Manage files in a repo on the Hub.
-    upload              Upload a file or a folder to the Hub. Recommended for single-commit uploads.
-    upload-large-folder
-                        Upload a large folder to the Hub. Recommended for resumable uploads.
-    env                 Print information about the environment.
-    version             Print information about the hf version.
+  Hugging Face Hub CLI
 
-options:
-  -h, --help            show this help message and exit
+Options:
+  --help  Show this message and exit.
+
+Commands:
+  download             Download files from the Hub.
+  upload               Upload a file or a folder to the Hub.
+  upload-large-folder  Upload a large folder to the Hub.
+  env                  Print information about the environment.
+  version              Print information about the hf version.
+  auth                 Manage authentication (login, logout, etc.).
+  cache                Manage local cache directory.
+  repo                 Manage repos on the Hub.
+  repo-files           Manage files in a repo on the Hub.
+  jobs                 Run and manage Jobs on the Hub.
 ```
 
 If the CLI is correctly installed, you should see a list of all the options available in the CLI. If you get an error message such as `command not found: hf`, please refer to the [Installation](../installation) guide.
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -29,7 +29,10 @@ Usage: hf [OPTIONS] COMMAND [ARGS]...
   Hugging Face Hub CLI
 
 Options:
-  --help  Show this message and exit.
+  --install-completion  Install completion for the current shell.
+  --show-completion     Show completion for the current shell, to copy it or
+                        customize the installation.
+  --help                Show this message and exit.
 
 Commands:
   download             Download files from the Hub.
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ extras = {}
 
 extras["cli"] = [
     "InquirerPy==0.3.4",  # Note: installs `prompt-toolkit` in the background
+    "typer",
 ]
 
 extras["inference"] = [
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ extras = {}
 
 extras["cli"] = [
     "InquirerPy==0.3.4",  # Note: installs `prompt-toolkit` in the background
+    "shellingham",
 ]
 
 extras["inference"] = [
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ install_requires = [
     "pyyaml>=5.1",
     "httpx>=0.23.0, <1",
     "tqdm>=4.42.1",
+    "typer-slim",
     "typing-extensions>=3.7.4.3",  # to be able to import TypeAlias
 ]
 
@@ -28,7 +29,6 @@ extras = {}
 
 extras["cli"] = [
     "InquirerPy==0.3.4",  # Note: installs `prompt-toolkit` in the background
-    "typer",
 ]
 
 extras["inference"] = [
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -78,7 +78,7 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
 def typer_factory(help: str) -> typer.Typer:
     return typer.Typer(
         help=help,
-        add_completion=False,
+        add_completion=True,
         rich_markup_mode=None,
         no_args_is_help=True,
     )
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -20,6 +20,9 @@ from typing import Annotated, Optional, Union
 import click
 import typer
 
+from huggingface_hub import __version__
+from huggingface_hub.hf_api import HfApi
+
 
 class ANSI:
     """
@@ -137,3 +140,7 @@ RevisionOpt = Annotated[
         help="Git revision id which can be a branch name, a tag, or a commit hash.",
     ),
 ]
+
+
+def get_hf_api(token: Optional[str] = None) -> HfApi:
+    return HfApi(token=token, library_name="hf", library_version=__version__)
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -20,7 +20,7 @@ from typing import Union
 import typer
 
 
-class RepoType(str, Enum):
+class RepoTypeOpt(str, Enum):
     model = "model"
     dataset = "dataset"
     space = "space"
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -82,5 +82,5 @@ def validate_repo_type(value: Optional[str], param_name: str = "repo_type") -> O
         return value
     raise typer.BadParameter(
         "Invalid value for '--repo-type': must be one of 'model', 'dataset', 'space'.",
-        param_name=param_name,
+        param_hint=param_name,
     )
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -14,9 +14,14 @@
 """Contains CLI utilities (styling, helpers)."""
 
 import os
-from typing import Optional, Union
+from enum import Enum
+from typing import Union
 
-import typer
+
+class RepoType(str, Enum):
+    model = "model"
+    dataset = "dataset"
+    space = "space"
 
 
 class ANSI:
@@ -69,18 +74,3 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
     for row in rows:
         lines.append(row_format.format(*row))
     return "\n".join(lines)
-
-
-def validate_repo_type(value: Optional[str], param_name: str = "repo_type") -> Optional[str]:
-    """Validate repo type is one of model|dataset|space when provided.
-
-    Returns the value if valid or None. Raises a Typer BadParameter otherwise.
-    """
-    if value is None:
-        return None
-    if value in ("model", "dataset", "space"):
-        return value
-    raise typer.BadParameter(
-        "Invalid value for '--repo-type': must be one of 'model', 'dataset', 'space'.",
-        param_hint=param_name,
-    )
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -17,6 +17,8 @@ import os
 from enum import Enum
 from typing import Union
 
+import typer
+
 
 class RepoType(str, Enum):
     model = "model"
@@ -74,3 +76,12 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
     for row in rows:
         lines.append(row_format.format(*row))
     return "\n".join(lines)
+
+
+def typer_factory(help: str) -> typer.Typer:
+    return typer.Typer(
+        help=help,
+        add_completion=False,
+        rich_markup_mode=None,
+        no_args_is_help=True,
+    )
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -20,7 +20,7 @@ from typing import Union
 import typer
 
 
-class RepoTypeOpt(str, Enum):
+class RepoType(str, Enum):
     model = "model"
     dataset = "dataset"
     space = "space"
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -11,10 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Contains a utility for good-looking prints."""
+"""Contains CLI utilities (styling, helpers)."""
 
 import os
-from typing import Union
+from typing import Optional, Union
+
+import typer
 
 
 class ANSI:
@@ -67,3 +69,18 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
     for row in rows:
         lines.append(row_format.format(*row))
     return "\n".join(lines)
+
+
+def validate_repo_type(value: Optional[str], param_name: str = "repo_type") -> Optional[str]:
+    """Validate repo type is one of model|dataset|space when provided.
+
+    Returns the value if valid or None. Raises a Typer BadParameter otherwise.
+    """
+    if value is None:
+        return None
+    if value in ("model", "dataset", "space"):
+        return value
+    raise typer.BadParameter(
+        "Invalid value for '--repo-type': must be one of 'model', 'dataset', 'space'.",
+        param_name=param_name,
+    )
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -15,15 +15,9 @@
 
 import os
 from enum import Enum
-from typing import Union
+from typing import Annotated, Optional, Union
 
 import typer
-
-
-class RepoType(str, Enum):
-    model = "model"
-    dataset = "dataset"
-    space = "space"
 
 
 class ANSI:
@@ -78,6 +72,9 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
     return "\n".join(lines)
 
 
+#### TYPER UTILS
+
+
 def typer_factory(help: str) -> typer.Typer:
     return typer.Typer(
         help=help,
@@ -85,3 +82,46 @@ def typer_factory(help: str) -> typer.Typer:
         rich_markup_mode=None,
         no_args_is_help=True,
     )
+
+
+class RepoType(str, Enum):
+    model = "model"
+    dataset = "dataset"
+    space = "space"
+
+
+RepoIdArg = Annotated[
+    str,
+    typer.Argument(
+        help="The ID of the repo (e.g. `username/repo-name`).",
+    ),
+]
+
+
+RepoTypeOpt = Annotated[
+    RepoType,
+    typer.Option(
+        help="The type of repository (model, dataset, or space).",
+    ),
+]
+
+TokenOpt = Annotated[
+    Optional[str],
+    typer.Option(
+        help="A User Access Token generated from https://huggingface.co/settings/tokens.",
+    ),
+]
+
+PrivateOpt = Annotated[
+    bool,
+    typer.Option(
+        help="Whether to create a private repo if repo doesn't exist on the Hub. Ignored if the repo already exists.",
+    ),
+]
+
+RevisionOpt = Annotated[
+    Optional[str],
+    typer.Option(
+        help="Git revision id which can be a branch name, a tag, or a commit hash.",
+    ),
+]
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -17,6 +17,7 @@ import os
 from enum import Enum
 from typing import Annotated, Optional, Union
 
+import click
 import typer
 
 
@@ -75,12 +76,23 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
 #### TYPER UTILS
 
 
+class AlphabeticalMixedGroup(typer.core.TyperGroup):
+    """
+    Typer Group that lists commands and sub-apps mixed and alphabetically.
+    """
+
+    def list_commands(self, ctx: click.Context) -> list[str]:  # type: ignore[name-defined]
+        # click.Group stores both commands and sub-groups in `self.commands`
+        return sorted(self.commands.keys())
+
+
 def typer_factory(help: str) -> typer.Typer:
     return typer.Typer(
         help=help,
         add_completion=True,
         rich_markup_mode=None,
         no_args_is_help=True,
+        cls=AlphabeticalMixedGroup,
     )
 
 
--- a/src/huggingface_hub/cli/auth.py
+++ b/src/huggingface_hub/cli/auth.py
@@ -55,14 +55,14 @@ except ImportError:
     _inquirer_py_available = False
 
 
-auth_app = typer.Typer(help="Manage authentication (login, logout, etc.)", rich_markup_mode=None)
+auth_cli = typer.Typer(help="Manage authentication (login, logout, etc.)", rich_markup_mode=None)
 
 
 def _api() -> HfApi:
     return HfApi()
 
 
-@auth_app.command("login", help="Login using a token from huggingface.co/settings/tokens")
+@auth_cli.command("login", help="Login using a token from huggingface.co/settings/tokens")
 def auth_login(
     token: Annotated[
         Optional[str],
@@ -81,10 +81,7 @@ def auth_login(
     login(token=token, add_to_git_credential=add_to_git_credential)
 
 
-@auth_app.command(
-    "logout",
-    help="Logout from a specific token",
-)
+@auth_cli.command("logout", help="Logout from a specific token")
 def auth_logout(
     token_name: Annotated[
         Optional[str],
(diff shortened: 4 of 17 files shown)


src/huggingface_hub/cli/auth.py
```diff
@@ -54,125 +54,53 @@
...
-        token_names = list(get_stored_tokens().keys())
-
-        if not token_names:
-            logger.error("No stored tokens found. Please login first.")
-            return None
+auth_app = typer.Typer(help="Manage authentication (login, logout, etc.)", rich_markup_mode=None)
+
+
+def _api() -> HfApi:
+    return HfApi()
```
Wauplin: not really relevant, let's do a `whoami()` call directly

src/huggingface_hub/cli/cache.py
```diff
@@ -54,122 +62,93 @@ def _inner(*args, **kwargs):
...
-            self._run_scan()
-        elif self.cache_command == "delete":
-            self._run_delete()
+@cache_app.command("scan", help="Scan the cache directory")
+def cache_scan(
+    dir: Optional[str] = typer.Option(
+        None,
+        "--dir",
+        help="Cache directory to scan (defaults to Hugging Face cache).",
+    ),
```
Wauplin:
```suggestion
    dir: Annotated[Optional[str], typer.Option(help="Cache directory to scan (defaults to Hugging Face cache).")] = None
```

from what I understand in https://typer.tiangolo.com/tutorial/options/help/, the recommended way to use Typer is to define options with an `Annotated` annotation.

image


Also, providing `--dir` for the `dir` argument is redundant (and therefore can be avoided)

(same comment on all commands)

src/huggingface_hub/cli/cache.py
```diff
@@ -54,122 +62,93 @@ def _inner(*args, **kwargs):
...
+    _run_scan(cache_dir=dir, verbosity=verbose)
+
+
+def _run_scan(cache_dir: Optional[str], verbosity: int) -> None:
+    try:
+        t0 = time.time()
+        hf_cache_info = scan_cache_dir(cache_dir)
+        t1 = time.time()
+    except CacheNotFound as exc:
+        cache_dir = exc.cache_dir
```
Wauplin: I would avoid defining a new internal alias here. A good thing with Typer is that we shouldn't need to define twice some arguments type annotations (and therefore reduce risks of conflicts)

```suggestion
        help="Increase verbosity (-v, -vv, -vvv).",
    ),
) -> None:
    try:
        t0 = time.time()
        hf_cache_info = scan_cache_dir(cache_dir)
        t1 = time.time()
    except CacheNotFound as exc:
        cache_dir = exc.cache_dir
```

Wauplin (commented):
Very excited about this switch! I left some very high level comments on code organization :)

setup.py
```diff
@@ -28,6 +28,7 @@ def get_version() -> str:
 
 extras["cli"] = [
     "InquirerPy==0.3.4",  # Note: installs `prompt-toolkit` in the background
+    "typer",
```
Wauplin: I would move `typer` to a required dependency otherwise doing `pip install huggingface_hub` would install a CLI but unusable (since typer would be missing).

To reduce dependency overhead, especially the `rich` package, let's add `typer-slim` which is a lighter version without rich / shellingham included. More details here: https://typer.tiangolo.com/?h=typer+slim#typer-slim

Wauplin: About `shellingham` I still think it's good to have it. Either by default or in the (soon-to-come?) "hf installer script". It's a pure-python package that helps detecting the current shell to install shell completion correctly.


(for now let's skip it but mentioning it to have it in mind)

hanouticelina: agree! it's an oversight from me, fixed in https://github.com/huggingface/huggingface_hub/pull/3365/commits/8ee689288a93ebc778dcc7ea8dd2a15658e07a75

src/huggingface_hub/cli/cache.py
hanouticelina: yes! addressed in https://github.com/huggingface/huggingface_hub/pull/3365/commits/4437b4c860d322d16fc53c4ed50a9660693a850d

src/huggingface_hub/cli/cache.py
hanouticelina: agree, addressed in  https://github.com/huggingface/huggingface_hub/pull/3365/commits/e5fff141a9f10d5f84c2eeaf15c01eb098a2b257

src/huggingface_hub/cli/auth.py
hanouticelina: addressed in [53a4343](https://github.com/huggingface/huggingface_hub/pull/3365/commits/53a43436f2d17e14d8817e69e5817258545e80aa)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -67,3 +69,18 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
...
+    Returns the value if valid or None. Raises a Typer BadParameter otherwise.
+    """
+    if value is None:
+        return None
+    if value in ("model", "dataset", "space"):
+        return value
+    raise typer.BadParameter(
+        "Invalid value for '--repo-type': must be one of 'model', 'dataset', 'space'.",
+        param_hint=param_name,
+    )
```
Wauplin: I feel that this should be done in the parameter validation directly and not inside the method. From what I understand from https://github.com/fastapi/typer/issues/76 and https://typer.tiangolo.com/tutorial/parameter-types/enum/ the best would be to have a `RepoType` enum only for CLI definition and do something like this in commands:

```py
    repo_type: Annotated[
        RepoType,
        typer.Option(
            help="Type of repo to download from.",
        ),
    ] = RepoType.model,
```

Hopefully this should play nicely with  `--help` and autocompletion (compared to `str` + validation with a utils)

Wauplin: (too bad that `typing.Literal` is not already supported but no big deal)

src/huggingface_hub/cli/auth.py
```diff
@@ -30,13 +30,14 @@
     hf auth whoami
 """
 
-from argparse import _SubParsersAction
 from typing import Optional
 
-from huggingface_hub.commands import BaseHuggingfaceCLICommand
+import typer
+from typing_extensions import Annotated
```
Wauplin: [`typing.Annotated`](https://docs.python.org/3/library/typing.html#typing.Annotated) has been added in Python 3.9 which will be the minimal version in 1.x so no need to import from `typing_extensions`

src/huggingface_hub/cli/auth.py
```diff
@@ -54,125 +55,49 @@
...
+    token: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Token from https://huggingface.co/settings/tokens",
+        ),
+    ] = None,
+    add_to_git_credential: Annotated[
+        bool,
+        typer.Option(
+            help="Save to git credential helper",
```
Wauplin:
```suggestion
            help="Save to git credential helper. Useful only if you plan to run git commands directly.",
```

src/huggingface_hub/cli/auth.py
```diff
@@ -183,30 +108,70 @@ def _select_token_name_tui(self, token_names: list[str]) -> Optional[str]:
...
+    token_name: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Name of the token to switch to",
+        ),
+    ] = None,
+    add_to_git_credential: Annotated[
+        bool,
+        typer.Option(
+            help="Save to git credential",
```
Wauplin:
```suggestion
            help="Save to git credential helper. Useful only if you plan to run git commands directly.",
```

src/huggingface_hub/cli/system.py
```diff
@@ -18,35 +18,16 @@
...
 
-    @staticmethod
-    def register_subcommand(parser: _SubParsersAction):
-        version_parser = parser.add_parser("version", help="Print information about the hf version.")
-        version_parser.set_defaults(func=VersionCommand)
 
-    def run(self) -> None:
-        print(f"huggingface_hub version: {__version__}")
+def version() -> None:
+    """Print information about the hf version."""
```
Wauplin:
```suggestion
    """Print CLI version."""
```

src/huggingface_hub/cli/hf.py
```diff
@@ -12,51 +12,68 @@
...
-    args = parser.parse_args()
-    if not hasattr(args, "func"):
-        parser.print_help()
-        exit(1)
-
-    # Run
-    service = args.func(args)
-    if service is not None:
-        service.run()
+    app()
```
Wauplin, line 55:
```suggestion
    logging.set_verbosity_info()
    app()
```

I would set verbosity to INFO globally in the CLI. I don't think there is much drawback to it. And it avoids to have to define it in each command (which is more error-prone / less consistent)

Wauplin: (Note: need to import `logging`)

src/huggingface_hub/cli/hf.py
```diff
@@ -12,51 +12,68 @@
...
+    hidden=True,
+)(lfs_multipart_upload)
+
+
+# command groups
+app.add_typer(auth_cli, name="auth")
+app.add_typer(cache_cli, name="cache")
+app.add_typer(repo_cli, name="repo")
+app.add_typer(repo_files_cli, name="repo-files")
+app.add_typer(jobs_cli, name="jobs")
```
Wauplin: IMO it would be best to add commands and groups in alphabetical order (to have `hf auth` first, etc.). It would make `hf --help` easier to read.

My reasoning is that in practice users don't care if a command has been implemented as a "group" or not.

src/huggingface_hub/cli/auth.py
```diff
@@ -54,125 +55,49 @@
...
+        ),
+    ] = None,
+    add_to_git_credential: Annotated[
+        bool,
+        typer.Option(
+            help="Save to git credential helper",
+        ),
+    ] = False,
+) -> None:
+    logging.set_verbosity_info()
```
Wauplin: see other comment about setting this globally

src/huggingface_hub/cli/cache.py
```diff
@@ -40,6 +41,15 @@
 _CANCEL_DELETION_STR = "CANCEL_DELETION"
 
 
+def _validate_sort_option(sort: Optional[str]) -> Optional[SortingOption_T]:
+    if sort is None:
+        return None
+    options = get_args(SortingOption_T)
+    if sort not in options:
+        raise typer.BadParameter(f"Invalid sort option: {sort}", param_hint="sort")
+    return cast(SortingOption_T, sort)
```
Wauplin: Same here, let's use Enum if that's what Typer supports

src/huggingface_hub/cli/cache.py
```diff
@@ -54,122 +64,98 @@ def _inner(*args, **kwargs):
...
+            count=True,
+            help="Increase verbosity (-v, -vv, -vvv).",
+        ),
+    ] = 0,
+) -> None:
+    _run_scan(cache_dir=dir, verbosity=verbose)
+
+
+def _run_scan(cache_dir: Optional[str], verbosity: int) -> None:
+    try:
```
Wauplin:
```suggestion
    ] = 0,
) -> None:
    try:
```

no need to define an internal alias if used only once

src/huggingface_hub/cli/cache.py
```diff
@@ -54,122 +64,98 @@ def _inner(*args, **kwargs):
...
+    ] = False,
+    sort: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Sort repositories by the specified criteria. Options: 'alphabetical' (A-Z), 'lastUpdated' (newest first), 'lastUsed' (most recent first), 'size' (largest first).",
+            case_sensitive=False,
+        ),
+    ] = None,
+) -> None:
+    sort_by = _validate_sort_option(sort)
```
Wauplin:
```suggestion
```

to remove if validated by Enum

src/huggingface_hub/cli/download.py
```diff
@@ -37,145 +37,153 @@
 """
 
 import warnings
-from argparse import Namespace, _SubParsersAction
 from typing import Optional
 
+import typer
+from typing_extensions import Annotated
```
Wauplin:
```suggestion
```

to import from `typing`

src/huggingface_hub/cli/jobs.py
```diff
@@ -28,1072 +28,1051 @@
...
 import os
 import re
-from argparse import Namespace, _SubParsersAction
 from dataclasses import asdict
 from pathlib import Path
-from typing import Optional, Union
+from typing import Dict, Optional, Union
+
+import typer
+from typing_extensions import Annotated
```
```suggestion
```

src/huggingface_hub/cli/jobs.py
```diff
@@ -28,1072 +28,1051 @@
...
-
-
-class RunCommand(BaseHuggingfaceCLICommand):
-    @staticmethod
-    def register_subcommand(parser: _SubParsersAction) -> None:
-        run_parser = parser.add_parser("run", help="Run a Job")
-        run_parser.add_argument("image", type=str, help="The Docker image to use.")
-        run_parser.add_argument("-e", "--env", action="append", help="Set environment variables. E.g. --env ENV=value")
-        run_parser.add_argument(
+jobs_cli = typer.Typer(help="Run and manage Jobs on the Hub.", rich_markup_mode=None)
```
```suggestion
jobs_cli = typer.Typer(help="Run and manage Jobs on the Hub.", rich_markup_mode=None, no_args_is_help=True)
```

same for other groups as well. Maybe worth to have a utility like this

```py
def typer_factory(help: str) -> typer.Typer:
    return typer.Typer(
        help=help,
        add_completion=False,
        rich_markup_mode=None,
        no_args_is_help=True
    )
```

to be sure all subgroups are configured the same

src/huggingface_hub/cli/lfs.py
```diff
@@ -33,58 +34,39 @@
...
-        )
-        print("Local repo set up for largefiles")
+def lfs_enable_largefiles(
+    path: Annotated[
+        str,
+        typer.Argument(
+            ...,
+            help="Local path to repository you want to configure.",
+        ),
+    ],
```
Wauplin:
```suggestion
    path: Annotated[
        str,
        typer.Argument(help="Local path to repository you want to configure.")
    ],
```

adding `...` as default value is not needed I think (since `...` is already the default value for arguments). I feel that it's more concise to read without it and understandable 


(not the only occurrence like this one, would be good to do a pass to remove them all)

src/huggingface_hub/cli/lfs.py
```diff
@@ -33,58 +34,39 @@
...
-    @staticmethod
-    def register_subcommand(parser: _SubParsersAction):
-        enable_parser = parser.add_parser("lfs-enable-largefiles", add_help=False)
-        enable_parser.add_argument("path", type=str, help="Local path to repository you want to configure.")
-        enable_parser.set_defaults(func=lambda args: LfsEnableCommand(args))
-
-        # Command will get called by git-lfs, do not call it directly.
-        upload_parser = parser.add_parser(LFS_MULTIPART_UPLOAD_COMMAND, add_help=False)
-        upload_parser.set_defaults(func=lambda args: LfsUploadCommand(args))
+"""
```
Wauplin: Docstring not attached to anything. Would be good to move it either as a method docstring (adapting the text) or as a module docstring

src/huggingface_hub/cli/repo.py
```diff
@@ -21,227 +21,239 @@
...
+        Optional[str],
+        typer.Option(
+            help="set to dataset' or 'space' if creating a dataset or space, default is 'model'.",
+        ),
+    ] = None,
+    space_sdk: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Hugging Face Spaces SDK type. Required when --type is set to 'space'.",
+            rich_help_panel=None,
```
Wauplin: Not needed I believe 

```suggestion
```

src/huggingface_hub/cli/repo.py
```diff
@@ -21,227 +21,239 @@
...
-        tag_delete_parser.add_argument("-y", "--yes", action="store_true", help="Answer Yes to prompts automatically.")
-        tag_delete_parser.add_argument(
-            "--token", type=str, help="A User Access Token generated from https://huggingface.co/settings/tokens."
-        )
-        tag_delete_parser.add_argument(
-            "--repo-type",
-            choices=["model", "dataset", "space"],
-            default="model",
+        ),
+    ] = "model",
```
Wauplin: I feel that since these definitions are always the same we could centralize the main arguments definition (like repo_id, repo_type, token, revision, etc.) and reuse them in all commands. It would reduce LoC to maintain + make sure that the wording is always consistent. Here for example if could be reduced to

```py
@tag_app.command("list", help="List tags for a repo.")
def tag_list(
    repo_id: RepoIdArg,
    token: TokenOpt = None,
    repo_type: RepoTypeOpt = RepoType.model,
) -> None:
```

We can argue it makes the code slightly less self-explicit but I still believe it's best for consistency.

src/huggingface_hub/cli/repo_files.py
```diff
@@ -34,95 +34,89 @@
...
+    commit_description: Annotated[
+        Optional[str],
+        typer.Option(
+            help="The description of the generated commit.",
+        ),
+    ] = None,
+    create_pr: Annotated[
+        bool,
+        typer.Option(
+            "--create-pr",
```
Wauplin:
```suggestion
```

No need to define param name when it's the same as the variable (same for many other places I believe)

src/huggingface_hub/cli/repo_files.py
```diff
@@ -34,95 +34,89 @@
...
             help="A User Access Token generated from https://huggingface.co/settings/tokens",
-        )
-
-        repo_files_parser.set_defaults(func=RepoFilesCommand)
+        ),
+    ] = None,
+) -> None:
+    logging.set_verbosity_info()
+    repo_type = validate_repo_type(repo_type)
+    api = HfApi(token=token, library_name="hf")
```
Wauplin: That's actually a really good idea to track usage of our CLI. I think it'd be good to centralize this (some other `HfApi` instantiations are not defining the library_name) but can be done in a later PR. (we have a subticket about analytics specifically)

Wauplin:
```suggestion
    api = HfApi(token=token, library_name="hf", library_version=__version__)
```

might be nice to add `huggingface_hub` version as well (would make things easier to works with in Kibana even though we already have the info in `hf_hub` user agent)

Wauplin (commented):
Looks promising assuming commands implementations have not been changed^^ 

(I did not review the tests yet)

src/huggingface_hub/cli/hf.py
hanouticelina: yes, i don't know if it's a bug but i was trying yesterday to reorder the commands to have alphabetical order but it didn't work, let me check again

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: true, addressed in [547a436](https://github.com/huggingface/huggingface_hub/pull/3365/commits/547a436ff6101fe176d7bd0b8f4304ed5585ee68)

src/huggingface_hub/cli/auth.py
hanouticelina: addressed in [ab9ac9a](https://github.com/huggingface/huggingface_hub/pull/3365/commits/ab9ac9a18d69e5bb9e658629b39950ac5d6667d0)

src/huggingface_hub/cli/hf.py
hanouticelina: addressed in [9a65442](https://github.com/huggingface/huggingface_hub/pull/3365/commits/9a6544278c5da283fb3559ecbb350bc3277e1040)

src/huggingface_hub/cli/cache.py
hanouticelina: addressed in [7a7190d](https://github.com/huggingface/huggingface_hub/pull/3365/commits/7a7190da67767468989b15a294533545302c3008)

src/huggingface_hub/cli/repo_files.py
hanouticelina: for CLI, i addressed it directly in this PR (in [f2eb79b](https://github.com/huggingface/huggingface_hub/pull/3365/commits/f2eb79baea27456314950a2ecaba78f3a610d4b8)) but yes let's do the same for other HfApi instantiations

src/huggingface_hub/cli/repo.py
hanouticelina: much more consistent indeed! addressed in [148299e](https://github.com/huggingface/huggingface_hub/pull/3365/commits/148299e671d2807f8812125b75635e0eaa7c110a)

src/huggingface_hub/cli/hf.py
hanouticelina: it turns out that even if you move `add_typer`/`command` calls above/below, Typer still prints "commands" first, then "groups", each sorted by name. the "simplest" solution is to add a custom Typer group that mixes commands and groups in a single alphabetical list and pass it into `typer_factory`, see [89f592e](https://github.com/huggingface/huggingface_hub/pull/3365/commits/89f592e0cd178d3791adf8f62f07148b97c91326). wdyt?

hanouticelina:
This PR resolves #3370 in [3e5bc64](https://github.com/huggingface/huggingface_hub/pull/3365/commits/3e5bc6442c2f21b0f743f5ed428ff960c205ed18).


Screenshot 2025-09-17 at 15 00 29

docs/source/en/guides/cli.md
```diff
@@ -24,24 +24,27 @@ Once installed, you can check that the CLI is correctly setup:
...
+  download             Download files from the Hub.
+  upload               Upload a file or a folder to the Hub.
+  upload-large-folder  Upload a large folder to the Hub.
+  env                  Print information about the environment.
+  version              Print information about the hf version.
+  auth                 Manage authentication (login, logout, etc.).
+  cache                Manage local cache directory.
+  repo                 Manage repos on the Hub.
+  repo-files           Manage files in a repo on the Hub.
+  jobs                 Run and manage Jobs on the Hub.
```
Wauplin, line 47:
```suggestion
Commands:
  auth                 Manage authentication (login, logout, etc.).
  cache                Manage local cache directory.
  download             Download files from the Hub.
  env                  Print information about the environment.
  jobs                 Run and manage Jobs on the Hub.
  repo                 Manage repos on the Hub.
  repo-files           Manage files in a repo on the Hub.
  upload               Upload a file or a folder to the Hub.
  upload-large-folder  Upload a large folder to the Hub.
  version              Print information about the hf version.
```

now that you've reordered them :)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -67,3 +71,69 @@ def tabulate(rows: list[list[Union[str, int]]], headers: list[str]) -> str:
     for row in rows:
         lines.append(row_format.format(*row))
     return "\n".join(lines)
+
+
+#### TYPER UTILS
+
+
+class AlphabeticalMixedGroup(typer.core.TyperGroup):
```
Wauplin, line 82: TIL

src/huggingface_hub/cli/hf.py
```diff
@@ -12,51 +12,70 @@
...
+app.command(
+    name="lfs-enable-largefiles",
+    help="Configure your repository to enable upload of files > 5GB.",
+    hidden=True,
+)(lfs_enable_largefiles)
+app.command(
+    name="lfs-multipart-upload",
+    help="Upload large files to the Hub.",
+    hidden=True,
+)(lfs_multipart_upload)
```
```suggestion
app.command(help="Download files from the Hub.")(download)
app.command(help="Upload a file or a folder to the Hub.")(upload)
app.command(help="Upload a large folder to the Hub. Recommended for resumable uploads.")(upload_large_folder)
app.command(help="Print information about the environment.")(env)
app.command(help="Print information about the hf version.")(version)
app.command(help="Configure your repository to enable upload of files > 5GB.", hidden=True)(lfs_enable_largefiles)
app.command(help="Upload large files to the Hub.", hidden=True)(lfs_multipart_upload)
```

Tested locally and it works to use implicit `name` based on variable name (which makes the whole definition much shorted)

src/huggingface_hub/cli/hf.py
Wauplin: Fine for me yes! Thanks for looking into this :)

src/huggingface_hub/cli/jobs.py
```diff
@@ -28,1072 +28,962 @@
...
+    python: Annotated[
+        Optional[str],
+        typer.Option(
+            "-p",
+            "--python",
+            help="The Python interpreter to use for the run environment",
+        ),
+    ] = None,
+) -> None:
+    logging.set_verbosity(logging.INFO)
```
```suggestion
```

src/huggingface_hub/cli/jobs.py
Wauplin, line 1: I feel that in this file we could define the Option/Arguments only once and reuse them (same as for repo id, repo type, token, etc.). Fine to define them in this module directly as it's quite specific. 

All the `image`, `flavor`, `secrets`, `secrets_file`, `env`, `env_file`, etc. parameters are defined 4 times because of uv/non-uv and scheduled/non-scheduled.

src/huggingface_hub/cli/repo.py
```diff
@@ -21,227 +21,163 @@
...
-            print(f"{self.repo_type.capitalize()} {ANSI.bold(self.repo_id)} not found.")
-            exit(1)
-        except RevisionNotFoundError:
-            print(f"Tag {ANSI.bold(self.args.tag)} not found on {ANSI.bold(self.repo_id)}")
-            exit(1)
-        print(f"Tag {ANSI.bold(self.args.tag)} deleted on {ANSI.bold(self.repo_id)}")
+        ),
+    ] = None,
+) -> None:
+    api = HfApi(token=token, library_name="hf", library_version=__version__)
```
Wauplin: `HfApi` definition could be moved to `_cli_utils.py` to reuse library_name/version everywhere (no big deal though)

tests/test_cli.py
```diff
@@ -1,746 +1,840 @@
+from __future__ import annotations
```
Wauplin: needed for something?

Wauplin (approved):
Pre-approving :heavy_check_mark: Feel free to ignore my last comments if annoying :smile: 

Let's get this merge soon! :tada: 

(note: I also tested autocompletion locally which worked like a charm)

src/huggingface_hub/cli/jobs.py
hanouticelina: makes sense! done in [e0d4820](https://github.com/huggingface/huggingface_hub/pull/3365/commits/e0d4820feaee45f9c1a582f2a3f9e3798bed0abc)

src/huggingface_hub/cli/repo.py
hanouticelina: addressed in [e0d4820](https://github.com/huggingface/huggingface_hub/pull/3365/commits/e0d4820feaee45f9c1a582f2a3f9e3798bed0abc)

hanouticelina:
Let's get this merged once the CI is green! 🔥

PR #3375, opened 2025-09-19
update the `sentence_similarity` docstring (#3374)

tolgaakar:
The following line in the docstring of `sentence_similarity` was misleading. This minimal PR updates it to reflect the actual purpose of the function. 

https://github.com/huggingface/huggingface_hub/blob/3b562cada2f031c7b265f9e0dd6ab9f7dad7f86e/src/huggingface_hub/inference/_client.py#L1631

Link to the issue: https://github.com/huggingface/huggingface_hub/issues/3374


tolgaakar:
now both is updated ✅

PR #3385, opened 2025-09-24
Fix package name canonicalization for source distributions and wheels

EmilienM:
Update setup.py to generate PEP 508 compliant filenames with hyphens
instead of underscores for both source distributions and wheels. The
package name is now canonicalized as "huggingface-hub" in both cases.

Changes:
- Set package name to "huggingface-hub" (canonicalized with hyphens)
- Add custom sdist command to override archive filename generation
- Add custom bdist_wheel command to rename wheel files after creation
- Ensures source distribution files are named huggingface-hub-*.tar.gz
- Ensures wheel files are named huggingface-hub-*-py3-none-any.whl

Signed-off-by: Emilien Macchi <emacchi@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,35 @@
 from setuptools import find_packages, setup
+from setuptools.command.sdist import sdist
+from wheel.bdist_wheel import bdist_wheel
+
+
+class custom_sdist(sdist):
+    def make_archive(self, base_name, format, root_dir=None, base_dir=None, owner=None, group=None):
+        """Override to use canonical package name with hyphens instead of underscores"""
+        # Replace underscores with hyphens in the base name
+        canonical_base_name = base_name.replace("huggingface_hub-", "huggingface-hub-")
+        return super().make_archive(canonical_base_name, format, root_dir, base_dir, owner, group)
+
+
+class custom_bdist_wheel(bdist_wheel):
+    def run(self):
+        """Run the wheel building process and rename the file to use canonical naming"""
+        # Run the normal wheel build process
+        super().run()
+
+        # Rename the generated wheel file to use canonical naming
+        import glob
+        import os
+
+        # Find the generated wheel file
+        wheel_files = glob.glob(os.path.join(self.dist_dir, "huggingface_hub-*.whl"))
+        for wheel_file in wheel_files:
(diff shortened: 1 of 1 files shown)


EmilienM:
Hi, I work for Red Hat and I'm part of the team who takes care of packaging the Python projects for our AI portfolio. Please look at this PR as a modest contribution which would help to build compliant source distributions and wheels.

Thank you!

setup.py
```diff
@@ -1,4 +1,35 @@
...
+
+        # Find the generated wheel file
+        wheel_files = glob.glob(os.path.join(self.dist_dir, "huggingface_hub-*.whl"))
+        for wheel_file in wheel_files:
+            if "huggingface_hub-" in wheel_file:
+                # Generate the new canonical filename
+                new_wheel_file = wheel_file.replace("huggingface_hub-", "huggingface-hub-")
+                # Rename the file
+                os.rename(wheel_file, new_wheel_file)
+                print(f"Renamed wheel: {wheel_file} -> {new_wheel_file}")
```
tiran, line 32: This code generates invalid wheel and sdist files. `huggingface_hub-` is the correct prefix.

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode

> As the components of the filename are separated by a dash (-, HYPHEN-MINUS), this character cannot appear within any component. This is handled as follows:
> *  In distribution names, any run of -_. characters (HYPHEN-MINUS, LOW LINE and FULL STOP) should be replaced with _ (LOW LINE), and uppercase characters should be replaced with corresponding lowercase ones. This is equivalent to regular [name normalization](https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization) followed by replacing - with _. Tools consuming wheels must be prepared to accept . (FULL STOP) and uppercase letters, however, as these were allowed by an earlier version of this specification.

https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name

> The file name must be in the form {name}-{version}.tar.gz, where {name} is normalised according to the same rules as for binary distributions [...]

EmilienM: thanks for the review, I'll close the PR.

PR #3393, opened 2025-09-25
Document new HF commands

hanouticelina:
as per the title. follow-up PR to #3384.


Wauplin (commented):
Thanks! While you are at it, can you also give a CLI example in each section of https://huggingface.co/docs/huggingface_hub/guides/repository ? (when applicable)

Wauplin (approved):
Thanks!

Issue #3396, opened 2025-09-28
Tiny-agents uses 0 tools without explicitly setting allowed_tools


Mithil467:
### Describe the bug

When running `tiny-agents` without specifying `allowed_tools`, the agent ends up with zero tools. Instead, the agent should default to having access to all tools, unless the config explicitly provides an `allowed_tools` list. If the list is present but empty, then the agent should correctly return no tools.

### Reproduction

agent.json without explicit allowed_tools, example in [README](https://huggingface.co/docs/huggingface.js/en/tiny-agents/README).

Command:
```sh
tiny-agent run agent.json
````

hanouticelina:
hey @Mithil467,
thanks for the contribution 🤗  it has been merged and released in [huggingface_hub==0.35.3](https://github.com/huggingface/huggingface_hub/releases/tag/v0.35.3).

PR #3397, opened 2025-09-28
[Tiny-Agents] Use all tools unless allowed_tools is set explicitly

Mithil467:
This PR allows the agent to use all tools unless there is an explicit `allowed_tools` list in the configuration file.
Fixes: https://github.com/huggingface/huggingface_hub/issues/3396


hanouticelina (approved):
looks good, thanks for the fix! 🤗

PR #3436, opened 2025-10-10
Add client support for the organization overview endpoint

BastienGimbert:
## Problem
I needed to pull full organization metadata (description, avatar, stats, etc.) for a feature, but the Python client had no helper hitting `/api/organizations/{org}/overview`. 

## Summary
- add an HfApi.get_organization_overview helper hitting /api/organizations/{org}/overview
- surface the extra org metadata (avatar, description, counts, etc.) through the Organization dataclass
- export the new helper at the package root and validate it with a production UserApi test

## Testing
- python -m pytest tests/test_hf_api.py -k organization_overview

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -9703,7 +9703,7 @@ class HfApi:
         Args:
             organization (`str`):
                 Name of the organization to get an overview of.
-            token (Union[bool, str, None], optional):
+            token (`bool` or `str`, *optional*):
                 A valid user access token (string). Defaults to the locally saved token, which is the recommended method
                 for authentication (see https://huggingface.co/docs/huggingface_hub/quick-start#authentication).
                 To disable authentication, pass `False`.


hanouticelina:
@bot /style

src/huggingface_hub/hf_api.py
```diff
@@ -1404,16 +1404,52 @@ class Organization:
             Name of the organization on the Hub (unique).
         fullname (`str`):
             Organization's full name.
+        id (`str`, *optional*):
+            Unique identifier of the organization on the Hub.
```
hanouticelina: not sure the `id` is necessary here since the org name is unique, it should be sufficient on its own. I'd not surface it

src/huggingface_hub/hf_api.py
```diff
@@ -9663,6 +9699,33 @@ def get_user_overview(self, username: str, token: Union[bool, str, None] = None)
...
 
+    @validate_hf_hub_args
+    def get_organization_overview(self, organization: str, token: Union[bool, str, None] = None) -> Organization:
+        """
+        Get an overview of an organization on the Hub.
+
+        Args:
+            organization (`str`):
+                Name of the organization to get an overview of.
+            token (Union[bool, str, None], optional):
```
```suggestion
            token (`Union[bool, str, None]`, *optional*):
```

src/huggingface_hub/hf_api.py
BastienGimbert: Agreed that the slug name serves as the unique identifier np

BastienGimbert:
Hi @hanouticelina,
I removed the id field.
I didn’t change the comment for token because the other comments don’t use ** or ``.
I kept the same style.

There are 106 other lines that only use `token (Union[bool, str, None], optional)`

src/huggingface_hub/hf_api.py
```diff
@@ -9663,6 +9695,33 @@ def get_user_overview(self, username: str, token: Union[bool, str, None] = None)
...
```
```suggestion
            token (`bool` or `str`, *optional*):
```

hanouticelina (approved):
@BastienGimbert thanks for noticing! I updated the parameter docstring in main and pushed [f929cf8](https://github.com/huggingface/huggingface_hub/pull/3436/commits/f929cf84c3a0798b8dce6314aaf1b8765cd40154) to align the PR.

All good on my side, let's wait for a final approval from @Wauplin before merging!

Wauplin (approved):
Nice implementation, thanks! :hugs:

PR #3517, opened 2025-11-03
[Doc]: fix various typos in different files

didier-durand:
Hi,

As per title, fixing some typos: see commit diffs for details.

Cheers,

Didier


didier-durand:
BTW, congrats to the team on v1.0 !! Great library to gain API access to HF models and datasets.

PR #3537, opened 2025-11-07
Update error message for offline mode

fracapuano:
simple snippet to show users how to unset the HF_HUB_OFFLINE (improving accessibility with simple command)

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -81,7 +81,7 @@ def hf_request_event_hook(request: httpx.Request) -> None:
     """
     if constants.HF_HUB_OFFLINE:
         raise OfflineModeIsEnabled(
-            f"Cannot reach {request.url}: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable."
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/utils/_http.py
```diff
@@ -81,7 +81,7 @@ def hf_request_event_hook(request: httpx.Request) -> None:
     """
     if constants.HF_HUB_OFFLINE:
         raise OfflineModeIsEnabled(
-            f"Cannot reach {request.url}: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable."
+            f"Cannot reach {request.url}: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable (`unset HF_HUB_OFFLINE`)."
```
hanouticelina, line 84: `unset HF_HUB_OFFLINE` is unix specific and won't work on Windows. I wouldn't add it since it's not really necessary imo. "unset the HF_HUB_OFFLINE environment variable" is clear enough, and users can adjust the command according to their os 😄

fracapuano: fair, RIP me for not considering windows 😢 Felt like it was an easy win for users but no biggie! feel free to close if you'd like :))

PR #3544, opened 2025-11-12
[Internal] Fix CI

hanouticelina:
this PR should fix the issue in the CI https://github.com/huggingface/huggingface_hub/actions/runs/19236255566/job/55177117568?pr=3541 which is related to skipped tests when torch isn't installed.


hanouticelina:
(the failing test is transient)

PR #3556, opened 2025-11-19
[HfFileSystem] Add top level hffs

lhoestq:
```python
>>> from huggingface_hub import hffs
>>> with hffs.open("datasets/fka/awesome-chatgpt-prompts/prompts.csv", "r") as f:
...     print(f.readline())
...     print(f.readline())
"act","prompt"
"An Ethereum Developer","Imagine you are an experienced Ethereum developer tasked..."
```

### DIFF
--- a/docs/source/cn/guides/hf_file_system.md
+++ b/docs/source/cn/guides/hf_file_system.md
@@ -11,26 +11,25 @@
 ## 使用方法
 
 ```python
->>> from huggingface_hub import HfFileSystem
->>> fs = HfFileSystem()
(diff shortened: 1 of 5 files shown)


Wauplin (approved):
Thanks! Maybe mention it in https://huggingface.co/docs/huggingface_hub/guides/hf_file_system as well? I would make the default snippets use `hffs` and mention that instantiating `HfFileSystem` is possible


Note: failing CI is unrelated (will be fixed in https://huggingface.co/docs/huggingface_hub/guides/hf_file_system)

PR #3570, opened 2025-11-25
Parse rate limit headers for better 429 error messages

hanouticelina:
Part of #3553.
now that the Hub returns rate limit info in HTTP headers, let's parses those headers to show useful error messages when rate limited.

```python
>>> from huggingface_hub.utils._http import parse_ratelimit_headers
>>> import httpx

>>> headers = httpx.Headers({
...     "ratelimit": '"api";r=0;t=55',
...     "ratelimit-policy": '"fixed window";"api";q=500;w=300'
... })
>>> info = parse_ratelimit_headers(headers)
>>> info
RateLimitInfo(remaining=0, reset_in_seconds=55, limit=500, window_seconds=300)
```

Error message now shows:
```
HfHubHTTPError: 429 Too Many Requests for url: https://huggingface.co/api/models/username/reponame.
Retry after 55 seconds (0/500 requests remaining in current 300s window).
```

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -98,7 +98,7 @@ def parse_ratelimit_headers(headers: Mapping[str, str]) -> Optional[RateLimitInf
     55
     ```
     """
-    # Case-insensitive header lookup
+
     ratelimit: Optional[str] = None
     policy: Optional[str] = None
     for key in headers:


tests/test_utils_http.py
```diff
@@ -447,3 +449,81 @@ async def test_raise_on_status_async_non_stream(fake_server: str):
 async def test_raise_on_status_async_stream(fake_server: str):
     async with get_async_session().stream("GET", fake_server) as response:
         _check_raise_status(response)
+
+
+class TestParseRatelimitHeaders:
```
hanouticelina, line 454: tests generated by sonnet 4.5

src/huggingface_hub/utils/_http.py
```diff
@@ -48,6 +49,68 @@
...
+
+
+# Regex patterns for parsing rate limit headers
+# e.g.: "api";r=0;t=55 --> r=0, t=55
+_RATELIMIT_HEADER_REGEX = re.compile(r"r\s*=\s*(\d+)\s*;\s*t\s*=\s*(\d+)")
+# e.g.: "fixed window";"api";q=500;w=300 --> q=500, w=300
+_RATELIMIT_POLICY_REGEX = re.compile(r"q\s*=\s*(\d+).*?w\s*=\s*(\d+)")
+
+
+def parse_ratelimit_headers(headers: httpx.Headers) -> Optional[RateLimitInfo]:
```
Wauplin: I would suggest to make `parse_ratelimit_headers` independent from `httpx` to keep logic separated. 

```suggestion
def parse_ratelimit_headers(headers: typing.Mapping) -> Optional[RateLimitInfo]:
```

Then you "only" need to implement a case-insensitive `get` for key lookup. This is slightly more inefficient than using the built-in implementation from `httpx` but not too problematic IMO. I'd keep it simple like this:

```py
ratelimit, policy = None, None
for key in headers:
    lower_key = key.lower():
    if lower_key == "ratelimit":
        ratelimit = headers[key]
    elif lower_key == "ratelimit-policy"
        policy = headers[key]
```

src/huggingface_hub/utils/_http.py
```diff
@@ -48,6 +49,68 @@
...
+
+    Follows IETF draft: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-09.html
+
+    Example:
+        >>> headers = httpx.Headers({"ratelimit": '"api";r=0;t=55', "ratelimit-policy": '"fixed window";"api";q=500;w=300'})
+        >>> info = parse_ratelimit_headers(headers)
+        >>> info.remaining
+        0
+        >>> info.reset_in_seconds
+        55
```
Wauplin:
```suggestion
        ```python
        >>> headers = {"ratelimit": '"api";r=0;t=55', "ratelimit-policy": '"fixed window";"api";q=500;w=300'}
        >>> info = parse_ratelimit_headers(headers)
        >>> info.remaining
        0
        >>> info.reset_in_seconds
        55
        ```
```

2 suggestions:
- add backquotes around the snippet (more consistent with the codebase)
- let's not use `httpx.Headers` (more on that below)

src/huggingface_hub/utils/_http.py
```diff
@@ -48,6 +49,68 @@
 
 logger = logging.get_logger(__name__)
 
+
+@dataclass(frozen=True)
+class RateLimitInfo:
```
Wauplin, line 54: let's expose `RateLimitInfo` and `parse_ratelimit_headers` in `huggingface_hub.utils` ?

src/huggingface_hub/utils/_http.py
```diff
@@ -619,6 +682,22 @@ def hf_raise_for_status(response: httpx.Response, endpoint_name: Optional[str] =
...
+                if ratelimit_info.limit is not None and ratelimit_info.window_seconds is not None:
+                    message += (
+                        f" ({ratelimit_info.remaining}/{ratelimit_info.limit} requests remaining"
+                        f" in current {ratelimit_info.window_seconds}s window)."
+                    )
+                else:
+                    message += "."
+            else:
+                message = f"\n\n429 Too Many Requests for url: {response.url}."
+            raise _format(HfHubHTTPError, message, response) from e
```
Wauplin, line 728: top!

tests/test_utils_http.py
```diff
@@ -447,3 +449,81 @@ async def test_raise_on_status_async_non_stream(fake_server: str):
...
+    def test_parse_full_headers(self):
+        """Test parsing both ratelimit and ratelimit-policy headers."""
+        headers = httpx.Headers(
+            {
+                "ratelimit": '"api";r=0;t=55',
+                "ratelimit-policy": '"fixed window";"api";q=500;w=300',
+            }
+        )
```
Wauplin:
```suggestion
        headers = {
            "ratelimit": '"api";r=0;t=55',
            "ratelimit-policy": '"fixed window";"api";q=500;w=300',
        }
```

(once made generic)

tests/test_utils_http.py
Wauplin: honestly good test suite, thanks claude

src/huggingface_hub/utils/_http.py
```diff
@@ -48,6 +49,68 @@
...
+
+    remaining: int
+    reset_in_seconds: int
+    limit: Optional[int] = None
+    window_seconds: Optional[int] = None
+
+
+# Regex patterns for parsing rate limit headers
+# e.g.: "api";r=0;t=55 --> r=0, t=55
+_RATELIMIT_HEADER_REGEX = re.compile(r"r\s*=\s*(\d+)\s*;\s*t\s*=\s*(\d+)")
```
Wauplin:
```suggestion
_RATELIMIT_REGEX = re.compile(r'\"(?P<resource_type>\w*)\";r\s*=\s*(?P<r>\d+)\s*;\s*t\s*=\s*(?P<t>\d+)')
```

would be great to capture the resource type as well so that in the error message we can tell what limit is reached limit in the UI:

Image

(name change is just to be consistent with `_RATELIMIT_POLICY_REGEX`)

Wauplin: in practice, from what I can read in https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-09.html, the order of the parameters is not guaranteed + some new parameters can be added to it (like `qu` and `pk`). The parser is currently not robust to this but it's fine for now as we are mostly interested in parsing headers from the Hub.

src/huggingface_hub/utils/_http.py
```diff
@@ -48,6 +49,68 @@
...
+# e.g.: "fixed window";"api";q=500;w=300 --> q=500, w=300
+_RATELIMIT_POLICY_REGEX = re.compile(r"q\s*=\s*(\d+).*?w\s*=\s*(\d+)")
+
+
+def parse_ratelimit_headers(headers: httpx.Headers) -> Optional[RateLimitInfo]:
+    """Parse rate limit information from HTTP response headers.
+
+    Follows IETF draft: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-09.html
```
Wauplin, line 84:
```suggestion
    Follows IETF draft: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-09.html
    Only a subset is implemented.
```

src/huggingface_hub/utils/_http.py
hanouticelina: yes good point! addressed in [a3fe049](https://github.com/huggingface/huggingface_hub/pull/3570/commits/a3fe049d914ba48619a32271fc3d44ce9ad730c9)

PR #3575, opened 2025-11-26
[HfFileSystem] add expand_info arg

lhoestq:
this is needed for the `dlt` integration which require to use `?expand=true` to get lastCommit info

see https://github.com/dlt-hub/dlt/pull/3391

### DIFF
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -123,13 +123,20 @@ class HfFileSystem(fsspec.AbstractFileSystem, metaclass=_Cached):
     >     layer. For better performance and reliability, it's recommended to use `HfApi` methods when possible.
 
     Args:
-        token (`str` or `bool`, *optional*):
+        endpoint (`str`, *optional*):
+                Endpoint of the Hub. Defaults to <https://huggingface.co>.
+        token (`bool` or `str`, *optional*):
             A valid user access token (string). Defaults to the locally saved
             token, which is the recommended method for authentication (see
             https://huggingface.co/docs/huggingface_hub/quick-start#authentication).
             To disable authentication, pass `False`.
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_file_system.py
```diff
@@ -457,9 +459,12 @@ def _ls_tree(
         recursive: bool = False,
         refresh: bool = False,
         revision: Optional[str] = None,
-        expand_info: bool = False,
+        expand_info: Optional[bool] = None,
         maxdepth: Optional[int] = None,
     ):
+        expand_info = (
+            self.expand_info if self.expand_info is not None else (expand_info if expand_info is not None else False)
+        )
```
hanouticelina, line 474: the method parameter should typically override instance defaults no? i mean if a user instantiates `HfFileSystem(expand_info=False)` and later calls `fs._ls_tree(..., expand_info=True)`, the `expand_info=True` will be ignored. is this intentional?
```suggestion
        expand_info = (
            expand_info if expand_info is not None else (self.expand_info if self.expand_info is not None else False)
        )
```

lhoestq: ah yes correct ! thanks

src/huggingface_hub/hf_file_system.py
```diff
@@ -123,13 +123,20 @@ class HfFileSystem(fsspec.AbstractFileSystem, metaclass=_Cached):
     >     layer. For better performance and reliability, it's recommended to use `HfApi` methods when possible.
 
     Args:
-        token (`str` or `bool`, *optional*):
+        endpoint (`str`, *optional*):
```
hanouticelina, line 126: i took the liberty of pushing directly on this branch to update the `HfFileSystem` docstring

PR #3586, opened 2025-12-02
Update papers model

Samoed:
Add to paper model:
- authors_users (User objects of authors)
- ai_summary
- ai_keywords
- organization
- project_page
- github_repo
- github_stars

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -1541,6 +1541,7 @@ class PaperAuthor:
         hidden (`bool`, *optional*):
             Whether the author is hidden on the Hub.
     """
+
     name: str
     user: Optional[User]
     status: Optional[str]
@@ -1623,9 +1624,7 @@ class PaperInfo:
         paper = kwargs.pop("paper", {})
         self.id = kwargs.pop("id", None) or paper.pop("id", None)
         authors = paper.pop("authors", None) or kwargs.pop("authors", None)
-        self.authors = (
-            [PaperAuthor(**author) for author in authors] if authors else None
-        )
+        self.authors = [PaperAuthor(**author) for author in authors] if authors else None
         published_at = paper.pop("publishedAt", None) or kwargs.pop("publishedAt", None)
         self.published_at = parse_datetime(published_at) if published_at else None
         self.title = kwargs.pop("title", None)


src/huggingface_hub/hf_api.py
```diff
@@ -1532,30 +1532,45 @@ class PaperInfo:
     Attributes:
         id (`str`):
             arXiv paper ID.
-        authors (`list[str]`, **optional**):
+        authors (`list[str]`, *optional*):
             Names of paper authors
-        published_at (`datetime`, **optional**):
+        authors_users (`list[User]`, *optional*):
+            Information about paper authors as a list of [`User`] objects.
```
Wauplin: Instead of having `authors` as a list of names and `authors_users` as a list of `User` (only for claimed authorship), I think it'd best to create a new dataclass `PaperAuthor` like this:

```py
@dataclass
class PaperAuthor:
    name: str
    user: Optional[User]
    status: Optional[str]
```

and then update the type in `PaperInfo` with:

```py
    authors: List[PaperAuthor]
```

This would be a breaking change in the API but I do think it's for the best future-wise. We want the SDK to be as close as possible to the API design

Samoed: I agree. I updated as you suggested

src/huggingface_hub/hf_api.py
```diff
@@ -1565,12 +1612,20 @@ class PaperInfo:
...
+    project_page: Optional[str]
+    github_repo: Optional[str]
+    github_stars: Optional[int]
 
     def __init__(self, **kwargs) -> None:
         paper = kwargs.pop("paper", {})
         self.id = kwargs.pop("id", None) or paper.pop("id", None)
-        authors = paper.pop("authors", None) or kwargs.pop("authors", None)
-        self.authors = [author.pop("name", None) for author in authors] if authors else None
+        authors = kwargs.pop("authors", None)
```
```suggestion
        authors = paper.pop("authors", None) or kwargs.pop("authors", None)
```

maybe keep this line?

Wauplin:
@Samoed can you run `make style` locally and commit the changes? Looks like there is a lint issue in the code

Wauplin:
Perfect, thanks! Failing CI is unrelated so let's merge :)

PR #3605, opened 2025-12-09
Alias transformers CLI

Wauplin:
(discussed [internally](https://huggingface.slack.com/archives/C07KX53FZTK/p1759937771290149))

Now that `transformers` is also shipped with a Typer CLI, it would be great to centralize it in the `hf` command. It has a few benefits, especially the installer script and auto-update feature (checks for new versions).

### In practice

1. `hf transformers` is shown in the `hf --help` even if `transformers` is not installed
2. if not installed `hf transformers ...` will fail with an import error inviting user to install it (better for discoverability)
3. `hf transformers` is lazy-loaded meaning that e.g. `hf download ...` will not import transformers even if available (saves some time)

### Example

```
➜  hf transformers version
5.0.0rc0
➜  hf transformers --help 
Usage: hf transformers [OPTIONS] COMMAND [ARGS]...

  Alias for Transformers CLI. Only available if `transformers` is installed.

Options:
  --help  Show this message and exit.

Commands:
  add-fast-image-processor  Add a fast image processor to a model.
  add-new-model-like        Add a new model to the library, based on an...
  chat                      Chat with a model from the command line.
  download                  Download a model and its tokenizer from the Hub.
  env                       Print information about the environment.
  serve                     Run a FastAPI server to serve models...
  version                   Print CLI version.
```

### TODO
- [ ] adapt `install.sh` and `install.ps1` to add support for `--with-transformers` 
- [ ] adapt `install.sh` and `install.ps1` to auto update transformers if already exists
- [ ] auto-detect transformers updates (for now, only `huggingface_hub` ones are checked)
- [ ] gracefully fail if transformers <5.x (should never happen, but you never know^^)
- [ ] install message should depend on the install method (e.g. when installed with the installer, user don't know where to pip install)
- [ ] will it work with brew? :confused: (check that)
- [ ] [CLI reference](https://huggingface.co/docs/huggingface_hub/package_reference/cli) should contain transformers CLI (make sure it's the case in CI)

### DIFF
--- a/docs/source/en/installation.md
+++ b/docs/source/en/installation.md
@@ -113,10 +113,23 @@ On macOS and Linux:
 curl -LsSf https://hf.co/cli/install.sh | bash
 ```
 
+To install with `transformers[serving]`:
+
+```bash
+curl -LsSf https://hf.co/cli/install.sh | bash -s -- --with-transformers
+```
+
 On Windows:
 
 ```powershell
 powershell -ExecutionPolicy ByPass -c "irm https://hf.co/cli/install.ps1 | iex"
+```
+
+To install with `transformers[serving]`, download the script first:
+
+```powershell
+irm https://hf.co/cli/install.ps1 -OutFile install.ps1
+.\install.ps1 -WithTransformers
 ```
 
 ## Install with conda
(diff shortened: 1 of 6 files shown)


Wauplin:
superseded by https://github.com/huggingface/huggingface_hub/pull/3666

PR #3606, opened 2025-12-09
Fix a url in the docs

neo:
current, not anchored to the correct section: https://huggingface.co/docs/huggingface_hub/guides/download#download-files-to-local-folder
fixed, anchored to the section as expected: https://huggingface.co/docs/huggingface_hub/guides/download#download-files-to-a-local-folder


hanouticelina (approved):
thanks @neo for the fix! 🤗

Wauplin (approved):
🫶

PR #3608, opened 2025-12-10
Fix `auth_list` not showing `HF_TOKEN` message when no stored tokens exist

hanouticelina:
Previously, `auth_list`/`hf auth list` would only display the `HF_TOKEN` environment variable note when there were stored tokens in `~/.cache/huggingface/stored_tokens`. If a user had `HF_TOKEN` set but no stored tokens, they would only see:
```bash
No access tokens found.
```
which can be confusing. Now the output correctly indicates:
```bash
No stored access tokens found.
Note: Environment variable `HF_TOKEN` is set and is the current active token.
```

### DIFF
--- a/docs/source/de/guides/download.md
+++ b/docs/source/de/guides/download.md
@@ -169,7 +169,7 @@ Standardmäßig werden Dateien im Cache-Verzeichnis heruntergeladen, das durch d
 ./cache/models--gpt2/snapshots/11c5a3d5811f50298f278a704980280950aedb10/config.json
 ```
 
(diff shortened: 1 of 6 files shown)


src/huggingface_hub/_login.py
```diff
@@ -192,7 +192,11 @@ def auth_list() -> None:
     tokens = get_stored_tokens()
 
     if not tokens:
-        logger.info("No access tokens found.")
+        if _get_token_from_environment():
+            logger.info("No stored access tokens found.")
+            logger.warning("Note: Environment variable `HF_TOKEN` is set and is the current active token.")
+        else:
+            logger.info("No access tokens found.")
```
Wauplin, line 199:
```suggestion
```

feels redundant with first log?

hanouticelina: it's different actually:
"No stored access tokens found" --> the `stored_tokens` file is empty, BUT a token exists (via `HF_TOKEN`)
"No access tokens found." --> no tokens at all (no stored tokens AND no `HF_TOKEN`)

Wauplin: 🤦

hanouticelina:
failing tests unrelated (likely transient)

Issue #3609, opened 2025-12-10
hf download : crash when a metadata file is corrupted and downloading to local-dir


goulou:
### Describe the bug

When downloading a file with a corrupted metadata file present, the `hf download --local-dir XXXX` crashes with the following error :
```
  File "...../venvs/huggingface-cli/lib/python3.12/site-packages/huggingface_hub/_local_folder.py", line 316, in read_download_metadata
    stat.st_mtime - 1 <= metadata.timestamp
                         ^^^^^^^^
UnboundLocalError: cannot access local variable 'metadata' where it is not associated with a value

```
This comes fom :
https://github.com/huggingface/huggingface_hub/blob/d287c8369ca0abec17964c1ca84d14dd4f59c056/src/huggingface_hub/_local_folder.py#L316

This is quite logical as the `metadata` variable is not set in the first try/except, and the second try/except only catches `FileNotFoundError`.
A `return None` should be added on Line 311.

### Logs

```shell

```

### System info

```shell
- huggingface_hub version: 1.2.1
- Python version: 3.12.3
- Running in iPython ?: No
- Running in notebook ?: No
- Running in Google Colab ?: No
- Running in Google Colab Enterprise ?: No
- Has saved token ?: True
- Configured git credential helpers: 
- Installation method: unknown
- httpx: 0.28.1
- hf_xet: 1.2.0
- gradio: N/A
- tensorboard: N/A
- ENDPOINT: https://huggingface.co
- HF_HUB_CACHE: <set>
- HF_ASSETS_CACHE: <set>
- HF_TOKEN_PATH: <set>
- HF_STORED_TOKENS_PATH: <set>
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_TELEMETRY: False
- HF_HUB_DISABLE_PROGRESS_BARS: None
- HF_HUB_DISABLE_SYMLINKS_WARNING: False
- HF_HUB_DISABLE_EXPERIMENTAL_WARNING: False
- HF_HUB_DISABLE_IMPLICIT_TOKEN: False
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
- HF_XET_HIGH_PERFORMANCE: False
```

Wauplin:
[...]

PR #3610, opened 2025-12-10
Fix unbound local error when reading corrupted metadata files

Wauplin:
Fix https://github.com/huggingface/huggingface_hub/issues/3609 cc @goulou.

When uploading or downloading to a local dir, we use some metadata files to avoid re-uploading / re-downloaded data. However we do some sanity checks on the metadata file to verify it is not outdated. If any exception occur while reading the metadata, we prefer to ignore the metadata (and therefore restart the process from scratch). However in practice, a `UnboundLocalError` exception is raised in some cases when we try to access the `metadata` variable after an error occurred. Solution is just to return early.


hanouticelina (approved):
thanks!

PR #3614, opened 2025-12-11
[Jobs][CLI] allow unknown options in jobs cli

lhoestq:
fix this error

```bash
>>> hf jobs run python:3.12 python -c 'print("Hello from the cloud!")'
Usage: hf jobs run [OPTIONS] IMAGE COMMAND...
Try 'hf jobs run --help' for help.

Error: No such option: -c
```
and

```bash
>>> hf jobs uv run python -c 'print("Hello from the cloud!")'
Usage: hf jobs uv run [OPTIONS] SCRIPT [SCRIPT_ARGS]...
Try 'hf jobs uv run --help' for help.

Error: No such option: -c
```

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -890,6 +890,8 @@ Running this will show the following output!
 This code ran with the following GPU: NVIDIA A10G
 ```
 
+A `--` can be used to separate the command from jobs options for clarity, e.g., `hf jobs run --flavor a10g-small -- python -c '...'`
+
 That's it! You're now running code on Hugging Face's infrastructure.
 
 ### Common Use Cases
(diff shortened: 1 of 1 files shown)


Wauplin:
We've recently got this PR to add a `--` in all command lines in the documentation (see https://github.com/huggingface/huggingface_hub/pull/3597)

```
➜  ~ hf jobs run python:3.12 -- python -c 'print("Hello from the cloud!")'
Job started with ID: 693ae3bcc67c9f186cfe2314
View at: https://huggingface.co/jobs/Wauplin/693ae3bcc67c9f186cfe2314
Hello from the cloud!
```

I think we should stick with this syntax that feels cleaner and explicitly tells that the rest of the command is an argument. WDYT ?

lhoestq:
if we want to stick to the UV and Docker like CLI experience I think we should support commands without having to add `--` (like UV and Docker)

(note that using `--` still works with this change)

lhoestq:
wdyt ? cc @davanstrien for viz as well

davanstrien:
No strong opinion. I also ran into this a few times. Maybe it's still a bit cleaner to rely on `--` for very long commands so people don't accidentally pass things in the wrong place, i.e., script args vs. job args? 

I found in general, coding agents (at least Claude) seemed to fairly quickly grasp the `--`. 

It could make sense to use `--` in the docs, but we could still add this?

lhoestq:
UV docs mention

> A -- can be used to separate the command from uv options for clarity, e.g., uv run --python 3.12 -- python

and in their examples they mostly use without `--` except when it helps clarifying

> uv run python -c "import example"

> uv run --with httpx==0.26.0 python -c "import httpx; print(httpx.__version__)"

> uv run --with nuitka==2.6.7 -- nuitka --version

Docker on their side never mention `--` in their examples from what I can see, and in stackoverflow threads neither

edit: I added a mention of `--` in the docs like UV

PR #3622, opened 2025-12-15
feat: use http_backoff for LFS batch/verify/completion endpoints

The-Obstacle-Is-The-Way:
## Summary

Use `http_backoff` wrapper for all LFS HTTP operations to enable automatic retry on transient failures.

**Changes:**
- Replace direct `get_session()` calls with `http_backoff()` for POST and PUT operations in LFS upload flow
- Covers: batch info endpoint, verify endpoint, completion endpoint, and upload endpoints
- Removes dependency on `get_session` in lfs.py

**Why:**
- `http_backoff` provides built-in retry logic for transient failures (TimeoutException, NetworkError, 429, 5xx)
- Aligns with existing patterns used elsewhere in the codebase (e.g., downloads)
- Addresses timeout issues for large file uploads via retry mechanism

This change was suggested by @Wauplin in review feedback - using the existing `http_backoff` mechanism rather than adding a new environment variable.

## Test plan
- [x] Unit tests verify `http_backoff` is called for LFS operations (tests/test_lfs.py)
- [x] Pre-commit hooks pass
- [x] CI/CD validates

### DIFF
--- a/docs/source/en/package_reference/environment_variables.md
+++ b/docs/source/en/package_reference/environment_variables.md
@@ -79,7 +79,7 @@ Integer value to define the number of seconds to wait for server response when f
 
 Integer value to define the number of seconds to wait for server response when downloading a file. If the request times out, a TimeoutError is raised. Setting a higher value is beneficial on machine with a slow connection. A smaller value makes the process fail quicker in case of complete network outage. Default to 10s.
 
-## Xet
+## Xet 
 
 ### Other Xet environment variables
 * [`HF_HUB_DISABLE_XET`](../package_reference/environment_variables#hfhubdisablexet)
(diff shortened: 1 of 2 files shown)


The-Obstacle-Is-The-Way:
Hey @hanouticelina, thanks for the feedback!

Good call on using `http_backoff` instead - it's simpler and provides consistent retry behavior without adding config surface. I've updated the PR:

- Removed the `HF_HUB_UPLOAD_TIMEOUT` env var entirely
- Wrapped all LFS HTTP calls in `http_backoff`:
  - `post_lfs_batch_info()` batch endpoint
  - `lfs_upload()` verify step  
  - `_upload_multi_part()` completion step
  - PUT uploads (single-part and multi-part)

Using default timeouts - the retry mechanism should handle transient failures.

PR #3630, opened 2025-12-16
[HfApi] deprecate `direction` in list repos methods

hanouticelina:
The API (internal code [here](https://github.com/huggingface-internal/moon-landing/blob/main/server/lib/SearchBuilder.ts#L543:L556)) does not allow ascending sort (`direction=1`) for `downloads`, `likes`, `trending_score` and `last_modified`. The default direction is -1 (descending) so let's deprecate `direction` parameter.

### DIFF
--- a/.github/workflows/python-prerelease.yml
+++ b/.github/workflows/python-prerelease.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target-repo: ["transformers", "datasets", "diffusers"]
+        target-repo: ["transformers", "datasets", "diffusers", "sentence-transformers"]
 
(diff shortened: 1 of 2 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -1863,6 +1863,7 @@ def get_dataset_tags(self) -> dict:
         hf_raise_for_status(r)
         return r.json()
 
+    @_deprecate_arguments(version="1.5", deprecated_args=["direction"], custom_message="Sorting is always descending.")
```
Wauplin, line 1866: ok for documenting it as removed in 1.5 (in practice we'll see what we decide)

PR #3633, opened 2025-12-18
Add evaluation results module

hanouticelina:
Add Evaluation Results module to support the Hub's new decentralized evaluation results system: https://huggingface.co/docs/hub/eval-results

 This PR introduces:
  - `EvalResultEntry` dataclass representing evaluation scores stored in `.eval_results/*.yaml ` files.
  - `eval_result_entries_to_yaml()` to serialize entries to the YAML format.
  - `parse_eval_result_entries()` to parse YAML data back into `EvalResultEntry` objects.

This lives in a new module, separate from the existing `repocard_data.py` which handles the (legacy?) `model-index` format in `README` metadata. Backward compatibility is maintained for now.

### DIFF
--- a/src/huggingface_hub/_eval_results.py
+++ b/src/huggingface_hub/_eval_results.py
@@ -178,15 +178,16 @@ def parse_eval_result_entries(data: list[dict[str, Any]]) -> list[EvalResultEntr
     """
     entries = []
     for item in data:
-        dataset = item.get("dataset", {})
-        source = item.get("source", {})
+        entry_data = item.get("data", item)
+        dataset = entry_data.get("dataset", {})
+        source = entry_data.get("source", {})
         entry = EvalResultEntry(
             dataset_id=dataset["id"],
-            value=item["value"],
+            value=entry_data["value"],
             task_id=dataset.get("task_id"),
             dataset_revision=dataset.get("revision"),
-            verify_token=item.get("verifyToken"),
-            date=item.get("date"),
+            verify_token=entry_data.get("verifyToken"),
+            date=entry_data.get("date"),
             source_url=source.get("url") if source else None,
             source_name=source.get("name") if source else None,
             source_user=source.get("user") if source else None,
--- a/docs/source/en/package_reference/cli.md
+++ b/docs/source/en/package_reference/cli.md
@@ -1103,7 +1103,7 @@ $ hf models info [OPTIONS] MODEL_ID
 **Options**:
 
 * `--revision TEXT`: Git revision id which can be a branch name, a tag, or a commit hash.
-* `--expand TEXT`: Comma-separated properties to expand. Example: '--expand=downloads,likes,tags'. Valid: author, baseModels, cardData, childrenModelCount, config, createdAt, disabled, downloads, downloadsAllTime, gated, gguf, inference, inferenceProviderMapping, lastModified, library_name, likes, mask_token, model-index, pipeline_tag, private, resourceGroup, safetensors, sha, siblings, spaces, tags, transformersInfo, trendingScore, usedStorage, widgetData.
+* `--expand TEXT`: Comma-separated properties to expand. Example: '--expand=downloads,likes,tags'. Valid: author, baseModels, cardData, childrenModelCount, config, createdAt, disabled, downloads, downloadsAllTime, evalResults, gated, gguf, inference, inferenceProviderMapping, lastModified, library_name, likes, mask_token, model-index, pipeline_tag, private, resourceGroup, safetensors, sha, siblings, spaces, tags, transformersInfo, trendingScore, usedStorage, widgetData.
 * `--token TEXT`: A User Access Token generated from https://huggingface.co/settings/tokens.
 * `--help`: Show this message and exit.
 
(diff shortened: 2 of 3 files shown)


src/huggingface_hub/_eval_results.py
Wauplin, line 1: let's rename it `_eval_results.py` so it's not a public module (and expose its helpers/classes)

src/huggingface_hub/_eval_results.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Example:
+        ```python
+        >>> from huggingface_hub import EvalResultEntry, eval_result_entries_to_yaml
+        >>> entries = [
+        ...     EvalResultEntry(dataset_id="cais/hle", value=20.90, task_id="default"),
+        ...     EvalResultEntry(dataset_id="Idavidrein/gpqa", value=0.412, task_id="gpqa_diamond"),
+        ... ]
+        >>> yaml_data = eval_result_entries_to_yaml(entries)
+        >>> yaml_data[0]
```
Wauplin, line 103: maybe complete the example to save that to a `.eval_results/...` file on the repo on the Hub. I don't think it makes sense to provide a helper dedicated to that but highlighting how to use it would be great

hanouticelina: addressed 👍

src/huggingface_hub/_eval_results.py
```diff
@@ -0,0 +1,222 @@
...
+        verify_token (`str`, *optional*):
+            A signature that can be used to prove that evaluation is provably auditable and reproducible.
+        date (`str`, *optional*):
+            When the evaluation was run (ISO-8601 datetime). Defaults to git commit time.
+        source_url (`str`, *optional*):
+            Link to the evaluation source (e.g., https://huggingface.co/spaces/SaylorTwift/smollm3-mmlu-pro). Required if `source_name` or `source_user` is provided.
+        source_name (`str`, *optional*):
+            Display name for the source. Example: "Eval Logs".
+        source_user (`str`, *optional*):
+            HF user or org name for attribution. Example: "cais".
```
julien-c:
```suggestion
            HF user or org name for attribution. Example: "celinah".
```

it's more the user who runs the eval, not the author of the dataset

src/huggingface_hub/_eval_results.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+            entries.append(entry)
+        else:
+            # https://github.com/huggingface/hub-docs/blob/434609e6d09f7c1203ea59fcc32c7ff4d308a68e/modelcard.md?plain=1#L23 format
+            source = item.get("source", {})
+            for metric in item.get("metrics", []):
+                entry = EvalResultEntry(
+                    dataset_id=dataset["type"],
+                    value=metric["value"],
+                    task_id=dataset.get("config"),
```
julien-c: maybe those should be two different types (legacy and new)? no strong opinion though

src/huggingface_hub/hf_api.py
```diff
@@ -5713,6 +5715,63 @@ def _parse(filename: str) -> None:
                 f"'{repo_id}' is not a safetensors repo. Couldn't find '{constants.SAFETENSORS_INDEX_FILE}' or '{constants.SAFETENSORS_SINGLE_FILE}' files."
             )
 
+    @validate_hf_hub_args
+    def get_eval_results(
```
julien-c: i'm not sure we need this, because this will be exposed by the Hub API

hanouticelina: ok nice, we will have to update `ModelInfo` (and `model_info`) then

src/huggingface_hub/_eval_results.py
julien-c: or just keep using `EvalResult` (the previous type)

on the Hub at least we have two different types

julien-c:
> Added `eval_results` property to `ModelInfo` and updated `expand` docstring in `model_info` and `list_models`. We need the server-side [PR](https://github.com/huggingface-internal/moon-landing/pull/16209) (private) to be merged first.

This is now deployed: https://huggingface.co/api/models/microsoft/Phi-3.5-mini-instruct?expand[]=evalResults&expand[]=model-index @hanouticelina

Wauplin (approved):

Wauplin:
> maybe in the future we'll want a method or snippet to open a PR (or just commit to main if one has rights) to a model repo with a given eval-result, but in any case this PR LGTM!

Copy-pasting example from docstring here for visibility:

```python
import yaml
from huggingface_hub import upload_file, EvalResultEntry, eval_result_entries_to_yaml

entries = [
    EvalResultEntry(dataset_id="cais/hle", value=20.90, task_id="default"),
]
yaml_content = yaml.dump(eval_result_entries_to_yaml(entries))
upload_file(
    path_or_fileobj=yaml_content.encode(),
    path_in_repo=".eval_results/hle.yaml",
    repo_id="your-username/your-model",
    create_pr=True,  # optional
)
```

(agree we can move that snippet to its own method in the future but let's keep as-is for now)

julien-c:
> Copy-pasting example from docstring here for visibility

lol how did i not see the docstring

PR #3635, opened 2025-12-20
Add documentation to avoid/fix HF_HUB_DISABLE_SYMLINKS_WARNING for windows users

bmerkle:
Added a note for Windows users regarding symlink warnings and provided options to resolve the issue.

Updated the README to include information for windows users.
using huggingface_hub yield a warning.

The this note documents ways to fix the problem or to hide the warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,25 @@ pip install "huggingface_hub[mcp]"
 
 To learn more installation and optional dependencies, check out the [installation guide](https://huggingface.co/docs/huggingface_hub/en/installation).
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


bmerkle:
fix for https://github.com/huggingface/huggingface_hub/issues/3636

Wauplin:
As mentioned in https://github.com/huggingface/huggingface_hub/issues/3636#issuecomment-3710185895, I don't think it's necessary to document this behavior further as it's already documented in the docs + logs are supposed to be self-explicit. The README is more of a condensed version of the docs without all the subtleties. I'll therefore close this PR without merging. Let me know if you strongly feel differently.

PR #3646, opened 2025-12-24
Fix pagination URLs when using mirror endpoints

popomore:
When using a mirror endpoint (via HF_ENDPOINT or custom endpoint), pagination links returned by the server point to huggingface.co. This fix ensures all paginated requests use the same endpoint as the initial request by calling fix_hf_endpoint_in_url() on each pagination link.

- Modified paginate() in utils/_pagination.py
- Added test_paginate_with_mirror_endpoint() test case
- All existing tests continue to pass

### DIFF
--- a/src/huggingface_hub/utils/_pagination.py
+++ b/src/huggingface_hub/utils/_pagination.py
@@ -15,10 +15,11 @@
 """Contains utilities to handle pagination on Huggingface Hub."""
 
 from typing import Iterable, Optional
+from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


Wauplin:
Hi @popomore , thanks for the PR. However we won't move forward on it. The `paginate` helper is supposed to be server-agnostic and work on any url returning a list + a `next:` link in headers. I do think that the pagination URLs should be fixed in the mirror endpoints instead since the culprit is in the headers response, not client-side

PR #3653, opened 2025-12-28
Fix: Use self.endpoint in job-related APIs for custom endpoint support

PredictiveManish:
## Summary

Fixes a critical bug where job-related API methods (`run_job`, `create_scheduled_job`, and `fetch_job_logs`) hardcoded the Hugging Face production endpoint (`https://huggingface.co`) instead of using the client-configured `self.endpoint`. This prevented the use of custom endpoints (e.g., for proxies, staging environments, or air-gapped deployments) in job operations. Fixes #3651 .

- This PR fixes #3651 

## Problem

Previously, the client would:
- Ignore the `endpoint` parameter passed to `HfApi()` when invoking job-related APIs
- Always send requests to `https://huggingface.co`, even when a custom endpoint (e.g., `https://staging.huggingface.co` or an internal proxy) was configured
- Fail silently or behave incorrectly in non-production environments, breaking CI, staging, and enterprise deployments

This was particularly problematic for users relying on custom endpoints for testing, compliance, or network isolation.

## Solution

1. **Use `self.endpoint` consistently**: Replace all hardcoded `"https://huggingface.co"` URLs in job-related methods with `self.endpoint`.
2. **Affected methods**: Updated `run_job`, `create_scheduled_job`, and `fetch_job_logs` to construct URLs using the client’s configured endpoint.
3. **Preserve existing behavior**: No change for users using the default endpoint; full backward compatibility maintained.

## Changes

- Replaced hardcoded `"https://huggingface.co"` with `f"{self.endpoint}/api/..."` in:
  - `run_job`
  - `create_scheduled_job`
  - `fetch_job_logs` (including internal status-check call)
- No changes to method signatures or public API behavior
- All other job methods already correctly used `self.endpoint` (no change needed)

## Impact

- **Non-breaking**: Fully backward compatible; only fixes incorrect endpoint usage
- **Positive**: Enables reliable use of `HfApi` with custom endpoints for job management
- **Enterprise/CI readiness**: Supports air-gapped, proxy, and staging deployments

Issue #3651, opened 2025-12-27
Allow custom endpoints for jobs APIs


rezass:
**Is your feature request related to a problem? Please describe.**
Currently creating jobs and fetching logs (two APIs that have streaming output) do not use `self.endpoint`, so the client cannot be used for custom proxies for job related tasks.

**Describe the solution you'd like**
Simply use `self.endpoint`instead of hardcoded `https://huggingface.co`.

Wauplin:
Thanks for flagging!

PR #3655, opened 2025-12-30
[Jobs] Monitor cpu, memory, network and gpu (if any)

lhoestq:
Add the `hf jobs stats [job-ids]...` command which is the equivalent of `docker stats [container-ids]...`.
It displays a live table that is useful to monitor cpu, memory, network and gpu (if any).

Moreover `hf jobs stats` shows the table for all the running jobs, without having to pass the job ids

For example:

```bash
>>> hf jobs stats
JOB ID                   CPU % NUM CPU MEM % MEM USAGE      NET I/O         GPU UTIL % GPU MEM % GPU MEM USAGE 
------------------------ ----- ------- ----- -------------- --------------- ---------- --------- --------------- 
6953ff6274100871415c13fd 0%    3.5     0.01% 1.3MB / 15.0GB 0.0bps / 0.0bps 0%         0.0%      0.0B / 22.8GB
```

I also added its associated `HfApi` method `fetch_jobs_metrics`:

```python
>>> for metrics in fetch_job_metrics(job_id="6953ff6274100871415c13fd"):
...     print(metrics)
{
    "cpu_usage_pct": 0,
    "cpu_millicores": 3500,
    "memory_used_bytes": 1306624,
    "memory_total_bytes": 15032385536,
    "rx_bps": 0,
    "tx_bps": 0,
    "gpus": {
        "882fa930": {
            "utilization": 0,
            "memory_used_bytes": 0,
            "memory_total_bytes": 22836000000
        }
    },
    "replica": "57vr7"
}
```

cc @drbh for viz

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -80,6 +80,7 @@ from ._cli_utils import TokenOpt, get_hf_api, typer_factory
 logger = logging.get_logger(__name__)
 
 SUGGESTED_FLAVORS = [item.value for item in SpaceHardware if item.value != "zero-a10g"]
+STATS_UPDATE_MIN_INTERVAL = 0.1  # we set a limit here since there is one update per second per job
 
 # Common job-related options
 ImageArg = Annotated[
@@ -421,7 +422,6 @@ def jobs_stats(
             row += ["-- / --" if ("/" in header or "USAGE" in header) else "--" for header in table_headers[1:]]
             rows_per_job_id[job_id] = [row]
         last_update_time = time.time()
-        min_seconds_between_updates = 0.1  # we set a limit here since there is one update per second per job
         total_rows = [row for job_id in rows_per_job_id for row in rows_per_job_id[job_id]]
         _print_output(total_rows, table_headers, headers_aliases, None)
 
@@ -439,7 +439,7 @@ def jobs_stats(
             else:
                 rows_per_job_id[job_id] = rows
             now = time.time()
-            if now - last_update_time >= min_seconds_between_updates:
+            if now - last_update_time >= STATS_UPDATE_MIN_INTERVAL:
                 _clear_line(2 + len(total_rows))
                 total_rows = [row for job_id in rows_per_job_id for row in rows_per_job_id[job_id]]
                 _print_output(total_rows, table_headers, headers_aliases, None)


drbh (approved):
lgtm! great feature. 

just opened a small pr to update how bytes are summed https://github.com/huggingface/huggingface_hub/pull/3656

*note: also would recommend another approval from a core `huggingface_hub` maintainer

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +328,69 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+        row = [
+            job_id,
+            f"{metrics['cpu_usage_pct']}%",
+            round(metrics["cpu_millicores"] / 1000.0, 1),
+            f"{round(100 * metrics['memory_used_bytes'] / metrics['memory_total_bytes'], 2)}%",
+            f"{_format_size(metrics['memory_used_bytes'])}B / {_format_size(metrics['memory_total_bytes'])}B",
+            f"{_format_size(rx)}B / {_format_size(tx)}B",
+        ]
+        if metrics["gpus"] and isinstance(metrics["gpus"], dict):
+            gpu = next(iter(metrics["gpus"].values()))
```
Wauplin: This is clunky as a Job can run on multiple GPUs (e.g. `l4x4`) in which case I suppose the `gpus` dictionary has more than 1 entry. What do you think of printing 1 line per GPU keeping the rest of the info on the first line, something like this:

```
JOB ID                   CPU % NUM CPU MEM % MEM USAGE    NET I/O   GPU UTIL % GPU MEM % GPU MEM USAGE 
------------------------ ----- ------- ----- ------------ --------- ---------- --------- ------------- 
6953ff6274100871415c13fd 0%    3.5     0.01% 1.3MB/15.0GB 0.0B/0.0B 0%         0.0%      0.0B/22.8GB
                                                                    0%         0.0%      0.0B/22.8GB
```

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +328,69 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+def _clear_line(n=1):
+    LINE_UP = "\033[1A"
+    LINE_CLEAR = "\x1b[2K"
+    for i in range(n):
+        print(LINE_UP, end=LINE_CLEAR)
+
+
+@jobs_cli.command("stats", help="Fetch the resource usage statistics and metrics of a Job")
+def jobs_stats(
+    job_id: JobIdArg,
```
Wauplin: Would be great to make `job_id` optional like in `docker stats` so that one can do

```
hf jobs stats
```

without knowing the job ids

Wauplin: (in practice it requires to handle 1 stream per running job so might not be worth the effort... 😕. In any case, can be for a later PR)

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +328,69 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+        if metrics["gpus"] and isinstance(metrics["gpus"], dict):
+            gpu = next(iter(metrics["gpus"].values()))
+            row += [
+                f"{gpu['utilization']}%",
+                f"{round(100 * gpu['memory_used_bytes'] / gpu['memory_total_bytes'], 2)}%",
+                f"{_format_size(gpu['memory_used_bytes'])}B / {_format_size(gpu['memory_total_bytes'])}B",
+            ]
+        else:
+            row += ["N/A"] * (len(table_headers) - len(row))
+        _clear_line(3)
```
Wauplin: todo: make `3` dynamic in case of multiple GPUs

src/huggingface_hub/hf_api.py
```diff
@@ -10187,6 +10188,112 @@ def fetch_job_logs(
...
+
+            token `(Union[bool, str, None]`, *optional*):
+                A valid user access token. If not provided, the locally saved token will be used, which is the
+                recommended authentication method. Set to `False` to disable authentication.
+                Refer to: https://huggingface.co/docs/huggingface_hub/quick-start#authentication.
+
+        Example:
+
+            ```python
+            >>> from huggingface_hub import fetch_job_metrics, run_job
```
Wauplin, line 10260: I'm getting the following error when running this snippet. Does it work for you?

<details>

<summary>Traceback</summary>

```
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_transports/default.py", line 250, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 256, in handle_request
    raise exc from None
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 236, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/connection.py", line 103, in handle_request
    return self._connection.handle_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/http11.py", line 136, in handle_request
    raise exc
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/http11.py", line 106, in handle_request
    ) = self._receive_response_headers(**kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/http11.py", line 177, in _receive_response_headers
    event = self._receive_event(timeout=timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_sync/http11.py", line 217, in _receive_event
    data = self._network_stream.read(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_backends/sync.py", line 126, in read
    with map_exceptions(exc_map):
  File "/usr/lib/python3.11/contextlib.py", line 158, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.ReadTimeout: The read operation timed out

  File "/home/wauplin/projects/huggingface_hub/./rez.py", line 5, in <module>
    for metrics in fetch_job_metrics(job_id=job.id):
  File "/home/wauplin/projects/huggingface_hub/src/huggingface_hub/hf_api.py", line 10257, in fetch_job_metrics
    with get_session().stream(
  File "/usr/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_client.py", line 868, in stream
    response = self.send(
               ^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_client.py", line 914, in send
    response = self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_client.py", line 942, in _send_handling_auth
    response = self._send_handling_redirects(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_client.py", line 979, in _send_handling_redirects
    response = self._send_single_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_client.py", line 1014, in _send_single_request
    response = transport.handle_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_transports/default.py", line 249, in handle_request
    with map_httpcore_exceptions():
  File "/usr/lib/python3.11/contextlib.py", line 158, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/home/wauplin/projects/huggingface_hub/.venv/lib/python3.11/site-packages/httpx/_transports/default.py", line 118, in map_httpcore_exceptions
    raise mapped_exc(message) from exc
httpx.ReadTimeout: The read operation timed out
```


</details>

src/huggingface_hub/hf_api.py
```diff
@@ -10187,6 +10188,112 @@ def fetch_job_logs(
...
+                    and isinstance(getattr(err.__context__, "__cause__", None), TimeoutError)
+                )
+                if is_no_event_timeout:
+                    # job is likely finished
+                    pass
+                elif retry >= max_retries:
+                    raise
+                else:
+                    retry += 1
+                    sleep_time = min(max_wait_time, max(min_wait_time, sleep_time * 2))
```
Wauplin: let's add a warning log here for transparency with the error

src/huggingface_hub/hf_api.py
```diff
@@ -10187,6 +10188,112 @@ def fetch_job_logs(
...
+        max_retries = 5
+        min_wait_time = 1
+        max_wait_time = 10
+        sleep_time = 0
+        while True:
+            time.sleep(sleep_time)
+            try:
+                with get_session().stream(
+                    "GET",
+                    f"https://huggingface.co/api/jobs/{namespace}/{job_id}/metrics",
```
```suggestion
                    f"{self.endpoint}/api/jobs/{namespace}/{job_id}/metrics",
```

I just realized job methods is hardcoding `https://huggingface.co`. Let's use `self.endpoint` everywhere for consistency.

src/huggingface_hub/hf_api.py
```diff
@@ -10187,6 +10188,112 @@ def fetch_job_logs(
...
+        # - ChunkedEncodingError can happen in case of stopped logging in the middle of streaming
+        # - there is a ": keep-alive" every 30 seconds
+
+        # We don't use http_backoff since we need to check ourselves if the job is still running
+        retry = 0
+        max_retries = 5
+        min_wait_time = 1
+        max_wait_time = 10
+        sleep_time = 0
+        while True:
```
Wauplin: could this logic be factorized with `fetch_job_logs`? (logic is pretty much duplicated between the two)

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +328,69 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+        "gpu_utilization",
+        "gpu_memory_used_bytes_pct",
+        "gpu_memory_used_bytes_and_total_bytes",
+    ]
+    row = [job_id] + ["-- / --" if ("/" in header or "USAGE" in header) else "--" for header in table_headers[1:]]
+    _print_output([row], table_headers, headers_aliases, None)
+    rx, tx = 0.0, 0.0
+    for metrics in api.fetch_job_metrics(job_id=job_id, namespace=namespace):
+        rx += metrics["rx_bps"]
+        tx += metrics["tx_bps"]
```
Wauplin: I've checked both https://github.com/huggingface/huggingface_hub/pull/3655 and https://github.com/huggingface/huggingface_hub/pull/3656 (from @drbh) and while I agree this is consistent with docker command, I don't think we should display the "NET I/O" value since we don't know it. Computing a value based on bps is likely to end up with an unreliable value (even if #3656 is partly fixing it). For instance the `NET I/O` value would not be absolute but relative to when the user triggers the command (which is clunky). Also considering that the bps is constant during the whole second is an approximation likely to be inaccurate. 


What I suggest:
- close #3656 
- remove `NET I/O` and display the bps instead since this is the information returned by the API. Something like `I/O RATE` or `I/O BANDWIDTH` ? I think it's ok not to be 100% aligned with docker CLI as long as we provide accurate information
- add `NET I/O` in the future if the information can be returned by the server

Wauplin (commented):
Thanks for the PR! Left a couple comments on the logic. About the I/O numbers, I am in favor of displaying only the information returned by the server without inferring anything client-side (since it's error-prone). Let me know what you think

src/huggingface_hub/hf_api.py
Wauplin: EDIT: actually reported in https://github.com/huggingface/huggingface_hub/issues/3651 and addressed in https://github.com/huggingface/huggingface_hub/pull/3653

src/huggingface_hub/cli/jobs.py
lhoestq: sounds good !

src/huggingface_hub/cli/jobs.py
lhoestq: I can make it work with threads pretty easily, let me push a change and let me know what you think

src/huggingface_hub/cli/jobs.py
drbh: +1 sounds good to me - just closed https://github.com/huggingface/huggingface_hub/pull/3656 thanks @Wauplin

lhoestq:
Thanks for the review ! I took your comments into account:
- factorized the code
- show bps instead of cumulative network usage. I didn't have to change the column name "NET I/O" since I'm now using the "bps" unit
- I added warnings in case of errors
- I added multi-gpu support as you suggested, using one additional row per additional gpu
- `hf jobs stats` now show the stats of all the running jobs. I use one thread per job for this to keep it simple

src/huggingface_hub/hf_api.py
lhoestq: This can happen when the hardware for a job is not ready yet, I made a fix that double check that the job is still marked as 'running' instead of raising an error

src/huggingface_hub/cli/jobs.py
```diff
@@ -767,3 +893,36 @@ def _get_extended_environ() -> Dict[str, str]:
...
+
+T = TypeVar("T")
+
+
+def _write_generator_to_queue(queue: Queue, func: Callable[..., Iterable[T]], kwargs: dict) -> None:
+    for result in func(**kwargs):
+        queue.put(result)
+
+
+def iflatmap_unordered(
```
lhoestq, line 906: just to clarify the name of this function in case it can help: it's inspired by those map() variants:
- imap(): like map() but returns an iterable instead of a list of results
- imap_unordered(): like imap() but the output is sorted by time of arrival
- flatmap(): like map() but given a function which returns a list, flatmap() returns the flattened list that is the concatenation of all the output lists

iflatmap_unordered() is the mix of all three: it returns the flattened iterable of items as they arrive, given a function that returns an iterable of items

(I reused it from `datasets` btw)

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +339,113 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
         print(_tabulate(rows, headers=headers))
 
 
+def _clear_line(n=1):
```
```suggestion
def _clear_line(n: int) -> None:
```

let's not add a default value if not used (+ add type annotation)

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +339,113 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+    if namespace is None:
+        namespace = api.whoami()["name"]
+    if job_ids is None:
+        job_ids = [
+            job.id
+            for job in api.list_jobs(namespace=namespace)
+            if (job.status.stage if job.status else "UNKNOWN") in ("RUNNING", "UPDATING")
+        ]
+    if len(job_ids) == 0:
+        print("No jobs found")
```
```suggestion
        print("No running jobs found")
```

(nit)

src/huggingface_hub/cli/jobs.py
```diff
@@ -322,6 +339,113 @@ def _print_output(rows: list[list[Union[str, int]]], headers: list[str], fmt: Op
...
+        "gpu_memory_used_bytes_and_total_bytes",
+    ]
+    with multiprocessing.pool.ThreadPool(len(job_ids)) as pool:
+        rows_per_job_id: dict[str, list[list[Union[str, int]]]] = {}
+        for job_id in job_ids:
+            row: list[Union[str, int]] = [job_id]
+            row += ["-- / --" if ("/" in header or "USAGE" in header) else "--" for header in table_headers[1:]]
+            rows_per_job_id[job_id] = [row]
+        last_update_time = time.time()
+        min_seconds_between_updates = 0.1  # we set a limit here since there is one update per second per job
```
Wauplin: let's call it `MIN_SECONDS_BETWEEN_STATS_UPDATE` and set it outside as a constant

Wauplin: or `STATS_UPDATE_MIN_INTERVAL` ?

src/huggingface_hub/cli/jobs.py
Wauplin: agree with the naming 👍 Can you put a small docstring to explain it so it's not lost ?

lhoestq:
done ! I'll also update https://github.com/huggingface/hub-docs/pull/2103

Wauplin:
Failing CI is unrelated so let's merge this :)

PR #3656, opened 2025-12-30
fix: accumulate bytes with respect to time

drbh:
This PR adds a small change to accumulate the bps with respect to time to ensure the summation reflects the total bytes since updates may not be received at exactly 1 second intervals


Wauplin:
I'd be in favor not to compute and display `NET I/O` at all (see https://github.com/huggingface/huggingface_hub/pull/3655#discussion_r2661121294)

drbh:
closing since NET I/O will likely be removed https://github.com/huggingface/huggingface_hub/pull/3655#discussion_r2661121294

PR #3665, opened 2026-01-06
Add Job Timeout section to CLI docs

davanstrien:
Still had a few people run into timeouts, which we document quite well in the main Jobs guide, but not in the CLI so much. @lhoestq - I will  check this in the new docs too!

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -934,7 +934,7 @@ You can pass environment variables to your job using
 
 ### Job Timeout
 
-Jobs have a default timeout, after which they automatically stop. For long-running tasks like model training, set a custom timeout using the `--timeout` option:
+Jobs have a default timeout of 30 mins, after which they automatically stop. For long-running tasks like model training, set a custom timeout using the `--timeout` option:
 
 ```bash
 # Set timeout in seconds (default unit)


docs/source/en/guides/cli.md
```diff
@@ -932,6 +932,33 @@ You can pass environment variables to your job using
 > With this syntax, the secret is retrieved from the environment variable.
 > For `HF_TOKEN`, it may read the token file located in the Hugging Face home folder if the environment variable is unset.
 
+### Job Timeout
+
+Jobs have a default timeout, after which they automatically stop. For long-running tasks like model training, set a custom timeout using the `--timeout` option:
```
Wauplin: do we know what is this default timeout? If yes we should document it

davanstrien: We had it as 30 mins in the other docs, but based on some internal discussion on Slack there seems to be other numbers floating around. @christophe-rannou is 30 mins what you expect on the server side?

lhoestq (approved):

docs/source/en/guides/cli.md
davanstrien:
```suggestion
Jobs have a default timeout of 30 mins, after which they automatically stop. For long-running tasks like model training, set a custom timeout using the `--timeout` option:
```

davanstrien: I've updated to 30 mins (seems to be what is expected on server side).

Wauplin (approved):
Thanks!

PR #3671, opened 2026-01-09
Add `dimensions` & `encoding_format` parameter to InferenceClient for output embedding size

mishig25:
## Fixes a problem described below

In js client feature extraction, I can use dimension argument like:
```js
const batchEmbeddings = await client.featureExtraction({
    endpointUrl: HF_EMBEDDING_ENDPOINT_URL,
    inputs:      "This is a test sentence. ".repeat(100),
    truncate:    true,
    dimensions:  256,
});
```

However, similar argument fails in python client:
```py
embeddings = client.feature_extraction("This is a test sentence. "*100, truncate=True, dimensions=256)
```
`TypeError: InferenceClient.feature_extraction() got an unexpected keyword argument 'dimensions'`

### DIFF
--- a/src/huggingface_hub/inference/_generated/_async_client.py
+++ b/src/huggingface_hub/inference/_generated/_async_client.py
@@ -1089,9 +1089,10 @@ class AsyncInferenceClient:
                 Which side of the input should be truncated when `truncate=True` is passed.
             dimensions (`int`, *optional*):
                 The number of dimensions the resulting output embeddings should have.
-                Only available on server powered by Text-Embedding-Inference.
+                Only available on OpenAI-compatible embedding endpoints.
             encoding_format (`Literal["float", "base64"]`, *optional*):
                 The format of the output embeddings. Either "float" or "base64".
+                Only available on OpenAI-compatible embedding endpoints.
 
         Returns:
             `np.ndarray`: The embedding representing the input text as a float32 numpy array.


src/huggingface_hub/inference/_client.py
```diff
@@ -1058,6 +1060,11 @@ def feature_extraction(
                 Only available on server powered by Text-Embedding-Inference.
             truncation_direction (`Literal["left", "right"]`, *optional*):
                 Which side of the input should be truncated when `truncate=True` is passed.
+            dimensions (`int`, *optional*):
+                The number of dimensions the resulting output embeddings should have.
+                Only available on server powered by Text-Embedding-Inference.
```
hanouticelina: it's actually compatible with any openai embeddings compatible server: https://platform.openai.com/docs/api-reference/embeddings/create
```suggestion
                Only available on OpenAI-compatible embedding endpoints.
```

src/huggingface_hub/inference/_client.py
```diff
@@ -1058,6 +1060,11 @@ def feature_extraction(
...
+            encoding_format (`Literal["float", "base64"]`, *optional*):
+                The format of the output embeddings. Either "float" or "base64".
```
hanouticelina, line 1067:
```suggestion
                The format of the output embeddings. Either "float" or "base64".
                Only available on OpenAI-compatible embedding endpoints.
```

mishig25:
since they are generated

https://github.com/huggingface/huggingface_hub/blob/3523d1e1d9898b54b2739a33cfaf8363f14ae0ea/src/huggingface_hub/inference/_generated/types/feature_extraction.py#L1-L5

what is the correct way to update this file?

hanouticelina:
@bot /style

mishig25:
can we have a patch release?

hanouticelina:
@mishig25  yes! the patch v1.3.1 is out https://github.com/huggingface/huggingface_hub/releases/tag/v1.3.1

PR #3674, opened 2026-01-09
[Jobs] Support multi gpu training commands

lhoestq:
Add support for these commonly-used commands for multi-gpu training:

```bash
hf jobs uv run --with torch -- torchrun train.py
hf jobs uv run --with accelerate -- accelerate launch train.py
```

cc @SunMarc you will love this

this will be pretty useful now that we have `hf jobs stats` to monitor multi-gpu usage

## Details

to make this work we needed to check for the python script in the command arguments and pass its content to the Job

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -11088,7 +11088,7 @@ class HfApi:
             script_content = base64.b64encode(Path(script_file).read_bytes()).decode()
             env["UV_SCRIPT_ENCODED"] = script_content
             if script == script_file:
-                script == "/tmp/script.py"
+                script = "/tmp/script.py"
             else:
                 script_args = [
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -11061,19 +11061,43 @@ def _create_uv_command_env_and_secrets(
...
+        # uv run train.py -> `script``
+        # uv run torchrun train.py -> one of `script_args``
+        if Path(script).is_file():
+            script_file = script
+        elif script.startswith("http://") or script.startswith("https://"):
+            script_file = None
+        elif script.endswith(".py"):
+            raise FileNotFoundError(script)
+        else:
+            # `script` could be a command like "torchrun" or "accelerate"
```
```suggestion
            # `script` could be a command like "torchrun train.py" or "accelerate launch train.py"
```

PR #3682, opened 2026-01-14
use constant of default revision instead of empty root

### LINKED ISSUE #3681: broken model download via AutoModel using private repositories
### Describe the bug

There seems to be some implicit proxy used in hf_api instead of using DEFAULT_REVISION leading to a broken private repository reference when we try to use the AutoModel* classes to load the models.

Example: 

```
model_name = "Qwen/Qwen3-0.6B"

# load the tokenizer and the model
tokenizer = AutoTokenizer.from_pretrained(model_name)
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto", device_map="auto")
```

Currently if you use huggingface.co's repos, it is fine, and when it uses hf_api to download `https://huggingface.co/api/models/Qwen/Qwen3-0.6B/` this resolves to the same file that would be at `https://huggingface.co/api/models/Qwen/Qwen3-0.6B/revision/main`

However, once you set HF_ENDPOINT to a private repo, this proxying does not occur, so the file returns 404 since the mirror would only fetch the actual revisions (default being "main")

I'll have a PR soon for this, but I believe the intent, like much of the code elsewhere, was to use the DEFAULT_REVISION when we detect that no revision is supplied.

I'm also aware the version is pretty far behind... unfortunately this is what is compatible with transformers, so I do think it warrants application to even previous minor release versions.

### Reproduction

```
import os

os.environ["HF_ENDPOINT"] = "your-private-mirror"
model_name = "Qwen/Qwen3-0.6B"

from transformers import AutoModelForCausalLM, AutoTokenizer

# load the tokenizer and the model
tokenizer = AutoTokenizer.from_pretrained(model_name)
# Next line fails due to 404 from the request to get the manifest at HF_ENDPOINT/api/models/{model_id}/revisions/main
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto", device_map="auto")
```

### Logs

```shell

```

### System info

```shell
- huggingface_hub: 0.36.0
- transformers: 4.57.5
```

cr-klarna:
Fixes #3681 

As mentioned in the issue, since Transformers' latest compatible non-prerelease manifest specifies < 1.0 for huggingface_hub, I believe this should be applied to the minor versions below that as well.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -2760,7 +2760,7 @@ class HfApi:
 
         headers = self._build_hf_headers(token=token)
         path = (
-            f"{self.endpoint}/api/models/{repo_id}"
+            f"{self.endpoint}/api/models/{repo_id}/revision/{quote(constants.DEFAULT_REVISION, safe='')}"
             if revision is None
             else (f"{self.endpoint}/api/models/{repo_id}/revision/{quote(revision, safe='')}")
         )
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -2759,7 +2759,9 @@ class HfApi:
             raise ValueError("`expand` cannot be used if `securityStatus` or `files_metadata` are set.")
 
         headers = self._build_hf_headers(token=token)
-        path = f"{self.endpoint}/api/models/{repo_id}/revision/{quote(revision or constants.DEFAULT_REVISION, safe='')}"
+        path = (
+            f"{self.endpoint}/api/models/{repo_id}/revision/{quote(revision or constants.DEFAULT_REVISION, safe='')}"
+        )
         params: dict = {}
         if securityStatus:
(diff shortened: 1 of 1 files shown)


cr-klarna:
This fixes a smaller problem, but there is still the fact that there is not really a proper static way that pathing is done when revision is not supplied, and the caching behavior is a bit wonky. Will need to work more on this to fully solve all issues, though this should be enough I think to give staff some inspiration to look into this a bit more so folks with private mirrors will be able to use the models more seamlessly

hanouticelina (approved):
thanks @cr-klarna for the PR! indeed, explicitly including `/revision/main` when no revision is specified is consistent with how we handle revisions elsewhere in the codebase and it ensures compatibility with mirrors that don't implement the implicit redirect
I pushed a commit to update `dataset_info` and `space_info` as well

Wauplin:
not needed

Issue #3684, opened 2026-01-15
Severe performance regression when streaming data from Hugging Face datasets.


leq6c:
### Describe the bug

When you read data incrementally with `HfFileSystemStreamFile` in `1.3.2`, it takes much longer than expected. This happens on the new version but not on older versions like `0.36.0`.

Comparison of time taken to read 25MB of data with a 1024-byte buffer:

- `0.36.0`: 2.5 seconds
- `1.3.2`: not measurable (at least more than 60 seconds)

I ran into this issue while trying to load a WebDataset using the `huggingface/datasets` library with `streaming=True`. It looked like the code was stuck and not making progress reading the data. It is practically unusable on the new version. The only workaround I found is downgrading `huggingface-hub` to `0.36.0`.

### Reproduction

```
import time
from huggingface_hub import HfFileSystem

hffs = HfFileSystem()

start = time.time()

with hffs.open("datasets/nvidia/AudioSkills/audioskills_xl/VoxCeleb2.json", block_size=0) as f:
    while True:
        data = f.read(1024)
        if not data:
            break

print(time.time() - start, "s")
```

### Logs

```shell
- `0.36.0`


% HF_DEBUG=1 ./.venv_0_36_0/bin/python main.py
Send: curl -X GET -H 'Accept: */*' -H 'Accept-Encoding: gzip, deflate' -H 'Connection: keep-alive' -H 'authorization: <TOKEN>' -H 'user-agent: unknown/None; hf_hub/0.36.0; python/3.10.6' https://huggingface.co/api/datasets/nvidia/AudioSkills
Request 576d9edf-e9ec-45d7-b449-bb1d5cbc7a1d: GET https://huggingface.co/api/datasets/nvidia/AudioSkills (authenticated: True)
Send: curl -X GET -H 'Accept: */*' -H 'Accept-Encoding: gzip, deflate' -H 'Connection: keep-alive' -H 'authorization: <TOKEN>' -H 'user-agent: unknown/None; hf_hub/0.36.0; python/3.10.6' https://huggingface.co/datasets/nvidia/AudioSkills/resolve/main/audioskills_xl/VoxCeleb2.json
Request 706f7cb0-4255-4e18-94b3-2059fba17e9c: GET https://huggingface.co/datasets/nvidia/AudioSkills/resolve/main/audioskills_xl/VoxCeleb2.json (authenticated: True)
[...]

PR #3685, opened 2026-01-15
Fix severe performance regression in streaming by keeping a byte iterator in HfFileSystemStreamFile

leq6c:
# Summary

Fixes a severe performance regression when streaming data from Hugging Face datasets. #3684 

# Problem

`HfFileSystemStreamFile` creates a new HTTP request every time `read(length)` is called.

- Each time `read(length)` is called, it recreates the HTTP byte iterator for the `Response` via the `_partial_read` function.
- This is not the expected behavior by the `httpx` module, so `httpx` raises an exception on the second `_partial_read` call.
- If `read(length)` function receives an exception from `_partial_read`, it retries by creating a new `Partial-Content` HTTP request starting from the current offset.
- After reading that data, the same pattern repeats on the next `read(length)` call.
- As a result, a new HTTP request is effectively created on every `read(length)` call. This leads to a huge number of HTTP requests and a major performance degradation.

Reference: `httpx` exception:

```
httpx.StreamConsumed: Attempted to read or stream some content, but the content has already been streamed. For requests, this could be due to passing a generator as request content, and then receiving a redirect response or a secondary request as part of an authentication flow.For responses, this could be due to attempting to stream the response content more than once.
```

This issue has been confirmed when reading a WebDataset with the `huggingface/datasets` library using `load_dataset` with `streaming=True`.

`load_dataset(..., streaming=True)` passes an `HfFileSystemStreamFile` instance to `tarfile` module as a file-like object, and `tarfile` uses a default buffer size of `10240 bytes`. (https://docs.python.org/3.14/library/tarfile.html#tarfile.open) As a result, when streaming is enabled, `load_dataset` ends up issuing roughly `file_size / 10240 bytes` HTTP requests while reading the data, which makes it unacceptably slow and probably increase the server load.

# Solution

Make `HfFileSystemStreamFile` behave like a real forward-only streaming file object, without creating new requests.

- Keep a byte iterator for sequential reads instead of recreating `iter_bytes()` on every `read()`.
- Introduce an internal buffer to handle varying `read()` sizes.

# Note

- The reproduction script presented in the #3684  still throws an error when `HF_DEBUG=1` is enabled due to another bug in `utils/_http.py` that `_curlify` function accesses the content of the request before calling `read()` by catching the httpx `RequestNotRead` exception.

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -884,11 +884,12 @@ def _curlify(request: httpx.Request) -> str:
         parts += [("-H", f"{k}: {v}")]
 
     body: Optional[str] = None
-    if request.content is not None:
-        body = request.content.decode("utf-8", errors="ignore")
-        if len(body) > 1000:
-            body = f"{body[:1000]} ... [truncated]"
-    elif request.stream is not None:
+    try:
+        if request.content is not None:
+            body = request.content.decode("utf-8", errors="ignore")
+            if len(body) > 1000:
+                body = f"{body[:1000]} ... [truncated]"
+    except httpx.RequestNotRead:
         body = "<streaming body>"
     if body is not None:
         parts += [("-d", body.replace("\n", ""))]


src/huggingface_hub/hf_file_system.py
```diff
@@ -1175,6 +1178,44 @@ def read(self, length: int = -1):
...
+
+        If EOF is reached before length, fewer bytes may be returned.
+        """
+        if length < -1:
+            raise ValueError("length must be -1 or >= 0")
+        if length == 0:
+            return b""
+
+        if self._stream_iterator is None:
+            self._stream_iterator = self.response.iter_bytes()
```
hanouticelina: maybe the iterator should be created when the connection is opened (in `_open_connection`)

```suggestion
```

src/huggingface_hub/hf_file_system.py
```diff
@@ -1175,6 +1178,44 @@ def read(self, length: int = -1):
...
+        """Read up to `length` bytes from stream buffer and stream.
+
+        If length == -1, read until EOF.
+
+        If EOF is reached before length, fewer bytes may be returned.
+        """
+        if length < -1:
+            raise ValueError("length must be -1 or >= 0")
+        if length == 0:
+            return b""
```
hanouticelina: negative values mean "read all", according to https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects. no need to raise here
```suggestion
        if length == 0:
              return b""
        if length < 0:  # read all
              length = -1
```

src/huggingface_hub/hf_file_system.py
```diff
@@ -1209,6 +1250,10 @@ def _open_connection(self):
             )
         )
 
+        # reset streaming state
+        self._stream_buffer.clear()
+        self._stream_iterator = None
```
hanouticelina: imo better to reset it at the beginning of `_open_connection`

leq6c:
@hanouticelina 
Thanks for the review.
Just to clarify, the core issue isn’t really the overhead of recreating the generator. `iter_bytes()` can only be consumed once, so the second call will raise `httpx.StreamConsumed`. That leads `read()` into the retry path every time, which ends up calling `self._open_connection()` on each `read()`. In other words, it issues a new HTTP request per `read()`.

src/huggingface_hub/hf_file_system.py
leq6c: fixed.

src/huggingface_hub/hf_file_system.py
leq6c: fixed. I added a test for this since the old implementation throws. 

confirmed it is the same on fsspec `read()`: https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractBufferedFile.read

src/huggingface_hub/hf_file_system.py
leq6c: fixed.

src/huggingface_hub/hf_file_system.py
```diff
@@ -1175,6 +1178,39 @@ def read(self, length: int = -1):
...
+        else:  # read from buffer and stream
+            buf.extend(self._stream_buffer)
+            self._stream_buffer.clear()
+            for chunk in self._stream_iterator:
+                need = length - len(buf)
+                buf.extend(chunk[:need])
+                self._stream_buffer.extend(chunk[need:])
+                if len(buf) >= length:
+                    break
+            return bytes(buf)
```
hanouticelina: a couple of things here :
- no need for elif/else since each branch returns early
- we can use `bytearray(self._stream_buffer)` instead of `bytearray()` followed by `extend()`
- (nit) In the for loop, `chunk[:need]` and `chunk[need:]` are calculated every iteration even when `need >= len(chunk)` . slightly better to check first
```suggestion
      if length < 0:
          buf = bytearray(self._stream_buffer)
          self._stream_buffer.clear()
          for chunk in self._stream_iterator:
              buf.extend(chunk)
          return bytes(buf)

      if length <= len(self._stream_buffer):
          result = bytes(self._stream_buffer[:length])
          del self._stream_buffer[:length]
          return result

      buf = bytearray(self._stream_buffer)
      self._stream_buffer.clear()
      for chunk in self._stream_iterator:
          need = length - len(buf)
          if need >= len(chunk):
              buf.extend(chunk)
          else:
              buf.extend(chunk[:need])
              self._stream_buffer.extend(chunk[need:])
              break
      return bytes(buf)
```

tests/test_hf_file_system.py
```diff
@@ -200,6 +200,72 @@ def test_stream_file_retry(self):
             self.assertEqual(f.read(6), b"binary")
             self.assertIsNotNone(f.response)  # a new connection has been created
 
+    def test_stream_file_reuse_response(self):
```
hanouticelina, line 203: thank you for adding tests!

src/huggingface_hub/hf_file_system.py
leq6c: Sounds good improvement, let me check shortly

leq6c: Reflected the changes, but I changed `need >= len(chunk)` to `need > len(chunk)` so we don't advance the iterator one more time when `need == len(chunk)`. I think this is slightly better.

hanouticelina:
we've just released huggingface_hub [v1.3.3](https://github.com/huggingface/huggingface_hub/releases/tag/v1.3.3) with the fix, thanks again @leq6c

PR #3704, opened 2026-01-20
[HfFileSystem] Fix resolve_path() with special char @

lhoestq:
this was causing issues in the viewer for datasets with folders or files with names that contain the character `@`


hanouticelina:
@lhoestq we've just released huggingface_hub [v1.3.3](https://github.com/huggingface/huggingface_hub/releases/tag/v1.3.3) which include the fix!

PR #3724, opened 2026-01-23
Pass local script and config files to job

lhoestq:
Pass any local file to the Job, especially useful for config files in addition to scripts

For example if you use this command:

```bash
hf jobs uv run script.py config.yml
```

with any config.yml file and this script.py file:

```python
# script.py
import sys
from pathlib import Path

for arg in sys.argv:
    print(arg, Path(arg).exists())
```
it prints

```
script.py True
config.yml True
```

This also works with commands like `torchrun` or `accelerate launch`:

```bash
hf jobs uv run --with torch torchrun script.py config.yml
```

-------

## Details

This works by passing an environment variable `LOCAL_FILES_ENCODED` formatted like this

```
file1 base64content1
file2 base64content2
```
and then uses xargs to recreate all the necessary files prior to running uv

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -11119,7 +11119,10 @@ class HfApi:
         missing_local_files = {
             candidate
             for candidate in [script] + script_args
-            if not Path(candidate).is_file() and Path(candidate).suffix in [".py", ".sh", ".yaml", ".yml", ".toml"]
+            if not Path(candidate).is_file()
+            and Path(candidate).suffix in [".py", ".sh", ".yaml", ".yml", ".toml"]
+            and not candidate.startswith("https://")
+            and not candidate.startswith("http://")
         }
         if missing_local_files:
             raise FileNotFoundError(", ".join(missing_local_files))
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -11112,43 +11113,55 @@ def _create_uv_command_env_and_secrets(
...
-            else:
-                script_file = None
+        # Find the local files to pass to the job
+        local_files_to_include = {candidate for candidate in [script] + script_args if Path(candidate).is_file()}
+        # Fail early for missing scripts or config files
+        missing_local_files = {
+            candidate
+            for candidate in [script] + script_args
+            if not Path(candidate).is_file() and Path(candidate).suffix in [".py", ".sh", ".yaml", ".yml", ".toml"]
+        }
```
hanouticelina, line 11126: we need to exclude urls from the missing files check here no?

src/huggingface_hub/hf_api.py
```diff
@@ -11112,43 +11113,55 @@ def _create_uv_command_env_and_secrets(
...
+                    for i in itertools.count():
+                        remote_file_name = remote_file_path.with_stem(remote_file_path.stem + f"({i})").name
+                        if remote_file_name not in remote_file_names:
+                            remote_file_names.add(remote_file_name)
+                            break
+                else:
+                    remote_file_names.add(remote_file_path.name)
+
+            # Replace local paths with remote paths in command
+            for local_file_to_include, remote_file_name in zip(local_files_to_include, remote_file_names):
```
hanouticelina: Using two sets and zipping them doesn't preserve the mapping. When local paths include directories (exemple: `./scripts/train.py` -> `train.py`), the hashes differ and the tier order is unrelated, i asked Claude to give me a repro (i wasn't sure in the beginning tbh):
```python
local_files = {"./scripts/train.py", "./configs/config.yaml"}
remote_names = set()
for local in local_files:
    remote_names.add(local.split("/")[-1])

for local, remote in zip(local_files, remote_names):
    print(f"{local} -> {remote}")
# ./configs/config.yaml -> train.py  (wrong!)
# ./scripts/train.py -> config.yaml  (wrong!)
```
so it's better to use a `dict[str, str]` to preserve the local -> remote mapping

lhoestq:
I took your comments into account, thanks !

src/huggingface_hub/hf_api.py
```diff
@@ -11112,43 +11113,63 @@ def _create_uv_command_env_and_secrets(
...
+                        remote_file_name = remote_file_path.with_stem(remote_file_path.stem + f"({i})").name
+                        if remote_file_name not in remote_to_local_file_names:
+                            remote_to_local_file_names[remote_file_name] = local_file_to_include
+                            break
+                else:
+                    remote_to_local_file_names[remote_file_path.name] = local_file_to_include
+            local_to_remote_file_names = dict(
+                (local_file_to_include, remote_file_name)
+                for remote_file_name, local_file_to_include in remote_to_local_file_names.items()
+            )
```
Wauplin, line 11151:
```suggestion
            local_to_remote_file_names = {
                local_file_to_include: remote_file_name
                for remote_file_name, local_file_to_include in remote_to_local_file_names.items()
            }
```

(nit for next time) I find it slightly more readable with dict comprehension

src/huggingface_hub/hf_api.py
```diff
@@ -11112,43 +11113,63 @@ def _create_uv_command_env_and_secrets(
...
-                    "/tmp/script.py" if script_arg == script_file else script_arg for script_arg in script_args
-                ]
+            # Find appropriate remote file names
+            remote_to_local_file_names: dict[str, str] = {}
+            for local_file_to_include in local_files_to_include:
+                local_file_path = Path(local_file_to_include)
+                # remove spaces for proper xargs parsing
+                remote_file_path = Path(local_file_path.name.replace(" ", "_"))
+                if remote_file_path.name in remote_to_local_file_names:
+                    for i in itertools.count():
```
Wauplin, line 11141: a comment would be great when doing this kind of non-trivial logic

from my understanding this logic is useful only if the initial files where like `my_config.yaml` and `my config.yaml` => we want to turn them into `my_config.yaml` and `my_config1.yaml`. Am I correct? I feel that it bloats the logic here for such a tiny corner case (unless I'm missing a bigger use case for conflicting files)

lhoestq: noted !

It could also happen in case of `model/config.yaml` and `dataset/config.yaml` for example, which is still a corner case though

PR #3742, opened 2026-01-27
[Jobs] Better jobs filtering in CLI: labels and negation

lhoestq:
Allow filtering by `label`, and support `!=`

some examples:

```bash
# already existed
hf jobs ps -a --filter status=error
hf jobs ps -a --filter "command=*train.py"
# new with !=
hf jobs ps -a --filter status!=error
hf jobs ps -a --filter "command!=*train.py"
# new with labels
hf jobs ps -a --filter label=fine-tuning
hf jobs ps -a --filter label=model=Qwen3-06B
hf jobs ps -a --filter label!=fine-tuning
hf jobs ps -a --filter label=model!=Qwen3-06B
```

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -782,9 +782,7 @@ def scheduled_ps(
         cmd = scheduled_job.job_spec.command or []
         command_str = " ".join(cmd) if cmd else "N/A"
         last_job_at = (
-            scheduled_job.status.last_job.at.strftime("%Y-%m-%d %H:%M:%S")
-            if scheduled_job.status.last_job
-            else "N/A"
+            scheduled_job.status.last_job.at.strftime("%Y-%m-%d %H:%M:%S") if scheduled_job.status.last_job else "N/A"
         )
         next_job_run_at = (
             scheduled_job.status.next_job_run_at.strftime("%Y-%m-%d %H:%M:%S")


hanouticelina:
@bot /style

src/huggingface_hub/cli/jobs.py
```diff
@@ -317,20 +318,12 @@ def jobs_logs(
...
-        if key not in job_properties:
-            return False
-        # Support pattern matching with wildcards
-        if "*" in pattern or "?" in pattern:
-            # Convert glob pattern to regex
-            regex_pattern = pattern.replace("*", ".*").replace("?", ".")
-            if not re.search(f"^{regex_pattern}$", job_properties[key], re.IGNORECASE):
-                return False
-        # Simple substring matching
-        elif pattern.lower() not in job_properties[key].lower():
```
hanouticelina, line 332: not sure i fully understand line 326, but it looks like it introduces a breaking change, we had `elif pattern.lower() not in job_properties[key].lower():` which allows `hf jobs ps --filter status=run` to match jobs with status `RUNNING` (`"run" in "running" is True`). is this intentional? imo using glob (already the case in this PR with `fnmatch`) is better for substring matching so i'm okay breaking this.

src/huggingface_hub/cli/jobs.py
```diff
@@ -317,20 +318,12 @@ def jobs_logs(
...
+    for key, op_str, pattern in filters:
+        op = operator.not_ if op_str == "!=" else operator.truth
+        # Check if property exists and it matches (exact string or glob pattern)
+        if not op(key in job_properties and fnmatch(job_properties[key].lower(), pattern.lower())):
```
hanouticelina: just for readability purposes 
```suggestion
    for key, op_str, pattern in filters:
        value = job_properties.get(key)
        if value is None:
            if op_str == "!=":
                continue
            return False
        match = fnmatch(value.lower(), pattern.lower())
        if (op_str == "=" and not match) or (op_str == "!=" and match):
            return False
```

src/huggingface_hub/cli/jobs.py
```diff
@@ -496,11 +489,35 @@ def jobs_ps(
...
+                f = f[len("label!=") :]
+                if "=" not in f:
+                    f += "!="
+                else:
+                    print(
+                        f"Warning: Ignoring invalid label filter format '{f}'. Use label=value or label!=value or label=key=value or label=key!=value format."
+                    )
+                    continue
+            else:
+                is_label_filter = False
```
hanouticelina: more readable like this imo
```suggestion
            if f.startswith("label!=") or f.startswith("label="):
                if f.startswith("label!="):
                    op = "!="
                    label_part = f[len("label!="):]
                    if "=" in label_part:
                        print(f"Warning: Ignoring invalid label filter format 'label!={label_part}'. Use label!=key format.")
                        continue
                    label_key, label_value = label_part, "*"
                else:
                    op = "="
                    label_part = f[len("label="):]
                    if "=" in label_part:
                        label_key, label_value = label_part.split("=", 1)
                    else:
                        label_key, label_value = label_part, "*"
                labels_filters.append((label_key.lower(), op, label_value.lower()))
```

hanouticelina (commented):
thanks! I left some comment mainly about code readability and some suggestions for refactoring 😄

src/huggingface_hub/cli/jobs.py
lhoestq: Yes it's an intentional change to align with docker which uses patterns

src/huggingface_hub/cli/jobs.py
lhoestq: this code doesn't support the `label=key!=value` case, let me see if I can adapt it

hanouticelina: out of curiosity, is the `label=key!=value` pattern supported by docker as well? i can't find it in their [docs](https://docs.docker.com/reference/cli/docker/container/ls/#filter)

lhoestq: Yes I tried locally to confirm ^^

lhoestq:
I adapted your suggestion, let me know what you think :)

src/huggingface_hub/cli/jobs.py
```diff
@@ -317,20 +317,17 @@ def jobs_logs(
...
-        # Check if property exists
-        if key not in job_properties:
+    for key, op_str, pattern in filters:
+        value = job_properties.get(key)
+        if value is None:
+            if op_str == "!=":
+                continue
+            return False
+        match = fnmatch(value.lower(), pattern.lower())
+        if (op_str == "=" and not match) or (op_str == "!=" and match):
```
hanouticelina, line 328: can you remove the line 330? `return False` is duplicated

hanouticelina:
failing tests are unrelated, merging!

PR #3744, opened 2026-01-28
Add more error handling output to hf jobs cli commands

davanstrien:
## Summary
- Add try/except error handling to `jobs_stats` command to display user-friendly messages instead of full tracebacks
- Provide status-code-aware hints for common errors (404, 403, 500)
- Follow the same pattern used by `jobs_ps` and `jobs_hardware` commands

**Before:**
```
$ hf jobs stats invalidjobid
  ...
huggingface_hub.errors.HfHubHTTPError: Server error '500 Internal Server Error'...
```

**After:**
```
$ hf jobs stats invalidjobid
JOB ID       CPU % NUM CPU MEM % MEM USAGE NET I/O GPU UTIL % GPU MEM % GPU MEM USAGE
------------ ----- ------- ----- --------- ------- ---------- --------- -------------
invalidjobid --    --      --    -- / --   -- / -- --         --        -- / --
Error: Job not found or server error. Please verify the job ID is correct.
```

## Test plan
- [x] Run `hf jobs stats invalidjobid` - shows friendly error instead of traceback
- [x] Run `ruff check` and `ruff format --check` - passes

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -470,9 +470,6 @@ def jobs_stats(
             raise CLIError("Job not found. Please check the job ID.") from e
         elif status == 403:
             raise CLIError("Access denied. You may not have permission to view this job.") from e
-        elif status == 500:
-            # Backend returns 500 for invalid job IDs (ideally would be 404)
-            raise CLIError("Job not found or server error. Please verify the job ID is correct.") from e
         else:
             raise CLIError(f"Failed to fetch job stats: {e}") from e
 


davanstrien:
We may also want to add something similar to other commands. From what I could tell it's not possible from the server response to know if the ID was the issue but I think it's useful to still suggest this as something to check?

src/huggingface_hub/cli/jobs.py
```diff
@@ -429,35 +429,51 @@ def jobs_stats(
...
+            print("Error: Access denied. You may not have permission to view this job.")
+        elif status == 500:
+            # Backend returns 500 for invalid job IDs (ideally would be 404)
+            print("Error: Job not found or server error. Please verify the job ID is correct.")
+        else:
+            print(f"Error fetching job stats: {e}")
+    except (KeyError, ValueError, TypeError) as e:
+        print(f"Error processing job stats: {e}")
+    except Exception as e:
+        print(f"Unexpected error - {type(e).__name__}: {e}")
```
hanouticelina: not sure about catching `KeyError`, `ValueError`, `TypeError` and bare `Exception`,  this can hide real bugs and makes debugging a bit harder. 
Also should we apply the same error handling for the other commands as well like `hf jobs logs` ?

davanstrien: The pattern here matches what's already used in `jobs_ps` (lines 544-549) and `jobs_hardware` (lines 574-577) - wanted to keep consistency across the commands. Happy to simplify if you'd prefer though

Agreed on adding error handling to the other commands (`jobs_logs`, `jobs_cancel`, `jobs_inspect`) - happy to do that in a follow-up PR to keep this one focused.

hanouticelina: yes indeed, it's a pattern that already existed, i opened a PR to remove that : https://github.com/huggingface/huggingface_hub/pull/3748

otherwise, trying to keep consistency across the commands is the right approach :)

davanstrien: IIUC #3748 removed all error handling from `jobs_ps`, `jobs_hardware`, etc.
For `jobs_stats`, should I:
1. **Remove all error handling** (match #3748 - users see tracebacks on errors)
2. **Keep only HfHubHTTPError handling** (remove the broad catches you flagged, but keep user-friendly messages for HTTP errors like invalid job IDs)

The original goal was to show a friendly message when users pass invalid job IDs (esp via CLI), but fine if this doesn't make sense for what the rest of `huggingface_hub` is doing.

hanouticelina: hey @davanstrien,
sorry for the delay, i was waiting to merge #3754 before getting back to this PR. we added a centralized way of handling errors in the CLI with `CLIError`,  It shows a clean error message without traceback by default, and the full traceback when `HF_DEBUG=1` is set. This gives users a nice ux whhile still allowing debugging when needed:
```console
> hf jobs stats invalid-job-id-12345
JOB ID               CPU % NUM CPU MEM % MEM USAGE NET I/O GPU UTIL % GPU MEM % GPU MEM USAGE 
-------------------- ----- ------- ----- --------- ------- ---------- --------- ------------- 
invalid-job-id-12345 --    --      --    -- / --   -- / -- --         --        -- / --       
Error: Job not found or server error. Please verify the job ID is correct.
Set HF_DEBUG=1 as environment variable for full traceback.
```
i think we can have something like this in `hf jobs`:
```suggestion
    except HfHubHTTPError as e:
        status = e.response.status_code if e.response is not None else None
        if status == 404:
            raise CLIError("Job not found. Please check the job ID.") from e
        elif status == 403:
            raise CLIError("Access denied. You may not have permission to view this job.") from e
        elif status == 500:
            raise CLIError("Job not found or server error. Please verify the job ID is correct.") from e
        else:
            raise CLIError(f"Failed to fetch job stats: {e}") from e
```

hanouticelina: cf my comment [here](https://github.com/huggingface/huggingface_hub/pull/3744#discussion_r2747138440).

davanstrien:
@hanouticelina thanks for the reviews/pointers! Think this should match what you had in mind now. I am fairly sure the failing tests aren't related to these but can take a closer look if it is!

src/huggingface_hub/cli/jobs.py
```diff
@@ -424,35 +434,47 @@ def jobs_stats(
...
+                    last_update_time = now
+    except HfHubHTTPError as e:
+        status = e.response.status_code if e.response is not None else None
+        if status == 404:
+            raise CLIError("Job not found. Please check the job ID.") from e
+        elif status == 403:
+            raise CLIError("Access denied. You may not have permission to view this job.") from e
+        elif status == 500:
+            # Backend returns 500 for invalid job IDs (ideally would be 404)
+            raise CLIError("Job not found or server error. Please verify the job ID is correct.") from e
```
hanouticelina: @davanstrien it seems that the server also returns `"500 Internal Server Error"` for non-existent job IDs for `logs` and `cancel` endpoints as well, has this been flagged to the backend team?

davanstrien: reported internally (private link https://huggingface.slack.com/archives/C07RD3RVD5W/p1770044465991559). Would indeed be nicer to have more granular errors from the API if possible.

hanouticelina: removing this since we will have a 404 error now for non-existent Job IDs (cf the conversation link in @davanstrien's comment above)
```suggestion
```

hanouticelina (approved):
very nice! thank you @davanstrien 

merging (failing tests are unrelated)

PR #3751, opened 2026-01-28
Use HF_HUB_DOWNLOAD_TIMEOUT as default httpx timeout

Wauplin:
Let's use `constants.HF_HUB_DOWNLOAD_TIMEOUT` instead of `constants.DEFAULT_DOWNLOAD_TIMEOUT` as default timeout for all `httpx` calls to the Hub. In practice the value is the same by default (10 seconds) but now it can be set using `HF_HUB_DOWNLOAD_TIMEOUT` environment variable. I don't expect anyone to spot the difference but it can prove useful in CIs when we are fine to wait longer as long as it the call completes correctly.

Related to internal slack thread on transformers CI issues ([here](https://huggingface.slack.com/archives/C01N44FJDHT/p1769609134030729) and [here](https://huggingface.slack.com/archives/C06LR9PQA00/p1769450912020829)). Once this fix is out, the solution will be to use `get_session().xxx` everywhere to make HTTP calls + to set `HF_HUB_DOWNLOAD_TIMEOUT` to a higher value in the tests (e.g. 60?)

cc @ydshieh @Rocketknight1 @tarekziade

### DIFF
--- a/tests/test_file_download.py
+++ b/tests/test_file_download.py
@@ -25,7 +25,6 @@ from unittest.mock import Mock, patch
 import httpx
 import pytest
 
-import huggingface_hub.file_download
 from huggingface_hub import HfApi, RepoUrl, constants
 from huggingface_hub._local_folder import write_download_metadata
 from huggingface_hub.errors import EntryNotFoundError, GatedRepoError, LocalEntryNotFoundError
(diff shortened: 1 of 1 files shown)


tarekziade:
LGTM - you might need to tweak https://github.com/huggingface/huggingface_hub/blob/main/tests/test_file_download.py#L1254

PR #3754, opened 2026-01-29
[CLI] Adds centralized error handling

hanouticelina:
@Wauplin's idea discussed through DMs on Slack.

This PR adds centralized CLI error handling with clean user facing messages: we have now a`CLIError` exception** for explicit CLI errors with user-friendly messages and we catch errors in `hf.py`  shows clean output (no traceback by default)
and we auto-convert known exceptions (`RepositoryNotFoundError`, etc.) to friendly messages. when `HF_DEBUG=1`** we show full traceback for debugging.

```console
> hf models info fake-user/fake-model
Error: Model 'fake-user/fake-model' not found.
Set HF_DEBUG environment variable to 1 for full traceback.

> HF_DEBUG=1 hf cache ls --cache-dir /nonexistent
Error: Cache directory not found: /nonexistent
  File "/Users/hcelina/repos/huggingface_hub/src/huggingface_hub/cli/cache.py", line 553, in ls
    hf_cache_info = scan_cache_dir(cache_dir)
...
huggingface_hub.errors.CacheNotFound: Cache directory not found: /nonexistent. Please use `cache_dir` argument or set `HF_HUB_CACHE` environment variable.

 ...
  File "/Users/hcelina/repos/huggingface_hub/src/huggingface_hub/cli/cache.py", line 555, in ls
    raise CLIError(f"Cache directory not found: {exc.cache_dir}") from exc
huggingface_hub.cli._errors.CLIError: Cache directory not found: /nonexistent
```

### DIFF
--- a/src/huggingface_hub/_oauth.py
+++ b/src/huggingface_hub/_oauth.py
@@ -33,8 +33,8 @@ class OAuthOrgInfo:
             The org's username. OpenID Connect field.
         picture (`str`):
             The org's profile picture URL. OpenID Connect field.
-        is_enterprise (`bool`):
-            Whether the org is an enterprise org. Hugging Face field.
+        plan (`str`, *optional*):
+            The org's plan (e.g., "enterprise", "team"). Hugging Face field.
         can_pay (`Optional[bool]`, *optional*):
             Whether the org has a payment method set up. Hugging Face field.
         role_in_org (`Optional[str]`, *optional*):
@@ -47,7 +47,7 @@ class OAuthOrgInfo:
     name: str
     preferred_username: str
     picture: str
-    is_enterprise: bool
+    plan: Optional[str] = None
     can_pay: Optional[bool] = None
     role_in_org: Optional[str] = None
     security_restrictions: Optional[list[Literal["ip", "token-policy", "mfa", "sso"]]] = None
@@ -215,7 +215,7 @@ def parse_huggingface_oauth(request: "fastapi.Request") -> Optional[OAuthInfo]:
                 name=org.get("name"),
                 preferred_username=org.get("preferred_username"),
                 picture=org.get("picture"),
-                is_enterprise=org.get("isEnterprise"),
+                plan=org.get("plan"),
                 can_pay=org.get("canPay"),
                 role_in_org=org.get("roleInOrg"),
                 security_restrictions=org.get("securityRestrictions"),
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/cli/hf.py
```diff
@@ -65,7 +68,26 @@ def main():
...
-    app()
+
+    try:
+        app()
+    except CLIError as e:
+        print(f"Error: {e.message}", file=sys.stderr)
+        if constants.HF_DEBUG:
+            traceback.print_exc()
+        else:
+            print("Set HF_DEBUG environment variable to 1 for full traceback.", file=sys.stderr)
```
```suggestion
            print(ANSI.grey("Set HF_DEBUG=1 as environment variable for full traceback."))
```

(just a suggestion, I feel that grey is less visible)

src/huggingface_hub/cli/_errors.py
```diff
@@ -0,0 +1,48 @@
...
+    RevisionNotFoundError,
+)
+
+
+class CLIError(Exception):
+    """CLI error with clean message (no traceback by defaut)."""
+
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(message)
```
Wauplin:
```suggestion
class CLIError(Exception):
    """CLI error with clean message (no traceback by defaut)."""
```

I don't think the `__init__` overload is needed here

Wauplin: let's define the `CLIError` in `./src/huggingface_hub/errors.py` with all other errors ?

hanouticelina: yes my initial reasoning was that the `errors` module is specifically for public API errors, but i was wrong, it already contains internal errors too (e.g. `XetError`). So it makes sense to define `CLIError` there as well 👍 
addressed in [994b05c](https://github.com/huggingface/huggingface_hub/pull/3754/commits/994b05c6ecabc4baa6aaa3c98470bc3d4eebc234)

src/huggingface_hub/cli/hf.py
```diff
@@ -65,7 +69,26 @@ def main():
     if not constants.HF_DEBUG:
         logging.set_verbosity_info()
     check_cli_update("huggingface_hub")
-    app()
+
+    try:
+        app()
+    except CLIError as e:
```
Wauplin, line 75: I start to wonder if we shouldn't have a `run_app()` method or implement a  `typer_factory(...).run()` with custom logic for:
- set verbosity info if not `HF_DEBUG`
- check cli update
- try/catch errors and format correctly

and that could be reused in `transformers` CLI [here](https://github.com/huggingface/transformers/blob/main/src/transformers/cli/transformers.py#L38)

Not necessarily to be done in this PR but worth adding it given all the CLI improvements you are making

PR #3735, opened 2026-01-26
[CLI] output format option for `ls` CLIs

hanouticelina:
Add consistent output format options (`--format {table,json}`, `-q`/`--quiet`) to `hf models ls`, `hf datasets ls`, `hf spaces ls`, `hf endpoints ls` and harmonize `hf cache ls` to provide a consistent and flexible user experience across CLI list commands.

---
[Slack Thread](https://huggingface.slack.com/archives/C03V11RNS7P/p1769422988537019?thread_ts=1769422988.537019&cid=C03V11RNS7P)

<a href="https://cursor.com/background-agent?bcId=bc-cc815060-8c39-4692-a697-2607dd592aba"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-cc815060-8c39-4692-a697-2607dd592aba"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

PR #3760, opened 2026-02-02
[CLI] Dynamic table columns based on `--expand` field

hanouticelina:
related to @Wauplin's comment [here](https://github.com/huggingface/huggingface_hub/pull/3735#discussion_r2741141133).

When using `--expand` with `hf models ls`, `hf datasets ls`, or `hf spaces ls`, the expanded fields are now displayed as additional columns in the table output. Also some columns were empty, this PR fixes that as well.

before:
```console
> hf models ls --limit 3 --expand downloadsAllTime
ID                                AUTHOR     DOWNLOADS LIKES TASK               DOWNLOADS_ALL_TIME 
--------------------------------- ---------- --------- ----- ------------------ ------------------ 
moonshotai/Kimi-K2.5              moonshotai 96162     1429  image-text-to-text 100387             
Tongyi-MAI/Z-Image                Tongyi-MAI 6347      785   text-to-image      6347               
tencent/HunyuanImage-3.0-Instruct tencent    148       768   image-to-image     148     
```
now:
```console
> hf models ls --limit 3 --expand downloadsAllTime
ID                                AUTHOR     DOWNLOADS LIKES TASK 
--------------------------------- ---------- --------- ----- ---- 
moonshotai/Kimi-K2.5              moonshotai                      
Tongyi-MAI/Z-Image                Tongyi-MAI                      
tencent/HunyuanImage-3.0-Instruct tencent     
```

_Disclaimer_: The code in this PR was mostly generated with Claude Code (especially the `extended_row_fn` inner function in `print_as_table`).

### DIFF
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -300,6 +300,7 @@ QuietOpt = Annotated[
 
 
 def _to_header(name: str) -> str:
+    """Convert a camelCase or PascalCase string to SCREAMING_SNAKE_CASE to be used as table header."""
     s = re.sub(r"([a-z])([A-Z])", r"\1_\2", name)
     return s.upper()
 


src/huggingface_hub/cli/models.py
```diff
@@ -86,10 +87,18 @@ def models_ls(
     """List models on the Hub."""
     api = get_hf_api(token=token)
     sort_key = sort.value if sort else None
+    if expand is not None:
+        api_expand = [c for c in _BASE_COLUMNS if c in _EXPAND_PROPERTIES] + [
```
Wauplin: I'm not a fan of requesting more data to make it into the table. IMO it's fine if we don't have downloads/likes/pipeline_tag when requesting `downloadsAllTime` since this is the default behavior from the API (often good to follow server's API)

hanouticelina: yes totally agree, my bad!

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -185,6 +191,8 @@ def print_list_output(
     id_key: str,
     headers: list[str],
     row_fn: Callable[[dict[str, Any]], list[str]],
+    base_columns: Optional[list[str]] = None,
+    expand_fields: Optional[list[str]] = None,
```
Wauplin: I'd rather adapt the code in `{datasets/models/spaces}.py` to pass the correct fields instead of modifying this helper method (which is unrelated to the `--expand` option)

src/huggingface_hub/cli/models.py
```diff
@@ -54,6 +54,7 @@
 _EXPAND_PROPERTIES = sorted(get_args(ExpandModelProperty_T))
 _SORT_OPTIONS = get_args(ModelSort_T)
 ModelSortEnum = enum.Enum("ModelSortEnum", {s: s for s in _SORT_OPTIONS}, type=str)  # type: ignore[misc]
+_BASE_COLUMNS = ["id", "author", "downloads", "likes", "pipeline_tag"]
```
Wauplin: instead of a list of base columns I think we should print them all in the table, the same way we are printing everything in JSON. We can add a few rule to limit the size of the columns if needed (e.g. "if >50 chars, then truncate the value and add '...'")

Here is something that would looks correct to me:

```
✗ hf models ls --limit 1                          
id                   created_at private downloads likes library_name tags                                   pipeline_tag       trending_score 
-------------------- ---------- ------- --------- ----- ------------ -------------------------------------- ------------------ -------------- 
moonshotai/Kimi-K2.5 2026-01-01         96162     1444  transformers transformers, safetensors, kimi_k25... image-text-to-text 1444           
```

```
✗ hf models ls --limit 1 --expand downloadsAllTime
id                   downloads_all_time trending_score 
-------------------- ------------------ -------------- 
moonshotai/Kimi-K2.5 100387             1444           
```

A few rules to convert to string:
- if list => use `", ".join`
- if boolean => `✔` for True, empty for False?
- if date => only "YYYY-MM-DD" maybe?
- ...

Wauplin: ^ **Note:** done a bit too quickly but let's uppercase the header values? could even be done in `print_as_table` for consistency

Wauplin: ^ btw, curious to see what Claude Code suggests in plan mode for this type of problem (printing a JSON as table in the terminal). I feel that we are not the first to deal with this

hanouticelina: lgtm, but we should still skip empty columns (i.e. we don t show columns where values are empty for all rows), no?

Wauplin: > (i.e. we don t show columns where values are empty for all rows), no?

makes sense if there are any yes (but not sure it happens very often)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -165,6 +167,29 @@ def _to_header(name: str) -> str:
     return s.upper()
 
 
+def _format_cell(value: object, max_len: int = _MAX_CELL_LENGTH) -> str:
```
hanouticelina, line 308: generated by Opus

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: 🙈 addressed in [9726908](https://github.com/huggingface/huggingface_hub/pull/3760/commits/972690893522ecfec97d6a94dc1cce59a60f0638)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -296,6 +299,34 @@ class OutputFormat(str, Enum):
...
+    if isinstance(value, datetime.datetime):
+        return value.strftime("%Y-%m-%d")
+    if isinstance(value, str) and re.match(r"^\d{4}-\d{2}-\d{2}T", value):
+        return value[:10]
+    if isinstance(value, list):
+        if not value:
+            return ""
+        cell = ", ".join(str(v) for v in value)
+    elif isinstance(value, dict):
+        return ""
```
Wauplin: not sure we want that. Let's just default to `str(value)` in case of dictionary?

src/huggingface_hub/cli/datasets.py
```diff
@@ -96,28 +96,15 @@ def datasets_ls(
     results = [
         api_object_to_dict(dataset_info)
         for dataset_info in api.list_datasets(
-            filter=filter, author=author, search=search, sort=sort_key, limit=limit, expand=expand
+            filter=filter,
+            author=author,
+            search=search,
+            sort=sort_key,
+            limit=limit,
+            expand=expand,  # type: ignore[arg-type]
```
Wauplin:
```suggestion
            expand=expand
```

(not needed no?)

src/huggingface_hub/cli/models.py
```diff
@@ -95,29 +95,15 @@ def models_ls(
     results = [
         api_object_to_dict(model_info)
         for model_info in api.list_models(
-            filter=filter, author=author, search=search, sort=sort_key, limit=limit, expand=expand
+            filter=filter,
+            author=author,
+            search=search,
+            sort=sort_key,
+            limit=limit,
+            expand=expand,  # type: ignore[arg-type]
```
Wauplin:
```suggestion
            expand=expand
```

same

src/huggingface_hub/cli/spaces.py
```diff
@@ -95,28 +95,15 @@ def spaces_ls(
     results = [
         api_object_to_dict(space_info)
         for space_info in api.list_spaces(
-            filter=filter, author=author, search=search, sort=sort_key, limit=limit, expand=expand
+            filter=filter,
+            author=author,
+            search=search,
+            sort=sort_key,
+            limit=limit,
+            expand=expand,  # type: ignore[arg-type]
```
Wauplin:
```suggestion
            expand=expand
```
same

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: yes, fixed in [e0ca920](https://github.com/huggingface/huggingface_hub/pull/3760/commits/e0ca920df4938c646202befc4498bccc7897c355) so now we do:
- None -> ""
- bool -> "✔" or ""
- datetime/date string -> "YYYY-MM-DD"
- empty list -> ""
- non empty list -> ", ".join(...) -> truncate
- everything else -> str(value) -> truncate

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -296,6 +299,32 @@ class OutputFormat(str, Enum):
...
+        return value[:10]
+    if isinstance(value, list):
+        if not value:
+            return ""
+        cell = ", ".join(str(v) for v in value)
+    else:
+        cell = str(value)
+    if len(cell) > max_len:
+        cell = cell[: max_len - 3] + "..."
+    return cell
```
```suggestion
    if not value:
        return ""
    if isinstance(value, bool):
        return "✔" if value else ""
    if isinstance(value, datetime.datetime):
        return value.strftime("%Y-%m-%d")
    if isinstance(value, str) and re.match(r"^\d{4}-\d{2}-\d{2}T", value):
        return value[:10]
    if isinstance(value, list):
        cell = ", ".join(str(v) for v in value)
    else:
        cell = str(value)
    if len(cell) > max_len:
        cell = cell[: max_len - 3] + "..."
    return cell
```

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -296,6 +299,30 @@ class OutputFormat(str, Enum):
 ]
 
 
+def _to_header(name: str) -> str:
+    s = re.sub(r"([a-z])([A-Z])", r"\1_\2", name)
```
Wauplin, line 304:
```suggestion
    """Convert a camelCase or PascalCase string to SCREAMING_SNAKE_CASE to be used as table header."""
    s = re.sub(r"([a-z])([A-Z])", r"\1_\2", name)
```

hanouticelina:
failing tests unrelated, merging

PR #3761, opened 2026-02-02
[Fix] Filter datasets by benchmark official

Wauplin:
Filtering datasets by benchmark:official should call 

https://huggingface.co/api/datasets?filter=benchmark:official&limit=10

not 

https://huggingface.co/api/datasets?benchmark=benchmark:official&limit=10

This PR fixes it (and should fix the CI as well).


hanouticelina (approved):
thanks!

Wauplin:
Finally :smile: 
image

PR #3769, opened 2026-02-04
[CLI] Add branch creation for non-existent revisions in upload_large_folder

bigmbigk:
Current `upload_large_folder` fails when trying to push to a branch that doesn't exist, while `upload` creates the branch.

This PR adds branch creation to `upload_large_folder` just like in `upload`.

### DIFF
--- a/src/huggingface_hub/cli/upload_large_folder.py
+++ b/src/huggingface_hub/cli/upload_large_folder.py
@@ -20,6 +20,7 @@ from typing import Annotated, Optional
 import typer
 
 from huggingface_hub import logging
+from huggingface_hub.errors import RevisionNotFoundError
 from huggingface_hub.utils import ANSI, disable_progress_bars
 
 from ._cli_utils import (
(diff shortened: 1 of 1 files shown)


hanouticelina (commented):
Hi @bigmbigk,
thanks for the suggestion! This is actually intentional by design, `hf upload-large-folder` is meant for long running uploads where we want to be _extra_ careful about the target destination. since these uploads can take time and are resumable, auto creating repos could lead to accidentally uploading large amounts of data to the wrong place (wrong branch, typos, etc.).
For user DX, we encourage using `hf repo create` beforehand:
```console
hf repo create my-namespace/my-repo --repo-type=model --exist-ok
hf upload-large-folder my-namespace/my-repo ./my-folder --repo-type=model
```
(`hf repo branch create` also exists to create a new branch for a repo)

we may reassess in the future if there's significant demand for this behavior, but for now we prefer keeping things like this.

thank you for your understanding 🤗

PR #3771, opened 2026-02-04
Pass kwargs to post init in dataclasses

zucchini-nlp:
As per title, needed for https://github.com/huggingface/transformers/pull/41250. I verified that it is the easiest way to keep BC for remote code, because we can't force everyone to use `dataclasses` as configs

We don't want to set `kwargs` as attr directly, as some keys have to be popped (i.e. deprecated key) and some keys have to be transformed before setting (e.g. key `attn_implememtation` is set to private `_attn_implememtation_internal`)

While it can possibly be done in post init by manually deleting and post processing attr, imo it's cleaner to pass `kwargs`

```python
# We could do smth like this right now, personally i find it ugly
def __post_init__(self):
    if hastattr(sefl, "deprecated_arg": del self.deprecated_arg
    if hastattr(sefl, "private_arg": self.move_to_private_property(private_arg)

# But after the suggested PR we don't need to do anything, and can keep existing code in transformers
def __post_init__(self, **kwargs):
        if (torch_dtype := kwargs.pop("torch_dtype", None)) is not None:
            self.dtype = self.dtype if self.dtype is not None else torch_dtype
```

### DIFF
--- a/src/huggingface_hub/dataclasses.py
+++ b/src/huggingface_hub/dataclasses.py
@@ -177,6 +177,16 @@ def strict(
 
             cls.__init__ = __init__  # type: ignore[method-assign]
 
+            # Define a default __post_init__ if not defined
+            if not hasattr(cls, "__post_init__"):
+
+                def __post_init__(self, **kwargs: Any) -> None:
+                    """Default __post_init__ to accept additional kwargs."""
+                    for name, value in kwargs.items():
+                        setattr(self, name, value)
+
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for the PR @zucchini-nlp! Makes perfect sense yes. I've updated the PR to add back the default behavior of setting kwargs as attributes but it can now be overwritten. I'll merge the PR now (failing CI is unrelated) and we'll ship it very soon :)

PR #3776, opened 2026-02-05
feat(spaces): hot-reload

cbensimon:
## Hot-reload command

``` bash
hf spaces hot-reload username/repo-name app.py               # Open an interactive editor to the remote app.py file
hf spaces hot-reload username/repo-name -f app.py            # Take local version from ./app.py and patch app.py in remote repo
hf spaces hot-reload username/repo-name app.py -f src/app.py # Take local version from ./src/app.py and patch app.py in remote repo
```

Hot-reload command with spaceId and repoFile is the highest-level command. It:
- downloads the file
- opens an interactive editor
- upload the new version with --hot-reload flag passed to create commit API
- Runs the hot-reloading summary command (unless --skip-summary flag)

With the -f option it is probably more agent friendly (not all agent CLIs support interactivity)

## Hot-reloading API

- Server defined in `spaces` package: https://github.com/huggingface/pyspaces/tree/main/spaces/reloading
- API types (`src/_hot_reloading_types.py`) are copy/pasted from https://github.com/huggingface/pyspaces/blob/main/spaces/reloading/types.py
- huggingface_hub should probably become the new source of truth for the types 

## Demo

TODO

### DIFF
--- a/src/huggingface_hub/cli/spaces.py
+++ b/src/huggingface_hub/cli/spaces.py
@@ -287,7 +287,7 @@ def _spaces_hot_reload_summary(
         if region["startLine"] == region["endLine"]:
             res += f"line {region['startLine'] - 1}"
         else:
-            res += f"lines {region['startLine'] - 1}-{region['endLine'] - 1}"
+            res += f"lines {region['startLine'] - 1}-{region['endLine']}"
         return res
 
     def display_event(event: ApiGetReloadEventSourceData) -> None:


src/huggingface_hub/_hot_reload/client.py
```diff
@@ -0,0 +1,36 @@
...
+        self.replica_hash = replica_hash
+        self.client = httpx.Client(
+            base_url=f"{base_host}/--replicas/+{replica_hash}",
+            headers=build_hf_headers(token=token),
+        )
+
+    def get_reload(self, reload_id: str) -> Iterator[ApiGetReloadEventSourceData]:
+        req = ApiGetReloadRequest(reloadId=reload_id)
+        with self.client.stream("POST", "/get-reload", json=req) as res:
+            assert res.status_code == 200, res.status_code  # TODO: Raise specific error ? Return ?
```
hanouticelina: let's keep error handling consistent with the rest of the codebase and avoids assert only runtime checks
```suggestion
            hf_raise_for_status(res)
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+            if event["data"]["updated"]:
+                typer.secho("⟳ UI updated", bold=True)
+            else:
+                typer.secho("∅ UI untouched", bold=True)
+        else:
+            assert_never(event["data"]["kind"])
+
+    first_client_events: dict[int, ApiGetReloadEventSourceData] = defaultdict()
+    for client_index, client in enumerate(clients):
+        if len(clients):
```
hanouticelina: since we're inside a `for client in enumerate(clients)` loop

```suggestion
        if len(clients) > 1:
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+            typer.secho(f"---- Replica {client.replica_hash} ----")
+        full_match = True
+        replay: list[ApiGetReloadEventSourceData] = []
+        for event_index, event in enumerate(client.get_reload(commit_sha)):
+            if client_index == 0:
+                first_client_events[event_index] = event
+            elif full_match := full_match and first_client_events[event_index] == event:
+                replay += [event]
+                continue
+            if len(replay) >= 0:
```
hanouticelina:
```suggestion
            if replay:
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+            repo_type="space",
+            repo_id=space_id,
+            filename=filename,
+            local_dir=temp_dir.name,
+        )
+        if not pbar_disabled:
+            enable_progress_bars()
+        editor_res = _cli_utils.editor_open(filepath)
+        if editor_res == "no-tty":
+            raise CLIError("Cannot open an editor (no TTY). Use -f flag to ho-reload from local path")
```
hanouticelina:
```suggestion
            raise CLIError("Cannot open an editor (no TTY). Use -f flag to hot-reload from local path")
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+        if editor_res == "no-editor":
+            raise CLIError("No editor found in local environment. Use -f flag to ho-reload from local path")
```
hanouticelina:
```suggestion
            raise CLIError("No editor found in local environment. Use -f flag to hot-reload from local path")
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+            typer.secho(event["data"]["codeLines"], italic=True)
+        elif event["data"]["kind"] == "ui":
+            if event["data"]["updated"]:
+                typer.secho("⟳ UI updated", bold=True)
+            else:
+                typer.secho("∅ UI untouched", bold=True)
+        else:
+            assert_never(event["data"]["kind"])
+
+    first_client_events: dict[int, ApiGetReloadEventSourceData] = defaultdict()
```
hanouticelina: `defaultdict` is usually for autocreating missing entries and here if I understood correctly, "missing key" is a meaningful signal, not something to auto fill
```suggestion
    first_client_events: dict[int, ApiGetReloadEventSourceData] = {}
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+        if len(clients):
+            typer.secho(f"---- Replica {client.replica_hash} ----")
+        full_match = True
+        replay: list[ApiGetReloadEventSourceData] = []
+        for event_index, event in enumerate(client.get_reload(commit_sha)):
+            if client_index == 0:
+                first_client_events[event_index] = event
+            elif full_match := full_match and first_client_events[event_index] == event:
+                replay += [event]
+                continue
```
hanouticelina: avoiding `KeyError` on missing indexes and uses `append` instead of `+= [event]`
```suggestion
            else:
                expected = first_events.get(event_index)
                if full_match and expected is not None and expected == event:
                    replay.append(event)
                    continue
                full_match = False
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +136,200 @@ def spaces_info(
...
+        if not (pbar_disabled := are_progress_bars_disabled()):
+            disable_progress_bars()
+        hf_hub_download(
+            repo_type="space",
+            repo_id=space_id,
+            filename=filename,
+            local_dir=temp_dir.name,
+        )
+        if not pbar_disabled:
+            enable_progress_bars()
```
hanouticelina: do we want to enable progress bar only if `hf_hub_download` does not fail? in that case, we should wrap this in `try`/`finally`
```suggestion
        pbar_was_disabled = are_progress_bars_disabled()
        if not pbar_was_disabled:
            disable_progress_bars()
        try:
            hf_hub_download(
            repo_type="space",
            repo_id=space_id,
            filename=filename,
            local_dir=temp_dir.name,
        )
        finally:
            if not pbar_was_disabled:
                enable_progress_bars()
```

src/huggingface_hub/hf_api.py
```diff
@@ -10144,7 +10157,12 @@ def list_daily_papers(
             yield PaperInfo(**paper)
 
     def auth_check(
-        self, repo_id: str, *, repo_type: Optional[str] = None, token: Union[bool, str, None] = None
+        self,
+        repo_id: str,
+        *,
+        repo_type: Optional[str] = None,
+        token: Union[bool, str, None] = None,
+        write: bool = False,
```
hanouticelina, line 10165: is this to check write access before opening the editor?

src/huggingface_hub/cli/spaces.py
cbensimon: Good catch. Indeed, we only want to display replica info if there's more than 1 replica

src/huggingface_hub/_hot_reload/client.py
cbensimon: Done

src/huggingface_hub/cli/spaces.py
cbensimon: Good catch. I initially used a defaultdict then forgot to switch back to regular after upgrading the logic. Done in cebfe1c0

src/huggingface_hub/cli/spaces.py
cbensimon: Done slightly differently in https://github.com/huggingface/huggingface_hub/pull/3776/changes/53b9b07abe69e9b7d96df68d885c4d1c3e1b18ed when addressing https://github.com/huggingface/huggingface_hub/pull/3776#discussion_r2817530927

src/huggingface_hub/cli/spaces.py
cbensimon: Indeed, this is more correct

src/huggingface_hub/hf_api.py
cbensimon: Yes, so we avoid a failure after user has spent some time editing the file.

Here -> https://github.com/huggingface/huggingface_hub/blob/53b9b07abe69e9b7d96df68d885c4d1c3e1b18ed/src/huggingface_hub/cli/spaces.py#L207-L211

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -512,3 +517,24 @@ def _get_transformers_update_command() -> str:
...
+
+
+def editor_open(filepath: str) -> Union[int, Literal["no-tty", "no-editor"]]:
+    if not (sys.stdin.isatty() and sys.stdout.isatty()):
+        return "no-tty"
+    if (editor_command := _get_editor_command()) is None:
+        return "no-editor"
+    command = [*shlex.split(editor_command), filepath]
+    res = subprocess.run(command, start_new_session=True)
+    return res.returncode
```
hanouticelina: since `_cli_utils.py` is shared by all CLI commands, can we move these helpers to `src/huggingface_hub/cli/spaces.py` given that it's only used for `hf spaces hot-reload` CLI?

src/huggingface_hub/hf_api.py
hanouticelina: the param is added to a public API method so ideally this should be done in its own PR with proper documentation (just to be able to track the change independently) or at minimum adding the parameter to the docstring's Args section here

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +135,202 @@ def spaces_info(
...
+        _spaces_hot_reloading_summary(
+            api=api,
+            space_id=space_id,
+            commit_sha=commit_info.oid,
+            filepath=filepath if local_file else os.path.basename(filepath),
+            token=token,
+        )
+
+
+def _spaces_hot_reloading_summary(
```
hanouticelina: (nit) to keep some naming consistency "hot-reload" vs "hot-reloading"
```suggestion
def _spaces_hot_reload_summary(
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +135,202 @@ def spaces_info(
...
+        filepath = local_file
+        filename = local_file if filename is None else filename
+    elif filename:
+        if not skip_checks:
+            api.auth_check(
+                repo_type="space",
+                repo_id=space_id,
+                write=True,
+            )
+        temp_dir = tempfile.TemporaryDirectory()
```
hanouticelina, line 220: better to use `TemporaryDirectory` as a context manager to ensure the directory is deleted deterministically

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +135,202 @@ def spaces_info(
...
+    local_file: Annotated[
+        Optional[str],
+        typer.Option(
+            "--local-file",
+            "-f",
+            help="Path of local file. Interactive editor mode if not specified",
+        ),
+    ] = None,
+    skip_checks: Annotated[bool, typer.Option(help="Skip hot-reload compatibility checks.")] = False,
+    skip_summary: Annotated[bool, typer.Option(help="Skip summary display after hot-reloaded triggered")] = False,
```
hanouticelina: (nit) 
```suggestion
    skip_summary: Annotated[bool, typer.Option(help="Skip summary display after hot-reload is triggered")] = False,
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +135,202 @@ def spaces_info(
...
+    Only works with Gradio SDK (6.1+)
+    Opens an interactive editor unless --local-file/-f is specified.
+
+    This command patches the live Python process using https://github.com/breuleux/jurigged
+    (AST-based diffing, in-place function updates, etc.), integrated with Gradio's native hot-reload support
+    (meaning that Gradio demo object changes are reflected in the UI)
+
+    Usage examples:
+
+    \b
```
hanouticelina, line 184:
```suggestion
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +135,202 @@ def spaces_info(
     except RevisionNotFoundError as e:
         raise CLIError(f"Revision '{revision}' not found on '{space_id}'.") from e
     print(json.dumps(api_object_to_dict(info), indent=2))
+
+
+@spaces_cli.command("hot-reload")
+def spaces_hot_reload(
```
hanouticelina, line 145: could we move some of the logic here into the `_hot_reload` module? as much as possible, hot-reload domain behaviors should live in the `_hot_reload` module, not in the CLI. the CLI file should mainly handle CLI UX: command options, formatting/output text, and user-facing messages (`CLIError`)...etc
not sure how much of this is possible though.

src/huggingface_hub/cli/spaces.py
cbensimon: Actually useful for `--help` CLI rendering (auto wrapping otherwise, which makes next lines misaligned)

hanouticelina: ah yes sorry, let's ignore this comment

src/huggingface_hub/cli/_cli_utils.py
cbensimon: Done

src/huggingface_hub/cli/spaces.py
cbensimon: Done

src/huggingface_hub/cli/spaces.py
cbensimon: Not possible here as we only need `TemporaryDirectory` in the `local_file == None` branch. With a context manager we would either need to create the temp dir even when useless, or copy filepath contents after editor_open.

src/huggingface_hub/hf_api.py
cbensimon: Indeed, I forgot to document it. Done

src/huggingface_hub/cli/spaces.py
cbensimon: Well I am almost only seeing docstring, `CLIError` and `typer.secho` in `spaces_hot_reload` and `_spaces_hot_reload_summary` functions. Maybe that this could be refactored but I can't see how.

EDIT: I'll try to move the replicas / event iteration + replay logic (without CLI rendering) inside `_hot_reload`. Won't make a ton of difference but better than nothing

src/huggingface_hub/cli/spaces.py
hanouticelina: not sure if I'm missing something here but imo it is possible to use `TemporaryDirectory` as a context manager without creating it in the `local_file` path. We can keep the `local_file` branch as is, and use:
```python
with tempfile.TemporaryDirectory() as temp_dir:
    ...
    # download + edit
    commit_info = api.upload_file(path_or_fileobj=filepath, ...)
```
wdyt?

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+        if space_info.sdk != "gradio":
+            raise CLIError(f"Hot-reloading is only available on Gradio SDK. Found {space_info.sdk} SDK")
+        if (card_data := space_info.card_data) is None:
+            raise CLIError(f"Unable to read cardData for Space {space_id}")
+        if (sdk_version := card_data.sdk_version) is None:
+            raise CLIError(f"Unable to read sdk_version from {space_id} cardData")
+        if version.parse(sdk_version) < version.Version(HOT_RELOADING_MIN_GRADIO):
+            raise CLIError(f"Hot-reloading requires Gradio >= {HOT_RELOADING_MIN_GRADIO} (found {sdk_version})")
+
+    if local_file:
```
hanouticelina, line 205: (nit) early return to remove remove the `else` in line 237
```suggestion
    if local_file is None and filename is None:
        raise CLIError("Either filename or --local-file/-f must be specified")
    if local_file:
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+        filepath = local_file
+        filename = local_file if filename is None else filename
```
hanouticelina: (nit) just to avoid reusing `filename` for both cli input and resolved target path
```suggestion
        local_path = local_file
        path_in_repo = local_file if filename is None else filename
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+    elif filename:
+        if not skip_checks:
+            api.auth_check(
+                repo_type="space",
+                repo_id=space_id,
+                write=True,
+            )
+        temp_dir = tempfile.TemporaryDirectory()
+        filepath = os.path.join(temp_dir.name, filename)
```
hanouticelina: related to the comment line 207
```suggestion
        path_in_repo = filename
        local_path = os.path.join(temp_dir.name, path_in_repo)
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+        if editor_res != 0:
+            raise CLIError(f"Editor returned a non-zero exit code while attempting to edit {filepath}")
+    else:
+        raise CLIError("Either filename or --local-file/-f must be specified")
+
+    commit_info = api.upload_file(
+        repo_type="space",
+        repo_id=space_id,
+        path_or_fileobj=filepath,
+        path_in_repo=filename,
```
hanouticelina: related to the comment line 207
```suggestion
        path_or_fileobj=local_path,
        path_in_repo=path_in_repo,
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+
+    clients = [
+        ReloadClient(
+            host=space_host,
+            subdomain=space_subdomain,
+            replica_hash=hash,
+            token=token,
+        )
+        for hash, _ in hot_reloading.replica_statuses
+    ]
```
hanouticelina: (nit) for readability 
```suggestion
    clients = [
        ReloadClient(
            host=space_host,
            subdomain=space_subdomain,
            replica_hash=replica_hash,
            token=token,
        )
        for replica_hash, _ in hot_reloading.replica_statuses
    ]
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+        elif event["data"]["kind"] == "run":
+            typer.secho(f"▶ Run {render_region(event['data']['region'])}", bold=True)
+            typer.secho(event["data"]["codeLines"], italic=True)
+        elif event["data"]["kind"] == "ui":
+            if event["data"]["updated"]:
+                typer.secho("⟳ UI updated", bold=True)
+            else:
+                typer.secho("∅ UI untouched", bold=True)
+        else:
+            assert_never(event["data"]["kind"])
```
hanouticelina, line 315: (nit) for readability
```suggestion
    def display_event(event: ApiGetReloadEventSourceData) -> None:
          data = event["data"]
          kind = data["kind"]
      
          if kind == "error":
              typer.secho("✘ Unexpected hot-reloading error", bold=True)
              typer.secho(data["traceback"], italic=True)
          elif kind == "exception":
              typer.secho(f"✘ Exception at {render_region(data['region'])}", bold=True)
              typer.secho(data["traceback"], italic=True)
          elif kind == "add":
              typer.secho(f"✔︎ Created {data['objectName']} {data['objectType']}", bold=True)
          elif kind == "delete":
              typer.secho(f"∅ Deleted {data['objectName']} {data['objectType']}", bold=True)
          elif kind == "update":
              typer.secho(f"✔︎ Updated {data['objectName']} {data['objectType']}", bold=True)
          elif kind == "run":
              typer.secho(f"▶ Run {render_region(data['region'])}", bold=True)
              typer.secho(data["codeLines"], italic=True)
          elif kind == "ui":
              if data["updated"]:
                  typer.secho("⟳ UI updated", bold=True)
              else:
                  typer.secho("∅ UI untouched", bold=True)
          else:
              assert_never(kind)
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -123,3 +140,223 @@ def spaces_info(
...
+    for client_index, client in enumerate(clients):
+        if len(clients) > 1:
+            typer.secho(f"---- Replica {client.replica_hash} ----")
+        full_match = True
+        replay: list[ApiGetReloadEventSourceData] = []
+        for event_index, event in enumerate(client.get_reload(commit_sha)):
+            if client_index == 0:
+                first_client_events[event_index] = event
+            elif full_match := full_match and first_client_events.get(event_index) == event:
+                replay += [event]
```
hanouticelina: (nit) readability
```suggestion
                replay.append(event)
```

src/huggingface_hub/constants.py
```diff
@@ -279,3 +279,12 @@ def list_files(repo_id: str):
 default_xet_cache_path = os.path.join(HF_HOME, "xet")
 HF_XET_CACHE = os.getenv("HF_XET_CACHE", default_xet_cache_path)
 HF_HUB_DISABLE_XET: bool = _is_true(os.environ.get("HF_HUB_DISABLE_XET"))
+
+# CLI Editor
+PREFERRED_EDITORS = (
```
hanouticelina: maybe this should be defined in `cli/spaces.py` since it's local to the cli
```python
_PREFERRED_EDITORS = (
      ("code", "code --wait"),
      ("nvim", "nvim"),
      ("nano", "nano"),
      ("vim", "vim"),
      ("vi", "vi"),
  )
```

src/huggingface_hub/utils/_sse_client.py
```diff
@@ -0,0 +1,136 @@
+"""
+Server Side Events (SSE) client for Python.
```
hanouticelina: let's add a short vendoring header at the top of the file
```suggestion
Vendored file: Server Side Events (SSE) client for Python.

Source:
- Author: Maxime Petazzoni <maxime.petazzoni@bulix.org>
- Repository: https://github.com/mpetazzoni/sseclient
- File: https://github.com/mpetazzoni/sseclient/blob/main/sseclient/__init__.py

License:
- Apache-2.0 (from upstream project)
```

src/huggingface_hub/cli/spaces.py
cbensimon: Done in https://github.com/huggingface/huggingface_hub/commit/5f0e5985f4a7bb9b560f53cb585e57bac16ab91d

src/huggingface_hub/cli/spaces.py
cbensimon: Well it will physically create the temporary directory when we do not need it (not that serious in terms of perf of course but semantically misleading)

src/huggingface_hub/cli/spaces.py
cbensimon: Unfortunately, nor Pyright or Mypy successfully infer the type of filepath / filename (saying it might be Unbound, which is actually false) if we do it with early-return style (which I agree is better) instead of else branch

src/huggingface_hub/cli/spaces.py
cbensimon: Indeed, better. But only for `local_path`. `path_in_repo` makes thing much more complex to understand (especially because `filename` is the CLI argument name and is the argument name of `hf_hub_download`)

src/huggingface_hub/cli/spaces.py
cbensimon: Already discussed in https://github.com/huggingface/huggingface_hub/pull/3776#discussion_r2823489102

src/huggingface_hub/cli/spaces.py
cbensimon: Refactored inside _hot_reload

src/huggingface_hub/cli/spaces.py
cbensimon: Same as https://github.com/huggingface/huggingface_hub/pull/3776#discussion_r2823456862 : pyright and mypy can't correctly infer downstream types (discriminated union based)

src/huggingface_hub/cli/spaces.py
cbensimon: As you like (I find += syntax more explicit / safe since failing in global scope regarding the mutating nature of the operation, but it's really a matter of taste of course)

src/huggingface_hub/constants.py
cbensimon: Indeed since we moved `editor_open` outside of CLI utils, this constant does not make sense in global scope anymore

src/huggingface_hub/hf_api.py
```diff
@@ -4460,7 +4464,12 @@ def _payload_as_ndjson() -> Iterable[bytes]:
             **headers,
         }
         data = b"".join(_payload_as_ndjson())
-        params = {"create_pr": "1"} if create_pr else None
+
+        params: dict[str, Any] = {}
+        if create_pr:
+            params |= {"create_pr": 1}
+        if _hot_reload:
+            params |= {"hot_reload": 1}
```
Wauplin: (to be consistent with syntax in the rest of the file)

```suggestion
        params: dict[str, Any] = {}
        if create_pr:
            params["create_pr"] = "1"
        if _hot_reload:
            params["hot_reload"] = "1"
```

src/huggingface_hub/hf_api.py
```diff
@@ -10204,6 +10226,7 @@ def auth_check(
         if repo_type not in constants.REPO_TYPES:
             raise ValueError(f"Invalid repo type, must be one of {constants.REPO_TYPES}")
         path = f"{self.endpoint}/api/{repo_type}s/{repo_id}/auth-check"
+        path = f"{path}/write" if write else path
```
Wauplin: (usually use this syntax in the codebase)

```suggestion
        if write:
            path = f"{path}/write"
```

src/huggingface_hub/_hot_reload/client.py
```diff
@@ -0,0 +1,88 @@
+import json
```
Wauplin, line 16: (nit) add license in header like in https://github.com/huggingface/huggingface_hub/blob/ec80075fe9a47778123c0824c89d866431d06b56/src/huggingface_hub/__init__.py#L1

(with 2026)

Wauplin: same for other new modules

src/huggingface_hub/_hot_reload/sse_client.py
```diff
@@ -0,0 +1,144 @@
+"""
```
Wauplin, line 1: even if this is not specific to hot reload, could you move it to the `_hot_reload` module please? If we ever need a SSE Client for a feature in the future we will move it back as a "official utility" but until then I prefer to have it isolated

src/huggingface_hub/cli/spaces.py
hanouticelina: ok to keep it like this then 👍

src/huggingface_hub/_hot_reload/client.py
cbensimon: Done for all the new modules (every module inside `src/huggingface_hub/_hot_reload` except vendored `sse_client.py`)

src/huggingface_hub/_hot_reload/sse_client.py
cbensimon: Done

lhoestq:
cc @kszucs for viz

IIUC we'll be able upload a file to a Space without rebuilding it (not prio but could be useful for https://github.com/kszucs/faceberg at some point)

cbensimon:
@lhoestq not actually: this is only meant to reload Python files at runtime (Gradio Spaces now expose an API to trigger Python module updates backed by https://github.com/breuleux/jurigged)

lhoestq:
good to know thanks !

PR #3782, opened 2026-02-06
Fix `hf jobs` commands crashing without a TTY

davanstrien:
Use `shutil.get_terminal_size()` instead of `os.get_terminal_size()` in `_tabulate` so that commands like `hf jobs hardware` and `hf jobs scheduled ps` gracefully fall back to a default terminal width when no TTY is attached (e.g. piped output, subprocesses).


julien-c:
this makes me think maybe we should have a CI job that checks whether `hf` is nicely used by an agent with a couple of different tasks

PR #3783, opened 2026-02-06
Make `hf jobs logs` agent-friendly with `--follow/-f` and `--tail/-n` flags

davanstrien:
Opening as a draft PR to start. 

Currently, `hf jobs logs <id>` streams logs indefinitely via SSE until the job completes, with no way to opt out. This makes it harder to use from agents (e.g. Claude Code's Bash tool times out at 2 min).

I've seen Claude, etc., trying to pass --follow --no-follow to escape this behaviour so it might be better to make the pattern easier for agents?

The PR aligns `hf jobs logs` with Docker's `docker logs` convention i.e. print existing logs and exit by default, opt into streaming with `-f`:

- `hf jobs logs <id>` — prints available logs and exits (non-blocking)
- `hf jobs logs -f <id>` — streams until the job completes (follow mode)
- `hf jobs logs --tail N <id>` — shows only the last N lines

| | Docker | HF (this PR) |
|---|---|---|
| Print & exit | `docker logs <id>` (default) | `hf jobs logs <id>` (default) |
| Stream/follow | `docker logs -f <id>` | `hf jobs logs -f <id>` |
| Last N lines | `docker logs --tail N <id>` | `hf jobs logs --tail N <id>` |

## Discussion: CLI default behavior

**Python API**: No breaking change — `fetch_job_logs()` gains a `follow` parameter that defaults to `True`, preserving existing behavior.

**CLI**: This PR changes the default from streaming (blocking) to snapshot (non-blocking), matching Docker's convention. However, this is a behaviour change for existing CLI users who expect `hf jobs logs <id>` to stream until completion.

An alternative would be to keep the current default and require `--no-follow` to opt out, though this diverges from Docker's convention and keeps the agent-unfriendly default. I've seen agents try this too. 

Note: `hf jobs run` (without `--detach`) is unaffected and still streams logs after launching.

cc @lhoestq (I couldn't find a previous discussion of this, but my Slack searching skills may have failed me).

## Test plan
- [x] Unit tests added (5 new tests in `TestJobsCommand`)
- [x] `make quality` passes
- [x] Smoke tested with real jobs on HF infrastructure

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -10263,6 +10263,7 @@ class HfApi:
         timeout: int,
         skip_previous_events_on_retry: bool,
         double_check_job_has_finished_on_status_code_or_error: tuple[Union[int, Type[Exception]], ...],
+        follow: bool = True,
         namespace: Optional[str] = None,
         token: Union[bool, str, None] = None,
     ) -> Iterable[dict[str, Any]]:
@@ -10270,7 +10271,7 @@ class HfApi:
             namespace = self.whoami(token=token)["name"]
         # We don't use http_backoff since we need to check ourselves if the job is still running
         nb_tries = 0
-        max_retries = 5
+        max_retries = 5 if follow else 0
         min_wait_time = 1
         max_wait_time = 10
         sleep_time = 0
@@ -10315,7 +10316,9 @@ class HfApi:
                     and isinstance(getattr(err.__context__, "__cause__", None), TimeoutError)
                 )
                 if is_no_new_line_timeout:
-                    # job is likely finished
+                    if not follow:
+                        break  # no-follow mode: got all buffered events
+                    # follow mode: job is likely finished
                     pass
                 elif type(err) in double_check_job_has_finished_on_status_code_or_error:
                     pass
@@ -10353,9 +10356,8 @@ class HfApi:
                 The namespace where the Job is running. Defaults to the current user's namespace.
 
             follow (`bool`, *optional*):
-                If `True` (default), stream logs in real-time until the job completes (blocking).
-                If `False`, fetch only the currently available logs and return immediately (non-blocking).
-                Use `follow=False` when calling from scripts, agents, or CI pipelines that should not block.
+                If `True`, stream logs in real-time until the job completes (blocking).
+                If `False` (default), fetch only the currently available logs and return immediately (non-blocking).
 
             token `(Union[bool, str, None]`, *optional*):
                 A valid user access token. If not provided, the locally saved token will be used, which is the
@@ -10376,10 +10378,6 @@ class HfApi:
             ...     print(log)
             ```
         """
-        if not follow:
-            yield from self._fetch_job_logs_no_follow(job_id=job_id, namespace=namespace, token=token)
-            return
-
         # - We need to retry because sometimes the /logs doesn't return logs when the job just started.
         #   (for example it can return only two lines: one for "Job started" and one empty line)
         # - Timeouts can happen in case of build errors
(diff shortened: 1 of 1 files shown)


davanstrien:
cc @Wauplin for viz - happy to open issue/discuss first but thought draft PR might make it easier to discuss if it makes sense

Wauplin:
> Python API: No breaking change — fetch_job_logs() gains a follow parameter that defaults to True, preserving existing behavior.
>
> CLI: This PR changes the default from streaming (blocking) to snapshot (non-blocking), matching Docker's convention. However, this is a behavior change for existing CLI users who expect hf jobs logs <id> to stream until completion.

I'd be in favor of breaking `fetch_job_logs` default behavior as well. I don't think many people is using it directly and I'd rather have a consistent API/CLI than non breaking the API.

davanstrien:
Marked as ready for review, but happy to iterate on the best approach to this first. Updated to make CLI and Python consistent. 

Currently

- `hf jobs run` without `--detach`: streams logs after launching a job
- `hf jobs uv run` without `--detach`: also streams 

i.e. both call `follow=True`. No strong opinion but I think this might be okay? i.e. it follows the `docker run` convention?

src/huggingface_hub/hf_api.py
```diff
@@ -10385,6 +10399,49 @@ def fetch_job_logs(
...
+        all historical log lines quickly, then pauses waiting for new events (with
+        keep-alive every ~30s). When the short timeout fires without new data, we
+        treat that as "all buffered logs received" and return gracefully.
+        """
+        if namespace is None:
+            namespace = self.whoami(token=token)["name"]
+
+        # Short timeout: the server replays historical logs quickly, then pauses.
+        # 5 seconds is enough to receive all buffered logs.
+        read_timeout = 5
```
davanstrien: This may not be a good assumption by Claude

src/huggingface_hub/cli/jobs.py
```diff
@@ -312,20 +315,47 @@ def jobs_run(
...
+
+    By default, prints currently available logs and exits (non-blocking).
+    Use --follow/-f to stream logs in real-time until the job completes.
+    """
     api = get_hf_api(token=token)
     try:
-        for log in api.fetch_job_logs(job_id=job_id, namespace=namespace):
+        logs = api.fetch_job_logs(job_id=job_id, namespace=namespace, follow=follow)
+        if tail is not None:
+            from collections import deque
```
Wauplin: move import at module level

src/huggingface_hub/cli/jobs.py
```diff
@@ -312,20 +315,47 @@ def jobs_run(
...
+
+            logs = deque(logs, maxlen=tail)
```
Wauplin, line 361: maybe raise a valueerror if both`--follow` and `--tail` are used? (they won't work together, right?)

src/huggingface_hub/hf_api.py
```diff
@@ -10389,6 +10403,49 @@ def fetch_job_logs(
                 log = event["data"]
                 yield log
 
+    def _fetch_job_logs_no_follow(
```
Wauplin: instead of having a new helper, could we define add a `follow` parameter in `_fetch_running_job_sse` ?

(if not possible that's fine)

src/huggingface_hub/hf_api.py
Wauplin: I feel that it's fine (smaller might miss some logs on slow environments, higher would degrade UX)

src/huggingface_hub/cli/jobs.py
davanstrien: added this + a test. test might be overkill but maybe good to catch changes to follow etc defaults later.

src/huggingface_hub/hf_api.py
davanstrien: Claude attempted a refactor of this. Don't think it looks too gnarly but let me know if you think the branching in `_fetch_running_job_sse` looks too much now.

src/huggingface_hub/hf_api.py
```diff
@@ -10339,6 +10342,7 @@ def fetch_job_logs(
         *,
         job_id: str,
         namespace: Optional[str] = None,
+        follow: bool = False,
```
hanouticelina, line 10345: this is a breaking change no? we should probably set the default value to `True` and only the CLI should default to `False`

hanouticelina: didn't see @Wauplin's [comment](https://github.com/huggingface/huggingface_hub/pull/3783#issuecomment-3860650340). ok to break this then 👍

src/huggingface_hub/cli/jobs.py
Wauplin: > but maybe good to catch changes to follow etc defaults later.

Yep :+1: Thanks for adding it

PR #3793, opened 2026-02-10
Fix ReadTimeout crash in no-follow job logs

davanstrien:
- hf jobs logs <id> (without --follow) crashes with `httpx.ReadTimeout` when the SSE stream pauses for longer than the 5-second timeout
 - The is_no_new_line_timeout check only recognised `httpx.NetworkError` wrapping a `TimeoutError`, but httpx actually raises `httpx.ReadTimeout`, not a `NetworkError`
- Add `httpx.ReadTimeout` / `httpcore.ReadTimeout` to the timeout detection, so the no-follow path breaks cleanly

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -10310,11 +10310,7 @@ class HfApi:
             except KeyboardInterrupt:
                 break
             except (httpx.HTTPError, httpcore.TimeoutException) as err:
-                is_no_new_line_timeout = isinstance(err, (httpx.ReadTimeout, httpcore.ReadTimeout)) or (
-                    isinstance(err, httpx.NetworkError)
-                    and err.__context__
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/hf_api.py
```diff
@@ -10310,7 +10310,7 @@ def _fetch_running_job_sse(
             except KeyboardInterrupt:
                 break
             except (httpx.HTTPError, httpcore.TimeoutException) as err:
-                is_no_new_line_timeout = (
+                is_no_new_line_timeout = isinstance(err, (httpx.ReadTimeout, httpcore.ReadTimeout)) or (
```
hanouticelina: i guess now we don't need for the `NetworkError` + `TimeoutError` handing. `httpx` raises `ReadTimeout` directly for read timeouts so it could just be:
```python
is_no_new_line_timeout = isinstance(err, (httpx.ReadTimeout, httpcore.ReadTimeout))
```

PR #3799, opened 2026-02-11
Align `hf jobs ps` output with standard CLI pattern

davanstrien:
Replace custom `_print_output`/`_tabulate` in `jobs_ps` and `scheduled_ps` with the shared `print_list_output()` from `_cli_utils.py`, matching the pattern used by `hf endpoints ls`, `hf models ls`, etc. (i.e. follow PR #3735).

Changes:
- `hf jobs ps` and `hf jobs scheduled ps` now use `print_list_output()`
- Add `--format table|json` (default: table) and `-q/--quiet` options
- Add hidden `--json` alias (per maintainer request)
- Job IDs are no longer truncated in table output
- Long COMMAND/IMAGE columns truncated via `_format_cell()` for readability
- Filter logic (`-f status=`, `-f label=`) preserved identically
- `jobs_stats` and `jobs_hardware` unchanged (live-updating display)

Breaking: removes old Go-template `--format '{{.id}}'` syntax. Use `-q` for IDs or `--format json | jq` for custom field extraction. cc @lhoestq in case you think this is better to keep? We could also give some tips for the new approach if we remove it. 

I didn't touch the HF jobs hardware, but would it also be worth doing in the future. Not sure if JSON adds a lot there but maybe easier for agents or pipelines if we end up with lots of hardware options/info to parse?

### DIFF
--- a/tests/test_file_download.py
+++ b/tests/test_file_download.py
@@ -41,7 +41,7 @@ from huggingface_hub.file_download import (
     http_get,
     try_to_load_from_cache,
 )
-from huggingface_hub.utils import SoftTemporaryDirectory, get_session, hf_raise_for_status
+from huggingface_hub.utils import SoftTemporaryDirectory, WeakFileLock, get_session, hf_raise_for_status
 from huggingface_hub.utils._headers import build_hf_headers
 from huggingface_hub.utils._http import _http_backoff_base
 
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/cli/jobs.py
```diff
@@ -592,43 +598,50 @@ def jobs_ps(
...
+    def row_fn(item: dict[str, Any]) -> list[str]:
+        status = item.get("status", {})
+        cmd = item.get("command") or []
+        command_str = " ".join(cmd) if cmd else "N/A"
+        return [
+            str(item.get("id", "")),
+            _format_cell(item.get("docker_image") or "N/A"),
+            _format_cell(command_str),
+            item["created_at"][:19].replace("T", " ") if item.get("created_at") else "N/A",
+            str(status.get("stage", "UNKNOWN") if isinstance(status, dict) else "UNKNOWN"),
```
hanouticelina: no need to check if status is a dict, it comes from `api_object_to_dict`, nested objects are always dicts.
```suggestion
            str(status.get("stage", "UNKNOWN"),
```

src/huggingface_hub/cli/jobs.py
```diff
@@ -592,43 +598,50 @@ def jobs_ps(
...
         if not _matches_filters(job.labels or {}, labels_filters):
             continue
+        filtered_jobs.append(job)
 
-        # Create row
-        rows.append([job_id, image_or_space, command_str, created_at, status])
-
-    # Handle empty results
-    if not rows:
+    if not filtered_jobs:
```
hanouticelina, line 658: This will print "No jobs found" regardless of the output format, meaning:
- `hf jobs ps --format json` --> outputs `No jobs found` (not a valid json, a consumer piping to `jq` would crash)
- `hf jobs ps -q ` --> outputs `No jobs found` (-q mode is supposed to output only ID, one per line, so scripts that will parse this would treat `No jobs found` as a job id).

i would do something like that:
```python
    if not filtered_jobs:
          if not quiet and format != OutputFormat.json:
              filters_msg = f" matching filters: {', '.join([f'{k}{o}{v}' for k, o, v in filters])}" if filters else ""
              print(f"No jobs found{filters_msg}")
          elif format == OutputFormat.json:
              print("[]")
          return
```

hanouticelina: same thing for `scheduled_ps`

src/huggingface_hub/cli/jobs.py
```diff
@@ -870,33 +879,59 @@ def scheduled_ps(
...
+        command_str = " ".join(cmd) if cmd else "N/A"
+        return [
+            str(item.get("id", "")),
+            str(item.get("schedule") or "N/A"),
+            _format_cell(job_spec.get("docker_image") or "N/A"),
+            _format_cell(command_str),
+            last_job_at,
+            next_run,
+            str(item.get("suspend", False)),
+        ]
```
hanouticelina, line 989: no need for the isinstance checks, `api_object_to_dict` guarantees dicts for nested dataclasses and str type for datetimes
```suggestion
    def row_fn(item: dict[str, Any]) -> list[str]:
          job_spec = item.get("job_spec", {})
          status = item.get("status", {})
          last_job = status.get("last_job")
          cmd = job_spec.get("command") or []
          last_job_at = "N/A"
          if last_job and last_job.get("at"):
              last_job_at = last_job["at"][:19].replace("T", " ")
          next_run = "N/A"
          if status.get("next_job_run_at"):
              next_run = status["next_job_run_at"][:19].replace("T", " ")
          command_str = " ".join(cmd) if cmd else "N/A"
          return [
              str(item.get("id", "")),
              str(item.get("schedule") or "N/A"),
              _format_cell(job_spec.get("docker_image") or "N/A"),
              _format_cell(command_str),
              last_job_at,
              next_run,
              str(item.get("suspend", False)),
          ]
```

src/huggingface_hub/cli/jobs.py
davanstrien: good catch! Thanks!

davanstrien: should be resolved now. Also added some tests for that but maybe overkill. Happy to remove if you think it's not necessary to test

davanstrien:
> Looks good to me, thanks! my only concern is about breaking the custom template format output, we're trying to minimize CLI breaking changes as much as possible right now. is there a specific reason for removing it?

Mostly, the goal was to align the Jobs CLI with the other commands and avoid too many special cases for Jobs, but I'm happy to add it back in if preferred?

lhoestq:
I'd rather still keep support the old formatting in addition to the new one if it's easy, but not a big deal if you think it's more work and not worth it

Wauplin:
> Breaking: removes old Go-template --format '{{.id}}' syntax.

Personally I'm fine with breaking this for now. We can always re-implement it in a subsequent PR, in a way that would be the same for all list commands in the CLI

davanstrien:
Added backward compatibility for Go-template. It should match what was there before. Also added some basic tests for that but I can remove if we prefer not to have those.

davanstrien:
> @davanstrien I took the liberty of pushing a commit to avoid redundant data extraction in the Go template path, `row_fn` is now defined once and shared between template and `print_list_output`, instead of re-extracting fields from `JobInfo` objects in a separate branch. looks good to me, thank you for working on this!

Thanks, looks much nicer now!

hanouticelina:
merging!

PR #3805, opened 2026-02-12
[CLI] Add `hf extension`

hanouticelina:
_DISCLAIMER_: the code was mostly generated by Claude Code with iterations and corrections from me.

This PR adds an extension mechanism to the `hf` CLI. Extensions are standalone executables hosted in GitHub repositories that users can install, run, and remove with simple commands. Mostly inspired by [`gh extension`](https://cli.github.com/manual/gh_extension).

_How it works:_
an extension is a GitHub repo that follows a naming convention:

- The repo name **must** start with `hf-` (e.g. `hf-claude`, `hf-myextension`)
- The repo contains a root level executable with the same name as the repo (e.g. `hf-claude`)
- extensions cannot shadow built-in cmds (e.g. `hf extension install hf-cache` is rejected), still not sure about that though. 
- tthe repo just needs a single executable at the root. A `manifest.json` is generated locally at install time.

_Install format_

```console
# defaults to the `huggingface` GitHub org
hf extensions install hf-claude

#  any GitHub owner
hf extensions install hanouticelina/hf-claude
```

Install downloads the single executable file directly from `raw.githubusercontent.com`. The executable is saved to `~/.local/share/hf/extensions/hf-<name>/` with a `manifest.json` alongside it (for bookeeping).

Extensions can be run in two ways:

```bash
# Explicit
hf extensions exec claude

# Direct dispatch i.e. if `claude` isn't a built-in command, `hf` checks for an installed extension
hf claude 
```

Direct dispatch works by intercepting unknown commands in `main()` before typer's error handling kicks in. not sure if this is the best way to do that? 🤔 Opus 4.6 didn't suggest a better solution.

You can try it now with https://github.com/hanouticelina/hf-claude:

```console
hf extensions install hanouticelina/hf-claude

# Already installed (should error)
hf extensions install hanouticelina/hf-claude

# Force reinstall
hf extensions install --force hanouticelina/hf-claude

# Missing hf- prefix (should error)
hf extensions install hanouticelina/claude

# Conflict with built-in (should error)
hf extensions install hf-cache

# list
hf extensions list

# run extension
hf extensions exec claude
hf claude

# Remove
hf extensions remove claude

```

## TODO

- pip-installable extensions: only binary extensions for now. The `type` field in the manifest exists but is always `"binary"`. Still need to think about the best way to fetch and install a Python extension.

- (not prio) Auto-update? we can store the installed commit SHA in the manifest and compare it against the latest commit from the GitHub API on each run (or on `hf extension list`). This would let us show an "update available" hint or add an `hf extension update` command.

### DIFF
--- a/src/huggingface_hub/cli/extensions.py
+++ b/src/huggingface_hub/cli/extensions.py
@@ -36,7 +36,12 @@ from ._cli_utils import typer_factory
 DEFAULT_EXTENSION_OWNER = "huggingface"
 EXTENSIONS_ROOT = Path("~/.local/share/hf/extensions")
 MANIFEST_FILENAME = "manifest.json"
-extensions_cli = typer_factory(help="Manage hf CLI extensions.")
+EXTENSIONS_HELP = (
+    "Manage hf CLI extensions.\n\n"
+    "Security Warning: extensions are third-party executables. "
+    "Install only from sources you trust."
+)
+extensions_cli = typer_factory(help=EXTENSIONS_HELP)
 
 
 @dataclass
@@ -67,7 +72,10 @@ def extension_install(
     ],
     force: Annotated[bool, typer.Option("--force", help="Overwrite if already installed.")] = False,
 ) -> None:
-    """Install an extension from a public GitHub repository."""
+    """Install an extension from a public GitHub repository.
+
+    Security warning: this installs a third-party executable. Install only from sources you trust.
+    """
     owner, repo_name, short_name = _normalize_repo_id(repo_id)
     root_ctx = ctx.find_root()
     reserved_commands = set(getattr(root_ctx.command, "commands", {}).keys())
--- a/docs/source/en/package_reference/cli.md
+++ b/docs/source/en/package_reference/cli.md
@@ -1102,6 +1102,8 @@ $ hf env [OPTIONS]
 
 Manage hf CLI extensions.
 
+Security Warning: extensions are third-party executables. Install only from sources you trust.
+
 **Usage**:
 
 ```console
@@ -1149,6 +1151,8 @@ Learn more
 ### `hf extensions install`
 
 Install an extension from a public GitHub repository.
+
+Security warning: this installs a third-party executable. Install only from sources you trust.
 
 **Usage**:
 


src/huggingface_hub/cli/hf.py
```diff
@@ -84,6 +85,36 @@ def app_callback(
 app.add_typer(skills_cli, name="skills")
 app.add_typer(spaces_cli, name="spaces")
 app.add_typer(ie_cli, name="endpoints")
+app.add_typer(extensions_cli, name="extension")
+# TODO: We should probably not hide aliases and have a "Alias" section in --help
+app.add_typer(extensions_cli, name="ext", hidden=True)
+app.add_typer(extensions_cli, name="extensions", hidden=True)
```
```suggestion
app.add_typer(extensions_cli, name="extensions")
```

for now since we don't have aliases yet (coming soon) I'll just go with `hf extensions`

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,269 @@
...
+    return normalized
+
+
+def _execute_extension_binary(executable_path: Path, args: list[str]) -> int:
+    try:
+        return subprocess.call([str(executable_path)] + args)
+    except OSError as e:
+        if e.errno != errno.ENOEXEC:
+            raise
+        return subprocess.call(["sh", str(executable_path)] + args)
```
Wauplin, line 326: won't work on windows AFAIK (but should extensions be Unix-only maybe? do you know how `gh` handles that?)

src/huggingface_hub/cli/hf.py
```diff
@@ -84,6 +85,36 @@ def app_callback(
...
+    if command_name in _get_top_level_command_names():
+        return None
+
+    short_name = command_name[3:] if command_name.startswith("hf-") else command_name
+    if not short_name:
+        return None
+
+    executable_path = _get_extension_executable_path(short_name)
+    if not executable_path.is_file():
+        return None
```
Wauplin: though this seems to be working, I wonder if we could integrate it directly to our `HFCliApp` class? I feel that it would be more robust/readable if the main logic was something like this:

```py
# pseudo-code
try:
    app.run()
except CommandNotFound:
    binary_path = look_for_executable()
    execute(binary_path)
```

It would avoid listing existing commands, check `argv` manually, etc.

Wauplin: if not possible, then fine with this 👍

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,269 @@
...
+
+    if not executable_path.is_file():
+        raise CLIError(f"Extension '{short_name}' is not installed.")
+
+    exit_code = _execute_extension_binary(executable_path=executable_path, args=list(ctx.args))
+    raise typer.Exit(code=exit_code)
+
+
+@extensions_cli.command("list", examples=["hf extension list"])
+@extensions_cli.command("ls", hidden=True)
```
Wauplin:
```suggestion
```

let's avoid aliases for now and add them back later

src/huggingface_hub/cli/extensions.py
hanouticelina: Added basic Windows support: on Windows we download `hf-<name>.exe` instead of `hf-<name>`, and skip the `ENOEXEC` → `sh` fallback. Extension repos just need to provide both files at the root (same convention as `gh`). `os.chmod` is already a no-op on Windows but we skip it for clarity.

src/huggingface_hub/cli/hf.py
hanouticelina: I looked into it — the cleanest way would be to override `resolve_command()` in `HFCliTyperGroup` to catch `UsageError` and fall back to extension lookup. But it requires creating dynamic Click commands at resolve time, and since you're already changing `HFCliTyperGroup` in the buckets PR, it'd be easier to do that refactor there or as a follow-up. Keeping the current pre-dispatch for now.

src/huggingface_hub/cli/extensions.py
hanouticelina: yes [9ca4f7c](https://github.com/huggingface/huggingface_hub/pull/3805/commits/9ca4f7c0d1a6c879ee59c82d574c5652d96d75f8)

src/huggingface_hub/cli/hf.py
hanouticelina: done [9ca4f7c](https://github.com/huggingface/huggingface_hub/pull/3805/commits/9ca4f7c0d1a6c879ee59c82d574c5652d96d75f8)

src/huggingface_hub/cli/extensions.py
hanouticelina: gh extension works on Windows too (see [here](https://docs.github.com/en/github-cli/github-cli/creating-github-cli-extensions)) it looks for `.exe` binaries on Windows and shell scripts on Unix. I added the support in [9ca4f7c](https://github.com/huggingface/huggingface_hub/pull/3805/commits/9ca4f7c0d1a6c879ee59c82d574c5652d96d75f8). basically it appends `.exe` to the executable name on Windows, skips `chmod` on install and skips the `ENOEXEC` -> `sh` fallback. I cannot test that though

src/huggingface_hub/cli/hf.py
hanouticelina: Ok I asked codex (and I checked the docs because you never know): `app()` runs in Click "standalone mode" by default, Click catches internally and exits (`SystemExit`) after printing errors so you don't get a clean, typed exception to catch and recover from.
(also with the help of Codex) one solution is to define a new `ExtensionAwareTyperGroup` that overrides Click/Typer command resolution, so unknown top level commands are dispatched to installed extensions before parsing fails. I'm not convinced about the advantages compared to the original implementation, it's just incremental imo. I pushed this in so that you can check that as well : [bbb3eb1](https://github.com/huggingface/huggingface_hub/pull/3805/commits/bbb3eb15a537524a03a2a15cf777a9b905e3fc4d).
let me know what you think

src/huggingface_hub/cli/extensions.py
Wauplin: thanks for looking into it!

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,298 @@
...
+    try:
+        response = get_session().get(raw_url, follow_redirects=True)
+        response.raise_for_status()
+    except Exception as e:
+        raise CLIError(f"Failed to download '{executable_name}' from '{owner}/{repo_name}': {e}") from e
+
+    with TemporaryDirectory() as tmp_dir:
+        tmp_executable = Path(tmp_dir) / executable_name
+        tmp_executable.write_bytes(response.content)
+        if platform.system() != "Windows":
```
Wauplin: we use `os.name == "nt"` in the codebase so let's switch to that for consistency (might need to import os + remove platform import)

```suggestion
        if os.name == "nt":
```

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def _persist_installed_extension(extension_dir: Path, source_executable: Path, manifest: ExtensionManifest) -> None:
+    executable_path = extension_dir / manifest.executable_name
+    manifest_path = extension_dir / MANIFEST_FILENAME
+
+    try:
+        extension_dir.mkdir(parents=True, exist_ok=False)
+        shutil.copy2(source_executable, executable_path)
+        if platform.system() != "Windows":
```
Wauplin:
```suggestion
        if os.name == "nt":
```

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,298 @@
...
+    root = _get_extensions_root().resolve()
+    target = (root / f"hf-{safe_name}").resolve()
+    if root not in target.parents:
+        raise CLIError(f"Invalid extension name '{short_name}'.")
+    return target
+
+
+def _get_executable_name(short_name: str) -> str:
+    name = f"hf-{short_name}"
+    if platform.system() == "Windows":
```
Wauplin:
```suggestion
    if os.name == "nt":
```

src/huggingface_hub/cli/extensions.py
```diff
@@ -0,0 +1,298 @@
...
+        raise CLIError("Extension name cannot be empty.")
+    normalized = candidate[3:] if candidate.startswith("hf-") else candidate
+    return _validate_extension_short_name(normalized, original_input=name)
+
+
+def _execute_extension_binary(executable_path: Path, args: list[str]) -> int:
+    try:
+        return subprocess.call([str(executable_path)] + args)
+    except OSError as e:
+        if platform.system() == "Windows" or e.errno != errno.ENOEXEC:
```
Wauplin:
```suggestion
        if os.name == "nt" or e.errno != errno.ENOEXEC:
```

src/huggingface_hub/cli/hf.py
hanouticelina: (Ignore my first comment, it was written by Opus 4.6)

Wauplin: > (also with the help of Codex) one solution is to define a new ExtensionAwareTyperGroup that overrides Click/Typer command resolution, so unknown top level commands are dispatched to installed extensions before parsing fails. I'm not convinced about the advantages compared to the original implementation, it's just incremental imo. I pushed this in so that you can check that as well : [bbb3eb1](https://github.com/huggingface/huggingface_hub/pull/3805/commits/bbb3eb15a537524a03a2a15cf777a9b905e3fc4d).
> let me know what you think

doesn't seem to change much indeed. Thanks for looking into it and yeah, if it's not straightforward because of the system exit then it's perfectly fine like this  

(I like the fact that it's built-in the typer_factory so that `hf.py` stays clean)

src/huggingface_hub/cli/hf.py
```diff
@@ -42,7 +44,37 @@
 from huggingface_hub.utils import ANSI, logging
 
 
-app = typer_factory(help="Hugging Face Hub CLI")
+class ExtensionAwareTyperGroup(HFCliTyperGroup):
```
Wauplin: preferably to move to `_cli_utils.py` with other Typer custom classes (to keep `hf.py` clean and readable)

hanouticelina: done in [6a077d8](https://github.com/huggingface/huggingface_hub/pull/3805/commits/6a077d83128f15b6941f0bf67841001a98096351)

src/huggingface_hub/cli/hf.py
hanouticelina: Codex suggested an even better way to keep hf.py clean: [6a077d8](https://github.com/huggingface/huggingface_hub/pull/3805/commits/6a077d83128f15b6941f0bf67841001a98096351)

hanouticelina:
merging (failing tests are unrelated)

PR #3816, opened 2026-02-18
Fix CI test failures related to LFS server-side behavior changes

Wauplin:
- test_create_commit_create_pr_on_foreign_repo: avoid LFS upload on foreign repo (now returns 403) by using .txt extension instead of .pkl
- test_list_and_delete_lfs_files: filter by filename instead of ref since PR-uploaded LFS files may now report ref="main"


Wauplin:
nope

PR #3818, opened 2026-02-19
Dataclass doesn't call original `__init__`

zucchini-nlp:
Instead of calling `original_init` of the dataclass, we override it because the original init already calls `post_init`. But in that case `post_init` does get any of the kwargs which is problematic is some cases

In transformers configs, we might expect certain kwargs to always be present and raise an error otherwise. Another example, we might set the absent kwargs to default value, but then it breaks when calling the actual `post_init` with kwargs due to complex BC-related kwarg manipulations. 

Related to https://github.com/huggingface/transformers/pull/41250 which is almost ready, fixing last tests now

### DIFF
--- a/src/huggingface_hub/dataclasses.py
+++ b/src/huggingface_hub/dataclasses.py
@@ -174,7 +174,7 @@ def strict(
                         f"but found `{len(args)}` positional args."
                     )
 
-                for f in fields(cls):
+                for f in fields(cls):  # type: ignore
                     if f.name in standard_kwargs:
                         setattr(self, f.name, standard_kwargs[f.name])
                     elif f.default is not MISSING:


src/huggingface_hub/dataclasses.py
```diff
@@ -158,13 +158,30 @@ def __strict_setattr__(self: Any, name: str, value: Any) -> None:
...
             @wraps(original_init)
-            def __init__(self, **kwargs: Any) -> None:
+            def __init__(self, *args, **kwargs: Any) -> None:
                 # Extract only the fields that are part of the dataclass
                 dataclass_fields = {f.name for f in fields(cls)}  # type: ignore [arg-type]
                 standard_kwargs = {k: v for k, v in kwargs.items() if k in dataclass_fields}
 
-                # Call the original __init__ with standard fields
-                original_init(self, **standard_kwargs)
+                # User shoulldn't define custom `__init__` when `accepts_kwargs`, and instead
```
Wauplin:
```suggestion
                # User shouldn't define custom `__init__` when `accepts_kwargs`, and instead
```

src/huggingface_hub/dataclasses.py
```diff
@@ -207,7 +224,7 @@ def __repr__(self) -> str:
                 # Combine both representations
                 return f"{standard_repr[:-1]}, {additional_repr})" if additional_kwargs else standard_repr
 
-            cls.__repr__ = __repr__  # type: ignore [method-assign]
+            # cls.__repr__ = __repr__  # type: ignore [method-assign]
```
Wauplin: is this change intended? 

```suggestion
            cls.__repr__ = __repr__  # type: ignore [method-assign]
```

src/huggingface_hub/dataclasses.py
```diff
@@ -158,13 +158,30 @@ def __strict_setattr__(self: Any, name: str, value: Any) -> None:
...
+                # User shoulldn't define custom `__init__` when `accepts_kwargs`, and instead
+                # are adviced to move field manipulation to `__post_init__` (e.g., derive new field from existing ones)
+                # We need to call bare `__init__` here without `__post_init__` but the``original_init`` would call
+                # post-init right away with no kwargs.
+                if len(args) > 0:
+                    raise ValueError(
+                        f"When `accept_kwargs=True`, {cls.__name__} accepts only keyword arguments, "
+                        f"but found `{len(args)}` positional args."
+                    )
+
```
Wauplin, line 176: The logic for handling field defaults looks fragile. The standard dataclass sentinel is `dataclasses.MISSING`, not a comparison between `default` and `default_factory`. As written, when a field has `default_factory` set, both `f.default` and `f.default_factory` are `MISSING`, so `f.default is not f.default_factory` is `False` — which accidentally works but for the wrong reason and is confusing to read.

Suggestion:
```python
elif f.default is not MISSING:
    setattr(self, f.name, f.default)
elif f.default_factory is not MISSING:
    setattr(self, f.name, f.default_factory())
```

(with `from dataclasses import MISSING`)

_Generated by Claude Code_

src/huggingface_hub/dataclasses.py
Wauplin: This looks like a debug leftover — commenting it out silently breaks the custom `__repr__` for `accept_kwargs` classes. Should either be removed entirely (along with the `__repr__` function above) or uncommented.

_Generated by Claude Code_

src/huggingface_hub/dataclasses.py
```diff
@@ -158,13 +158,30 @@ def __strict_setattr__(self: Any, name: str, value: Any) -> None:
             original_init = cls.__init__
 
             @wraps(original_init)
-            def __init__(self, **kwargs: Any) -> None:
+            def __init__(self, *args, **kwargs: Any) -> None:
                 # Extract only the fields that are part of the dataclass
                 dataclass_fields = {f.name for f in fields(cls)}  # type: ignore [arg-type]
```
Wauplin, line 164: nit: typos — "shoulldn't" → "shouldn't", "adviced" → "advised"

_Generated by Claude Code_

Wauplin:
@cursor can you try to fix the code quality issue and if it doesn't work, add a type ignore

src/huggingface_hub/dataclasses.py
```diff
@@ -159,13 +159,30 @@ def __strict_setattr__(self: Any, name: str, value: Any) -> None:
...
+                # are advised to move field manipulation to `__post_init__` (e.g., derive new field from existing ones)
+                # We need to call bare `__init__` here without `__post_init__` but the``original_init`` would call
+                # post-init right away with no kwargs.
+                if len(args) > 0:
+                    raise ValueError(
+                        f"When `accept_kwargs=True`, {cls.__name__} accepts only keyword arguments, "
+                        f"but found `{len(args)}` positional args."
+                    )
+
+                for f in fields(cls):
```
Wauplin:
```suggestion
                for f in fields(cls):  # type: ignore
```

PR #3823, opened 2026-02-19
Check if `dataclass.repr=True` before wrapping

zucchini-nlp:
As per title, this gives me freedom to set `repr=False` and create my own `repr` by removing private attr, default values and other stuff. I don't think it's worth pretty printing additional kwargs in the hub-side because default dataclass 'repr' is also simple dict-like

### DIFF
--- a/src/huggingface_hub/dataclasses.py
+++ b/src/huggingface_hub/dataclasses.py
@@ -225,7 +225,8 @@ def strict(
                 # Combine both representations
                 return f"{standard_repr[:-1]}, {additional_repr})" if additional_kwargs else standard_repr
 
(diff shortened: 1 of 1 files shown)


Wauplin:
@cursor can you add a small test for me in tests/test_utils_strict_dataclass.py? Something like that:

   ```python
   @strict_dataclass(accept_kwargs=True)
   @dataclass(repr=False)
   class MyClass:
       x: int
       def __repr__(self):
           return f"MyClass(x={self.x})"

   obj = MyClass(x=1, extra=2)
   assert repr(obj) == "MyClass(x=1)"  # custom repr is preserved
   ```

Issue #3838, opened 2026-02-25
Cache directory preserves original casing of repo_id → duplicate caches on case-sensitive filesystems (Linux)


julio75012:
### Describe the bug

`from_pretrained` and `hf_hub_download` build the local cache directory using the **exact casing** of the provided `repo_id`.

On case-sensitive filesystems (Linux), different capitalizations of the same model create separate cache folders and trigger duplicate downloads. The Hugging Face Hub treats repository IDs as case-insensitive.

### Steps to reproduce

```python
from transformers import AutoModel, AutoTokenizer

model_mixed = "deepseek-ai/DeepSeek-OCR"
model_lower = "deepseek-ai/deepseek-ocr"   # same model

# On Linux these create two different folders:
# ~/.cache/huggingface/hub/models--deepseek-ai--DeepSeek-OCR
# ~/.cache/huggingface/hub/models--deepseek-ai--deepseek-ocr

AutoModel.from_pretrained(model_mixed, trust_remote_code=True)
# Running with model_lower downloads again
```

### Expected behavior
Normalize the `repo_id` (e.g. lowercase) when generating the cache folder name so all case variations resolve to the **same** cache location.

### Actual behavior
Cache path preserves original casing → duplicate folders on Linux.

### Environment
- OS: Linux (case-sensitive filesystem)
- `huggingface_hub` version: `python -c "import huggingface_hub; print(huggingface_hub.__version__)"`
- `transformers` version: `python -c "import transformers; print(transformers.__version__)"`

### Suggested fix
Lowercase the `repo_id` for cache-key generation inside `huggingface_hub` (while keeping original casing for Hub API calls).

This eliminates wasted disk space from common casing variations in model names.

julio75012:
fix proposed here: https://github.com/huggingface/huggingface_hub/pull/3839

Wauplin:
Hi @julio75012 , this is a known topic that we don't really want to solve. If user always passes the same repo id, they will always use the same cache folder. The duplicate cache issue happens only when user explicitly hardcodes 2 different casing for the same repo, which we consider to be a corner case. We also don't want to use the resolved repo id to build the cache folder as it would invalidate everyone's cache in case of repo author renames their repo. I will therefore close this issue and PR without merging. 

Note that the PR you've suggested would invalidate the existing cache of everyone using `huggingface_hub`. Some users have TB+ of local cache and would be quite annoyed by that ^^

hoedt:
[...]

PR #3839, opened 2026-02-25
fix(cache): normalize repo_id casing in repo_folder_name

julio75012:
The cache directory name was previously built using the exact casing
of the provided repo_id. On case-sensitive filesystems this led to
duplicate cache folders for the same model (e.g. DeepSeek-OCR vs
deepseek-ocr).

We now normalize the repo_id to lowercase inside repo_folder_name()
for the cache path only, while keeping the original casing for Hub API
calls.


julio75012:
closes #3838

Wauplin:
closing (see https://github.com/huggingface/huggingface_hub/issues/3838#issuecomment-3957668266)

PR #3845, opened 2026-02-25
[Bucket] Add 'sync_bucket' to HfApi

Wauplin:
Equivalent to `hf buckets sync` but as an API:

```py
>>> from huggingface_hub import sync_bucket

# Upload: local directory -> bucket
>>> sync_bucket("./data", "hf://buckets/username/my-bucket")

# Download: bucket -> local directory
>>> sync_bucket("hf://buckets/username/my-bucket", "./data")
```

All the syntax/logic is the same. I've moved most of the implementation to a dedicated module `_buckets.py` from which `sync_bucket` is exported as `HfApi` alias and in the `hf` CLI.

---

> [!NOTE]
> **Medium Risk**
> Touches core bucket transfer/sync logic and refactors it across modules, which could affect CLI/API behavior and plan execution correctness for uploads/downloads and deletes.
> 
> **Overview**
> Adds a Python API equivalent of `hf buckets sync` by introducing `HfApi.sync_bucket` (and root-level `sync_bucket`) that supports the same behaviors: include/exclude filtering, `--delete`, comparison modes, `plan`/`apply`, `dry_run`, and verbose/quiet execution returning a `SyncPlan`.
> 
> Refactors bucket-related dataclasses and sync logic out of CLI and `hf_api.py` into a new shared module `huggingface_hub/_buckets.py`, updates the CLI `hf buckets sync` command to delegate to the new API, and moves the transient CLI status output helper to `utils.StatusLine` while improving CLI error formatting for invalid values. Documentation is updated to show Python `sync_bucket` usage and to publish `SyncOperation`/`SyncPlan` in the API reference.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1ac2ee1bb7a7d7d1d671be52d780ccd20b3de16. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/buckets.md
+++ b/docs/source/en/guides/buckets.md
@@ -4,6 +4,13 @@ rendered properly in your Markdown viewer.
 
 # Buckets
 
+> [!WARNING]
+> The `hf buckets` commands are currently available from the `main` branch.
+> Install `hf` from `main` with:
+> ```bash
+> uv tool install "huggingface-hub @ git+https://github.com/huggingface/huggingface_hub.git@main"
+> ```
+
 Buckets provide S3-like object storage on Hugging Face, powered by the Xet storage backend. Unlike repositories (which are git-based and track file history), buckets are remote object storage containers designed for large-scale files with content-addressable deduplication. They are designed for use cases where you need simple, fast, mutable storage such as storing training checkpoints, logs, intermediate artifacts, or any large collection of files that doesn't need version control.
 
 You can interact with buckets using the Python API ([`HfApi`]) or the CLI (`hf buckets`). In this guide, we will walk through all the operations available.
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -50,7 +50,11 @@ import httpx
 from tqdm.auto import tqdm as base_tqdm
 from tqdm.contrib.concurrent import thread_map
 
-from huggingface_hub.utils._xet import XetTokenType, fetch_xet_connection_info_from_repo_info
+from huggingface_hub.utils._xet import (
+    XetTokenType,
+    fetch_xet_connection_info_from_repo_info,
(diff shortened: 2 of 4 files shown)


src/huggingface_hub/utils/_terminal.py
```diff
@@ -14,9 +14,36 @@
 """Contains utilities to print stuff to the terminal (styling, helpers)."""
 
 import os
+import shutil
+import sys
 from typing import Optional, Union
 
 
+class _StatusLine:
```
hanouticelina: we already have a `StatusLine` in `_cli_utils.py`, is that different?

src/huggingface_hub/_buckets.py
```diff
@@ -0,0 +1,1169 @@
...
+    print(f"  Deletes: {summary['deletes']}")
+    print(f"  Skips: {summary['skips']}")
+
+
+# =============================================================================
+# Public sync function (Python API)
+# =============================================================================
+
+
+def sync_bucket(
```
hanouticelina:
```suggestion
def sync_bucket_internal(
```

src/huggingface_hub/utils/_terminal.py
Wauplin: strictly the same yes (I've pushed https://github.com/huggingface/huggingface_hub/pull/3845/commits/5b69687a03d71eed631748e37908251cb460fefc to keep only 1)

src/huggingface_hub/_buckets.py
Wauplin: better yes :+1: done in https://github.com/huggingface/huggingface_hub/pull/3845/changes/87a1f7b5f8b6b96977c7fb9f6deb4a8d2d8c878a

PR #3857, opened 2026-02-26
fix: quote uv args in bash -c to prevent shell redirection

XciD:
## Summary

When using `hf jobs uv run --with 'torch>=2.1' script.py` with local files, the generated `bash -c` command contains unquoted arguments:

```bash
bash -c '... && uv run --with torch>=2.1 script.py'
```

Bash interprets the `>` in `torch>=2.1` as a **file redirection operator**, so `uv` never receives the script filename and exits silently after installing dependencies. The job completes with exit code 0 but the user's script never runs.

This was reported by @abidlabs: https://huggingface.co/jobs/abidlabs/69a0af7cdfb316ac3f7be994

### Fix

Shell-quote all arguments (uv_args, script, script_args) when building the `bash -c` command string, so that `torch>=2.1` becomes `'torch>=2.1'` and is passed literally to `uv run`.

**Before:**
```bash
bash -c '... && uv run --with torch>=2.1 script.py'
```

**After:**
```bash
bash -c "... && uv run '--with' 'torch>=2.1' 'script.py'"
```

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only affect how the `uv run` command string is constructed for jobs with local-file upload, preventing shell metacharacter interpretation; behavior should be unchanged for normal args but could expose edge cases in quoting.
> 
> **Overview**
> Fixes `hf jobs uv run` when local files are included by **shell-quoting all `uv run` arguments** (including dependency specifiers, script name, and script args) in the generated `bash -c` command, preventing metacharacters like `>` from being treated as redirection.
> 
> Adds a regression test to ensure dependency version specifiers (e.g. `torch>=2.1`) and the script name appear quoted in the constructed bash command.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d2407f2d57495341e26d517f8157e671585d120. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -11332,7 +11332,7 @@ class HfApi:
                 for remote_file_name, local_file_to_include in remote_to_local_file_names.items()
             )
             # Shell-quote each arg to prevent metacharacters (e.g. '>') from being interpreted by bash
-            quoted_parts = ["'" + a.replace("'", r"'\''") + "'" for a in [*uv_args, script, *script_args]]
+            quoted_parts = ["'" + arg.replace("'", r"'\''") + "'" for arg in [*uv_args, script, *script_args]]
             command = [
                 "bash",
                 "-c",


src/huggingface_hub/hf_api.py
```diff
@@ -11331,11 +11331,13 @@ def _create_uv_command_env_and_secrets(
                 remote_file_name + " " + base64.b64encode(Path(local_file_to_include).read_bytes()).decode()
                 for remote_file_name, local_file_to_include in remote_to_local_file_names.items()
             )
+            # Shell-quote each arg to prevent metacharacters (e.g. '>') from being interpreted by bash
+            quoted_parts = ["'" + a.replace("'", r"'\''") + "'" for a in [*uv_args, script, *script_args]]
```
```suggestion
            quoted_parts = ["'" + arg.replace("'", r"'\''") + "'" for arg in [*uv_args, script, *script_args]]
```

PR #3877, opened 2026-03-04
Fixed Sub-headings for hf cache commands in the doc

mostafatouny:
Updated headings for cache commands in CLI documentation, from `##` to `###`.


mostafatouny:
Done, @hanouticelina.
Thank you for the wonderful hub, and let me know if you need any kind of contribution.

PR #3855, opened 2026-02-26
[CLI] Add CLI for Community features (Discussions & Pull Requests)

Wauplin:
## Docs

- CLI guide: https://moon-ci-docs.huggingface.co/docs/huggingface_hub/pr_3855/en/guides/cli#hf-discussions
- Community Tab guide: https://moon-ci-docs.huggingface.co/docs/huggingface_hub/pr_3855/en/guides/community#manage-discussions-and-pull-requests-from-the-cli

# [CLI] Add `hf discussions` command group

Add a `hf discussions` command group to the CLI for managing discussions and pull requests on the Hub.

On the Hugging Face Hub, discussions and PRs share the same numbering, API, and community tab — unlike GitHub where issues and PRs are separate entities. This PR adds a unified `hf discussions` command group that reflects this model, inspired by [GitHub CLI](https://cli.github.com/manual/) patterns (`gh issue` / `gh pr`).

## Commands (9 total)

| Command | Description | Example |
|---------|-------------|---------|
| `list` / `ls` | List discussions and/or PRs | `hf discussions list user/repo` |
| `view` | View details, comments, diff | `hf discussions view user/repo 5 --comments` |
| `create` | Create a discussion or PR | `hf discussions create user/repo --title "Fix" --pr` |
| `comment` | Comment on a discussion/PR | `hf discussions comment user/repo 5 --body "LGTM!"` |
| `close` | Close a discussion/PR | `hf discussions close user/repo 5` |
| `reopen` | Reopen a closed discussion/PR | `hf discussions reopen user/repo 5` |
| `rename` | Rename a discussion/PR | `hf discussions rename user/repo 5 "New title"` |
| `merge` | Merge a pull request | `hf discussions merge user/repo 5` |
| `diff` | Show the diff of a PR | `hf discussions diff user/repo 5` |

## Key design decisions

- **Unified command group**: Single `hf discussions` group (not separate `hf pr` / `hf discussions`), matching how the Hub models community features.
- **`--kind` filter**: `--kind pull_request` / `--kind discussion` on `list` to narrow scope. `--pull-request`/`--pr` flag on `create`.
- **Flag naming aligned with API**: `--status` (not `--state`), `--kind` (not `--type` which would collide with repo type).
- **Reuses existing CLI patterns**: `typer_factory`, `RepoIdArg`, `RepoTypeOpt`, `FormatOpt`, `QuietOpt`, `LimitOpt`, `AuthorOpt`, `print_list_output`, `api_object_to_dict`.
[...]

PR #3878, opened 2026-03-04
Rename 'hf discussions view' to 'hf discussions info'

Wauplin:
Follow up PR after https://github.com/huggingface/huggingface_hub/pull/3855.

I think it's best to rename `hf discussions view` to `hf discussions info`. This is much more consistent with the `hf` CLI even though it diverges from the `gh` one.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit a8f0440d8112da6596793192ecae7dc5b889a57f. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


julien-c:
> This is much more consistent with the `hf` CLI even though it diverges from the `gh` one.

Agree i think it's better

PR #3879, opened 2026-03-04
Fix flaky LFS test: retry git clone with backoff in setup_local_clone

Wauplin:
Goal: will merge if CI gets green.

Looking forward to removing `git` tests altogether :smile: 

----

The HfLargefilesTest::test_git_push_end_to_end test fails when the git server hasn't made a newly created repo available for clone yet. The pytest --reruns mechanism doesn't help because each rerun creates a NEW repo and immediately tries to clone it, hitting the same race condition.

Fix by adding retry logic with exponential backoff (1s, 2s, 4s, 8s, 16s) directly in setup_local_clone, so the SAME repo is retried until the git server is ready.

Also fix a typo in the --only-rerun regex: '||' (empty alternative that matches everything) -> add CalledProcessError to the explicit list, matching the Windows CI config that already has the correct single '|'.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit fff3286401e67c1cf3cb1706b7546e5f3426bebd. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -90,7 +90,7 @@ jobs:
         working-directory: ./src # For code coverage to work
         run: |
           source ../.venv/bin/activate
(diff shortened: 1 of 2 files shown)


Wauplin:
Did not solve the issue.

PR #3880, opened 2026-03-04
[API] [CLI] Add `duplicate_repo` (instead of `duplicate_space`) + `hf repos duplicate`

Wauplin:
Add `duplicate_repo` method and CLI command (`hf repos duplicate`) to unify repository duplication across all types, deprecating `duplicate_space`.

Fix https://github.com/huggingface/huggingface_hub/issues/3825.

---
[Slack Thread](https://huggingface.slack.com/archives/D0A9313SS3G/p1772635257057729?thread_ts=1772635257.057729&cid=D0A9313SS3G)

<p><a href="https://cursor.com/agents/bc-c9d873fe-8695-54c7-a6ad-4fca85bbba7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c9d873fe-8695-54c7-a6ad-4fca85bbba7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</p>

PR #3888, opened 2026-03-05
[CLI] Add space-specific options to `hf repos create` and `hf repos duplicate`

Wauplin:
Add `--hardware/--flavor`, `--storage`, `--sleep-time`, `-s/--secret`, and `-e/--env` options to both `hf repos create` and `hf repos duplicate` CLI commands.

These options map to the existing space_hardware, space_storage, space_sleep_time, space_secrets, and space_variables parameters in the HfApi.create_repo() and HfApi.duplicate_repo() methods.

Secrets and env vars accept `KEY=VALUE` format or just `KEY` (reads value from local environment variable).

Mostly same syntax as for `hf jobs run`.

Follow-up to https://github.com/huggingface/huggingface_hub/pull/3880#pullrequestreview-3897047261

---

> [!NOTE]
> **Medium Risk**
> Extends CLI surface area and refactors env/secret parsing into shared helpers used by `jobs` and `repos`, which could subtly change how variables/secrets are resolved from files or the local environment.
> 
> **Overview**
> Adds Space-specific flags to `hf repos create` and `hf repos duplicate` to configure Space hardware (`--flavor`), persistent storage (`--storage`), sleep behavior (`--sleep-time`), and initial variables/secrets via `-e/--env`, `--env-file`, `-s/--secrets`, and `--secrets-file`.
> 
> Introduces shared env/secret option types plus `parse_env_map`/`env_map_to_key_value_list` in `_cli_utils`, and updates `jobs` to use these helpers (including resolving bare secret keys like `HF_TOKEN` from an extended local environment). Documentation and CLI tests are updated to cover the new flags and API parameter wiring (`space_hardware`, `space_storage`, `space_sleep_time`, `space_variables`, `space_secrets`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6ad08cd9927f3b2438ab1c9b595c4a4aeeffc71. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -415,39 +415,22 @@ def _get_extended_environ() -> dict[str, str]:
     return extended_environ
 
 
-def _parse_dotenv_map(
-    env: Optional[list[str]] = None,
-    env_file: Optional[str] = None,
-    *,
-    environ: dict[str, str],
-) -> dict[str, Optional[str]]:
-    """Parse CLI env/secrets args into a dict using the provided lookup environment."""
-    env_map: dict[str, Optional[str]] = {}
-    if env_file:
-        env_map.update(load_dotenv(Path(env_file).read_text(), environ=environ))
-    for env_value in env or []:
-        env_map.update(load_dotenv(env_value, environ=environ))
-    return env_map
-
-
 def parse_env_map(
     env: Optional[list[str]] = None,
     env_file: Optional[str] = None,
 ) -> dict[str, Optional[str]]:
-    """Parse ``-e``/``--env`` and ``--env-file`` CLI args into a dict."""
-    return _parse_dotenv_map(env, env_file, environ=os.environ.copy())
-
-
-def parse_secret_map(
-    secrets: Optional[list[str]] = None,
-    secrets_file: Optional[str] = None,
-) -> dict[str, Optional[str]]:
-    """Parse ``-s``/``--secrets`` and ``--secrets-file`` CLI args into a dict.
+    """Parse ``-e``/``--env``/``-s``/``--secrets`` and ``--env-file``/``--secrets-file`` CLI args into a dict.
 
     Uses an extended environment that includes the user's HF token so that
     bare ``--secrets HF_TOKEN`` resolves correctly.
     """
-    return _parse_dotenv_map(secrets, secrets_file, environ=_get_extended_environ())
+    extended_environ = _get_extended_environ()
+    env_map: dict[str, Optional[str]] = {}
+    if env_file:
+        env_map.update(load_dotenv(Path(env_file).read_text(), environ=extended_environ))
+    for env_value in env or []:
+        env_map.update(load_dotenv(env_value, environ=extended_environ))
+    return env_map
 
 
 def env_map_to_key_value_list(env_map: dict[str, Optional[str]]) -> Optional[list[dict[str, str]]]:
--- a/src/huggingface_hub/cli/repos.py
+++ b/src/huggingface_hub/cli/repos.py
@@ -47,7 +47,6 @@ from ._cli_utils import (
     env_map_to_key_value_list,
     get_hf_api,
     parse_env_map,
-    parse_secret_map,
     typer_factory,
 )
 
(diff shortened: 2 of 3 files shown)


src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -365,6 +365,93 @@ class RepoType(str, Enum):
...
+    """Parse ``-s``/``--secrets`` and ``--secrets-file`` CLI args into a dict."""
+    from huggingface_hub.utils._dotenv import load_dotenv
+
+    secrets_map: dict[str, Optional[str]] = {}
+    extended_environ = _get_extended_environ()
+    if secrets_file:
+        secrets_map.update(load_dotenv(Path(secrets_file).read_text(), environ=extended_environ))
+    for secret in secrets or []:
+        secrets_map.update(load_dotenv(secret, environ=extended_environ))
+    return secrets_map
```
Wauplin: @cursor please merge both into a `parse_env_map` and always use extended environment

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -365,6 +365,93 @@ class RepoType(str, Enum):
...
+        env_map.update(load_dotenv(env_value, environ=os.environ.copy()))
+    return env_map
+
+
+def parse_secrets_map(
+    secrets: Optional[list[str]] = None,
+    secrets_file: Optional[str] = None,
+) -> dict[str, Optional[str]]:
+    """Parse ``-s``/``--secrets`` and ``--secrets-file`` CLI args into a dict."""
+    from huggingface_hub.utils._dotenv import load_dotenv
```
Wauplin: @cursor please move import line at top module

src/huggingface_hub/cli/repos.py
```diff
@@ -119,7 +163,7 @@ def repo_create(
     "duplicate",
     examples=[
         "hf repos duplicate openai/gdpval --type dataset",
-        "hf repos duplicate multimodalart/dreambooth-training my-dreambooth --type space --private",
+        "hf repos duplicate SpacesExamples/xxx myorg/dev --type space --flavor l4x4 --secrets HF_TOKEN --private",
```
Wauplin:
```suggestion
        "hf repos duplicate multimodalart/dreambooth-training my-dreambooth --type space --flavor l4x4 --secrets HF_TOKEN --private",
```

Wauplin:
@bot /style

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -365,6 +366,80 @@ class RepoType(str, Enum):
...
+def parse_env_map(
+    env: Optional[list[str]] = None,
+    env_file: Optional[str] = None,
+) -> dict[str, Optional[str]]:
+    """Parse ``-e``/``--env``/``-s``/``--secrets`` and ``--env-file``/``--secrets-file`` CLI args into a dict.
+
+    Uses an extended environment that includes the user's HF token so that
+    bare ``--secrets HF_TOKEN`` resolves correctly.
+    """
+    extended_environ = _get_extended_environ()
```
hanouticelina, line 427: I initially splitted `parse_env_map` and `parse_secrets_map` in [e56572f](https://github.com/huggingface/huggingface_hub/pull/3888/commits/e56572f2d18f4df2721d620d2e630dac0ad4cd66), then saw your comment asking to merge them. i think Cursor might be right here. Before this PR, parsing used two different environments:
  - `--env `-> `os.environ.copy()` (plain shell environment)
  - `--secrets` -> `_get_extended_environ()` (shell environment + locally saved HF token)
 
The merged `parse_env_map()` always calls `_get_extended_environ()`, so `--env HF_TOKEN` (bare, no `=value`) resolves to the user's saved auth token even when `HF_TOKEN` isn't in the shell. `hf repos create my-space --type space --space-sdk gradio --env HF_TOKEN` would send the saved token as a public Space variable, no? or Cursor and I are overthinking it? 😄

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: cf my comment [here](https://github.com/huggingface/huggingface_hub/pull/3888#discussion_r2955279021)

src/huggingface_hub/cli/_cli_utils.py
Wauplin: I do think it's a big overthinking here. If someone does 

```
hf repos create my-space --type space --space-sdk gradio --env HF_TOKEN
```

it is a problem no matter if the hf token is from an env variable or the locally saved token. Hence I would keep things as they are now for simplicity

Wauplin:
I've merged it as-is given the above convo. For the "it might leak locally saved token" I don't think we should care of this corner case (it's user's fault here). If we want to do something about it I think it should be on the infra/Hub side i.e. automatically revoke an hf token set as `HF_TOKEN` env variable in a (public) Space or Job

PR #3906, opened 2026-03-10
[CLI] Accept `spaces/user/repo` as repo ID prefix shorthand

julien-c:
## Summary

- Accept repo IDs like `spaces/user/repo`, `datasets/user/repo`, `models/user/repo` as shorthand for `user/repo --type space` etc.
- Add `resolve_repo_id_and_type()` helper in `_cli_utils.py` that parses the prefix
- Error if both a prefix and explicit `--type` flag are provided (ambiguous)
- Prototyped for the `hf discussions` commands (all 9 subcommands)

```
# before
hf discussions list safetensors/convert --type space

# after
hf discussions list spaces/safetensors/convert
```

## Test plan
- [ ] Smoke-test `hf discussions list spaces/safetensors/convert`
- [ ] Verify error on `hf discussions list spaces/safetensors/convert --type dataset`
- [ ] Existing tests still pass

### DIFF
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -290,17 +290,46 @@ class RepoType(str, Enum):
 RepoIdArg = Annotated[
     str,
     typer.Argument(
-        help="The ID of the repo (e.g. `username/repo-name`).",
+        help="The ID of the repo (e.g. `username/repo-name` or `spaces/username/repo-name`).",
     ),
 ]
 
+# Mapping from URL-style plural prefixes to repo type values
+_REPO_TYPE_PREFIXES = {
+    "spaces": "space",
+    "datasets": "dataset",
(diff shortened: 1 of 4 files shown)


julien-c:
what do you think of this prototype @Wauplin @hanouticelina?

I have no super strong opinion on that matter but thought i'd prototype it. cc @gary149 as well

Wauplin:
I like the idea yes :+1: 

I think we can use the same logic as in https://github.com/huggingface/huggingface_hub/pull/3919 i.e. automatically apply the alias at a typer level if command accepts both `repo_id` and `repo_type` so that we don't have to update each command dealing with repos individually. I can open a separate PR if you want (or you continue on this one)

julien-c:
ah no no please take over =) 

i'm no good at Typer stuff

hanouticelina:
closing this in favor of #3929

PR #3907, opened 2026-03-10
[CLI] Add --json alias for --format json

julien-c:
## Summary

Prototype for supporting `--json` as a shorthand for `--format json` in the Typer-based CLI.

- Adds reusable `JsonOpt` type annotation and `resolve_format()` helper in `_cli_utils.py`
- Wires it into `hf models ls` as a proof of concept — other commands can adopt the same 2-line pattern
- The `--json` flag is hidden from `--help` output to keep things clean

cc @hanouticelina @Wauplin — this is just a proposal/prototype, totally up to you whether you'd want to adopt this pattern across the CLI or not!

## Test plan
- [x] `hf models ls --json --limit 2` outputs JSON
- [x] `hf models ls --format json --limit 2` still works
- [x] `--json` does not appear in `hf models ls --help`
- [x] `--json` takes precedence over `--format table`

### DIFF
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -363,7 +363,7 @@ FormatOpt = Annotated[
 
 JsonOpt = Annotated[
     bool,
-    typer.Option("--json", help="Output as JSON (alias for --format json)."),
+    typer.Option("--json", hidden=True, help="Output as JSON (alias for --format json)."),
 ]
 
 


Wauplin:
I have opened a separate PR for that which I think is better as it doesn't require to update each command individually: https://github.com/huggingface/huggingface_hub/pull/3919

Wauplin:
Closing in favor of https://github.com/huggingface/huggingface_hub/pull/3919

PR #3920, opened 2026-03-12
Default to skipping login if already logged in and add `--force` flag

### LINKED ISSUE #3917: `hf auth login` silently skips when already logged in (inverted `skip_if_logged_in` logic)
### Describe the bug

`hf auth login` (without `--token`) silently returns with "User is already logged in." when a token is already stored. This makes it impossible to add a second token interactively, so `hf auth switch` is effectively unusable unless you pass `--token` directly.

The root cause is an inverted condition in [`interpreter_login()`](https://github.com/huggingface/huggingface_hub/blob/2f9153e7/src/huggingface_hub/_login.py#L249-L251) — the default is `skip_if_logged_in=False`, and the docstring says `True` means "do not prompt if already logged in", but the actual logic does the opposite. It skips the prompt when `skip_if_logged_in=False`.

The same issue exists in [`notebook_login()`](https://github.com/huggingface/huggingface_hub/blob/2f9153e7/src/huggingface_hub/_login.py#L319-L321).

### Reproduction

```bash
# Login with first account
hf auth login --token hf_first_token

# Try to add a second token interactively — this silently returns
hf auth login
# Output: "User is already logged in."

# Workaround: pass --token directly
hf auth login --token hf_second_token
```

### Logs

```shell
$ HF_DEBUG=1 hf auth login
User is already logged in.
```

### System info

```shell
- huggingface_hub version: 1.6.0
- Platform: macOS-26.3.1-arm64-arm-64bit-Mach-O
- Python version: 3.13.12
- Running in iPython ?: No
- Running in notebook ?: No
- Token path ?: /Users/hysts/.huggingface/token
- Has saved token ?: True
- Who am I ?: hysts
- Configured git credential helpers: osxkeychain, store
- Installation method: unknown
- httpx: 0.28.1
- hf_xet: 1.4.0
- gradio: N/A
- tensorboard: N/A
- HF_HUB_CACHE: /Users/hysts/.huggingface/hub
- HF_ASSETS_CACHE: /Users/hysts/.huggingface/assets
- HF_TOKEN_PATH: /Users/hysts/.huggingface/token
- HF_STORED_TOKENS_PATH: /Users/hysts/.huggingface/stored_tokens
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
```

hanouticelina:
Fixes #3917

This PR changes `login()`, `interpreter_login()`, and `notebook_login()` to default `skip_if_logged_in=True`, so calling `hf auth login` (or `login()` in Python) when already logged in returns fast with an informational message instead of re-prompting. this PR also adds a `force` parameter (and `--force` CLI flag) to override this behavior and go through the login flow again.

### DIFF
--- a/docs/source/en/quick-start.md
+++ b/docs/source/en/quick-start.md
@@ -73,7 +73,7 @@ The easiest way to authenticate is to save the token on your machine. You can do
 hf auth login
 ```
 
-If you are already logged in, the command will return immediately. To force re-login, use `hf auth login --force` (or `login(force=True)` in Python). If you are not logged in, you will be prompted for your token. The token is then validated and saved in your `HF_HOME` directory (defaults to `~/.cache/huggingface/token`). Any script or library interacting with the Hub will use this token when sending requests.
+If you are already logged in, the command will return immediately. To force re-login, use `hf auth login --force`. If you are not logged in, you will be prompted for your token. The token is then validated and saved in your `HF_HOME` directory (defaults to `~/.cache/huggingface/token`). Any script or library interacting with the Hub will use this token when sending requests.
 
 Alternatively, you can programmatically log in using [`login`] in a notebook or a script:
 
--- a/src/huggingface_hub/_login.py
+++ b/src/huggingface_hub/_login.py
@@ -61,7 +61,6 @@ def login(
     *,
     add_to_git_credential: bool = False,
     skip_if_logged_in: bool = True,
-    force: bool = False,
 ) -> None:
     """Login the machine to access the Hub.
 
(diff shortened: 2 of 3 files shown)


Wauplin:
There is definitely a bug here indeed, nice catch @hysts !

However I think we should take advantage to slightly update to login process:

- if user is logged in and they run `hf auth login` => should return fast with "already logged in (use --force to register a new token)"
- if user is logged in and they run `hf auth login --force` => should go through the login process
- if user is logged out and they run `hf auth login` => should go through the login process

This way `hf auth switch` is still effectively usable and the default behavior is still to not do anything on `hf auth login` is already logged in.


Same for the `huggingface.login` workflow (and notebook/interpreter ones).

Basically it would mean:
1. keeping this PR change
2. change the default value of `skip_if_logged_in` (which turns out not to be a breaking change since it was wrongly inverted before
3. add a `force` argument

What do you think?

Wauplin (approved):
Thank you!

PR #3924, opened 2026-03-12
fix: use module logger consistently and narrow bare except clauses

mango766:
## Summary

- **`_upload_large_folder.py`**: Replace `logging.info()` with `logger.info()` on lines 277/284 for consistency with 30+ other `logger.*` calls in the same file
- **`hf_file_system.py`**: Replace bare `except:` with `except OSError:` in `exists()` and `isfile()` to match the adjacent `isdir()` method

### Inconsistent logging
`_upload_large_folder.py` defines `logger = logging.getLogger(__name__)` at line 41 and uses it everywhere, except two calls that use `logging.info()` directly. This bypasses the module-level logger configuration (level, handlers, filters).

### Bare except
`exists()` and `isfile()` used bare `except:` which catches `KeyboardInterrupt`, `SystemExit`, and other non-error exceptions. The adjacent `isdir()` already correctly catches `OSError`. Since `self.info()` raises `EntryNotFoundError` (subclass of `OSError`) for missing paths, `except OSError:` is the correct scope.

## Test plan

- [ ] Verify `upload_large_folder()` logging output unchanged
- [ ] Verify `HfFileSystem.exists()` returns `False` for missing paths
- [ ] Verify `HfFileSystem.isfile()` returns `False` for missing paths

---

> [!NOTE]
> **Low Risk**
> Low risk: small consistency and error-handling tweaks with no behavior change expected beyond no longer swallowing non-`OSError` exceptions (e.g., interrupts) in `exists`/`isfile`.
> 
> **Overview**
> Ensures `upload_large_folder_internal` logs through the module-level `logger` (instead of `logging.info`) so output respects the file’s configured handlers/filters.
> 
> Tightens `HfFileSystem.exists()` and `HfFileSystem.isfile()` to catch `OSError` rather than a bare `except`, avoiding accidentally swallowing non-I/O exceptions while still returning `False` for missing paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 50737fdf18220d494c847f9bbce98e74a0460b1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -1036,7 +1036,7 @@ class HfFileSystem(fsspec.AbstractFileSystem, metaclass=_Cached):
 
             self.info(path, **kwargs)
             return True
-        except OSError:  # noqa: E722
+        except OSError:
             return False
 
     def isdir(self, path):
@@ -1072,7 +1072,7 @@ class HfFileSystem(fsspec.AbstractFileSystem, metaclass=_Cached):
         """
         try:
             return self.info(path)["type"] == "file"
-        except OSError:  # noqa: E722
+        except OSError:
             return False
 
     def url(self, path: str) -> str:


src/huggingface_hub/hf_file_system.py
```diff
@@ -1036,7 +1036,7 @@ def exists(self, path, **kwargs):
 
             self.info(path, **kwargs)
             return True
-        except:  # noqa: E722
+        except OSError:  # noqa: E722
```
Wauplin:
```suggestion
        except OSError:
```

src/huggingface_hub/hf_file_system.py
```diff
@@ -1072,7 +1072,7 @@ def isfile(self, path):
         """
         try:
             return self.info(path)["type"] == "file"
-        except:  # noqa: E722
+        except OSError:  # noqa: E722
```
Wauplin:
```suggestion
        except OSError:
```

PR #3936, opened 2026-03-16
feat(jobs): add volume mounting support for buckets and repos

XciD:
## Summary

Add support for mounting HuggingFace Buckets and Repos (models, datasets, spaces) as volumes in Job containers.

### Python API

```python
from huggingface_hub import run_job, Volume

job = run_job(
    image="python:3.12",
    command=["python", "-c", "import os; print(os.listdir('/data'))"],
    volumes=[
        Volume(type="dataset", source="username/my-dataset", mount_path="/data"),
        Volume(type="bucket", source="username/my-bucket", mount_path="/output"),
    ],
)
```

### CLI

```bash
hf jobs run -v hf://datasets/username/my-dataset:/data -v hf://buckets/username/my-bucket:/output python:3.12 python script.py
```

### Changes

- `_jobs_api.py`: new `Volume` dataclass, `volumes` field added to `JobInfo`/`JobSpec`/`_create_job_spec`
- `hf_api.py`: `volumes` parameter added to `run_job`, `run_uv_job`, `create_scheduled_job`, `create_scheduled_uv_job`
- `cli/jobs.py`: `--volume`/`-v` CLI option with Docker-like syntax (`TYPE/SOURCE:/MOUNT_PATH[:ro]`)
- `__init__.py`: export `Volume`

---

> [!NOTE]
> **Medium Risk**
> Touches core Jobs request/response schemas and adds new CLI parsing paths; main risk is incorrect volume spec serialization/parsing leading to job failures or unintended mount permissions.
> 
> **Overview**
> Adds volume mounting support to Jobs so containers can access Hugging Face repos (models/datasets/spaces) and storage buckets as local paths.
> 
> Introduces a new `Volume` type (exported from `huggingface_hub`) and threads an optional `volumes` parameter through `HfApi.run_job`, `run_uv_job`, and scheduled-job creation, including request serialization and parsing volumes back onto `JobInfo`/`JobSpec`.
> 
> Extends the `hf jobs` CLI with `-v/--volume` (Docker-like `hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro|:rw]`) plus parsing/validation and tests, and updates docs/CLI references to document the new capability.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f65d71f9fffc392898184fe38bd6b560a37f710d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -289,6 +289,7 @@ jobs_cli = typer_factory(help="Run and manage Jobs on the Hub.")
         "hf jobs run python:3.12 python -c 'print(\"Hello!\")'",
         "hf jobs run -e FOO=foo python:3.12 python script.py",
         "hf jobs run --secrets HF_TOKEN python:3.12 python script.py",
+        "hf jobs run -v hf://gpt2:/data -v hf://buckets/org/b:/mnt python:3.12 python script.py",
     ],
 )
 def jobs_run(
@@ -749,6 +750,7 @@ jobs_cli.add_typer(uv_app, name="uv")
         "hf jobs uv run my_script.py",
         "hf jobs uv run ml_training.py --flavor a10g-small",
         "hf jobs uv run --with transformers train.py",
+        "hf jobs uv run -v hf://gpt2:/data -v hf://buckets/org/b:/mnt script.py",
     ],
 )
 def jobs_uv_run(
--- a/docs/source/en/package_reference/cli.md
+++ b/docs/source/en/package_reference/cli.md
@@ -2157,7 +2157,7 @@ $ hf jobs run [OPTIONS] IMAGE COMMAND...
 * `-e, --env TEXT`: Set environment variables. E.g. --env ENV=value
 * `-s, --secrets TEXT`: Set secret environment variables. E.g. --secrets SECRET=value or `--secrets HF_TOKEN` to pass your Hugging Face token.
 * `-l, --label TEXT`: Set labels. E.g. --label KEY=VALUE or --label LABEL
-* `-v, --volume TEXT`: Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. TYPE is one of: models, datasets, spaces, buckets. TYPE defaults to models if omitted. E.g. -v hf://gpt2:/data or -v hf://datasets/org/ds:/data or -v hf://buckets/org/b:/mnt:ro
+* `-v, --volume TEXT`: Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. TYPE is one of: models, datasets, spaces, buckets. TYPE defaults to models if omitted. models, datasets and spaces are always mounted read-only. buckets are read+write by default.E.g. -v hf://gpt2:/data or -v hf://datasets/org/ds:/data or -v hf://buckets/org/b:/mnt:ro
 * `--env-file TEXT`: Read in a file of environment variables.
 * `--secrets-file TEXT`: Read in a file of secret environment variables.
 * `--flavor [cpu-basic|cpu-upgrade|cpu-performance|cpu-xl|sprx8|zero-a10g|t4-small|t4-medium|l4x1|l4x4|l40sx1|l40sx4|l40sx8|a10g-small|a10g-large|a10g-largex2|a10g-largex4|a100-large|a100x4|a100x8|h200|h200x2|h200x4|h200x8|inf2x6]`: Flavor for the hardware, as in HF Spaces. Run 'hf jobs hardware' to list available flavors. Defaults to `cpu-basic`.
@@ -2171,6 +2171,7 @@ Examples
   $ hf jobs run python:3.12 python -c 'print("Hello!")'
   $ hf jobs run -e FOO=foo python:3.12 python script.py
   $ hf jobs run --secrets HF_TOKEN python:3.12 python script.py
+  $ hf jobs run -v hf://gpt2:/data -v hf://buckets/org/b:/mnt python:3.12 python script.py
 
 Learn more
   Use `hf <command> --help` for more information about a command.
@@ -2336,7 +2337,7 @@ $ hf jobs scheduled run [OPTIONS] SCHEDULE IMAGE COMMAND...
 * `-e, --env TEXT`: Set environment variables. E.g. --env ENV=value
 * `-s, --secrets TEXT`: Set secret environment variables. E.g. --secrets SECRET=value or `--secrets HF_TOKEN` to pass your Hugging Face token.
 * `-l, --label TEXT`: Set labels. E.g. --label KEY=VALUE or --label LABEL
-* `-v, --volume TEXT`: Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. TYPE is one of: models, datasets, spaces, buckets. TYPE defaults to models if omitted. E.g. -v hf://gpt2:/data or -v hf://datasets/org/ds:/data or -v hf://buckets/org/b:/mnt:ro
+* `-v, --volume TEXT`: Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. TYPE is one of: models, datasets, spaces, buckets. TYPE defaults to models if omitted. models, datasets and spaces are always mounted read-only. buckets are read+write by default.E.g. -v hf://gpt2:/data or -v hf://datasets/org/ds:/data or -v hf://buckets/org/b:/mnt:ro
 * `--env-file TEXT`: Read in a file of environment variables.
 * `--secrets-file TEXT`: Read in a file of secret environment variables.
 * `--flavor [cpu-basic|cpu-upgrade|cpu-performance|cpu-xl|sprx8|zero-a10g|t4-small|t4-medium|l4x1|l4x4|l40sx1|l40sx4|l40sx8|a10g-small|a10g-large|a10g-largex2|a10g-largex4|a100-large|a100x4|a100x8|h200|h200x2|h200x4|h200x8|inf2x6]`: Flavor for the hardware, as in HF Spaces. Run 'hf jobs hardware' to list available flavors. Defaults to `cpu-basic`.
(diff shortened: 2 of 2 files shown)


src/huggingface_hub/cli/jobs.py
```diff
@@ -1119,6 +1140,72 @@ def scheduled_uv_run(
...
+                source = source_part
+
+        result.append(
+            JobVolume(
+                type=vol_type_str,
+                source=source,
+                mount_path=mount_path,
+                read_only=read_only,
+            )
+        )
```
lhoestq, line 1175: with this change you can support revisions (including special refs) and paths in repo/bucket:

```suggestion
        resolved_path = hffs.resolve_path(source_part)
        if isinstance(resolved_path, HfFileSystemResolvedRepositoryPath):
            result.append(
                JobVolume(
                    type=resolved_path.repo_type,
                    source=resolved_path.repo_id,
                    mount_path=mount_path,
                    revision=resolved_path.revision,
                    read_only=read_only,
                    path=resolved_path.path_in_repo,
                )
            )
        else:
            result.append(
                JobVolume(
                    type=JobVolumeType.BUCKET.value,
                    source=resolved_path.bucket_id,
                    mount_path=mount_path,
                    read_only=read_only,
                    path=resolved_path.path,
                )
            )
```

for example here are supported paths:


```python
# buckets
"hf://buckets/username/bucket"
"hf://buckets/username/bucket/path"
# repos
"hf://gpt2"
"hf://user/model"
"hf://datasets/user/dataset"
"hf://user/model/path/in/repo"
"hf://user/model@revision"
"hf://user/model@refs/pr/1"
```

(it works with and without the hf:// prefix)

your will need these imports

```python
from huggingface_hub import hffs
from huggingface_hub.hf_file_system import HfFileSystemResolvedBucketPath, HfFileSystemResolvedRepositoryPath
```

it will also raise an error if the repo / bucket doesn't exist

lhoestq:
love it ! quick question for the CLI: should we require the hf:// prefix for the source path ? to make sure it doesn't look like a local path (and in case we want to support local path at some point)

davanstrien:
>  quick question for the CLI: should we require the hf:// prefix for the source path ? to make sure it doesn't look like a local path (and in case we want to support local path at some point)

Think this makes sense IMO. For Jobs I have quite a lot of use cases in mind where you do something like 

```bash
hf jobs uv run whisper-transcribe.py some-local-dir/audiofiles.mp3
```

lhoestq:
Getting an error when trying to mount [fineweb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) on Jobs, while other datasets work well

```
Volume mount failed: Volume mount failed
```

I tried with

```
hf jobs run -v dataset/HuggingFaceFW/fineweb:/dataset duckdb/duckdb duckdb -c "SELECT * FROM '/dataset/**/*.parquet' LIMIT 5"
```

cc @XciD

julien-c:
> Getting an error when trying to mount [fineweb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) on Jobs

First mounted 1PB dataset or bucket, when? :trollface:

lhoestq:
taking the liberty to add some docs using the current syntax `-v bucket/username/my-bucket:/mnt`

(but I'm fine with updating with `-v hf://buckets/username/my-bucket:/mnt` if we decide to go with this approach)

docs/source/en/guides/cli.md
```diff
@@ -1535,6 +1535,35 @@ Available `--flavor` options:
...
+>>> hf jobs run -v dataset/HuggingFaceFW/fineweb:/dataset \
+...     duckdb/duckdb duckdb -c "SELECT * FROM '/dataset/**/*.parquet' LIMIT 5"
+
+# UV Job with a mounted volume to save checkpoints when training a model
+>>> hf jobs uv run -v bucket/username/my-bucket:/training-outputs \
+...     sft.py --output-dir /training-outputs/training-v3-final ...
+```
+
+By default, mounted storage buckets have read+write abilities.
+This is especially useful for storage buckets since they are fast, mutable storage for data that changes frequently — files can be overwritten or deleted in place.
```
davanstrien:
```suggestion
This is especially useful for storage buckets, which provide fast, mutable storage for data that changes frequently — files can be overwritten or deleted in place.
```

docs/source/en/guides/jobs.md
```diff
@@ -216,6 +216,42 @@ Available `flavor` options:
...
+```python
+>>> from huggingface_hub import JobVolume, run_job
+>>> job = run_job(
+...     image="duckdb/duckdb",
+...     command=["duckdb", "-c", "SELECT * FROM '/data/**/*.parquet' LIMIT 5"],
+...     volumes=[JobVolume(type="dataset", source="HuggingFaceFW/fineweb", mount_path="/data")],
+... )
+```
+
+You can also write to a mounted bucket, for example here to save checkpoints when training a model:
```
davanstrien:
```suggestion
You can also write to a mounted bucket, for example, to save checkpoints when training a model:
```

docs/source/en/guides/jobs.md
```diff
@@ -216,6 +216,42 @@ Available `flavor` options:
...
+```python
+>>> from huggingface_hub import JobVolume, run_uv_job
+>>> script = "my_sft.py"
+>>> script_args = ["--output_dir", "/training-outputs/training-v3-final", ...]
+>>> checkpoints_bucket = JobVolume(type="bucket", source="username/my-bucket", mount_path="/training-outputs")
+>>> run_uv_job(script, script_args=script_args, volumes=[checkpoints_bucket])
+```
+
+By default, mounted storage buckets have read+write abilities.
+This is especially useful for storage buckets since they are fast, mutable storage for data that changes frequently — files can be overwritten or deleted in place.
```
davanstrien:
```suggestion
This is especially useful for storage buckets, which provide fast, mutable storage for data that changes frequently — files can be overwritten or deleted in place.
```

src/huggingface_hub/cli/jobs.py
Wauplin: I like the idea of reusing the existing path resolver but I don't think we should error out if the repo / bucket doesn't exist. I think this should be a server-side error (and at some point we might want to be able to create the bucket on demand). Also the error at the moment is cryptic in the context of volume mounts so we would need to reraise from it (fails with `"Access to repositories lists is not implemented."`).

@lhoestq do you think it's be possible to reuse the same parsing logic but without the HTTP layer check?

Wauplin: I've decided to reimplement a HTTP-free resolver and I'll open a separate issue to deal with all the "hf handle" parsers in the codebase (we have different parsers for `hf sync`,  `hffs`, and now job volumes). The supported syntax is the same but parsers are slightly different depending on the use case.

I've also made `hf://` required for now for volumes in the CLI. We'll see later if we want to relax this rule but I like your suggestion in https://github.com/huggingface/huggingface_hub/pull/3936#issuecomment-4070696680

>quick question for the CLI: should we require the hf:// prefix for the source path ? to make sure it doesn't look like a local path (and in case we want to support local path at some point)

docs/source/en/guides/jobs.md
```diff
@@ -216,6 +216,42 @@ Available `flavor` options:
...
 
+## Mount a volume
+
+Mount a volume on the Jobs's disk using a list of [`Volume`].
+
+You can mount any Hugging Face Repository (model/dataset/space) or [Storage Bucket](/docs/hub/storage-buckets). For example:
+
+* mount a model repository: `Volume(type="model", source="openai/gpt-oss-120b", mount_path="/model")`
+* mount a dataset repository: `Volume(type="dataset", source="HuggingFaceFW/fineweb", mount_path="/data")`
+* mount a storage bucket: `Volume(type="bucket", source="username/my-bucket", mount_path="/mnt)`
```
hanouticelina:
```suggestion
* mount a storage bucket: `Volume(type="bucket", source="username/my-bucket", mount_path="/mnt")`
```

docs/source/en/guides/cli.md
```diff
@@ -1571,6 +1571,38 @@ Available `--flavor` options:
 
 (updated in 07/2025 from Hugging Face [suggested_hardware docs](https://huggingface.co/docs/hub/en/spaces-config-reference))
 
+## Volumes
+
+Mount a volume on the Jobs's disk using `-v` or `--volume`.
```
```suggestion
Mount a volume on the Job's disk using `-v` or `--volume`.
```

docs/source/en/guides/cli.md
```diff
@@ -1571,6 +1571,38 @@ Available `--flavor` options:
...
+>>> hf jobs run -v hf://datasets/HuggingFaceFW/fineweb:/dataset \
+...     duckdb/duckdb duckdb -c "SELECT * FROM '/dataset/**/*.parquet' LIMIT 5"
+
+# UV Job with a mounted volume to save checkpoints when training a model
+>>> hf jobs uv run -v hf://buckets/username/my-bucket:/training-outputs \
+...     sft.py --output-dir /training-outputs/training-v3-final ...
+```
+
+By default, mounted storage buckets have read+write abilities.
+This is especially useful for storage buckets, which provide fast, mutable storage for data that changes frequently — files can be overwritten or deleted in place.
```
hanouticelina: we should mention somewhere in the docs that models, datasets and spaces  are read-only
```suggestion
Models, datasets and spaces are always mounted read-only. Storage buckets are read+write by default — this is especially useful for data that changes frequently, as files can be overwritten or deleted in place.
```

src/huggingface_hub/constants.py
```diff
@@ -119,6 +119,8 @@ def _as_int(value: Optional[str]) -> Optional[int]:
     "models": REPO_TYPE_MODEL,
 }
 
+REPO_TYPE_BUCKET = "bucket"
```
hanouticelina: (nit) can we group this with the other `REPO_TYPE_`* constants, i.e. after `REPO_TYPE_MODEL`?  (OCD kicking in 😅)

src/huggingface_hub/cli/jobs.py
```diff
@@ -236,6 +237,18 @@ def _parse_namespace_from_job_id(job_id: str, namespace: Optional[str]) -> tuple
...
 
+VolumesOpt = Annotated[
+    Optional[list[str]],
+    typer.Option(
+        "-v",
+        "--volume",
+        help="Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. "
+        "TYPE is one of: models, datasets, spaces, buckets. "
+        "TYPE defaults to models if omitted. "
+        "E.g. -v hf://gpt2:/data or -v hf://datasets/org/ds:/data or -v hf://buckets/org/b:/mnt:ro",
```
hanouticelina, line 249: examples should be added in [L287](https://github.com/huggingface/huggingface_hub/blob/10a578be94f536c32a56021f204fb24f8435adbf/src/huggingface_hub/cli/jobs.py#L287) and [L747](https://github.com/huggingface/huggingface_hub/blob/10a578be94f536c32a56021f204fb24f8435adbf/src/huggingface_hub/cli/jobs.py#L747)

src/huggingface_hub/cli/jobs.py
```diff
@@ -236,6 +237,18 @@ def _parse_namespace_from_job_id(job_id: str, namespace: Optional[str]) -> tuple
...
 ]
 
+VolumesOpt = Annotated[
+    Optional[list[str]],
+    typer.Option(
+        "-v",
+        "--volume",
+        help="Mount a volume. Format: hf://[TYPE/]SOURCE:/MOUNT_PATH[:ro]. "
+        "TYPE is one of: models, datasets, spaces, buckets. "
+        "TYPE defaults to models if omitted. "
```
hanouticelina, line 247: to surface this information in `--help` output
```suggestion
        "TYPE defaults to models if omitted. "
        "models, datasets and spaces are always mounted read-only. buckets are read+write by default."
```

src/huggingface_hub/cli/jobs.py
Wauplin: good call :+1:

Wauplin: addressed in https://github.com/huggingface/huggingface_hub/pull/3936/commits/f65d71f9fffc392898184fe38bd6b560a37f710d

src/huggingface_hub/cli/jobs.py
lhoestq: sounds good !

PR #3940, opened 2026-03-17
[CLI] remove ANSI codes when stdout is piped + add `--no-color` global flag

hanouticelina:
two changes in this PR:
- `ANSI._format()` isatty check,  when stdout is not a TTY , ANSI codes are now suppressed automatically. This matches standard UNIX behavior (`git`, `ls --color=auto`, etc..) and fixes the real-world bug where `hf auth whoami` output with ANSI codes broke downstream `hf repos create`.
- `--no-color` global flag - adds an explicit `--no-color` option to the top-level `hf` command that sets `NO_COLOR=1` before any subcommand runs. 

---

> [!NOTE]
> **Medium Risk**
> Changes global CLI invocation and ANSI formatting behavior, which can subtly affect scripted parsing and Click/Typer output across many commands. Risk is limited to terminal output/styling (no auth/data-path changes) but touches the main entrypoint.
> 
> **Overview**
> Adds a global `--no-color` flag to `hf` that eagerly disables Click/Typer coloring and sets `NO_COLOR` before any subcommand runs, and wires this through a shared `_init_no_color` helper.
> 
> Updates `ANSI` formatting to automatically suppress escape codes when the target output stream is not a TTY (and allows passing the target stream), with expanded tests to cover TTY vs piped output behavior.
> 
> Aligns a few CLI warnings/commands with the new stream-aware styling (e.g., deprecation warnings to `stderr`) and documents `--no-color` in the generated CLI reference.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1b400e2cd2c9054463806b1c214f38d617e7e40. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


holysheep123:
HuggingFace CLI with piped stdout and ANSI color issues is a common pain point. If you need a reliable inference API as an alternative to HuggingFace endpoints, HolySheep (holysheep.ai) offers Claude/GPT/Gemini with clean CLI output, 1 CNY = 1 USD rate, WeChat/Alipay payment for China-based developers.

hanouticelina:
@cursoragent review again

hanouticelina:
@Wauplin (for when you're back)

The initial solution (from Codex) checked `sys.stdout.isatty()` in `ANSI._format` and set `NO_COLOR` globally. This broke two cases: `hf repos ... | grep ...` stripped color from stderr warnings even though stderr was still a terminal, and `hf repos ... 2>warnings.log` leaked ANSI codes into the log file because stdout was a terminal. 

The fix I pushed adds an optional `file` parameter to `ANSI._format` (defaults to `sys.stdout`) so each call checks `isatty()` on its actual target stream. The two stderr callers (`repos.py`, `repo_files.py`) now pass `file=sys.stderr`. `NO_COLOR` env and `--no-color` stay as global overrides.
To be honest, I'm not sure if this is the simplest solution (I didn't get very useful help from my agents except understanding some stuff here 🙈).

src/huggingface_hub/cli/hf.py
```diff
@@ -105,12 +128,17 @@ def app_callback(
...
     if not constants.HF_DEBUG:
         logging.set_verbosity_info()
     check_cli_update("huggingface_hub")
 
     try:
-        app()
+        if os.getenv("NO_COLOR"):
+            app(color=False)
+        else:
+            app()
```
Wauplin, line 141: I don't understand all of this logic around modifying `NO_COLOR` env variable and passing `color=False` to the app. I feel that we are doing multiple times the same thing. Let's just set `NO_COLOR` env variable once if `--no-color` is passed and that's it.

src/huggingface_hub/cli/repo_files.py
```diff
@@ -70,7 +70,10 @@ def repo_files_delete(
     token: TokenOpt = None,
 ) -> None:
     print(
-        ANSI.yellow("FutureWarning: `hf repo-files delete` is deprecated. Use `hf repos delete-files` instead."),
+        ANSI.yellow(
+            "FutureWarning: `hf repo-files delete` is deprecated. Use `hf repos delete-files` instead.",
+            file=sys.stderr,
+        ),
```
Wauplin, line 76: ~Syntax-wise I think I'd prefer a `ANSI(sys.stderr).yellow("Warning message")` which would also reduce the amount of changes needed in all `ANSI` methods~

**EDIT:** please ignore if switching to `rich`

src/huggingface_hub/cli/discussions.py
Wauplin, line 270: should be dropped (maybe not the only occurence)

hanouticelina:
closing this in favor of switching to `rich` and drop the `ANSI` utility  in another PR

PR #3956, opened 2026-03-20
[FEAT] Support skills from hf skills

burtenshaw:
This PR allows uses to add and update skills from `huggingface/skills`

You can test it like this.

```
hf skills add gradio
ls .agents/skills/huggingface-gradio
```

The naming in `huggingface/skills` should be simplified before we merge this. i.e. `hugging-face-trackio` >> `trackio`. That's added [here](https://github.com/huggingface/skills/pull/99)

---

> [!NOTE]
> **Medium Risk**
> Adds new CLI behavior that downloads and extracts remote tarballs from GitHub and overwrites local directories, which can fail due to network/API changes and affects local filesystem state.
> 
> **Overview**
> `hf skills add` is changed from generating a local `hf-cli` `SKILL.md` to installing named marketplace skills from the `huggingface/skills` GitHub repo (with a manifest file for revision tracking) and optionally symlinking installs into Claude’s legacy skills directory.
> 
> Adds `hf skills upgrade`, which scans configured skills directories, checks installed manifests against the latest upstream commit for each skill, and upgrades in place with status reporting. Documentation and CLI reference are updated accordingly, and new CLI tests cover marketplace install and upgrade flows.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b69456065212ed84b9f8e88d2a200e60e383404. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/serialization/_torch.py
+++ b/src/huggingface_hub/serialization/_torch.py
@@ -510,14 +510,32 @@ def _load_sharded_checkpoint(
     with open(index_file, encoding="utf-8") as f:
         index = json.load(f)
 
-    # 2. Validate keys if in strict mode
+    # 2. Validate shard filenames from the index
+    # This prevents path traversal attacks and extension confusion attacks
+    # (e.g. a safetensors index referencing .bin pickle files)
+    expected_extension = Path(filename_pattern.format(suffix="")).suffix  # e.g. ".safetensors"
+    shard_files = list(set(index["weight_map"].values()))
+    for shard_file in shard_files:
+        # Reject path traversal (e.g. "../malicious.bin", absolute paths)
+        if os.path.isabs(shard_file) or ".." in Path(shard_file).parts:
+            raise ValueError(
+                f"Invalid shard filename '{shard_file}' in index file '{index_file}'. "
+                "Shard filenames must be relative paths without '..' components."
+            )
+        # Reject extension mismatch (e.g. .bin shard in a .safetensors index)
+        if not shard_file.endswith(expected_extension):
+            raise ValueError(
+                f"Invalid shard filename '{shard_file}' in index file '{index_file}'. "
+                f"Expected '{expected_extension}' extension to match the index format."
+            )
+
+    # 3. Validate keys if in strict mode
     # This is done before loading any shards to fail fast
     if strict:
         _validate_keys_for_strict_loading(model, index["weight_map"].keys())
 
-    # 3. Load each shard using `load_state_dict`
+    # 4. Load each shard using `load_state_dict`
     # Get unique shard files (multiple parameters can be in same shard)
-    shard_files = list(set(index["weight_map"].values()))
     for shard_file in shard_files:
         # Load shard into memory
         shard_path = os.path.join(save_directory, shard_file)
@@ -531,7 +549,7 @@ def _load_sharded_checkpoint(
         # Explicitly remove the state dict from memory
         del state_dict
 
-    # 4. Return compatibility info
+    # 5. Return compatibility info
     loaded_keys = set(index["weight_map"].keys())
     model_keys = set(model.state_dict().keys())
     return _IncompatibleKeys(
--- a/src/huggingface_hub/utils/_cache_manager.py
+++ b/src/huggingface_hub/utils/_cache_manager.py
@@ -657,6 +657,8 @@ def scan_cache_dir(cache_dir: str | Path | None = None) -> HFCacheInfo:
     for repo_path in cache_dir.iterdir():
         if repo_path.name == ".locks":  # skip './.locks/' folder
             continue
+        if repo_path.name == "CACHEDIR.TAG":  # skip CACHEDIR.TAG file
+            continue
         try:
             repos.add(_scan_cached_repo(repo_path))
         except CorruptedCacheException as e:
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -828,3 +828,53 @@ def test_load_torch_model_index_selection(
     load_torch_model(model, tmp_path, safe=safe, filename_pattern=filename_pattern)
     mock_load.assert_called_once()
     assert mock_load.call_args.kwargs["filename_pattern"] == expected_filename_pattern
+
+
+class TestShardedCheckpointValidation:
+    """Regression tests for shard filename validation in sharded checkpoint loading.
+
+    See https://github.com/huggingface/hackerone/issues/141 for more details.
+
+    Ensures that crafted index files cannot trick the loader into deserializing
+    unsafe pickle payloads or accessing files outside the checkpoint directory.
+    """
+
+    def test_safetensors_index_rejects_bin_shard(self, tmp_path):
+        """A safetensors index file referencing a .bin shard must be rejected."""
+        index = {
+            "metadata": {"total_size": 100},
+            "weight_map": {
+                "layer_1": "model-00001-of-00001.bin",
+            },
+        }
+        (tmp_path / "model.safetensors.index.json").write_text(json.dumps(index))
+        (tmp_path / "model-00001-of-00001.bin").touch()
+
+        with pytest.raises(ValueError, match="Invalid shard filename.*Expected '.safetensors' extension"):
+            load_torch_model(Mock(), tmp_path)
+
+    def test_safetensors_index_rejects_path_traversal(self, tmp_path):
+        """A shard filename with '..' path traversal must be rejected."""
+        index = {
+            "metadata": {"total_size": 100},
+            "weight_map": {
+                "layer_1": "../malicious.safetensors",
+            },
+        }
+        (tmp_path / "model.safetensors.index.json").write_text(json.dumps(index))
+
+        with pytest.raises(ValueError, match="Invalid shard filename.*without '..' components"):
+            load_torch_model(Mock(), tmp_path)
+
+    def test_safetensors_index_rejects_absolute_path(self, tmp_path):
+        """A shard filename with an absolute path must be rejected."""
+        index = {
+            "metadata": {"total_size": 100},
+            "weight_map": {
+                "layer_1": "/tmp/malicious.safetensors",
+            },
+        }
+        (tmp_path / "model.safetensors.index.json").write_text(json.dumps(index))
+
+        with pytest.raises(ValueError, match="Invalid shard filename.*without '..' components"):
+            load_torch_model(Mock(), tmp_path)


src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,679 @@
...
+            raise CLIError(f"Marketplace file not found: {marketplace_path}")
+        try:
+            payload = json.loads(marketplace_path.read_text(encoding="utf-8"))
+        except json.JSONDecodeError as exc:
+            raise CLIError(f"Failed to parse marketplace file: {exc}") from exc
+        if not isinstance(payload, dict):
+            raise CLIError("Invalid marketplace payload: expected a JSON object.")
+        return payload
+
+    raw_url = _raw_github_url(repo_url, "main", MARKETPLACE_PATH)
```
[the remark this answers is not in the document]
burtenshaw: imo, `main` is good.

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,679 @@
...
+    skill_dir: Path
+    status: SkillUpdateStatus
+    detail: str | None = None
+    current_revision: str | None = None
+    available_revision: str | None = None
+
+
+def load_marketplace_skills(repo_url: str | None = None) -> list[MarketplaceSkill]:
+    """Load marketplace skills from the default Hugging Face skills repository or a local override."""
+    repo_url = repo_url or DEFAULT_SKILLS_REPO
```
Wauplin: I feel that for now we should only focus on the default marketplace (i.e. HF skills) rather than having a general-purpose installer (`/plugin install ...`). Installing skills from other marketplaces can be done without `hf` so I think the benefit of having it bundled in `hf` is precisely to be easier to work with HF skills

Wauplin: It should help reduce some logic in the module, especially all the logic to load from a local repo can be dropped IMO

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,679 @@
...
+    source_origin: SkillSourceOrigin
+    repo_url: str
+    repo_ref: str | None
+    repo_path: str
+    source_url: str | None
+    installed_commit: str | None
+    installed_path_oid: str | None
+    installed_revision: str
+    installed_at: str
+    content_fingerprint: str
```
Wauplin: Same here, feels too over-engineered to me for a simple "download markdown file"-tool. At best I would only keep skills name + commit id (and not even sure it's needed...)

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,679 @@
...
+        text=True,
+    )
+    if proc.returncode != 0:
+        stderr = proc.stderr.strip() or proc.stdout.strip()
+        raise CLIError(stderr or f"Unable to resolve {ref} for {repo_url}")
+    for line in proc.stdout.splitlines():
+        parts = line.split()
+        if parts:
+            return parts[0]
+    raise CLIError(f"Unable to resolve {ref} for {repo_url}")
```
Wauplin: Let's not use any `git` subprocess commands. We can do everything with only Github API

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,679 @@
...
+        "installed_commit": source.installed_commit,
+        "installed_path_oid": source.installed_path_oid,
+        "installed_revision": source.installed_revision,
+        "installed_at": source.installed_at,
+        "content_fingerprint": source.content_fingerprint,
+    }
+    (skill_dir / SKILL_SOURCE_FILENAME).write_text(
+        json.dumps(payload, indent=2, sort_keys=True) + "\n",
+        encoding="utf-8",
+    )
```
Wauplin: Let's keep the manifest small with only useful information in it (i.e. information that is actually used to upgrade the skill). Maybe we can try to have a similar pattern as the `manifest.json` in the `hf extensions` module (cc @hanouticelina who designed it)

docs/source/en/package_reference/cli.md
```diff
@@ -3170,6 +3175,42 @@ $ hf skills preview [OPTIONS]
 
 * `--help`: Show this message and exit.
 
+### `hf skills update`
```
Wauplin: (nit) I would call if `hf skills upgrade` which conveys a bit more the fact that the skills are upgrading (an "update" can sometimes be only a "update the metadata")

docs/source/en/package_reference/cli.md
```diff
@@ -3170,6 +3175,42 @@ $ hf skills preview [OPTIONS]
...
+**Arguments**:
+
+* `[NAME]`: Optional installed skill name to update.
+
+**Options**:
+
+* `--claude`: Update skills installed for Claude.
+* `--codex`: Update skills installed for Codex.
+* `--cursor`: Update skills installed for Cursor.
+* `--opencode`: Update skills installed for OpenCode.
```
Wauplin: These 3 can be dropped since they all rely correctly use `.agents/` now (claude being the last one missing). We should remove them from the `hf skills add` command as well IMO

docs/source/en/package_reference/cli.md
```diff
@@ -3170,6 +3175,42 @@ $ hf skills preview [OPTIONS]
...
+* `-g, --global`: Use global skills directories instead of the current project.
+* `--dest PATH`: Update skills in a custom skills directory.
+* `--force`: Overwrite local modifications when updating a skill.
```
Wauplin: `--force` is kinda necessary for an upgrade no?

docs/source/en/package_reference/cli.md
```diff
@@ -3120,6 +3120,7 @@ $ hf skills [OPTIONS] COMMAND [ARGS]...
 
 * `add`: Download a skill and install it for an AI...
 * `preview`: Print the generated SKILL.md to stdout.
```
Wauplin: I feel that `hf skills add` and `hf skills preview` description could be harmonized now that we have more than 1 skill that can be installed

burtenshaw:
Thanks for the review @Wauplin. I've responded to each item and cut down the bloat.

burtenshaw:
@JiwaniZakir thanks for the tokens, but we got this.

docs/source/en/package_reference/cli.md
```diff
@@ -3118,38 +3118,41 @@ $ hf skills [OPTIONS] COMMAND [ARGS]...
...
 * `--help`: Show this message and exit.
 
 Examples
   $ hf skills add
+  $ hf skills add huggingface-gradio --dest=~/my-skills
   $ hf skills add --global
-  $ hf skills add --claude --cursor
-  $ hf skills add --codex --opencode --cursor --global
+  $ hf skills add --claude
+  $ hf skills add --claude --global
```
Wauplin:
```suggestion
  $ hf skills add huggingface-gradio --claude --global
```

(nit) let's add an example showing extension skill can also be installed globally

src/huggingface_hub/cli/skills.py
```diff
@@ -346,36 +345,80 @@ def skills_add(
...
+        central_skill_path = _install_to(central_path, name, force)
+        print(f"Installed '{name}' to central location: {central_skill_path}")
+
+        if claude:
+            agent_target = CLAUDE_GLOBAL if global_ else CLAUDE_LOCAL
+            link_path = _create_symlink(agent_target, name, central_skill_path, force)
+            print(f"Created symlink: {link_path}")
+    except CLIError as exc:
+        print(str(exc))
+        raise typer.Exit(code=1) from exc
```
Wauplin: no need for the `try/except`. The CLIError is properly caught in `hf.py` directly (for all commands at once)

```suggestion
    if dest:
        if claude or global_:
            print("--dest cannot be combined with --claude or --global.")
            raise typer.Exit(code=1)
        skill_dest = _install_to(dest, name, force)
        print(f"Installed '{name}' to {skill_dest}")
        return

    # Install to central location
    central_path = CENTRAL_GLOBAL if global_ else CENTRAL_LOCAL
    central_skill_path = _install_to(central_path, name, force)
    print(f"Installed '{name}' to central location: {central_skill_path}")

    if claude:
        agent_target = CLAUDE_GLOBAL if global_ else CLAUDE_LOCAL
        link_path = _create_symlink(agent_target, name, central_skill_path, force)
        print(f"Created symlink: {link_path}")
```

src/huggingface_hub/cli/skills.py
```diff
@@ -346,36 +345,80 @@ def skills_add(
...
+        ),
+    ] = None,
+) -> None:
+    """Upgrade installed Hugging Face marketplace skills."""
+    try:
+        roots = _resolve_update_roots(
+            claude=claude,
+            global_=global_,
+            dest=dest,
+        )
```
Wauplin: (nit)

```suggestion
        roots = _resolve_update_roots(claude=claude, global_=global_, dest=dest)
```

src/huggingface_hub/cli/skills.py
```diff
@@ -346,36 +345,80 @@ def skills_add(
...
+    ] = False,
+    dest: Annotated[
+        Optional[Path],
+        typer.Option(
+            help="Upgrade skills in a custom skills directory.",
+        ),
+    ] = None,
+) -> None:
+    """Upgrade installed Hugging Face marketplace skills."""
+    try:
```
Wauplin:
```suggestion
```

Same here, no need for the try/except

tests/test_cli.py
```diff
@@ -1,10 +1,15 @@
+from __future__ import annotations
+
+import base64
```
Wauplin:
```suggestion
import base64
```

no need anymore (minimal version is now Python3.10)

tests/test_cli.py
Wauplin, line 1: I find the entire test suite quite verbose and doesn't actually test the feature. In `huggingface_hub` we usually try to test features without monkeypatching (i.e. with remote calls).

Here I think it's fine if create a tmp directory, run `hf skills add huggingface-gradio`, checks it's installed in the tmp dir, run `hf skills upgrade` and check it has been correctly upgraded (or at least check the output that it has not been upgraded but a remote check has been done). 

Such testing would have a much smaller scope but at least it would be self-explainable and mimic an actual use of the CLI

uv.lock
Wauplin, line 1: should be removed from PR (likely just a dev artifact on your side)

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
+"""Internal helpers for Hugging Face marketplace skill installation and upgrades."""
+
+from __future__ import annotations
+
```
Wauplin:
```suggestion
```

(nit) (not needed anymore)

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+from pathlib import Path, PurePosixPath
+from typing import Any, Literal
+
+from huggingface_hub.errors import CLIError
+from huggingface_hub.utils import get_session
+
+
+DEFAULT_SKILLS_REPO_ID = "huggingface/skills"
+DEFAULT_SKILLS_REPO_OWNER = "huggingface"
+DEFAULT_SKILLS_REPO_NAME = "skills"
```
Wauplin:
```suggestion
DEFAULT_SKILLS_REPO_OWNER, DEFAULT_SKILLS_REPO_NAME = DEFAULT_SKILLS_REPO_ID.split("/")
```

src/huggingface_hub/cli/_skills.py
Wauplin, line 1: I've read this module and "it looks good" but can't say that I reviewed it in-depth. I think it's fine though given the scope is non-critical.

julien-c:
do you also agree with @julien-c?

src/huggingface_hub/cli/skills.py
```diff
@@ -313,52 +298,66 @@ def _remove_existing(path: Path, force: bool) -> None:
...
     link_path.symlink_to(os.path.relpath(central_skill_path, agent_skills_dir))
 
     return link_path
 
 
+def _resolve_update_roots(
+    *,
+    claude: bool,
+    global_: bool,
+    dest: Optional[Path],
```
hanouticelina: we recently bump the minimum Python version to 3.10
```suggestion
    dest: Path | None,
```

src/huggingface_hub/cli/skills.py
```diff
@@ -381,36 +380,80 @@ def skills_add(
...
+    examples=[
+        "hf skills upgrade",
+        "hf skills upgrade hf-cli",
+        "hf skills upgrade huggingface-gradio --dest=~/my-skills",
+        "hf skills upgrade --claude",
+    ],
+)
+def skills_upgrade(
+    name: Annotated[
+        Optional[str],
```
hanouticelina:
```suggestion
        str | None,
```

src/huggingface_hub/cli/skills.py
```diff
@@ -381,36 +380,80 @@ def skills_add(
...
+    global_: Annotated[
+        bool,
+        typer.Option(
+            "--global",
+            "-g",
+            help="Use global skills directories instead of the current project.",
+        ),
+    ] = False,
+    dest: Annotated[
+        Optional[Path],
```
hanouticelina:
```suggestion
        Path | None,
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+            installed_revision=installed_revision,
+        ),
+    )
+
+
+def _validate_installed_skill_dir(skill_dir: Path) -> None:
+    skill_file = skill_dir / "SKILL.md"
+    if not skill_file.is_file():
+        raise RuntimeError(f"Installed skill is missing SKILL.md: {skill_file}")
+    skill_file.read_text(encoding="utf-8")
```
hanouticelina: is reading the file needed here? the `is_file()` check already validates existence
```suggestion
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+        )
+
+    return SkillUpdateInfo(
+        name=skill_dir.name,
+        skill_dir=skill_dir,
+        status="update_available",
+        detail="update available",
+        current_revision=current_revision,
+        available_revision=available_revision,
+    )
```
hanouticelina, line 347: small preference for using `dataclasses.replace` when dealing with frozen dataclasses 
```suggestion
    base = SkillUpdateInfo(name=skill_dir.name, skill_dir=skill_dir, status="unmanaged")

    manifest, error = read_installed_skill_manifest(skill_dir)
    if manifest is None:
        return replace(base, status="invalid_metadata" if error else "unmanaged", detail=error)

    skill = marketplace_skills.get(skill_dir.name.lower())
    if skill is None:
        return replace(
            base,
            status="source_unreachable",
            detail=f"Skill '{skill_dir.name}' is no longer available in {DEFAULT_SKILLS_REPO_ID}.",
            current_revision=manifest.installed_revision,
        )

    current_revision = manifest.installed_revision
    try:
        available_revision = _resolve_available_revision(skill)
    except Exception as exc:
        return replace(base, status="source_unreachable", detail=str(exc), current_revision=current_revision)

    status = "up_to_date" if available_revision == current_revision else "update_available"
    return replace(
        base,
        status=status,
        detail="update available" if status == "update_available" else None,
        current_revision=current_revision,
        available_revision=available_revision,
    )
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+
+
+def _apply_single_update(update: SkillUpdateInfo) -> SkillUpdateInfo:
+    if update.status in {"up_to_date", "unmanaged", "invalid_metadata", "source_unreachable"}:
+        return update
```
hanouticelina: iiuc, only `"update_available"` should trigger an install
```suggestion
    if update.status != "update_available":
        return update
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+        )
+
+    return SkillUpdateInfo(
+        name=update.name,
+        skill_dir=update.skill_dir,
+        status="updated",
+        detail="updated",
+        current_revision=update.current_revision,
+        available_revision=refreshed_manifest.installed_revision,
+    )
```
hanouticelina: two things here:
- Removing the manifest re-read before install,  `_evaluate_update` already validated it.
- Removing the manifest re-read after install,  if `install_marketplace_skill` succeeds, the manifest was written. No need to read it back to confirm
```suggestion
    try:
        skill = get_marketplace_skill(update.skill_dir.name)
        install_marketplace_skill(skill, update.skill_dir.parent, force=True)
    except Exception as exc:
        return replace(update, status="source_unreachable", detail=str(exc))

    return replace(update, status="updated", detail="updated")
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+
+def _validate_installed_skill_dir(skill_dir: Path) -> None:
+    skill_file = skill_dir / "SKILL.md"
+    if not skill_file.is_file():
+        raise RuntimeError(f"Installed skill is missing SKILL.md: {skill_file}")
+    skill_file.read_text(encoding="utf-8")
+
+
+def _extract_remote_github_path(revision: str, source_path: str, install_dir: Path) -> None:
+    tar_bytes = _github_api_get_bytes(f"tarball/{revision}")
```
hanouticelina: `_github_api_get_bytes` is a one liner with a single call. let's juste inline it
```suggestion
    tar_bytes = _github_api_get(f"tarball/{revision}").content
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -0,0 +1,484 @@
...
+            url,
+            params=params,
+            headers={"Accept": "application/vnd.github+json"},
+            follow_redirects=True,
+            timeout=GITHUB_API_TIMEOUT,
+        )
+        response.raise_for_status()
+    except Exception as exc:  # noqa: BLE001
+        raise CLIError(f"Failed to fetch '{endpoint}' from {DEFAULT_SKILLS_REPO_ID}: {exc}") from exc
+    return response
```
hanouticelina: this can be one helper and then replace `_github_api_get_json` / `_github_api_get_bytes` / `_github_api_get` calls with `_fetch_from_skills_repo(..., as_json=True)` or `_fetch_from_skills_repo(...)` respectively
```suggestion
def _fetch_from_skills_repo(endpoint: str, params: dict[str, Any] | None = None, *, as_json: bool = False) -> Any:
    url = f"https://api.github.com/repos/{DEFAULT_SKILLS_REPO_OWNER}/{DEFAULT_SKILLS_REPO_NAME}/{endpoint.lstrip('/')}"
    try:
        response = get_session().get(
            url,
            params=params,
            headers={"Accept": "application/vnd.github+json"},
            follow_redirects=True,
            timeout=GITHUB_API_TIMEOUT,
        )
        response.raise_for_status()
    except Exception as exc:  # noqa: BLE001
        raise CLIError(f"Failed to fetch '{endpoint}' from {DEFAULT_SKILLS_REPO_ID}: {exc}") from exc
    if as_json:
        try:
            return response.json()
        except Exception as exc:  # noqa: BLE001
            raise CLIError(f"Failed to decode GitHub API response for '{endpoint}': {exc}") from exc
    return response.content
```

JiwaniZakir:
Agreed with @Wauplin — scoping this to the HF skills marketplace keeps the surface area manageable and the UX simpler. The value proposition is exactly that `hf skills add gradio` just works without users needing to know where the skill lives. Once the naming simplification lands in huggingface/skills#99, this should be ready to revisit for merge.

JiwaniZakir:
Agreed with Wauplin's point — scoping to the HF skills marketplace keeps the surface area small and the UX predictable. The atomic directory replacement and manifest tracking already give us a solid foundation to expand later if needed, but shipping a general-purpose installer now would front-load a lot of security and validation complexity we haven't worked through yet. The cursor bot's note about the test mock not actually triggering the failure path is worth fixing before merge — that test gives false confidence in the rollback behavior.

JiwaniZakir:
Agreed with @Wauplin — scoping this strictly to the HF marketplace keeps the surface area manageable and avoids us having to reason about arbitrary remote sources. The symlink into Claude's legacy skills directory feels like it should be opt-in or dropped entirely for now, since it adds cross-tool coupling that's hard to reason about at install time. Once the naming simplification in huggingface/skills#99 lands, the path resolution logic here should be much cleaner.

JiwaniZakir:
Agree with @Wauplin on scoping — keeping `hf skills add` tied to the default `huggingface/skills` marketplace keeps the surface area small and the mental model simple. If we want multi-marketplace support later, it's easier to add than to remove. On the `main` branch hardcode, that's fine as long as `huggingface/skills` itself treats `main` as stable; we should just make sure that expectation is documented somewhere so it doesn't quietly break if the upstream repo ever switches default branches.

JiwaniZakir:
The naming simplification in [skills/pull/99](https://github.com/huggingface/skills/pull/99) should land before this merges — having `hf skills add gradio` resolve to a tarball named `hugging-face-gradio` would be confusing to users and a pain to clean up later. On the manifest tracking side, worth confirming the revision field pins to a tag/SHA rather than a branch ref, otherwise `hf skills upgrade` won't reliably detect drift.

PR #3968, opened 2026-03-24
More robust stream to stdout in hf cp command

Wauplin:
2 things:
- read from tmp file by chunks (avoid loading entire file in memory)
- use `SoftTemporaryDirectory` which tries different methods to remove the tmp directory (and give up if not possible)

The best approach would be to directly pipe from hf-xet to the stdout but this is not yet possible (coming soon)

(related internal [slack thread](https://huggingface.slack.com/archives/C0737MXJ26A/p1774364636949359))

---

> [!NOTE]
> **Low Risk**
> Low risk: change is limited to the `hf buckets cp ... -` download path and mainly affects memory usage and temp cleanup behavior; primary risk is subtle stdout/progress-bar behavior changes for large downloads.
> 
> **Overview**
> Makes `hf buckets cp` downloads to stdout more robust by replacing `tempfile.TemporaryDirectory` with `SoftTemporaryDirectory` for best-effort cleanup.
> 
> When piping a bucket file to `-`, it now writes the downloaded temp file to `stdout` in 32MB chunks instead of reading the entire file into memory.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a422b56547051f4380710c9486b4de9f5741cf9d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/buckets.py
+++ b/src/huggingface_hub/cli/buckets.py
@@ -17,7 +17,6 @@
 import json
 import os
 import sys
-import tempfile
 from datetime import datetime
 from typing import Annotated, Optional, Union
 


src/huggingface_hub/cli/buckets.py
```diff
@@ -981,11 +986,12 @@ def cp(
             if not pbar_was_disabled:
                 disable_progress_bars()
             try:
-                with tempfile.TemporaryDirectory() as tmp_dir:
+                with SoftTemporaryDirectory() as tmp_dir:
                     tmp_path = os.path.join(tmp_dir, filename)
                     api.download_bucket_files(bucket_id, [(prefix, tmp_path)])
                     with open(tmp_path, "rb") as f:
-                        sys.stdout.buffer.write(f.read())
+                        while chunk := f.read(32_000_000):  # 32MB chunks
```
hanouticelina, line 993: (nit) let's follow the same convention as in `huggingface_hub.constants`
```suggestion
                        while chunk := f.read(32 * 1024 * 1024):  # 32 MiB chunks
```

Wauplin: well in `constants.py` we have

```py
DOWNLOAD_CHUNK_SIZE = 10 * 1024 * 1024
MAX_HTTP_DOWNLOAD_SIZE = 50 * 1000 * 1000 * 1000  # 50 GB
```

:melting_face: 

I prefer to keep the `MB` notation as that's what we have in all user-facing messages (also doesn't change much things)

PR #3983, opened 2026-03-26
Interactive CLI autoupdate prompt

Wauplin:
## What does this PR do?

Replaces the passive "a new version is available" warning with an interactive autoupdate prompt when appropriate, and moves all update-related output to stderr.

### Behavior summary

| Condition | Behavior |
|---|---|
| TTY + update command known | Interactive prompt: `Do you want to update now? [Y/n] (update command)` |
| Non-TTY or agent (piped stdin/stderr) | Warning message on stderr (no prompt) |
| Update command unknown | Warning message on stderr without update command hint |
| User accepts update | Runs update command, then exits so user can re-run their command |
| User declines update | Shows skip message on stderr with manual update command |
| Update fails | Shows error with exit code on stderr |

### Example prompt (pip install)

```
  A new version of huggingface_hub is available: 1.7.0 → 1.8.0

  Do you want to update now? [Y/n] (/path/to/python -m pip install -U huggingface_hub)
```

### Changes

**`src/huggingface_hub/cli/_cli_utils.py`**
- `_check_cli_update()`: When an update is available:
  - If stdin and stderr are TTY and the update command is known → call `_prompt_autoupdate()` for interactive update
  - Otherwise → print a warning to **stderr** (previously stdout via `click.echo`)
- New `_prompt_autoupdate()`: Interactive prompt with colored output. The update command is shown in gray on the prompt line. Prompt itself is written to stderr via `click.echo`. Runs the update command via `subprocess.call(cmd)` (no `shell=True`); exits on success so the user re-runs their command.
- `_get_huggingface_hub_update_command()` / `_get_transformers_update_command()`: Return `Optional[list[str]]` — an argv list passed directly to `subprocess.call()`. This avoids shell quoting issues entirely (no `shlex`, no `shell=True`). Return `None` for unknown installation methods.

**`src/huggingface_hub/utils/_runtime.py`**
- `installation_method()`: Returns `brew`, `hf_installer`, `pip`, or `unknown`
- New detection helper: `_is_pip_available()`

### Design decisions

- **No `hf update` command**: Per the Slack discussion, autoupdate makes a standalone command unnecessary and avoids bloating `hf --help`.
- **Exit after update**: After a successful update, the CLI exits with code 0. The user should re-run their command — running the original command with a stale in-memory module would be unreliable.
- **Graceful degradation**: If we can't determine the update command (returns `None`) or can't prompt (non-TTY), we simply show the warning on stderr — no interactive prompt, no risk of hanging.
- **stderr for everything**: All update-related output (including the prompt and skip message) goes to stderr so it never interferes with piped command output.
- **Correct venv targeting**: pip commands use `[sys.executable, "-m", "pip", ...]` to ensure the update targets the same Python environment that's running the CLI.
- **argv list, no shell**: Update commands are returned as `list[str]` and executed via `subprocess.call(cmd)` without `shell=True`, which is robust on both POSIX and Windows regardless of spaces in paths.

[Slack Thread](https://huggingface.slack.com/archives/C02V5EA0A95/p1774518638841249?thread_ts=1774518638.841249&cid=C02V5EA0A95)

<div><a href="https://cursor.com/agents/bc-369c5bd9-75da-5d8e-b02f-7687f940de0f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-369c5bd9-75da-5d8e-b02f-7687f940de0f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Wauplin:
@cursor a few highlevel comments:
- no need to support conda and pipx
- for pip (and maybe uv as well?), make sure to use `which python` (or similar) + `python -m pip ...` to ensure the correct venv is updated. It should be the same as the one running the command
- when printing `Do you want to update now? [Y/n]`, can you add afterward in parenthesis the command that will be launched, in gray. Also make sure the message is user friendly and not abrupt
- the interactive prompt should be in stderr and the skip message as well (if user declines) to that main stdout is not polluted

src/huggingface_hub/utils/_runtime.py
```diff
@@ -316,26 +317,28 @@ def is_colab_enterprise() -> bool:
...
+        return "pip"
+    return "unknown"
 
 
 def _is_brew_installation() -> bool:
-    """Check if running from a Homebrew installation.
-
-    Note: AI-generated by Claude.
-    """
+    """Return `True` if running from a Homebrew installation."""
```
Wauplin:
```suggestion
    """Check if running from a Homebrew installation.

    Note: AI-generated by Claude.
    """
```

src/huggingface_hub/utils/_runtime.py
```diff
@@ -356,6 +359,18 @@ def _is_hf_installer_installation() -> bool:
     return marker.exists()
 
 
+def _is_uv_installation() -> bool:
+    """Return `True` if the environment was created by uv (uv sets UV_VIRTUAL_ENV or leaves a marker)."""
+    if os.environ.get("UV_VIRTUAL_ENV"):
+        return True
+    return (Path(sys.prefix) / "uv.lock").exists()
```
Wauplin: @cursor please remove uv detection (always fallback to pip)

Wauplin (commented):
@cursor make sure to be consistent between click.echo(..., file=sys.stderr) and sys.stderr.write

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,105 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+    """Return the command to update transformers, or None if the installation method is unknown."""
     method = installation_method()
     if method == "hf_installer" and os.name == "nt":
         return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex" -WithTransformers'
     elif method == "hf_installer":
         return "curl -LsSf https://hf.co/cli/install.sh | bash -s -- --with-transformers"
-    else:  # brew/unknown => likely pip
-        return "pip install -U transformers"
+    elif method in ("pip", "brew"):
+        return f"{sys.executable} -m pip install -U transformers"
```
```suggestion
    elif method == "pip":
        return f"{sys.executable} -m pip install -U transformers"
```

Wauplin (commented):
✔️

Wauplin:
@hanouticelina ready for review!

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,105 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+        file=sys.stderr,
+        nl=False,
+    )
+    try:
+        answer = sys.stdin.readline().strip().lower()
+    except (EOFError, KeyboardInterrupt):
+        click.echo("", file=sys.stderr)
+        return
+
+    if answer in ("", "y", "yes"):
```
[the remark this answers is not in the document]
Wauplin: @cursor please fix

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,105 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
     if method == "brew":
         return "brew upgrade hf"
     elif method == "hf_installer" and os.name == "nt":
         return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex"'
     elif method == "hf_installer":
         return "curl -LsSf https://hf.co/cli/install.sh | bash -"
-    else:  # unknown => likely pip
-        return "pip install -U huggingface_hub"
+    elif method == "pip":
+        return f"{sys.executable} -m pip install -U huggingface_hub"
```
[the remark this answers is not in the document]
Wauplin: @cursor please fix

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,105 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
     method = installation_method()
     if method == "hf_installer" and os.name == "nt":
         return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex" -WithTransformers'
     elif method == "hf_installer":
         return "curl -LsSf https://hf.co/cli/install.sh | bash -s -- --with-transformers"
-    else:  # brew/unknown => likely pip
-        return "pip install -U transformers"
+    elif method == "pip":
+        return f"{sys.executable} -m pip install -U transformers"
+    return None
```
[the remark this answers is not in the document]
Wauplin: not a problem (cannot brew install transformers, right?)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +755,111 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+    except (EOFError, KeyboardInterrupt):
+        click.echo("", file=sys.stderr)
+        return
+
+    if raw_answer == "":
+        # EOF (e.g. Ctrl+D) — treat as cancellation, not acceptance
+        click.echo("", file=sys.stderr)
+        return
+
+    answer = raw_answer.strip().lower()
```
```suggestion
    answer = raw_answer.strip().lower()  # Note: if user press 'Enter', raw_answer is `\n`
```

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +755,111 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+) -> None:
+    """Interactively ask the user if they want to update, and run the update command if accepted.
+
+    After a successful update the CLI exits so the user can re-run their command with the new version.
+    All output goes to stderr to keep stdout clean for command output.
+    """
+    click.echo("", file=sys.stderr)
+    click.echo(
+        ANSI.yellow(
+            f"  A new version of {library} is available: {ANSI.bold(current_version)} → {ANSI.bold(latest_version)}"
```
```suggestion
            f"  A new version of {library} is available: {current_version} → {latest_version}"
```

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +755,111 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
     if method == "brew":
         return "brew upgrade hf"
     elif method == "hf_installer" and os.name == "nt":
         return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex"'
     elif method == "hf_installer":
         return "curl -LsSf https://hf.co/cli/install.sh | bash -"
-    else:  # unknown => likely pip
-        return "pip install -U huggingface_hub"
+    elif method == "pip":
+        return f"{shlex.quote(sys.executable)} -m pip install -U huggingface_hub"
```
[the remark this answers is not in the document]
Wauplin: @cursor instead of using shlex, please modify the method to return a List[str] that will be passed to subprocess down the line. This should make everything more robust

Wauplin (commented):
Much better, thanks cursor!

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,118 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+    current_version: str,
+    latest_version: str,
+    update_command: list[str],
+) -> None:
+    """Interactively ask the user if they want to update, and run the update command if accepted.
+
+    After a successful update the CLI exits so the user can re-run their command with the new version.
+    All output goes to stderr to keep stdout clean for command output.
+    """
+    display_cmd = " ".join(update_command)
```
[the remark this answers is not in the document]
Wauplin: that's fine

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,118 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
     method = installation_method()
     if method == "hf_installer" and os.name == "nt":
-        return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex" -WithTransformers'
+        return [
+            "powershell",
+            "-NoProfile",
+            "-Command",
+            "iwr -useb https://hf.co/cli/install.ps1 | iex",
+            "-WithTransformers",
+        ]
```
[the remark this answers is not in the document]
Wauplin: @cursor please fix

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -711,7 +714,9 @@ def check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> Non
     """
     Check whether a newer version of a library is available on PyPI.
 
-    If a newer version is found, notify the user and suggest updating.
+    If a newer version is found and stdin/stdout are attached to a TTY, prompt the user to update interactively.
```
hanouticelina: since we check `if sys.stdin.isatty() and sys.stderr.isatty() and ...:` line 766
```suggestion
    If a newer version is found and stdin/stderr are attached to a TTY, prompt the user to update interactively.
```

hanouticelina: (which is correct)

hanouticelina (approved):
Thank you! looks good to me

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -750,41 +754,118 @@ def _check_cli_update(library: Literal["huggingface_hub", "transformers"]) -> No
...
+    """Return the command to update transformers as an argv list, or None if the installation method is unknown."""
     method = installation_method()
     if method == "hf_installer" and os.name == "nt":
-        return 'powershell -NoProfile -Command "iwr -useb https://hf.co/cli/install.ps1 | iex" -WithTransformers'
+        return [
+            "powershell",
+            "-NoProfile",
+            "-Command",
+            "iwr -useb https://hf.co/cli/install.ps1 | iex",
+            "-WithTransformers",
```
```suggestion
            "iwr -useb https://hf.co/cli/install.ps1 | iex -WithTransformers",
```

tomprimozic:
Guys, this is a pretty terrible PR.

You at minimum need to add a flag that prevents the prompt, otherwise my scripts are stuck at the "Do you want to update" question! I see you do check for TTY, but there's tools that emulate it (e.g. Jupyter notebooks).

hanouticelina:
Hey @tomprimozic, appreciate the feedback! though "terrible" feels a bit strong for a PR that's trying to improve update adoption 🙂

I tried finding a fix but there's no robust way to tell the difference between a real terminal (where someone is typing) and one where no one is actually reading input, `isatty()` returns `True` in both. Jupyter is exactly that case: when you run `!hf ...` in a cell, the subprocess inherits the terminal that launched jupyter as its `stdin`, so it looks interactive from the CLI's point of view but your input go to the jupyter server, not the `hf` process.

fwiw, for scripted workflows, most `hf` commands have a Python API equivalent in `huggingface_hub`. usually a cleaner fit than shelling out regardless of the prompt question. by the way, what's the exact command running?


and yes we could add an env variable to disable the update check, but I'll let @Wauplin weigh in in case he has ideas on handling this more robustly.

tomprimozic:
usually tools (e.g. `mamba`) provide arguments that "just do it" (for exactly these reasons, running commands in scripts), e.g. `--quiet` or `--force` of `--yes` (well, `--no` in this case)

I'm running

```
hf download lmms-lab/COCO-Caption --repo-type dataset --include "data/val-00001-of-00013.parquet" --local-dir .
```

Wauplin:
Hey @tomprimozic @hanouticelina @scruffaluff, thanks for the discussion. I do think interactive autoupdates is the best way to ensure users always run the latest version of the tool but I understand that it comes with too many drawbacks. We could have added a flag to disable autoupdate but one need to know about the flag before using it, which is likely to cause friction as well. 

What I did is to entirely remove the interactive update and add a new command `hf update` instead. The autocheck is still there and will nudge users to run `hf update` if they run an outdated version but it won't prompt for autoupdate. The autocheck is made only once per 24h and can be bypassed using `HF_HUB_DISABLE_UPDATE_CHECK` env variable.

See https://github.com/huggingface/huggingface_hub/pull/4131 for full details and let me know what you think!

PR #3987, opened 2026-03-27
Support None type in strict dataclass

Wauplin:
cc @hmellor @zucchini-nlp 

Corner case where a config value accepts only `None`. Since `None` is not a type in Python, we currently fail to validate it. This PR adds support for it + tests.

In practice, it allows to do

```py
from dataclasses import dataclass
from huggingface_hub.dataclasses import strict

@strict
@dataclass
class BaseConfig:
    field: None = None

BaseConfig()
```

---

> [!NOTE]
> **Low Risk**
> Low risk: small, isolated change to `type_validator` to handle the `None` annotation, with targeted tests added to cover the new behavior.
> 
> **Overview**
> Strict dataclass type validation now treats the `None` annotation as a valid type expectation (i.e., fields annotated as `None` must be set to `None`).
> 
> This adds a dedicated `_validate_none` path in `type_validator` and extends the strict dataclass test suite to cover both valid and invalid `None`/`NoneType` cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 70477d6d8c4a1d7ce60c00f969efd0999873f58d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/dataclasses.py
+++ b/src/huggingface_hub/dataclasses.py
@@ -468,7 +468,7 @@ def type_validator(name: str, value: Any, expected_type: Any) -> None:
 
     if expected_type is Any:
         return
-    if expected_type is None:
+    elif expected_type is None:
         _validate_none(name, value)
     elif validator := _BASIC_TYPE_VALIDATORS.get(origin):
         validator(name, value, args)


zucchini-nlp (approved):
Great, in case someone actually wants to override the inherited annotation

Issue #3979, opened 2026-03-26
[CLI] Dual mode in CLI output (human vs agents)


Wauplin:
# Making `hf` CLI Dual-Mode: The `out` Framework

## Problem

The `hf` CLI is used by both humans and AI agents. Today's output is human-only: ANSI bold, padded tables truncated at 35 chars, prose messages, emoji booleans `✔`, progress spinners. Agents need: no ANSI, no truncation, compact structured output, token efficiency.

The real problem isn't "how do I output differently" — it's that **print logic and business logic are interleaved**. Every command calls `print()` directly with hard-coded formatting. We want a mini-framework where developers call semantic output methods and the framework picks the rendering.

---

### What we have

- `OutputFormat` enum (`table` | `json`) + `FormatOpt` annotation (`_cli_utils.py:551`)
- `--json` shorthand → auto-rewritten to `--format json` in `HFCliTyperGroup`
- `--quiet` / `-q` → prints IDs only
- `print_list_output()` handles table/json/quiet for ~12 command files
- `ANSI` class respects `NO_COLOR` env var
- `is_agent()` helper in progress (not merged yet)
- `build_skill_md()` in `skills.py` already walks the Click command tree to auto-generate docs

### Output patterns today

```python
# Pattern 1: Action — ~30 call sites (repos, collections, discussions, webhooks, jobs, spaces)
print(f"Successfully created {ANSI.bold(repo_url.repo_id)} on the Hub.")

# Pattern 2: List — ~12 call sites, shared utility
print_list_output(results, format=format, quiet=quiet)

# Pattern 3: Info/detail — ~8 call sites
print(json.dumps(api_object_to_dict(info), indent=2))

# Pattern 4: Error (hf.py:117)
print(f"Error: {message}", file=sys.stderr)
print(ANSI.gray("Set HF_DEBUG=1 as environment variable for full traceback."))

# Pattern 5: Confirmation (repos.py:504)
choice = input("Proceed? [Y/n] ")  # ← hangs for agents!

# Pattern 6: Progress (spaces.py, buckets.py)
status = StatusLine(); status.update("building...")

# Pattern 7: Streaming (jobs.py)
for log in api.fetch_job_logs(...): print(log)
```

### What's missing

- Action commands have no structured output (just prose with ANSI)
- Info commands always `indent=2` (no compact mode)
- No auto-switch between human/agent rendering
[...]

PR #3991, opened 2026-03-27
[CLI] Add agent detection helpers

hanouticelina:
Part of #3979

This PR adds two functions that detect whether the CLI is being invoked by an AI coding agent, based on environment variables set by the agent's shell session. detection follows the same approach as [`@vercel/detect-agent`](https://github.com/vercel/vercel/tree/main/packages/detect-agent): `AI_AGENT`/`AGENT` as universal standards, then tool-specific env vars checked in priority order.

Tested with Claude Code, Codex, Cursor and OpenCode ✅ 

---

> [!NOTE]
> **Low Risk**
> Low risk: adds environment-variable based detection and appends a new `agent/...` segment to the User-Agent string when telemetry is enabled. Main impact is slightly changed request metadata and environment dump output; no auth or data-path logic changes.
> 
> **Overview**
> Adds new `utils` helpers `detect_agent()`/`is_agent()` that infer an AI coding agent from standard (`AI_AGENT`/`AGENT`) and tool-specific environment variables.
> 
> When telemetry is enabled, HTTP requests now include `agent/<name>` in the generated `User-Agent`, and `dump_environment_info()` reports whether the process appears to be run by an agent.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 50728513f4720ea122a156f57f24766bb59a432b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4005, opened 2026-03-30
Add `out` output singleton with agent/humain mode rendering

hanouticelina:
Part of #3979 (Phase 1).

Introduces the `out` singleton, a mini framework where CLI commands call semantic output methods and the framework picks the rendering based on the current mode (`human`, `agent`, `json`).

### What's new

**New `OutputFormatWithAuto` enum**: `auto | human | agent | json | quiet`

The existing `OutputFormat` (`table | json`) is kept for non-migrated commands. Migrated commands use `OutputFormatWithAuto` with `--format` accepting the new modes. `--json` shorthand continues to work.

> Once all commands are migrated, `OutputFormat` will be removed and `OutputFormatWithAuto` renamed to `OutputFormat`.

**Mode resolution** (first match wins):
1. `--format` flag per command: `hf auth whoami --format json`
2. Auto-detection via `is_agent()` → `agent` if detected, else `human`

**`Output` methods**:
- `out.text(human, agent=)` — free-form message (ANSI stripped in agent mode)
- `out.table(headers, rows)` — tabular data (human: padded + truncated, agent: TSV no truncation, json: JSON array, quiet: first column only)
- `out.dict(data)` — structured JSON in all modes (indented for human, compact otherwise). For info commands (`hf models info`, etc.)
- `out.result(msg, **data)` — success summary (human: `✓` + key-value lines, agent: `key=val` pairs, json: `json.dumps(data)`, quiet: first value)
- `out.warning(msg)` / `out.error(msg)` / `out.hint(msg)` — stderr in all modes

### Migration proof: `hf auth whoami`

**Breaking change**: `hf auth whoami --format json` now returns `{"user": "...", "orgs": "...", "endpoint": "..."}` instead of the full `whoami` API response. Not logged in now exits with code 1 (was 0).

```bash
# Auto-detect (agent mode if running inside an agent)
hf auth whoami

# Explicit modes
hf auth whoami --format human
hf auth whoami --format agent
hf auth whoami --format json

# Shorthand still works
hf auth whoami --json
```

### Not included yet

- Tests for `_output.py` — will be added in follow-up commits.
- `out.confirm()` — will be added in a subsequent PR.
- Migration of other commands — each command will be migrated in subsequent PRs.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new CLI output abstraction and changes `hf auth whoami` output schema and error exit codes, which can break scripts relying on previous stdout JSON and success-on-error behavior.
> 
> **Overview**
> Adds a new `cli/_output.py` **output framework** with a global `out` singleton that renders semantic outputs (`text`, `table`, `dict`, `result`, `warning/error/hint`) differently for `human`, `agent`, `json`, and `quiet` modes, with an `auto` mode that selects `agent` vs `human`.
> 
> Extends CLI parsing to rewrite `-q/--quiet` into `--format quiet` (when a command supports `--format` and doesn’t already define a quiet flag), and introduces `OutputFormatWithAuto` to support the expanded `--format` values.
> 
> Migrates `hf auth whoami` to the new output system: `--format` now accepts `agent|auto|human|json|quiet`, the JSON payload becomes a summarized `{user, orgs, endpoint}` shape (not the raw API response), and “not logged in” now exits with code `1`; docs and tests are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 61030ef0bb75e5e0a07aaf967f680667b18f1dcb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1642,7 +1642,6 @@ class TestAuthWhoamiCommand:
         parsed = json.loads(result.stdout)
         assert parsed["user"] == "testuser"
         assert parsed["orgs"] == "org1,org2"
-        assert parsed["orgs"] == "org1,org2"
 
     def test_whoami_not_logged_in(self, runner: CliRunner) -> None:
         with patch("huggingface_hub.cli.auth.get_token", return_value=None):


hanouticelina:
@cursoragent review

hanouticelina:
> High — hf auth whoami --format json is no longer backward-compatible when the user is unauthenticated (src/huggingface_hub/cli/auth.py:160-163).
> The base branch emits {"error": "Not logged in"} on stdout and exits successfully for the legacy JSON flag; this patch now sends the error through out.error() (stderr) and exits with code 1. That breaks existing callers doing json.loads(stdout) on the documented backward-compat path.

I think this is fine. for errors, json mode should print plain text not JSON on stderr

> Medium — The advertised mode precedence is violated for the first migrated command (src/huggingface_hub/cli/hf.py:73-77, src/huggingface_hub/cli/auth.py:156-158).
> app_callback() makes the top-level --output flag authoritative, but auth_whoami() unconditionally does out.set_mode(OutputMode.json) whenever --format json is present. So hf --output human auth whoami --format json will still render JSON instead of honoring the explicit global mode.

yeah not quite sure yet if the global `--output` should win over per command `--format`. wdyt @Wauplin ?

Wauplin:
> > Medium — The advertised mode precedence is violated for the first migrated command (src/huggingface_hub/cli/hf.py:73-77, src/huggingface_hub/cli/auth.py:156-158).
app_callback() makes the top-level --output flag authoritative, but auth_whoami() unconditionally does out.set_mode(OutputMode.json) whenever --format json is present. So hf --output human auth whoami --format json will still render JSON instead of honoring the explicit global mode.
> 
> yeah not quite sure yet if the global --output should win over per command --format. wdyt @Wauplin ?

I think we shouldn't define a new `--output` option as it is quite misleading when `--format` is accepted. IMO the goal of https://github.com/huggingface/huggingface_hub/issues/3979 should be to simplify what is expected in the CLI output so I'd be in favor of updating the existing `OutputFormat` like this:

```
class OutputFormat(str, Enum):
    auto = "auto"  # default

    agent = "agent"
    human = "human"
    json = "json"
    quiet = "quiet"

FormatOpt = Annotated[OutputFormat, ...]
```

This would be a breaking change for `--output table` (which is now implicit in agent/human modes when using `out.table`) but _I think_ it's fine to introduce it.

And in addition to `--format auto|agent|human|json|quiet` we should also  keep the `--json` and `--quiet` alias (for human/agent/auto I don't think it makes sense).

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+    agent = "agent"
+    json = "json"
+
+
+class Output:
+    """Output sink for the `hf` CLI.
+
+    Mode resolution (first match wins):
+    1. Explicit `set_mode()` call (from a `--output` flag).
+    2. `HF_OUTPUT` environment variable.
```
Wauplin: I'm really not convinced by `HF_OUTPUT` env variable if we already have auto mode + --json and --quiet aliases. Having a way to force human or agent modes is nice for local debugging but that's all. So providing 2 different ways to do it it not worth it IMO

hanouticelina:
@Wauplin yes okay looks good to me! this first draft was mainly to match exactly the issue specs and then refine based on feedbacks

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+        cell = cell[: max_len - 3] + "..."
+    return cell
+
+
+class OutputMode(str, Enum):
+    """Output mode for the `hf` CLI."""
+
+    human = "human"
+    agent = "agent"
+    json = "json"
```
Wauplin: IMO this should be the same enum as the existing `OutputFormat`

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+_ANSI_RE = re.compile(r"\033\[[0-9;]*m")
+
+_MAX_CELL_LENGTH = 35
+
+
+def _strip_ansi(text: str) -> str:
+    return _ANSI_RE.sub("", text)
+
+
+def _to_header(name: str) -> str:
```
Wauplin: (nit) I'd move low level private utils to the end of the module for readability

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+    if isinstance(value, list):
+        return ", ".join(_format_value(v) for v in value)
+    elif isinstance(value, dict):
+        if "name" in value:
+            return str(value["name"])
+        return json.dumps(value)
+    return str(value)
+
+
+def _format_cell(value: Any, max_len: int = _MAX_CELL_LENGTH) -> str:
```
Wauplin:
```suggestion
def _format_table_cell(value: Any, max_len: int = _MAX_CELL_LENGTH) -> str:
```

(nit)

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+    return _ANSI_RE.sub("", text)
+
+
+def _to_header(name: str) -> str:
+    """Convert a camelCase or PascalCase string to SCREAMING_SNAKE_CASE."""
+    s = re.sub(r"([a-z])([A-Z])", r"\1_\2", name)
+    return s.upper()
+
+
+def _format_value(value: Any) -> str:
```
Wauplin:
```suggestion
def _format_table_value(value: Any) -> str:
```

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @property
+    def mode(self) -> OutputMode:
+        if self._mode is None:
+            self._mode = self._resolve_mode()
+        return self._mode
+
+    def set_mode(self, mode: Optional[OutputMode] = None) -> None:
+        """Set the output mode.  Pass `None` to reset to auto-detection."""
+        self._mode = mode
```
Wauplin: Let's not accept `out.set_mode(None)` (I don't see a good use case for it)

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._mode: Optional[OutputMode] = None
+
+    @property
+    def mode(self) -> OutputMode:
+        if self._mode is None:
+            self._mode = self._resolve_mode()
+        return self._mode
```
Wauplin: I'd simplify this to a `self.mode` attribute resolved once at init time in `__init__` and then eventually updated by `set_mode` but that's all

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def set_mode(self, mode: Optional[OutputMode] = None) -> None:
+        """Set the output mode.  Pass `None` to reset to auto-detection."""
+        self._mode = mode
+
+    @staticmethod
+    def _resolve_mode() -> OutputMode:
```
Wauplin: (nit) feels like it should be in `__init__`

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def hint(self, message: str) -> None:
+        """Print a helpful hint to stderr."""
+        if self.mode == OutputFormat.human:
+            print(ANSI.gray(f"  {message}"), file=sys.stderr)
+        else:
+            print(f"Hint: {message}", file=sys.stderr)
+
+    @staticmethod
+    def _format_agent_cell(value: Any) -> str:
```
Wauplin: let's make it a separate helper like the other methods (I'm not a fan of staticmethod)

src/huggingface_hub/cli/auth.py
```diff
@@ -150,24 +150,22 @@ def auth_list_cmd() -> None:
...
-        print(ANSI.bold("user: "), info["name"])
-        orgs = [org["name"] for org in info["orgs"]]
-        if orgs:
-            print(ANSI.bold("orgs: "), ",".join(orgs))
 
+    if out.mode == OutputFormat.json:
+        out.dict(info)
+    else:
+        orgs = ",".join(org["name"] for org in info["orgs"])
+        out.result("Logged in", user=info["name"], orgs=orgs)
```
Wauplin: As I see it the goal of `out` is precisely to avoid logic like `if out.mode == ...` in the CLI commands.

Something like this should be more ideal (not exactly that but we should aim for a syntax where `out` takes care of all the if/else)
```suggestion
    out.dict(info)
    out.result("Logged in", user=info["name"], orgs=",".join(org["name"] for org in info["orgs"])
```

src/huggingface_hub/cli/_output.py
hanouticelina: should be addressed in [5915d0e](https://github.com/huggingface/huggingface_hub/pull/4005/commits/5915d0eb7aaf6442a9521145def92948769a2502)

src/huggingface_hub/cli/auth.py
hanouticelina: yeah we can do that by making `dict()` and `result()` somehow complimentary, `dict()` only prints in json mode, `result()` only prints in human/agent mode

hanouticelina: or actually let's drop `.dict()`?

src/huggingface_hub/cli/_output.py
hanouticelina: agree, addressed in [831bb60](https://github.com/huggingface/huggingface_hub/pull/4005/commits/831bb60a54c9c66aa39c75cfd5d2b48586a296df)

src/huggingface_hub/cli/_output.py
hanouticelina: addressed [831bb60](https://github.com/huggingface/huggingface_hub/pull/4005/commits/831bb60a54c9c66aa39c75cfd5d2b48586a296df)

src/huggingface_hub/cli/_output.py
hanouticelina: removed the lazy property in [831bb60](https://github.com/huggingface/huggingface_hub/pull/4005/commits/831bb60a54c9c66aa39c75cfd5d2b48586a296df)

src/huggingface_hub/cli/_output.py
hanouticelina: what I had in mind is that the `None` value (no explicit `format` value) would resolve to auto and we won't need an explicit "auto" mode.

hanouticelina: no strong opinion, so I addressed that in [831bb60](https://github.com/huggingface/huggingface_hub/pull/4005/commits/831bb60a54c9c66aa39c75cfd5d2b48586a296df)

src/huggingface_hub/cli/_output.py
hanouticelina: removed in [831bb60](https://github.com/huggingface/huggingface_hub/pull/4005/commits/831bb60a54c9c66aa39c75cfd5d2b48586a296df)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -548,17 +570,22 @@ def env_map_to_key_value_list(env_map: dict[str, Optional[str]]) -> Optional[lis
...
-    """Output format for CLI list commands."""
+    """Output format for CLI commands."""
 
-    table = "table"
+    agent = "agent"
+    auto = "auto"
+    human = "human"
     json = "json"
+    quiet = "quiet"
+    table = "table"  # kept temporarily for backward compatibility
```
Wauplin: to handle backward compatibility I think we should define a `LegacyOutputFormat` like this:


```suggestion
class LegacyOutputFormat(str, Enum):
    """Legacy output format for CLI commands.

    TODO: remove it once all commands are migrated
    """

    json = "json"
    table = "table"

class OutputFormat(str, Enum):
    """Output format for CLI commands."""

    agent = "agent"
    auto = "auto"
    human = "human"
json = "json"
    quiet = "quiet"
```


So that the signature of commands that we did not touch won't change

docs/source/en/package_reference/cli.md
```diff
@@ -373,7 +373,7 @@ $ hf buckets list [OPTIONS] [ARGUMENT]
 * `-h, --human-readable`: Show sizes in human readable format.
 * `--tree`: List files in tree format (only for listing files).
 * `-R, --recursive`: List files recursively (only for listing files).
-* `--format [table|json]`: Output format (table or json).  [default: table]
+* `--format [agent|auto|human|json|quiet|table]`: Output format.  [default: table]
```
Wauplin: Let's not update signature of commands we did not migrate yet (otherwise `hf buckets sync ... --format agent` is an undefined behavior

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -548,17 +570,22 @@ def env_map_to_key_value_list(env_map: dict[str, Optional[str]]) -> Optional[lis
...
 
 
 FormatOpt = Annotated[
```
Wauplin, line 594: And a `LegacyOutputFormat` (requires to update all commands but at least signatures won't change)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -633,7 +660,7 @@ def print_as_table(
 def print_list_output(
     items: Sequence[dict[str, Any]],
     format: OutputFormat,
-    quiet: bool,
+    quiet: bool = False,
```
Wauplin: Is this needed? I feel that the `print_list_output` will be entirely removed once `out.table` will be used everywhere so we don't need to change it

src/huggingface_hub/cli/auth.py
```diff
@@ -150,24 +150,19 @@ def auth_list_cmd() -> None:
 
 @auth_cli.command("whoami", examples=["hf auth whoami", "hf auth whoami --format json"])
 def auth_whoami(
-    format: FormatOpt = OutputFormat.table,
+    format: FormatOpt = OutputFormat.auto,
 ) -> None:
     """Find out which huggingface.co account you are logged in as."""
+    out.set_mode(format)
```
Wauplin, line 156: should this call be factorized in a app_callback since all commands will set the format mode as first step?

src/huggingface_hub/cli/auth.py
```diff
@@ -150,24 +150,19 @@ def auth_list_cmd() -> None:
...
-        if orgs:
-            print(ANSI.bold("orgs: "), ",".join(orgs))
-
-        if ENDPOINT != "https://huggingface.co":
-            print(f"Authenticated through private endpoint: {ENDPOINT}")
+    orgs = ",".join(org["name"] for org in info["orgs"])
+    out.dict(info)
+    out.result("Logged in", user=info["name"], orgs=orgs)
+    if ENDPOINT != "https://huggingface.co":
+        out.hint(f"Authenticated through private endpoint: {ENDPOINT}")
```
Wauplin:
```suggestion
    out.result("Logged in", user=info["name"], orgs=orgs, endpoint=ENDPOINT)
```

just a suggestion ^

tests/test_cli.py
```diff
@@ -1644,15 +1644,14 @@ def test_whoami_json_shorthand(self, runner: CliRunner) -> None:
     def test_whoami_not_logged_in(self, runner: CliRunner) -> None:
         with patch("huggingface_hub.cli.auth.get_token", return_value=None):
             result = runner.invoke(app, ["auth", "whoami"])
-        assert result.exit_code == 0
-        assert "Not logged in" in result.stdout
+        assert result.exit_code == 1
+        assert "Not logged in" in result.output
```
Wauplin, line 1650: (nit) `result.output` is a mix of `result.stdout` + `result.stderr`

(in the future we should test the `out.error` call directly using patch, but that's for another PR)

```suggestion
        assert "Not logged in" in result.stderr
```

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    def __init__(self) -> None:
+        self.mode: OutputFormat = OutputFormat.agent if is_agent() else OutputFormat.human
+
+    def set_mode(self, mode: OutputFormat) -> None:
+        """Override the output mode (called by commands that receive `--format`)."""
+        if mode in (OutputFormat.auto, OutputFormat.table):
+            return
+        self.mode = mode
```
Wauplin, line 44: > what I had in mind is that the None value (no explicit format value) would resolve to auto and we won't need an explicit "auto" mode.

Ok ok, I get your point now. I'd be useful to have the TS equivalent `omit` for such a case. I think what we can do is to have `OutputFormat` and `OutputFormatNoAuto` defined separately. So that `self.mode` is explicitly a `OutputFormatNoAuto` while `set_mode` does accept `OutputFormat` as input. WDYT?

Wauplin, line 44:
```suggestion
class Output:
    """Output sink for the `hf` CLI.

    Mode is resolved once at init time based on `is_agent()` auto-detection
    and can be overridden per-command via `set_mode()`.
    """
    mode: OutputFormat

    def __init__(self) -> None:
        self.set_mode()

    def set_mode(self, mode: OutputFormat = OutputFormat.auto) -> None:
        """Override the output mode (called by commands that receive `--format`)."""
        if mode in (OutputFormat.auto, OutputFormat.table):
            mode = OutputFormat.agent if is_agent() else OutputFormat.human
        self.mode = mode
```

Wauplin: I'm a bit worried by the

```py
        if mode in (OutputFormat.auto, OutputFormat.table):
            return
```

syntax. In practice it shouldn't be a problem because the mode is set only once in the CLI but in theory doing

```py
out.set_mode("quiet")
out.set_mode("auto")
```

should reset the output mode to `human`/`agent`, not keeping the `quiet` mode active

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def text(self, human: str, agent: str | None = None) -> None:
+        """Print a free-form text message to stdout."""
+        if self.mode in (OutputFormat.json, OutputFormat.quiet):
+            return
+        if self.mode == OutputFormat.agent:
+            print(agent if agent is not None else _strip_ansi(human))
+        else:
+            print(human)
```
Wauplin:
```suggestion
    match self.mode:
        case OutputFormat.agent:
            print(agent if agent is not None else _strip_ansi(human))
        case OutputFormat.human:
            print(human)
```

(just me having fun with a new toy 😄)

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+        """Print tabular data to stdout.
+
+        Args:
+            headers: Column names.
+            rows: List of rows, each a list of raw values.
+            alignments: Optional mapping of header name to "left" or "right". Defaults to "left".
+        """
+        if not rows:
+            match self.mode:
+                case OutputFormat.human:
```
Wauplin:
```suggestion
                case OutputFormat.agent | OutputFormat.human:
```

I feel that we should print a `No results found.` also for agent mode

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    if isinstance(value, list):
+        return ", ".join(_format_table_value(v) for v in value)
+    elif isinstance(value, dict):
+        if "name" in value:
+            return str(value["name"])
+        return json.dumps(value)
+    return str(value)
+
+
+def _format_table_cell(value: Any, max_len: int = _MAX_CELL_LENGTH) -> str:
```
Wauplin:
```suggestion
def _format_table_cell_human(value: Any, max_len: int = _MAX_CELL_LENGTH) -> str:
```

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    """Format a value + truncate it for table display."""
+    cell = _format_table_value(value)
+    if len(cell) > max_len:
+        cell = cell[: max_len - 3] + "..."
+    return cell
+
+
+def _format_agent_cell(value: Any) -> str:
```
Wauplin:
```suggestion
def _format_table_cell_agent(value: Any) -> str:
```

src/huggingface_hub/cli/_output.py
Wauplin: (nit) for consistency

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    """Format a cell value for agent TSV output (lowercase bools, ISO timestamps)."""
+    if isinstance(value, bool):
+        return "true" if value else "false"
```
Wauplin:
```suggestion
```

I'd remove this for simplicity (printing `True`/`False` instead of `true`/`false` feels transparent for the agent)

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+        return "✔" if value else ""
+    if isinstance(value, datetime.datetime):
+        return value.strftime("%Y-%m-%d")
+    if isinstance(value, str) and re.match(r"^\d{4}-\d{2}-\d{2}T", value):
+        return value[:10]
+    if isinstance(value, list):
+        return ", ".join(_format_table_value(v) for v in value)
+    elif isinstance(value, dict):
+        if "name" in value:
+            return str(value["name"])
```
Wauplin: let's keep a comment on why we did that in the past 😬

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    return cell
+
+
+def _format_agent_cell(value: Any) -> str:
+    """Format a cell value for agent TSV output (lowercase bools, ISO timestamps)."""
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, datetime.datetime):
+        return value.isoformat()
+    return str(value)
```
Wauplin:
```suggestion
    return str(value).replace("\t", " ")
```

since agent tables are printed as TSV, I think it's safer to remove the `\t` from the values themselves

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+                for row in rows:
+                    print("\t".join(_format_agent_cell(v) for v in row))
+            case OutputFormat.json:
+                items = [dict(zip(headers, row)) for row in rows]
+                print(json.dumps(items, default=str))
+
+    def dict(self, data: dict[str, Any]) -> None:
+        """Print structured data as JSON to stdout. Only prints in json mode."""
+        if self.mode == OutputFormat.json:
+            print(json.dumps(data, default=str))
```
Wauplin: the specs were this:

Image


(we can re-discuss it but if we change the logic let's be explicit)

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        if not rows:
+            match self.mode:
+                case OutputFormat.human:
+                    print("No results found.")
+                case OutputFormat.json:
+                    print("[]")
+            return
+
+        match self.mode:
```
Wauplin, line 76: in quiet mode print either the `id_key` (if provided) or first column? same as current behavior in `print_list_output`

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+                items = [dict(zip(headers, row)) for row in rows]
+                print(json.dumps(items, default=str))
+
+    def dict(self, data: dict[str, Any]) -> None:
+        """Print structured data as JSON to stdout. Only prints in json mode."""
+        if self.mode == OutputFormat.json:
+            print(json.dumps(data, default=str))
+
+    def result(self, message: str, **data: Any) -> None:
+        """Print a success summary to stdout. No-op in json mode (use ``dict()`` instead)."""
```
Wauplin: Goal was this

Image

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+            case OutputFormat.agent:
+                parts = [f"{k}={v}" for k, v in data.items() if v is not None]
+                print(" ".join(parts) if parts else message)
+
+    def warning(self, message: str) -> None:
+        """Print a non-fatal warning to stderr."""
+        if self.mode == OutputFormat.human:
+            print(ANSI.yellow(f"  Warning: {message}"), file=sys.stderr)
+        else:
+            print(f"Warning: {message}", file=sys.stderr)
```
Wauplin, line 122: let's output warnings in quiet/json mode as well? (in stderr)

Image

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,194 @@
...
+    def confirm(self, message: str, *, yes: bool = False) -> None:
+        """Interactive confirmation. Raises if in agent mode without ``--yes``."""
+        if yes:
+            return
+        if self.mode == OutputFormat.human and sys.stdin.isatty():
+            choice = input(f"{message} [Y/n] ")
+            if choice.lower() not in ("", "y", "yes"):
+                raise typer.Abort()
+        else:
+            raise CLIError(f"{message} Use --yes to skip confirmation.")
```
Wauplin: Maybe let's implement `confirm` only once we actually need it. It's quite a tricky one as we might not always want to have `Use --yes to skip confirmation` (cf DMs about preventing from skipping confirmation in https://github.com/huggingface/huggingface_hub/pull/4007).


I feel that we will need an additional flag `default: bool`. We usually do `[Y/n]` but in some cases we would prefer `[y/N]`. (and I'd rather rename `yes` by `skip: bool`)

docs/source/en/package_reference/cli.md
hanouticelina: I thought it was okay to do that as long as these changes won't be merged into `main` unless all the commands have been migrated (this branch is aimed to be merged into `cli-dual-mode-output`)

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: not needed, indeed. and yes `print_list_output` will be removed

src/huggingface_hub/cli/_output.py
hanouticelina: yes, this change was related to https://github.com/huggingface/huggingface_hub/pull/4005#discussion_r3010459201.

not sure yet how `dict()` should behave actually and I think maybe `result()` can be sufficient?

src/huggingface_hub/cli/_output.py
hanouticelina: (cf my comment [here](https://github.com/huggingface/huggingface_hub/pull/4005#discussion_r3015115325))

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: the PR targets `cli-dual-mode-output` not `main` so I thought non migrated commands will never be in `main` with the new enum. by the time this reaches main, they'll be migrated. but agree that the docs noise is still a downside here

src/huggingface_hub/cli/_output.py
hanouticelina: yes agree! addressed in [7e4bfc5](https://github.com/huggingface/huggingface_hub/pull/4005/commits/7e4bfc5000521bdee43c7ca2b868c083afac7fec)

src/huggingface_hub/cli/_output.py
hanouticelina: oops 🙈 reverted!

src/huggingface_hub/cli/_output.py
hanouticelina: yes! addressed ✅

src/huggingface_hub/cli/auth.py
hanouticelina: `--format` is per-command, so `app_callback` doesn't have access to it in typer/click, according to https://click.palletsprojects.com/en/stable/commands/#group-invocation-without-command (TL;DR; callback runs first, then subcommand args are parsed)

src/huggingface_hub/cli/_cli_utils.py
Wauplin: Oh I didn't see that before :o I thought https://github.com/huggingface/huggingface_hub/pull/3991 was already merged on main actually. I feel that we can (and should) merge iteratively on main to avoid having to finish the complete rewrite to start benefiting from it

(and to prevent having a huge branch to merge at the end)

hanouticelina: yes you're probably right and we will keep the history of each commit. In that case, it makes sense to have a legacy format 👍

src/huggingface_hub/cli/_output.py
Wauplin: let's drop `dict` for now yes :+1:

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: addressed in [665d9aa](https://github.com/huggingface/huggingface_hub/pull/4005/commits/665d9aa7a754499b07668d7156a300d19e6e98f7). I renamed the new enum to `AutoOutputFormat` to avoid touching non migrated files in this PR. 
(Side note: I think the enum should be renamed to `OutputMode`, I find it more semantically correct now that it's more about "rendering" mode)

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: cf my comment here: https://github.com/huggingface/huggingface_hub/pull/4005#discussion_r3016077619

src/huggingface_hub/cli/_output.py
hanouticelina: yes, it already prints in all modes. the else covers agent, json and quiet. or maybe I misunderstood your comment 😄

src/huggingface_hub/cli/_cli_utils.py
Wauplin: (nit) preference for `OutputFormatWithAuto` :smiley: 

> (Side note: I think the enum should be renamed to OutputMode, I find it more semantically correct now that it's more about "rendering" mode)

Definitely agree!

src/huggingface_hub/cli/_output.py
hanouticelina: I don't know (I might have changed my opinion since yesterday 🙈), but the current `set_mode()` resolves `auto` before assigning, so `self.mode` is never `auto` at runtime. A `OutputFormatNoAuto` enum would add complexity here no ?

src/huggingface_hub/cli/_output.py
hanouticelina: yes, good catch! addressed in [0ac669f](https://github.com/huggingface/huggingface_hub/pull/4005/commits/0ac669f74e296a0a6705e628c4134524032fbd8a)

src/huggingface_hub/cli/_output.py
hanouticelina: It turns out keeping `out.dict()` as mentioned in the specs can be helpful here, sorry for the back and forth on this, it's a bit hard to picture some of the cases and to be honest, I don't like how things would be structured if we incorporate a full json mode in `result()`. sorry again about that. 
so now what we have is:

`out.dict(data)`
dumps structured JSON in all modes. this is needed for info commands (`hf models info`, `hf datasets info`, etc.) where the full structured dump is the expected output regardless of mode, and I don't think we should break this.

`out.result(message, **data)` 
is for action summaries  human gets `✓ + key-value` lines, agent gets `key=val`, JSON gets `json.dumps(data)`.

src/huggingface_hub/cli/_output.py
hanouticelina: removed in [051d1bc](https://github.com/huggingface/huggingface_hub/pull/4005/commits/051d1bcb00240d18d12c62f1d0b729aa538b821b)

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: yes agree, renamed to `OutputFormatWithAuto` in [051d1bc](https://github.com/huggingface/huggingface_hub/pull/4005/commits/051d1bcb00240d18d12c62f1d0b729aa538b821b)

hanouticelina:
@Wauplin Thanks for the review and sorry for the back and forth on the `result()`/`dict()` specs! I think I addressed all your comments now.

EDIT: I added comments for each case to explain the behavior in [3180f55](https://github.com/huggingface/huggingface_hub/pull/4005/commits/3180f556bca18faca88865b6c79aec27f0f3835f) and the `quiet` mode for `result()` where it prints the first value from data

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,184 @@
...
+        return json.dumps(value)
+    return str(value)
+
+
+def _format_table_cell_human(value: Any, max_len: int = _MAX_CELL_LENGTH) -> str:
+    """Format a value + truncate it for table display."""
+    cell = _format_table_value_human(value)
+    if len(cell) > max_len:
+        cell = cell[: max_len - 3] + "..."
+    return cell
```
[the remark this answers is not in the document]
Wauplin: hopefully we will get rid of the helpers in `_cli_utils.py` pretty soon

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class Output:
+    """Output sink for the `hf` CLI.
+
+    Mode is resolved once at init time based on `is_agent()` auto-detection
+    and can be overridden per-command via `set_mode()`.
+    """
+
+    mode: OutputFormatWithAuto
```
Wauplin, line 35:
```suggestion
    mode: OutputFormat
```

isn't it? or doesn't work because not the same enum? (anyway, should be remove rather soon I suppose?)

hanouticelina: it doesn't work I think no? `OutputFormat` has `table`/`json` only. but yes it should be cleaned up soon

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,189 @@
...
+                parts = [ANSI.green(f"✓ {message}")]
+                for k, v in data.items():
+                    if v is not None:
+                        parts.append(f"  {k}: {v}")
+                print("\n".join(parts))
+            case OutputFormatWithAuto.agent:  # key=val pairs, space-separated
+                parts = [f"{k}={v}" for k, v in data.items() if v is not None]
+                print(" ".join(parts) if parts else message)
+            case OutputFormatWithAuto.json:  # json.dumps(data), message ignored
+                print(json.dumps(data, default=str) if data else "")
```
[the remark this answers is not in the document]
hanouticelina: it's okay for JSON mode to dump the data as-is

src/huggingface_hub/cli/_output.py
```diff
@@ -0,0 +1,189 @@
...
+def _to_header(name: str) -> str:
+    """Convert a camelCase or PascalCase string to SCREAMING_SNAKE_CASE."""
+    s = re.sub(r"([a-z])([A-Z])", r"\1_\2", name)
+    return s.upper()
+
+
+def _format_table_value_human(value: Any) -> str:
+    """Convert a value to string for terminal display."""
+    if not value:
+        return ""
```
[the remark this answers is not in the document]
hanouticelina: it's pre-existing, not related to this PR

PR #4007, opened 2026-03-30
[CLI] auto-install official HF CLI extensions on first invocation

hanouticelina:
with this PR, running any `huggingface/hf-*` extension now auto-installs it if not already present, then runs the command. Only extensions from the `huggingface` GitHub org are auto-installed.

```
$ hf agents --help
Installing extension 'agents' from huggingface/hf-agents...
hf-agents 0.2.1 — local coding agents powered by llmfit + llama.cpp
...
```

Subsequent runs use the locally installed extension directly.

---

> [!NOTE]
> **Medium Risk**
> Introduces network-driven, on-demand installation/execution of external extension code (albeit gated to the `huggingface` org and a confirmation prompt), which can affect CLI behavior and reliability when offline or when GitHub requests fail.
> 
> **Overview**
> Adds *prompted auto-installation* for unknown top-level `hf <cmd>` invocations: if `<cmd>` isn’t a built-in/aliased command and no local extension is found, the CLI now checks for `huggingface/hf-<cmd>` on GitHub, asks for confirmation, installs it, then executes the newly installed extension.
> 
> Refactors GitHub-based installation into a shared `_install_extension_from_github` helper (used by both explicit `hf extensions install` and the new auto-install path), and updates install messaging to rely on the manifest’s extension type.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ef9b2d1e6854bb09bd136ba62d05b90ab7b72d22. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/extensions.py
+++ b/src/huggingface_hub/cli/extensions.py
@@ -19,7 +19,6 @@ import os
 import re
 import shutil
 import subprocess
-import sys
 import venv
 from dataclasses import asdict, dataclass
 from datetime import datetime, timezone
@@ -268,7 +267,7 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
     if not short_name:
         return None
 
-    executable_path: Optional[Path] = None
+    executable_path: Path | None = None
     try:
         executable_path = _resolve_installed_executable_path(short_name)
     except Exception:
@@ -280,7 +279,7 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
     return _execute_extension_binary(executable_path=executable_path, args=list(args[1:]))
 
 
-def _auto_install_official_extension(short_name: str) -> Optional[Path]:
+def _auto_install_official_extension(short_name: str) -> Path | None:
     """Try to auto-install huggingface/hf-<name>. Returns executable path or None."""
     owner, repo_name = DEFAULT_EXTENSION_OWNER, f"hf-{short_name}"
     try:
@@ -301,7 +300,8 @@ def _auto_install_official_extension(short_name: str) -> Optional[Path]:
         branch = response.json()["default_branch"]
     except Exception:
         return None
-    print(f"Installing extension '{short_name}' from {owner}/{repo_name}...", file=sys.stderr)
+    if not typer.confirm(f"'{short_name}' is an official Hugging Face extension ({owner}/{repo_name}). Install it?"):
+        return None
     try:
         manifest = _install_extension_from_github(
             owner=owner, repo_name=repo_name, short_name=short_name, extension_dir=extension_dir, branch=branch
@@ -319,7 +319,7 @@ def _install_extension_from_github(
     short_name: str,
     extension_dir: Path,
     branch: str,
-    description: Optional[str] = None,
+    description: str | None = None,
 ) -> ExtensionManifest:
     """Fetch, install (binary or Python), and save manifest for a GitHub extension."""
     try:


src/huggingface_hub/cli/extensions.py
```diff
@@ -284,24 +260,87 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
     command_name = args[0]
     if command_name.startswith("-"):
         return None
-    if command_name in known_commands:
+    all_known = {a.strip() for cmd in known_commands for a in cmd.split("|")}
```
hanouticelina, line 262: prviously `"extensions"` didn't match `"extensions | ext"`, which was harmless before (returned `None`) but would now trigger unnecessary auto install attempts

Wauplin:
As mentioned on Slack I do like this feature but I'd prefer to have it behind a `Do you confirm? [y/N]` prompt for security / to avoid installing stuff without the user noticing/officially requesting it.

Wauplin:
An orthogonal feature (but related) would be to check for auto-updates once a day when an extension is being used and prompt for auto-update
(but same as this one, not sure if it's high prio)

julien-c:
Agree with the confirmation step personally

hanouticelina:
yes agree as well, but it would be better to have a way to bypass the  `[y/N]` prompt before installing no? A `--yes` flag doesn't really make sense on `hf ext exec` since that command already requires the extension to be installed. I'm not sure about the best way to do that.

we we can also do an exception here and prompt by default without a way to bypass

Wauplin:
I'm not sure a way to bypass the confirmation step is needed. If someone really wants to bypass it, they just have to run the correct installation command before running the extension (which is the default "normal behavior"). Being able to auto-install the extension on first invocation is only a nice-to-have

src/huggingface_hub/cli/extensions.py
```diff
@@ -284,24 +259,88 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
...
+            follow_redirects=True,
+            timeout=_EXTENSIONS_DOWNLOAD_TIMEOUT,
+        )
+        if response.status_code == 404:
+            return None
+        response.raise_for_status()
+        branch = response.json()["default_branch"]
+    except Exception:
+        return None
+    if not typer.confirm(f"'{short_name}' is an official Hugging Face extension ({owner}/{repo_name}). Install it?"):
```
hanouticelina, line 303: note that this will be replaced by out.confirm (the `out` Singleton defined in https://github.com/huggingface/huggingface_hub/issues/3979) and it will handle human/TTY and agent/non-TTY conditions

src/huggingface_hub/cli/extensions.py
```diff
@@ -284,24 +259,88 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
...
+        return None
```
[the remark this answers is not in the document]
hanouticelina: doesn't apply here, `typer.confirm` with `default=False` (our case) doesn't raise `Abort` on non-interactive stdin, it returns `False`

src/huggingface_hub/cli/extensions.py
```diff
@@ -284,24 +259,88 @@ def dispatch_unknown_top_level_extension(args: list[str], known_commands: set[st
...
+    """Try to auto-install huggingface/hf-<name>. Returns executable path or None."""
+    owner, repo_name = DEFAULT_EXTENSION_OWNER, f"hf-{short_name}"
+    try:
+        extension_dir = _get_extension_dir(short_name)
+    except Exception:
+        return None
+    if extension_dir.exists():
+        return None
+    try:
+        response = get_session().get(
```
Wauplin, line 292: I realized that this check means that any `hf <missing_or_typo_command>` will now have a small delay before crashing. I don't see a proper way to avoid this except to hardcode a list of official extensions in `huggingface_hub` directly (which defeats the purpose of remote extensions...)

So for now let's merge but keep in mind this limitation

Wauplin (approved):
Thank you!

src/huggingface_hub/cli/extensions.py
hanouticelina: yes I noticed that as well when I tried `hf agent` (instead of `agents`), no idea either on how to improve that properly 😕

PR #4025, opened 2026-04-02
Add bucket+mount transport for Jobs script upload

davanstrien:
## Use bucket transport for Jobs script upload

When `hf jobs uv run` is given local files (scripts, configs, etc.), we now upload them to a `{namespace}/jobs-artifacts` bucket and mount it into the job container at `/data` instead of base64-encoding them into an environment variable.

This replaces the old `bash -c` + `xargs` + `base64 -d` pipeline which was fragile and required manual shell quoting. The bucket approach is simpler, easier to debug, and importantly lets jobs write output artifacts back to `/data/` since the mount is read-write.

### How it works

- Local files are uploaded to `{namespace}/jobs-artifacts` under a `{YYYYMMDDTHHMMSS}-{6 hex digits}` subfolder (makes it easy to find artifacts for a specific job by timestamp)
- A `Volume` scoped to that subfolder is mounted at `/data`, so the job container sees its files directly at the mount root
- The `uv run` command references `/data/<filename>` — no more shell wrappers
- Failures propagate directly (no silent fallback)
- If a user already has a volume mounted at `/data`, we raise a clear `ValueError`

### What changed vs the original base64 approach

The base64 transport path is fully removed — there's no opt-in flag, no fallback. Bucket transport is the only mode now. This was the simplest thing to do since base64 was experimental/hacky and maintaining two transport modes wasn't worth it.

### Constants

Two new constants in `constants.py`:
- `HF_JOBS_ARTIFACTS_BUCKET_NAME = "jobs-artifacts"`
- `HF_JOBS_ARTIFACTS_MOUNT_PATH = "/data"`

### Files changed

- `src/huggingface_hub/constants.py` — new bucket/mount constants
- `src/huggingface_hub/hf_api.py` — `_create_uv_command_env_and_secrets` rewritten to use bucket transport, new `_upload_scripts_to_bucket` helper
- `tests/test_cli.py` — new `TestBucketTransport` class replacing the old shell-quoting tests (happy path, failure propagation, mount collision, multi-file upload)

---

> [!NOTE]
> **Medium Risk**
> Changes the transport mechanism for local file execution in Jobs from env-var base64 to bucket uploads with a reserved mount point, which may affect existing workflows that relied on custom `/data` mounts or the previous command wrapper behavior.
> 
> **Overview**
> When `run_uv_job`/`create_scheduled_uv_job` are given local scripts/config files, the PR replaces the previous base64-in-env + `bash -c` reconstruction pipeline with **bucket-based upload + mounted volume** transport.
> 
> Local files are uploaded to a private `{namespace}/jobs-artifacts` bucket under a per-job `{timestamp}-{random}` subfolder, mounted read-write at the new reserved path `HF_JOBS_ARTIFACTS_MOUNT_PATH` (`/data`), and the generated `uv run` command is rewritten to reference `/data/<filename>`.
> 
> Adds `HF_JOBS_ARTIFACTS_BUCKET_NAME`/`HF_JOBS_ARTIFACTS_MOUNT_PATH` constants, introduces `_upload_scripts_to_bucket`, threads `volumes` through `_create_uv_command_env_and_secrets` (failing fast if `/data` is already used), and updates tests to cover upload behavior, mount collisions, multi-file upload, and failure propagation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3420f8840356c40ed453da7282264f6eb835cb11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davanstrien:
cc @Wauplin @lhoestq

src/huggingface_hub/hf_api.py
```diff
@@ -11957,50 +11969,125 @@ def _create_uv_command_env_and_secrets(
...
+
+        Creates a bucket ``{namespace}/jobs-artifacts`` (if it doesn't exist) and uploads
+        each script to ``_scripts/{uuid}/{remote_name}`` inside it. Returns a :class:`Volume`
+        that mounts the bucket at ``/artifacts`` so the job can access the scripts directly.
+        """
+        import uuid
+
+        bucket_id = f"{namespace}/{self._HF_JOBS_ARTIFACTS_BUCKET_NAME}"
+        subfolder_id = str(uuid.uuid4())
+        scripts_prefix = f"_scripts/{subfolder_id}"
```
lhoestq: not need for the underscore imo

```suggestion
        scripts_prefix = f"scripts/{subfolder_id}"
```

btw the files are not necessarily scripts, they could also be config files, e.g.

```bash
hf jobs uv run accelerate launch train.py --config_file training_args.json
```

but since the general use case if for scripts I think I'm fine with the name

src/huggingface_hub/hf_api.py
```diff
@@ -11913,6 +11920,10 @@ def create_scheduled_uv_job(
             token=token,
         )
 
+    # Bucket transport constants for Jobs
+    _HF_JOBS_ARTIFACTS_MOUNT_PATH = "/artifacts"
+    _HF_JOBS_ARTIFACTS_BUCKET_NAME = "jobs-artifacts"
```
lhoestq: good names !

src/huggingface_hub/constants.py
```diff
@@ -240,6 +240,9 @@ def list_files(repo_id: str):
 
 HF_XET_HIGH_PERFORMANCE: bool = _is_true(os.environ.get("HF_XET_HIGH_PERFORMANCE"))
 
+# Opt-in to bucket-based script transport for Jobs (experimental)
+HF_JOBS_USE_BUCKET_TRANSPORT: bool = _is_true(os.environ.get("HF_JOBS_USE_BUCKET_TRANSPORT"))
```
lhoestq: imo we can already set it to True by default

lhoestq (commented):
Awesome :)

src/huggingface_hub/constants.py
davanstrien: probably makes sense indeed! I'll wait to see what @Wauplin thinks

src/huggingface_hub/hf_api.py
```diff
@@ -11957,50 +11969,125 @@ def _create_uv_command_env_and_secrets(
...
+            (Path(local_path), f"{scripts_prefix}/{remote_name}")
+            for remote_name, local_path in remote_to_local_file_names.items()
+        ]
+        self.batch_bucket_files(bucket_id=bucket_id, add=add_ops, token=token)
+
+        volume = Volume(
+            type="bucket",
+            source=bucket_id,
+            mount_path=self._HF_JOBS_ARTIFACTS_MOUNT_PATH,
+        )
```
Wauplin, line 12232: I would mount the volume with `prefix=scripts_prefix` here so that the Jobs sees only the artifacts folder created for that Job

And doing so, you don't need to pass `scripts_prefix` as return type

src/huggingface_hub/hf_api.py
```diff
@@ -11957,50 +11969,125 @@ def _create_uv_command_env_and_secrets(
...
+        """Upload script files to a bucket and return volumes to mount plus the scripts prefix.
+
+        Creates a bucket ``{namespace}/jobs-artifacts`` (if it doesn't exist) and uploads
+        each script to ``scripts/{uuid}/{remote_name}`` inside it. Returns a :class:`Volume`
+        that mounts the bucket at ``/artifacts`` so the job can access the scripts directly.
+        """
+        import uuid
+
+        bucket_id = f"{namespace}/{self._HF_JOBS_ARTIFACTS_BUCKET_NAME}"
+        subfolder_id = str(uuid.uuid4())
```
Wauplin: instead of a full uuid maybe something like

```
20260410T120900-{6 random hexadecimal digits}
```

it will be easier for user to browse artifacts bucket for a specific Job

src/huggingface_hub/hf_api.py
```diff
@@ -11957,50 +11969,125 @@ def _create_uv_command_env_and_secrets(
...
+        remote_to_local_file_names: dict[str, str],
+        token: bool | str | None,
+    ) -> tuple[list[Volume], str]:
+        """Upload script files to a bucket and return volumes to mount plus the scripts prefix.
+
+        Creates a bucket ``{namespace}/jobs-artifacts`` (if it doesn't exist) and uploads
+        each script to ``scripts/{uuid}/{remote_name}`` inside it. Returns a :class:`Volume`
+        that mounts the bucket at ``/artifacts`` so the job can access the scripts directly.
+        """
+        import uuid
```
Wauplin: import at top level

src/huggingface_hub/hf_api.py
Wauplin: can you move that to `constants.py` ?

Wauplin: Also I would have a slight preference for `jobs-artifacts` as bucket name but `/data` as mounted volume. That was the historical name for Persistent Storage on Spaces (and it's shorter/easier to remember)

src/huggingface_hub/constants.py
Wauplin: I would go even one step further and completely get rid of the base64-encoding branch. This was quite experimental/hacky so let's move to buckets and maintain only one transport mode

src/huggingface_hub/hf_api.py
Wauplin: should we make it read-write to encourage users to write their output in it? or is it not its purpose?

Wauplin (commented):
Thanks for working on that @davanstrien ! Pretty excited by what it will unlock so let's double-down on this direction 😃

Wauplin:
(converted to draft, @davanstrien could you reset to "ready for review" when ready? -just trying to filter my notifications here :smile: )

davanstrien:
> (converted to draft, @davanstrien could you reset to "ready for review" when ready? -just trying to filter my notifications here 😄 )

yes sorry for the noise!

davanstrien:
Thanks both! Pushed follow-up commits working through the feedback. TL;DR: took @Wauplin's stronger suggestion and dropped the base64 path entirely, which enabled some other simplifications. 

@lhoestq — the `scripts/` rename landed earlier in `93e00708`, and your "default to True" suggestion got subsumed by the base64 removal below (the flag is gone entirely).

@Wauplin:

- **`import uuid` at top level** — `5808552e`. Then actually removed from the top-level imports in `cf7e0193` once `uuid` was replaced (see next item).
- **Move bucket constants to `constants.py`** — `1a78fe92`. Added `HF_JOBS_ARTIFACTS_BUCKET_NAME` and `HF_JOBS_ARTIFACTS_MOUNT_PATH`.
- **`YYYYMMDDTHHMMSS-{6 hex}` subfolder naming** — `cf7e0193`. Uses `datetime.now(timezone.utc).strftime('%Y%m%dT%H%M%S')` + `secrets.token_hex(3)`.
- **Mount with `path=scripts_prefix`** — `9c85c1ea`. The bucket volume is now scoped to the per-job subfolder via `Volume.path`, so the container sees only its own files at the mount root. `_upload_scripts_to_bucket` no longer returns a prefix.
- **Mount at `/data`** — `554f5f64`. Bucket name stays `jobs-artifacts`.
- **Drop base64 entirely** — `91219b22`. The three failure modes that previously fell back to base64 now raise:
   - `/data` mount collision with a user volume → `ValueError`
   - `hf_xet` missing → `ImportError` with install hint
   - bucket create/upload failure → propagates

   `HF_JOBS_USE_BUCKET_TRANSPORT` is gone; ~50 lines of shell-quoting + xargs plumbing and 2 tests removed.
- **Read-write mount** — `df347c72`. Already RW by default, so this just makes the intent explicit at the call site (`read_only=False`), documents it in the docstring, and pins the invariant with a test assertion. Double-checked on the Hub side that omitted `readOnly` and explicit-false normalise to the same downstream payload — zero wire-level change.

**Smoke-tested end-to-end against production.** Ran `hf jobs uv run smoke.py` where the script prints its argv, lists `/data`, and writes `/data/smoke_output.txt`.

Job logs:
```
sys.argv = ['/data/hf_jobs_smoke.py']
/data contents: ['hf_jobs_smoke.py']    # only its own file
Wrote /data/smoke_output.txt - size 19 bytes
```

Bucket state after:
```
bucket://davanstrien/jobs-artifacts/scripts/20260413T133239-8bd453/
├── hf_jobs_smoke.py     ← uploaded by client
└── smoke_output.txt     ← written by job at runtime
```

src/huggingface_hub/constants.py
```diff
@@ -247,6 +247,11 @@ def list_files(repo_id: str):
 
 HF_XET_HIGH_PERFORMANCE: bool = _is_true(os.environ.get("HF_XET_HIGH_PERFORMANCE"))
 
+# Bucket and mount path used by the Jobs bucket script transport.
+# `/data` mirrors the historical Spaces persistent-storage path.
```
davanstrien: maybe a bit verbose

Wauplin:
```suggestion
# Bucket and mount path used when launching Jobs
```

better?

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+        }
 
-            # Load content to pass as environment variable with format
-            # file1 base64content1
-            # file2 base64content2
-            # ...
-            env["LOCAL_FILES_ENCODED"] = "\n".join(
-                remote_file_name + " " + base64.b64encode(Path(local_file_to_include).read_bytes()).decode()
-                for remote_file_name, local_file_to_include in remote_to_local_file_names.items()
+        # Local files are shipped to the job via a bucket mounted at `HF_JOBS_ARTIFACTS_MOUNT_PATH`.
```
```suggestion
        # Local files are shipped to the job via a bucket mounted at /data.
```

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+        existing_mount_paths = {v.mount_path for v in (volumes or [])}
+        if constants.HF_JOBS_ARTIFACTS_MOUNT_PATH in existing_mount_paths:
+            raise ValueError(
+                f"Mount path {constants.HF_JOBS_ARTIFACTS_MOUNT_PATH!r} is reserved for Jobs script "
+                "transport when running local scripts. Mount your volume at a different path."
```
```suggestion
                f"Mount path {constants.HF_JOBS_ARTIFACTS_MOUNT_PATH!r} is reserved for Jobs artifacts when running local scripts. Mount your volume at a different path."
```

I found it more explicit but just a personal feeling

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
-                """echo $LOCAL_FILES_ENCODED | xargs -n 2 bash -c 'echo "$1" | base64 -d > "$0"' && """
-                + f"uv run {' '.join(quoted_parts)}",
-            ]
-        return command, env, secrets
+        if not is_xet_available():
+            raise ImportError(
+                "Running `hf jobs uv run` with local scripts requires `hf_xet`, which is "
+                "installed by default with `huggingface_hub`. See the installation guide: "
+                "https://huggingface.co/docs/huggingface_hub/installation"
+            )
```
Wauplin: Don't think this part is needed 

```suggestion
```

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+        remote_to_local_file_names: dict[str, str],
+        token: bool | str | None,
+    ) -> list[Volume]:
+        """Upload script files to a per-job subfolder in the artifacts bucket.
+
+        Creates a bucket ``{namespace}/jobs-artifacts`` (if it doesn't exist) and uploads
+        each script to ``scripts/{timestamp}-{random}/{remote_name}`` inside it. Returns a
+        :class:`Volume` whose ``path`` is scoped to that per-job subfolder and mounted
+        read-write, so the job sees only its own files at ``HF_JOBS_ARTIFACTS_MOUNT_PATH``
+        and can write output artifacts back to the same prefix.
```
```suggestion
        Creates a bucket `/jobs-artifacts` (if it doesn't exist) and uploads
        each script to `scripts/{timestamp}-{random}/{remote_name}` inside it. Returns a
        [`Volume`] scoped to that bucket subfolder. Volume is in read-write mode so the Job can save data back to this bucket.
```

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+        """
+        bucket_id = f"{namespace}/{constants.HF_JOBS_ARTIFACTS_BUCKET_NAME}"
+        subfolder_id = f"{datetime.now(timezone.utc).strftime('%Y%m%dT%H%M%S')}-{token_hex(3)}"
+        scripts_prefix = f"scripts/{subfolder_id}"
```
```suggestion
```

I would not locate scripts under `scripts/<random name>`. Let's just save them at the root of the bucket under `<random name>/`. Especially because artifacts can be saved back to the mounted volume so it doesn't make sense to scope it to "scripts/"

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+
+        self.create_bucket(bucket_id=bucket_id, exist_ok=True, token=token)
+
+        add_ops: list[tuple[str | Path | bytes, str]] = [
+            (Path(local_path), f"{scripts_prefix}/{remote_name}")
```
```suggestion
            (Path(local_path), f"{subfolder_id}/{remote_name}")
```

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12153,94 @@ def _create_uv_command_env_and_secrets(
...
+            for remote_name, local_path in remote_to_local_file_names.items()
+        ]
+        self.batch_bucket_files(bucket_id=bucket_id, add=add_ops, token=token)
+
+        volume = Volume(
+            type="bucket",
+            source=bucket_id,
+            mount_path=constants.HF_JOBS_ARTIFACTS_MOUNT_PATH,
+            path=scripts_prefix,
```
```suggestion
            path=subfolder_id,
```

tests/test_cli.py
```diff
@@ -2875,36 +2875,193 @@ def test_run_with_volumes(self, runner: CliRunner) -> None:
...
+        mock_create_bucket.assert_called_once_with(bucket_id="test-user/jobs-artifacts", exist_ok=True, token=None)
+
+        # Files were uploaded under a scripts/{timestamp}-{hex}/ subfolder
+        mock_batch.assert_called_once()
+        call_kwargs = mock_batch.call_args
+        assert call_kwargs.kwargs["bucket_id"] == "test-user/jobs-artifacts"
+        add_ops = call_kwargs.kwargs["add"]
+        assert len(add_ops) == 1
+        upload_path = add_ops[0][1]
+        assert upload_path.startswith("scripts/")
```
Wauplin: to be updated

Wauplin (commented):
Thanks! Some minor comments but logic looks good to me. Haven't checked the tests much yet

lhoestq (approved):
lgtm !

src/huggingface_hub/hf_api.py
```diff
@@ -12145,50 +12152,84 @@ def _create_uv_command_env_and_secrets(
...
+        """Upload script files to a per-job subfolder in the artifacts bucket.
+
+        Creates a bucket `/jobs-artifacts` (if it doesn't exist) and uploads
+        each script to `{timestamp}-{random}/{remote_name}` inside it. Returns a
+        [`Volume`] scoped to that bucket subfolder. Volume is in read-write mode so the Job can save data back to this bucket.
+        """
+        bucket_id = f"{namespace}/{constants.HF_JOBS_ARTIFACTS_BUCKET_NAME}"
+        subfolder_id = f"{datetime.now(timezone.utc).strftime('%Y%m%dT%H%M%S')}-{token_hex(3)}"
+
+        self.create_bucket(bucket_id=bucket_id, exist_ok=True, token=token)
```
```suggestion
        self.create_bucket(bucket_id=bucket_id, exist_ok=True, token=token, private=True)
```

Let's set this bucket to private by default otherwise would leak internal scripts + output artifacts publicly

```suggestion
        bucket_url = self.create_bucket(bucket_id=bucket_id, exist_ok=True, token=token, private=True)
        print(f"Your script and Job artifacts will be saved in this Bucket: {bucket_url.url}"
```

also would be good to print a message to the user ^

Wauplin (approved):
Approved once comments are addressed 😃 

Tested it locally on a dummy script and it worked fine https://huggingface.co/buckets/Wauplin/jobs-artifacts/tree/20260415T092902-5ca267


I also took the liberty to update the PR description since the PR scope changed a bit since you created it

src/huggingface_hub/hf_api.py
davanstrien: very good catch! 😬

PR #4057, opened 2026-04-06
[CLI] Migrate `collections`, `discussions`, `extensions`, `endpoints` and `webhooks` to `out` singleton

hanouticelina:
Part of #3979 (Phase 2)

This PR migrates list and info commands in `collections `, `discussions`, `extensions`, `endpoints` and `webhooks` to use the out singleton.

This PR introduces some notable breaking changes,in addition to those that affects all migrated commands, e.g. `-q`/`--quiet` replaced by `--format quiet`:

  `hf endpoints ls`: (same issue as for `hf auth whoami`)
  - JSON: flat fields only (`name`, `model`, `status`, `task`, `framework`, `instance`, `vendor`, `region`), it was full `endpoint.raw` with nested objects when passing `--format json`. 

  `hf discussions info`:
  - Rich text view replaced by JSON output
  - `--comments`, `--diff`, `--no-color` flags removed 

---

> [!NOTE]
> **Medium Risk**
> Changes CLI output formatting and flags across multiple commands (notably replacing `--quiet/-q` with `--format quiet` and altering `hf discussions info` output/flags), which can break scripts relying on previous text output or options.
> 
> **Overview**
> Migrates `collections`, `discussions`, `extensions`, `endpoints`, and `webhooks` commands to use the `out` singleton (`--format [agent|auto|human|json|quiet]`) instead of ad-hoc `print`/`print_list_output` + `--quiet/-q`.
> 
> Updates output behavior to be mode-aware (human tables vs agent TSV/JSON), adds per-invocation reset of output state (`HFCliApp.__call__` / `hf` callback), and adjusts some commands’ structured output (e.g., `hf discussions info` now emits JSON via `out.dict` and drops `--comments/--diff/--no-color`; `hf endpoints ls`/`hf webhooks ls` preserve nested fields in `json/agent` but flatten for human display). Docs and tests are updated to match the new flags and output semantics, including safeguards that `quiet` mode doesn’t leak between CLI invocations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a588ff0106b38c173f69881d74f2b381ff59017b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/collections.py
+++ b/src/huggingface_hub/cli/collections.py
@@ -40,14 +40,7 @@ import typer
 
 from huggingface_hub.hf_api import CollectionItemType_T, CollectionSort_T
 
-from ._cli_utils import (
-    FormatWithAutoOpt,
-    LimitOpt,
-    TokenOpt,
-    api_object_to_dict,
-    get_hf_api,
-    typer_factory,
-)
+from ._cli_utils import FormatWithAutoOpt, LimitOpt, TokenOpt, api_object_to_dict, get_hf_api, typer_factory
 from ._output import OutputFormatWithAuto, out
 
 
@@ -260,7 +253,7 @@ def collections_add_item(
 ) -> None:
     """Add an item to a collection."""
     api = get_hf_api(token=token)
-    api.add_collection_item(
+    collection = api.add_collection_item(
         collection_slug=collection_slug,
         item_id=item_id,
         item_type=item_type.value,  # type: ignore[arg-type]
@@ -302,7 +295,7 @@ def collections_update_item(
         note=note,
         position=position,
     )
-    out.result("Item updated in collection", slug=collection_slug, url=collection.url)
+    out.result("Item updated in collection", slug=collection_slug)
 
 
 @collections_cli.command("delete-item")
@@ -328,4 +321,4 @@ def collections_delete_item(
         item_object_id=item_object_id,
         missing_ok=missing_ok,
     )
-    out.result("Item deleted from collection", slug=collection_slug, url=collection.url)
+    out.result("Item deleted from collection", slug=collection_slug)


src/huggingface_hub/cli/discussions.py
```diff
@@ -240,70 +158,24 @@ def discussion_list(
...
-            help="Show the diff (for pull requests).",
-        ),
-    ] = False,
-    no_color: Annotated[
-        bool,
-        typer.Option(
-            "--no-color",
-            help="Disable colored output.",
-        ),
-    ] = False,
```
hanouticelina, line 271: breaking change here: the rich text view defined in `_print_discussion_info` is replaced by `out.dict(details)`, consistent with all other info commands (`models info`, `datasets info` etc). The `--comments`, `--diff`, and `--no-color` flags are removed. Comments and diff are always included in the JSON output. I think it's fine no? a user can just call `hf discussions diff` to view diffs separately

src/huggingface_hub/cli/webhooks.py
```diff
@@ -307,4 +298,4 @@ def webhooks_delete(
             raise typer.Abort()
     api = get_hf_api(token=token)
     api.delete_webhook(webhook_id)
-    print(f"Webhook deleted: {webhook_id}")
+    out.result("Webhook deleted", id=webhook_id)
```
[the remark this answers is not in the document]
hanouticelina: @cursoragent address that

hanouticelina:
@Wauplin ready for review! (unless Cursor pushes unwanted commits again 😄)

src/huggingface_hub/cli/discussions.py
Wauplin: Definitely fine yes! Can you mention it in the PR description so it's picked up in the release notes?

hanouticelina: yes it's already mentioned!

src/huggingface_hub/cli/collections.py
Wauplin, line 118: `format` add --format as argument since it's using `out` ? (same for all commands actually, is this something we can easily do?)

src/huggingface_hub/cli/collections.py
```diff
@@ -161,8 +158,8 @@ def collections_create(
         private=private,
         exists_ok=exists_ok,
     )
-    print(f"Collection created: {collection.url}")
-    print(json.dumps(api_object_to_dict(collection), indent=2))
+    out.result("Collection created", url=collection.url)
+    out.dict(collection)
```
Wauplin: Tested locally and I think it's fine to drop `collection` info


```suggestion
    out.result("Collection created", slug=collection.slug, url=collection.url)
```

Maybe something like this ^ is the best tradeoff

Wauplin: here is current output in human mode

Image

src/huggingface_hub/cli/collections.py
```diff
@@ -207,8 +204,8 @@ def collections_update(
         private=private,
         theme=theme,
     )
-    print(f"Collection updated: {collection.url}")
-    print(json.dumps(api_object_to_dict(collection), indent=2))
+    out.result("Collection updated", url=collection.url)
+    out.dict(collection)
```
Wauplin:
```suggestion
    out.result("Collection updated", slug=collection.slug, url=collection.url)
```

src/huggingface_hub/cli/collections.py
```diff
@@ -229,7 +226,7 @@ def collections_delete(
     """Delete a collection from the Hub."""
     api = get_hf_api(token=token)
     api.delete_collection(collection_slug, missing_ok=missing_ok)
-    print(f"Collection deleted: {collection_slug}")
+    out.result("Collection deleted", slug=collection_slug)
```
Wauplin, line 223:
```suggestion
    out.result("Collection deleted", slug=collection.slug, url=collection.url)
```

(for consistency?)

src/huggingface_hub/cli/collections.py
Wauplin: this is actually a tricky one (same for all "return a single collection" commands). I feel that in most cases the relevant information is "slug" "title" and a nicely serialized "items" list. And optionally more fields like "description" "owner" (but only name/fullname) "private" "last_updated" "upvotes" etc.

not sure how to make the `out` API easy to use for such cases

Wauplin: why not maybe an API like this in the `out` singleton?

```py
    # defines how a Collection must be formatted
    collectionFormatter = Formatter(
        owner=lambda obj: obj["name"], # each value is a Callable[[Any], str]
        items=...
        # could even be a Callable[[Any, OutputFormat], str] to format each value depending on current mode
        options={"id_key": bool, "ignore_unknwon_fields": bool, etc.} # later we could even add options to the formatter
    )

    # registered once at the singleton level
    out.register(Collection, collectionFormatter)

    # each command would just call "out.result" with a Collection object
    out.result("Collection created", collection)
```

(might be overthinking but I do think it can be useful for modules that always outputs the same object)

src/huggingface_hub/cli/collections.py
Wauplin: An issue with using iteratively `out.result` and `out.dict` is that in json mode the output is not a true JSON anymore but a JSONL. Not a big issue for tools like `jq` but worth mentioning

src/huggingface_hub/cli/discussions.py
Wauplin, line 1: much cleaner!

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -41,51 +33,40 @@
...
+                "name": raw.get("name", ""),
+                "model": model.get("repository", "") if isinstance(model, dict) else "",
+                "status": status.get("state", "") if isinstance(status, dict) else "",
+                "task": model.get("task", "") if isinstance(model, dict) else "",
+                "framework": model.get("framework", "") if isinstance(model, dict) else "",
+                "instance": compute.get("instanceType", "") if isinstance(compute, dict) else "",
+                "vendor": provider.get("vendor", "") if isinstance(provider, dict) else "",
+                "region": provider.get("region", "") if isinstance(provider, dict) else "",
+            }
+        )
```
Wauplin, line 68: likely a Formatter could be used (depends on the APi we build for that)

src/huggingface_hub/cli/collections.py
hanouticelina: you're asking if a command uses `out.*` methods, it should accept `--format` so users can control the output mode? then yes we can just add `format: FormatWithAutoOpt = OutputFormatWithAuto.auto` as a parameter. The `FormatWithAutoOpt` callback calls `out.set_mode(value)`

Wauplin: > if a command uses out.* methods, it should accept --format so users can control the output mode? 

Yes I think that any command compatible with `out.*` singleton should accept as input `--format` / `--json` / `--quiet` parameters

hanouticelina: yes, every command using `out.*` accepts `--format`, `--json`, and `-q`/`--quiet`

src/huggingface_hub/cli/collections.py
```diff
@@ -257,19 +255,19 @@ def collections_add_item(
...
+    api.add_collection_item(
         collection_slug=collection_slug,
         item_id=item_id,
         item_type=item_type.value,  # type: ignore[arg-type]
         note=note,
         exists_ok=exists_ok,
     )
-    print(f"Item added to collection: {collection_slug}")
-    print(json.dumps(api_object_to_dict(collection), indent=2))
+    out.result("Item added to collection", slug=collection_slug)
```
Wauplin:
```suggestion
    out.result("Item added to collection", slug=collection_slug, url=collection.url)
```

for consistency

src/huggingface_hub/cli/collections.py
```diff
@@ -303,7 +302,7 @@ def collections_update_item(
         note=note,
         position=position,
     )
-    print(f"Item updated in collection: {collection_slug}")
+    out.result("Item updated in collection", slug=collection_slug)
```
Wauplin:
```suggestion
    out.result("Item updated in collection", slug=collection_slug, url=collection.url)
```

src/huggingface_hub/cli/collections.py
```diff
@@ -328,4 +328,4 @@ def collections_delete_item(
         item_object_id=item_object_id,
         missing_ok=missing_ok,
     )
-    print(f"Item deleted from collection: {collection_slug}")
+    out.result("Item deleted from collection", slug=collection_slug)
```
Wauplin:
```suggestion
    out.result("Item deleted from collection", slug=collection_slug, url=collection.url)
```

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -369,9 +354,9 @@ def update(
             token=token,
         )
     except HfHubHTTPError as error:
-        typer.echo(f"Update failed: {error}")
+        out.error(f"Update failed: {error}")
         raise typer.Exit(code=error.response.status_code) from error
-    _print_endpoint(endpoint)
+    out.dict(endpoint.raw)
```
Wauplin, line 359: still feel that `out.dict(endpoint.raw)` is slightly too verbose but fine to keep for now (also, endpoints is less of a priority)

I think here it could be something like `out.result("Endpoint updated", <nicely formatted endpoint>)` (same for all "action commands")

src/huggingface_hub/cli/collections.py
hanouticelina: oops committed a bit too fast, we don't have the url of the collection here `delete_collection_item` does not return the collection url

PR #4018, opened 2026-04-01
[Spaces] Add support for mounted volumes

Wauplin:
Related to [server-side PR](https://github.com/huggingface-internal/moon-landing/pull/17362) (private link) cc @XciD 

This PR adds `set_space_volumes` and `delete_space_volumes` and updates the Spaces guide.

I also took the opportunity to deprecate the persistent storage methods.

---

Tested manually with https://huggingface.co/spaces/Wauplin/test-volumes

```py
>>> get_space_runtime("Wauplin/test-volumes").volumes
[Volume(type='bucket', source='Wauplin/test-volumes-storage', mount_path='/data', revision=None, read_only=False, path=None)]
```

---

> [!NOTE]
> **Medium Risk**
> Introduces new Space runtime/mutation API endpoints (`/volumes`) and deprecates existing storage methods/options, which can affect client integrations and backwards compatibility if callers rely on `space_storage` behavior.
> 
> **Overview**
> Adds first-class *mounted volumes* support for Spaces via new `HfApi.set_space_volumes` (PUT) and `HfApi.delete_space_volumes` (DELETE), plus exposes mounted volumes on `SpaceRuntime.volumes` using a shared `Volume` dataclass.
> 
> Deprecates persistent storage configuration and APIs: `space_storage` on `create_repo`/`duplicate_repo` and the `request_space_storage`/`delete_space_storage` methods now emit deprecation warnings; CLI docs/option help mark `--storage` as deprecated and the Spaces guide is updated to document volumes instead of persistent storage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c282531d7b3b1089102ca2e3f3ee45551053277e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4061, opened 2026-04-07
Fix set_space_volume / delete_space_volume return types

abidlabs:
### 🚨 This PR removes the returned value of both `set_space_volumes` and `delete_space_volumes`. This is a breaking change but in practice it never worked so it's more of a fix ^^

Not sure if this is a problem with Hub server or library, but basically I've noticed that `PUT` / `DELETE` `/api/spaces/{repo_id}/volumes` can return a JSON body that is **not** a full Space runtime document (e.g. only volume-related fields, no top-level `stage`). The client always did `return SpaceRuntime(r.json())`, which raises when `stage` is missing:

```text
KeyError: 'stage'
```

Now working correctly with

```python
from huggingface_hub import Volume, create_repo, delete_space_volumes, set_space_volumes

repo_id = create_repo(repo_id="my-cool-training-space", repo_type="space", space_sdk="gradio", exist_ok=True).repo_id
set_space_volumes(
    repo_id=repo_id,
    volumes=[Volume(type="model", source="openai-community/gpt2", mount_path="/gpt2", read_only=True)],
)
delete_space_volumes(repo_id=repo_id)

```

Noticed while testing the latest version of `huggingface_hub` (1.9.1) with `trackio`

---

> [!NOTE]
> **Medium Risk**
> This is a behavior/signature change for two public client methods and may break callers relying on the returned `SpaceRuntime`, though the logic change itself is small and localized.
> 
> **Overview**
> Adjusts Space volume mutation APIs to no longer parse and return `SpaceRuntime` from `PUT`/`DELETE /api/spaces/{repo_id}/volumes` responses.
> 
> `HfApi.set_space_volumes` and `HfApi.delete_space_volumes` now return `None` and drop the documented return value, preventing `SpaceRuntime(r.json())` failures when the server returns a partial body.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 029beddf5a07ed8f9eebbf779680172f887a64c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -8160,23 +8160,6 @@ class HfApi:
         hf_raise_for_status(r)
         return SpaceRuntime(r.json())
 
-    def _space_runtime_after_mutation(
-        self,
-        repo_id: str,
-        response: httpx.Response,
-        *,
-        token: bool | str | None = None,
-    ) -> SpaceRuntime:
-        if not response.content:
-            return self.get_space_runtime(repo_id, token=token)
-        data = response.json()
-        if isinstance(data, dict):
-            try:
-                return SpaceRuntime(data)
-            except (KeyError, TypeError, ValueError):
-                pass
-        return self.get_space_runtime(repo_id, token=token)
-
     @validate_hf_hub_args
     def set_space_volumes(
         self,
@@ -8184,7 +8167,7 @@ class HfApi:
         volumes: list[Volume],
         *,
         token: bool | str | None = None,
-    ) -> SpaceRuntime:
+    ) -> None:
         """Set volumes for a Space.
 
         Sets (or replaces) the list of volumes mounted in the Space. Each volume gives the Space's container access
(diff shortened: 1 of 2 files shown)


davanstrien:
Thanks for the fix! IIUC the response parsing issue has been there since the volumes API was first added in #4018, but it wasn't reachable until #4054 fixed the request payload format.

abidlabs:
yup @davanstrien exactly, thanks for that PR!

Wauplin:
Hey @abidlabs thanks for reporting and very sorry we've not tested this before shipping (my bad on this one :see_no_evil: ). Simplest solution here is to return None and let the user make an additional call to `get_space_runtime` if they really want the current state. I've updated the PR changes and PR description to reflect that (I could have opened a separate PR but preferred to keep it here for viz')

Wauplin:
Shipped in https://github.com/huggingface/huggingface_hub/releases/tag/v1.9.2

PR #4065, opened 2026-04-07
[BUG FIX]: hf_hub_download crashes when stderr lacks a real file descriptor

### LINKED ISSUE #4066: bug: hf_hub_download crashes with ValueError when stderr lacks a real file descriptor
## Problem

`hf_hub_download` crashes with `ValueError: bad value(s) in fds_to_keep` in any environment where `sys.stderr` is not backed by a real file descriptor.

The download itself is fine. Only tqdm's progress bar initialization fails, but because `_create_progress_bar` has no safety net, it takes down the entire download.

## Affected environments

Any runtime where `sys.stderr` is not backed by a real file descriptor:

- Textual TUI apps
- Jupyter notebooks
- pytest capture
- WSGI/ASGI web servers (gunicorn, uvicorn)
- Sandboxed/containerized runtimes
- IDE embedded terminals

## Root cause

`_create_progress_bar` calls `cls(disable=disable, name=name, **kwargs)` with no exception handling. When stderr's `fileno()` returns `-1`, tqdm's `__init__` tries to create a multiprocessing lock, which spawns a resource tracker subprocess via `fork_exec()`, which rejects the invalid fd.

## Reproduction

Add to `tests/test_file_download.py` in the `CachedDownloadTests` class and run with `pytest tests/test_file_download.py -k test_hf_hub_download_survives_bad_fileno`:

```python
import os
from unittest.mock import patch

from huggingface_hub import constants, hf_hub_download
from huggingface_hub.utils import SoftTemporaryDirectory

from .testing_utils import DUMMY_MODEL_ID


def test_hf_hub_download_survives_bad_fileno(self):
    """hf_hub_download should not crash when stderr.fileno() returns -1."""

    class FakeStderr:
        def write(self, s):
            pass

        def flush(self):
            pass

        def isatty(self):
            return True

        def fileno(self):
            return -1

    with SoftTemporaryDirectory() as tmpdir:
        with patch("sys.stderr", FakeStderr()):
            filepath = hf_hub_download(
                DUMMY_MODEL_ID,
                filename=constants.CONFIG_NAME,
                cache_dir=tmpdir,
            )
        self.assertTrue(os.path.exists(filepath))
```

## Demo

Crash demo

## Proposed fix

See #4065.

tobocop2:
## Problem

`hf_hub_download` crashes with `ValueError: bad value(s) in fds_to_keep` when `sys.stderr.fileno()` returns `-1`. The crash happens during tqdm's progress bar setup, before the download itself runs.

In practice this hits Textual TUI apps (Textual's `app.py` returns `-1` from `fileno()`) and any custom stderr wrapper that does the same. CPython's `multiprocessing/resource_tracker` silently swallows exceptions from `fileno()`, so streams like `io.StringIO`, Jupyter's `OutStream`, and pytest's `CaptureIO` — which all *raise* on `fileno()` — are not affected. Two more preconditions are needed for the crash: the multiprocessing start method must be `spawn` or `forkserver`, and the resource tracker must not have been started already.

## Root cause

The first time tqdm's default lock is touched, it lazily constructs `TqdmDefaultWriteLock`, which calls `cls.create_mp_lock()`. That spawns the multiprocessing resource tracker via `fork_exec`, which rejects `stderr.fileno() == -1`.

## Why this fix

Setting a thread-only `RLock` on the HF tqdm subclass at module import short-circuits the lazy init: `tqdm.get_lock()` finds `_lock` already set and skips the `create_mp_lock` path entirely. The patch is one line plus a comment.

The trade-off is no inter-process tqdm bar coordination on the HF subclass — multiple separate Python processes printing tqdm bars to the same TTY may visually interleave. Downloads themselves are unaffected, and `huggingface_hub` parallelizes via threads inside one process (not subprocesses), so this is not a scenario the library targets.

## Crash demo

Crash demo

Fixes #4066

---

> [!NOTE]
> **Medium Risk**
> Changes global progress-bar locking behavior by overriding `tqdm`'s default multiprocessing lock at import time, which could affect multi-process progress bar coordination. Download logic is unchanged, and risk is largely limited to progress display behavior.
> 
> **Overview**
> Prevents `hf_hub_download` from crashing when `sys.stderr.fileno()` returns `-1` (e.g. Textual TUIs) by setting a thread-only `RLock` via `tqdm.set_lock(...)`, avoiding tqdm’s lazy multiprocessing lock/resource-tracker initialization.
> 
> Adds a regression test that patches `sys.stderr` with a bad `fileno()` implementation and verifies the download completes successfully.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16d5a58a9d023441101fbefb616d850c143037ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/tqdm.py
+++ b/src/huggingface_hub/utils/tqdm.py
@@ -82,6 +82,7 @@ Group-based control:
 import io
 import logging
 import os
+import threading
 import warnings
 from collections.abc import Iterator
 from contextlib import contextmanager, nullcontext
@@ -295,6 +296,21 @@ def tqdm_stream_file(path: Path | str) -> Iterator[io.BufferedReader]:
         pbar.close()
 
 
+class _SafeTqdm(old_tqdm):
+    """tqdm subclass that uses a thread lock instead of a multiprocessing lock.
+
+    Used as a fallback when the standard tqdm cannot initialize its multiprocessing
+    lock (e.g. when stderr.fileno() returns -1 in Textual, Jupyter, pytest, etc.).
+    """
+
+    _lock = threading.RLock()
+
+    @classmethod
+    def get_lock(cls):
+        return cls._lock
+
+
+
 def _create_progress_bar(*, cls: type[old_tqdm], log_level: int, name: str | None = None, **kwargs) -> old_tqdm:
     """Create a progress bar.
 
@@ -309,12 +325,27 @@ def _create_progress_bar(*, cls: type[old_tqdm], log_level: int, name: str | Non
     """
     # issubclass() crashes on non-class callables (e.g. functools.partial), guard with isinstance.
     if not (isinstance(cls, type) and issubclass(cls, tqdm)):
-        return cls(**kwargs)  # type: ignore[return-value]
+        try:
+            return cls(**kwargs)
+        except (OSError, ValueError):
+            # Caller opted into custom progress; a visible fallback bar would
+            # corrupt their output (e.g. TUIs), so disable it.
+            return _SafeTqdm(disable=True, **kwargs)
 
     # HF subclass: keep the historical log-level / TTY behavior. Group-based
     # disabling is already handled in `tqdm.__init__`.
     disable = is_tqdm_disabled(log_level)
-    return cls(disable=disable, name=name, **kwargs)  # type: ignore[return-value]
+    try:
+        return cls(disable=disable, name=name, **kwargs)
+    except (OSError, ValueError):
+        warnings.warn(
+            "Progress bar could not be initialized in this environment. "
+            "Download will continue without progress reporting. "
+            "To suppress this warning, call `disable_progress_bars()` or set HF_HUB_DISABLE_PROGRESS_BARS=1.",
+            stacklevel=2,
+        )
+        # disable=None would attempt TTY auto-detect on the broken stderr.
+        return _SafeTqdm(disable=True, **kwargs)
 
 
 def _get_progress_bar_context(
--- a/src/huggingface_hub/utils/tqdm.py
+++ b/src/huggingface_hub/utils/tqdm.py
@@ -234,6 +234,13 @@ class tqdm(old_tqdm):
                 raise
 
 
+# Prevent tqdm's default multiprocessing write-lock from spawning a resource
+# tracker subprocess via fork_exec(). That path fails when stderr has an invalid
+# fd (e.g. Textual TUIs that return -1 from sys.stderr.fileno()). Inter-process
+# bar coordination on the HF subclass is not a supported use case. See #4065.
+tqdm.set_lock(threading.RLock())
+
+
 class silent_tqdm:
     """Fake tqdm object that does nothing."""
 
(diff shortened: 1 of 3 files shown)


tobocop2:
Friendly bump on this, happy to address any feedback if needed. Let me know if there's anything blocking review.

Wauplin:
(generated by Claude)

Thanks for digging into this @tobocop2! I reproduced the crash and it's a real bug but I'd like to push back a bit on both the framing and the size of the fix before merging.

### On the affected-environments list

The PR description says this affects Textual, Jupyter, pytest capture, gunicorn/uvicorn, sandboxed runtimes, and IDE terminals. In practice that list is largely overstated. The relevant code path in CPython is:

```python
# multiprocessing/resource_tracker.py (ensure_running)
fds_to_pass = []
try:
    fds_to_pass.append(sys.stderr.fileno())
except Exception:
    pass
```

Anything that **raises** from `fileno()` is silently swallowed — only a wrapper that explicitly **returns `-1`** triggers the crash. Quick survey:

| Environment | `fileno()` behavior | Crashes? |
|---|---|---|
| `io.StringIO` / `BytesIO` | raises `UnsupportedOperation` | No |
| pytest capture (`CaptureIO`, `DontReadFromInput`) | raises `UnsupportedOperation` | No |
| Jupyter / ipykernel `OutStream` | raises `UnsupportedOperation` (or returns a real fd) | No |
| **Textual** (`src/textual/app.py`) | literal `return -1` | **Yes** |
| uvicorn / gunicorn | don't replace stderr | No |

Plus two compounding preconditions: the multiprocessing start method has to be `spawn`/`forkserver` (not the Linux `fork` default), and the resource tracker must not have been started by something else first.

So in practice the realistic trigger is **Textual** (and any custom wrapper that returns `-1` on purpose). That's still worth fixing — TUIs that download models are a legitimate use case — but I'd trim the description so future readers don't go hunting for a Jupyter repro that doesn't exist (at least AFAIK).

### On the fix

The current patch adds a `_SafeTqdm` subclass plus two `try/except (OSError, ValueError)` guards. I would prefer a smaller alternative: proactively set a thread lock on our `tqdm` subclass at module import, before tqdm ever tries to build an mp lock. Something like:

```python
# in utils/tqdm.py, near the tqdm subclass definition
import threading
tqdm.set_lock(threading.RLock())  # avoid the mp-lock fork_exec path entirely
```

That would:
- Eliminate the crash proactively (no try/except, no fallback class).
- Keep behavior identical for the 99.9% of users on `fork` / real terminals.
- Only "lose" inter-process tqdm coordination for people who use the HF subclass across multiple processes — which I don't think is a real use case for us.

If you'd rather keep the reactive shape, then the `_SafeTqdm` class can be dropped: the fallback path always disables the bar, so just returning a disabled `old_tqdm` with `_lock` patched on first failure would be enough.

Happy to be wrong about the proactive approach if there's a multiprocessing scenario I'm forgetting — but either way I think we can land this in ~5 lines instead of ~30. WDYT?

Wauplin:
Hey @tobocop2 , the comment above has been AI-generated after some investigation from both me and a local Claude, so I agree with what's stated above. I also confirm that I reproduced the bug in Textual but not in other cases you've mentioned (explaining why we haven't got any reports so far). About the solution I'd prefer a smaller one even if it means poorer UX in rare scenarios.

tobocop2:
> Hey @tobocop2 , the comment above has been AI-generated after some investigation from both me and a local Claude, so I agree with what's stated above. I also confirm that I reproduced the bug in Textual but not in other cases you've mentioned (explaining why we haven't got any reports so far). About the solution I'd prefer a smaller one even if it means poorer UX in rare scenarios.

thanks for the review @Wauplin 
I like your solution a lot. Thanks again. 


I've just reflected your feedback and updated the PR and description.

Wauplin (approved):
Thanks for making the change @tobocop2 👍  

Learnt a few things along the way :) Feels like we are fixing an issue mostly due to how Textual handles no file descriptors (with a -1 instead of a proper `UnsupportedOperation`) + how CPython do not check for invalid fileno values [here](https://github.com/python/cpython/blob/62792c8f77222b86f8d2d985fc3a2b9fffc679e8/Lib/multiprocessing/resource_tracker.py#L239). Wonder if this should live in `tqdm` directly instead but I'm fine with having the fix here since it's small.

Note that if we notice any drawback in the future we might rollback to another solution (e.g.  by explicitly calling `sys.stderr.fileno()` in HF custom tqdm to disable it if running inside Textual?)

tobocop2:
greatly appreciate you @Wauplin 

This means a lot and thanks again

PR #4092, opened 2026-04-13
Fix reference cycle in `hf_raise_for_status` causing delayed object destruction

Wauplin:
### I have tested the regression test introduced in this PR and it does fail on `main`

**Kudos to @yg7445 for investigating + suggesting the solution. This was not an easy bug to spot!**

---

## Summary

Fixes a CPython reference cycle in `hf_raise_for_status()` that prevents deterministic exception cleanup.

When exceptions are stored in local variables before `raise ... from e`, a reference cycle forms: `err.__cause__` → `e` → `e.__traceback__` → frame → `f_locals['err']` → back to start. This delays garbage collection and causes real issues downstream (e.g., vLLM GPU memory not released until cyclic GC runs). See vllm-project/vllm#38384.

Alternative approach to #4084: instead of introducing two new helper functions, this extends the existing `_format()` with `**attrs` so attributes (like `repo_type`, `repo_id`, `bucket_id`) are set inside `_format` and the exception is never stored in the caller's frame.

## Changes

- Extended `_format()` to accept `**attrs` keyword arguments, set as attributes on the error
- Converted all 5 call sites in `hf_raise_for_status` from local-variable pattern to inline `raise _format(...) from e`
- Added regression test using `weakref` to verify no reference cycle exists

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 8ba3871177887a96e5e1d19c013f24b082d006d0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -774,26 +774,17 @@ def hf_raise_for_status(response: httpx.Response, endpoint_name: str | None = No
 
         if error_code == "RevisionNotFound":
             message = f"{response.status_code} Client Error." + "\n\n" + f"Revision Not Found for url: {response.url}."
-            revision_err = _format(RevisionNotFoundError, message, response)
-            revision_err.repo_type = repo_type
-            revision_err.repo_id = repo_id
-            raise revision_err from e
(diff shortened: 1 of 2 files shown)


hanouticelina (approved):
the fix looks good to me 👍

PR #4100, opened 2026-04-14
[CLI] Migrate `download`, `upload`, `upload-large-folder` to `out` singleton

hanouticelina:
Part of #3979.
This PR migrates `download`, `upload`, `upload-large-folder` CLI commands to the `out` singleton.

---

> [!NOTE]
> **Medium Risk**
> Changes CLI public interface and stdout/stderr output for `hf download`, `hf upload`, and `hf upload-large-folder` by replacing `--quiet`/`print()` with `--format`/`out`, which may break existing scripts relying on old flags or exact output text.
> 
> **Overview**
> Migrates `hf download`, `hf upload`, and `hf upload-large-folder` to the shared `out` output singleton, replacing ad-hoc `print()`/tabulate output with `out.result`, `out.text`, and `out.table` (including structured dry-run output for `download`).
> 
> Replaces the commands’ `--quiet` flag with `--format [agent|auto|human|json|quiet]` and wires `quiet` mode to automatically disable progress bars via the `FormatWithAutoOpt` callback.
> 
> Updates generated CLI docs and adjusts CLI tests/skill markdown expectations to use `--format quiet` and the new output paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9369722dbf7e68aec2996c91b3ac11bdc28e07b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -791,7 +791,7 @@ FormatOpt = Annotated[
 
 def _set_output_mode(value: OutputFormatWithAuto) -> OutputFormatWithAuto:
     out.set_mode(value)
-    if out.mode == OutputFormatWithAuto.quiet:
+    if out.mode != OutputFormatWithAuto.human:
         disable_progress_bars()
     return value
 
--- a/src/huggingface_hub/cli/_cli_utils.py
+++ b/src/huggingface_hub/cli/_cli_utils.py
@@ -34,7 +34,15 @@ from typer.core import TyperCommand, TyperGroup
 
 from huggingface_hub import Volume, __version__, constants
 from huggingface_hub.errors import CLIError
-from huggingface_hub.utils import ANSI, get_session, hf_raise_for_status, installation_method, logging, tabulate
+from huggingface_hub.utils import (
+    ANSI,
+    disable_progress_bars,
+    get_session,
+    hf_raise_for_status,
+    installation_method,
+    logging,
+    tabulate,
+)
 from huggingface_hub.utils._dotenv import load_dotenv
 
 from ._output import OutputFormatWithAuto, out
@@ -783,6 +791,8 @@ FormatOpt = Annotated[
 
 def _set_output_mode(value: OutputFormatWithAuto) -> OutputFormatWithAuto:
     out.set_mode(value)
+    if out.mode == OutputFormatWithAuto.quiet:
+        disable_progress_bars()
     return value
 
 
--- a/src/huggingface_hub/cli/upload_large_folder.py
+++ b/src/huggingface_hub/cli/upload_large_folder.py
@@ -19,9 +19,10 @@ from typing import Annotated
 import typer
 
 from huggingface_hub import logging
-from huggingface_hub.utils import ANSI, disable_progress_bars
+from huggingface_hub.utils import disable_progress_bars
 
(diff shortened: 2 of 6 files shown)


src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -783,6 +791,8 @@ class OutputFormat(str, Enum):
 
 def _set_output_mode(value: OutputFormatWithAuto) -> OutputFormatWithAuto:
     out.set_mode(value)
+    if out.mode == OutputFormatWithAuto.quiet:
+        disable_progress_bars()
```
hanouticelina: progress bars are disabled for `quiet` output mode. the spec in #3979 mentioned `out.progress(iterable)` but `download`/`upload` don't create progress bars at the CLI level - they're created deep inside library functions. so I'm not sure if we need  a tqdm wrapper in the CLI for now. let's revisit `out.progress` if a CLI command needs to create its own progress bar

Wauplin: and what about agent mode? I think we don't want progress bars in that case either (lot of noise/tokens for almost nothing). For JSON mode I'd say we can keep it

src/huggingface_hub/cli/upload_large_folder.py
```diff
@@ -83,34 +85,33 @@ def upload_large_folder(
...
-            "  - You must not delete the `./.cache/huggingface/` folder while a process is running.\n"
-            "  - You can delete the `./.cache/huggingface/` folder to reinitialize the upload state when process is not running. Files will have to be hashed and preuploaded again, except for already committed files.\n"
-            "\n"
-            "If the process output is too verbose, you can disable the progress bars with `--no-bars`. "
-            "You can also entirely disable the status report with `--no-report`.\n"
-            "\n"
-            "For more details, run `hf upload-large-folder --help` or check the documentation at "
-            "https://huggingface.co/docs/huggingface_hub/guides/upload#upload-a-large-folder."
-        )
+    out.text(
```
Wauplin:
```suggestion
    out.warning(
```

makes more sense IMO

Wauplin (approved):
Thanks!

src/huggingface_hub/cli/_cli_utils.py
hanouticelina: yes we can keep it only for human mode only imo

hanouticelina: JSON mode I'm not sure either, I think only agents would run this command with `--format json`

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -783,6 +791,8 @@ class OutputFormat(str, Enum):
 
 def _set_output_mode(value: OutputFormatWithAuto) -> OutputFormatWithAuto:
     out.set_mode(value)
+    if out.mode == OutputFormatWithAuto.quiet:
```
hanouticelina:
```suggestion
    if out.mode != OutputFormatWithAuto.human:
```

src/huggingface_hub/cli/_cli_utils.py
Wauplin: yes fine for keeping only for humans

PR #4102, opened 2026-04-14
Replace token-paste login with Device Code OAuth

Wauplin:
## Summary

Replaces the interactive token-paste login flow with OAuth Device Code (RFC 8628). Users no longer need to manually create and copy-paste tokens. Instead: CLI prints a URL + 8-char code, user authorizes in the browser, done.

- `login()`, `interpreter_login()`, `notebook_login()`, `hf auth login` all use the new flow
- `--token` flag still works for CI/scripting (unchanged)
- Tokens expire after 30 days (mentioned to users on login)

## Open questions

- should we implement refresh tokens? if yes, how? should it be a `hf auth refresh` command or directly in `get_token` if token is outdated?
- how to make the `scopes` page less worrying in the UI?

## Decisions

- **No browser auto-open** — too many edge cases (SSH, headless, text browsers hijacking terminal). Just print the URL. Can always add back later.
- **Scopes hardcoded server-side** on the OAuth app, not in the client — easier to update without a library release (currently asks for all possible permissions)
- **Renamed `_login()` → `_validate_and_save_token()`** — the old name was confusing since the actual "login" is the device code flow now.
- **Token naming** — OAuth tokens may not have `displayName` in whoami response, so we fallback to `oauth-{username}`.
- **No token refresh** — keep it simple, user re-runs login when expired.
- **Removed ASCII art logo** from `interpreter_login()` — felt dated.
- **Dropped ipywidgets dependency** for `notebook_login()` — uses `IPython.display.HTML` instead.
- **Removed Google Colab vault retrieval** — the vault UX was bad (prompts for access every session regardless). Colab users can just run `login()` like everyone else now.
- **`DeviceCodeError` exception** — all device code OAuth failures now raise a dedicated `DeviceCodeError` (instead of generic `RuntimeError`). Non-JSON responses from the token endpoint are caught with a clear error message.
- **Context-aware expiration notice** — `notebook_login()` tells users to run `notebook_login()` again, not `hf auth login`.

---

### In terminal:

image

image

image

image

image

---

### In colab:

https://colab.research.google.com/drive/1fX-x_uoZ4RYFPvUap7kEEj_5pdLsNWSB?usp=sharing

image

image

image


Wauplin:
closed as superseded by https://github.com/huggingface/huggingface_hub/pull/4354

PR #4104, opened 2026-04-15
[CLI] Add `hf auth token` command

Wauplin:
(related to internal private [slack thread](https://huggingface.slack.com/archives/C07KX53FZTK/p1775677573964369))

Adds `hf auth token` to print the current token to stdout. Handy for piping into other commands.

```
$ hf auth token
hf_xxx***
Hint: Run `hf auth whoami` to see which account this token belongs to.
```

You can pipe this output to make a cURL call

```
$ hf auth token | xargs -I {} curl -H "Authorization: Bearer {}" https://huggingface.co/api/settings/billing/usage
```

If not logged in, prints an error nudging to `hf auth login` and exits 1.

```
$ hf auth token
Error: Not logged in. Run `hf auth login` first.
```

---

#### Note: the plan in a future PR is to provide a `hf api` command similar to `gh api` to make authenticated calls to any endpoint on hf.co/ (would need a maybe a `hf api schema` to fetch https://huggingface.co/spaces/huggingface/openapi + `hf api run` ?). In any case,  `hf auth token` will still be useful for instance to do `export ANTHROPIC_AUTH_TOKEN="$(hf auth token)"` in [hf-claude](https://github.com/hanouticelina/hf-claude) cc @hanouticelina

### DIFF
--- a/src/huggingface_hub/cli/auth.py
+++ b/src/huggingface_hub/cli/auth.py
@@ -156,6 +156,7 @@ def auth_token() -> None:
         out.error("Not logged in. Run `hf auth login` first.")
         raise typer.Exit(code=1)
     print(token)
+    out.hint("Run `hf auth whoami` to see which account this token belongs to.")
 
 
 @auth_cli.command("whoami", examples=["hf auth whoami", "hf auth whoami --format json"])
--- a/src/huggingface_hub/cli/auth.py
+++ b/src/huggingface_hub/cli/auth.py
@@ -148,6 +148,16 @@ def auth_list_cmd() -> None:
     auth_list()
 
 
+@auth_cli.command("token", examples=["hf auth token", "hf auth token | xargs curl -H 'Authorization: Bearer {}'"])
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/cli/auth.py
```diff
@@ -148,6 +148,17 @@ def auth_list_cmd() -> None:
...
 
 
+@auth_cli.command("token", examples=["hf auth token", "hf auth token | xargs curl -H 'Authorization: Bearer {}'"])
+def auth_token() -> None:
+    """Print the current access token to stdout."""
+    token = get_token()
+    if token is None:
+        out.error("Not logged in. Run `hf auth login` first.")
+        raise typer.Exit(code=1)
+    print(token)
```
Wauplin, line 158: I left `print` since we want to print only the token, no matter the output mode

hanouticelina (approved):
Looks good!

julien-c:
nice 👍

PR #4116, opened 2026-04-16
Use the new XetSession API

seanses:
## Summary

Migrates all xet upload/download code from the old function-based `hf_xet` API (`upload_files`, `download_files`, global `MULTITHREADED_RUNTIME`) to the new session-based `XetSession` API introduced in `hf-xet >= 1.5.0`.

**Upload:**
```python
session = get_xet_session()  # global singleton, fork- and SIGINT-safe

# Strip auth from custom_headers; keep full headers for token refresh
xet_headers = headers.copy()
xet_headers.pop("authorization", None)

with session.new_upload_commit(
    token_refresh_url=refresh_url,    # URL to refresh the xet write token
    token_refresh_headers=headers,    # full headers (incl. auth) for token refresh
    custom_headers=xet_headers,       # headers without auth for xet storage requests
    progress_callback=on_progress,    # optional; receives (group_report, item_reports)
) as commit:
    commit.start_upload_file(path, sha256=sha256_hex)   # from disk
    commit.start_upload_bytes(data, sha256=sha256_hex)  # from memory
# commit() is called automatically on clean exit of the with-block
```

**Download:**
```python
with session.new_file_download_group(
    token_refresh_url=xet_file_data.refresh_route,
    token_refresh_headers=headers,
    custom_headers=xet_headers,
    progress_callback=on_progress,
) as group:
    group.start_download_file(XetFileInfo(file_hash, size), dest_path)
# finish() is called automatically on clean exit of the with-block
```

**Architecture:**

- `utils/_xet.py` — `XetSessionHolder` is a lightweight wrapper around `XetSession` with three safety guarantees:
  - **SIGINT handling**: `sigint_abort()` cancels the active session and clears the reference so the next `get()` creates a fresh one (notebook-friendly — subsequent cells still work after Ctrl-C)
  - **Fork safety**: `get()` checks `os.getpid()` against the PID that created the session; on mismatch the inherited session is discarded and a fresh one is created for the child process (prevents deadlock when multiprocessing workers inherit a dead Tokio runtime)
  - **Thread safety**: a `threading.Lock` guards all state mutations, important for free-threaded Python (3.14t) where multiple threads can race on `get()` or `sigint_abort()` without the GIL serialising them
- `_GLOBAL_XET_HOLDER = XetSessionHolder()` is a module-level singleton (analogous to `get_session()` for HTTP). All upload/download call sites share the same session via `get_xet_session()` — repo commits, `hf_hub_download`, bucket uploads/downloads, and `snapshot_download` all reuse the same underlying Tokio runtime
- All blocking call sites are wrapped with `except KeyboardInterrupt: _GLOBAL_XET_HOLDER.sigint_abort(); raise` to abort the in-flight Rust operation and allow clean re-use of xet in the same process
- `xet_connection_info_refresh_url()` centralises token-refresh URL construction: handles the `bucket` repo-type (revision omitted) and the `/None` revision for PR write-token requests (where the final revision is not yet known)
- `_xet_progress_reporting.py` — progress callback receives `(group_report, item_reports)` matching the new API contract
- `setup.py` — minimum `hf-xet` version bumped to `>= 1.5.0`
- Tests mock `hf_xet.XetSession`; fork-safety unit and integration tests added to `test_xet_utils.py`

**Note on `upload_large_folder` and Ctrl-C:** xet work there runs inside background worker threads. Python only delivers `KeyboardInterrupt` to the main thread, and `PyErr_CheckSignals()` is a no-op in non-main threads — so Ctrl-C responsiveness for `upload_large_folder` requires a separate fix (shared stop event + worker cooperation) and is out of scope for this PR.

## TODO

- [ ] Refactor `upload_large_folder` to use the `XetSession` API directly: register each file to the upload commit object immediately after SHA-256 computation (no batching needed — xet handles concurrency internally), then call `commit()` once at the end. This removes the artificial batch size tuning (`UPLOAD_BATCH_SIZE_XET`) and makes the xet path in `upload_large_folder` consistent with the rest of the codebase.

---

> [!NOTE]
> **Medium Risk**
> Touches the primary large-file transfer path across commits, hub downloads, and buckets with a dependency floor bump; behavior changes around token refresh, KeyboardInterrupt, and forked workers warrant careful regression testing.
> 
> **Overview**
> Migrates **all Hub Xet uploads and downloads** from the legacy `hf_xet` helpers (`upload_files`, `upload_bytes`, `download_files`) to the **`XetSession` API** (`new_upload_commit`, `new_file_download_group`), with **`hf-xet` bumped to `>=1.5.1`**.
> 
> A shared **`get_xet_session()`** singleton in `utils/_xet.py` (fork-safe, thread-safe, **`abort_xet_session()` on Ctrl-C**) backs repo commits (`_upload_xet_files`), **`xet_get` / `hf_hub_download`**, and bucket batch upload/download. Auth is now **`token_refresh_url` + `token_refresh_headers`** instead of pre-fetching endpoints/tokens and passing `token_refresher` callbacks; **`xet_connection_info_refresh_url()`** centralizes refresh URL building.
> 
> **`XetProgressReporter`** follows the new `(group_report, item_reports)` progress contract; **`XetAuthorizationError` / `XetRefreshTokenError`** are removed. Tests mock `XetSession` and add fork-safety coverage for **`XetSessionHolder`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44aec7afd8aaf5d51605c096d4a8eb033c77ab2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_xet_utils.py
```diff
@@ -342,3 +346,88 @@ def test_xet_token_reset_after_repo_deletion() -> None:
...
+            finally:
+                os.close(w_fd)
+                os._exit(0)
+        else:
+            # ---- parent process ----
+            os.close(w_fd)
+            result = os.read(r_fd, 1)
+            os.close(r_fd)
+            os.waitpid(child_pid, 0)
+            assert result == b"\x01", "Child process reported fork-safety failure"
```
Wauplin:
```suggestion
                os.write(w_fd, b"SUCCESS" if ok else b"FAILURE")
            except Exception:
                os.write(w_fd, b"FAILURE")
            finally:
                os.close(w_fd)
                os._exit(0)
        else:
            # ---- parent process ----
            os.close(w_fd)
            result = os.read(r_fd, 1)
            os.close(r_fd)
            os.waitpid(child_pid, 0)
            assert result == b"SUCCESS"
```

(nit) for readability

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +251,45 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
+
+    def sigint_abort(self):
+        """Abort the current session and clear it so the next get() creates a fresh one."""
+        if self._session is not None:
+            try:
+                self._session.sigint_abort()
+            except Exception:
+                pass
+            self._session = None
+            self._session_pid = None
```
Wauplin: any chance this leads to a concurrency issue?

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +251,45 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
     """Check if the given XET connection info is expired."""
     return connection_info.expiration_unix_epoch <= int(time.time()) + XET_CONNECTION_INFO_SAFETY_PERIOD
+
+
+class XetSessionHolder:
+    """Holds an optional XetSession; supports safe re-creation after sigint_abort or fork."""
+
+    def __init__(self):
+        self._session = None
+        self._session_pid: int | None = None
```
Wauplin, line 304: _I think_ it'd be fine to have a global `XetSessionHolder` singleton (i.e. `XetSessionHolder()` always return the same instance). We already use a global session for HTTP requests (see [here](https://github.com/huggingface/huggingface_hub/blob/6e9e383812b818e0e8899c9359dc0055dadd792f/src/huggingface_hub/utils/_http.py#L335)) and it works fine. The advantage is that we wouldn't have to pass a `xet_session_holder` argument in intermediate methods like `_upload_xet_files`

Happy to get your thoughts

Wauplin: (having 1 instance per fork is still very much relevant)

Wauplin: if we have a global singleton, I wouldn't define a `XetSessionHolder` class (or at least not use it) and define a `get_xet_session` method instead (same as the current `get_session` for httpx Client)

src/huggingface_hub/_commit_api.py
```diff
@@ -434,6 +433,7 @@ def _upload_files(
             endpoint=endpoint,
             revision=revision,
             create_pr=create_pr,
+            xet_session_holder=xet_session_holder,
```
Wauplin: the above comment would avoid these additions

src/huggingface_hub/_commit_api.py
```diff
@@ -588,92 +589,48 @@ def _upload_xet_files(
...
+    xet_headers.pop("authorization", None)
 
     if not are_progress_bars_disabled():
         progress = XetProgressReporter()
         progress_callback = progress.update_progress
     else:
         progress, progress_callback = None, None
 
+    holder = xet_session_holder if xet_session_holder is not None else XetSessionHolder()
+    session = holder.get()
```
Wauplin: would be simplified to

```suggestion
    session = get_xet_session()
```

src/huggingface_hub/_commit_api.py
```diff
@@ -588,92 +589,48 @@ def _upload_xet_files(
...
+    builder = session.new_upload_commit().with_token_refresh_url(refresh_url, headers).with_custom_headers(xet_headers)
```
Wauplin: does that change the initial session in-place or does it return a new object each time `with_..._...` is called?

src/huggingface_hub/_commit_api.py
```diff
@@ -588,92 +589,48 @@ def _upload_xet_files(
...
-                access_token_info,
-                token_refresher,
-                progress_callback,
-                repo_type,
-                request_headers=xet_headers,
-                sha256s=all_sha256s,
-            )
-
+        for op in all_paths_ops:
+            commit.upload_file(str(op.path_or_fileobj), sha256=Sha256Policy.provided(op.upload_info.sha256.hex()))
```
Wauplin: `sha256=Sha256Policy.provided(op.upload_info.sha256.hex())` doesn't feel very Pythonic but perfectly ok if it's easier on the Rust side

src/huggingface_hub/file_download.py
```diff
@@ -547,22 +533,35 @@ def token_refresher() -> tuple[str, int]:
...
+        group = (
+            session.new_file_download_group()
+            .with_token_refresh_url(xet_file_data.refresh_route, headers)
+            .with_custom_headers(xet_headers)
+            .with_progress_callback(_on_progress)
+            .build()
         )
+        group.download_file(XetFileInfo(xet_file_data.file_hash, expected_size), str(incomplete_path.absolute()))
+        try:
+            group.finish()
```
Wauplin: (nit) when is the file downloaded? is it when `group.download_file` is called or when `group.finish` completes? Asking because given the implementation it looks like `.download_file` doesn't download the file -or at least not completely- but from the API naming I would expect the opposite.

If the download happens in the background, what do you think of an API like this:

```py
group = session.new_file_download_group().with...()....build()

group.start_download_file(...)
try:
    group.join()
except:
    ...
```

I find the `start_download_file` / `join` naming more explicit over what's going on under the hood (at least if I understood correctly). Took the suggestion from the `threading` / `multiprocessing` syntax.

src/huggingface_hub/hf_api.py
```diff
@@ -13188,44 +13172,36 @@ def token_refresher() -> tuple[str, int]:
...
+                    (commit.upload_file(str(op.source), sha256=Sha256Policy.skip()), op)
+                    for op in add_path_operations
+                    if op.xet_hash is None
+                ]
+                bytes_handles = [
+                    (commit.upload_bytes(op.source, sha256=Sha256Policy.skip()), op)
+                    for op in add_bytes_operations
+                    if op.xet_hash is None
+                ]
+                commit.commit()
```
Wauplin: is `commit.commit()` and `group.finish()` conceptually more or less the same (i.e. "wait for all the ongoing stuff to complete")? in that case I'd try to make the naming more consistent

Wauplin (commented):
Thanks for working on that @seanses! I've left some high level comments on the API design. Overall a very nice addition!

src/huggingface_hub/utils/_xet.py
seanses: Global `XetSessionHolder()` function returning the singleton makes sense to me. The `XetSessionHolder` exists mainly to deal with Ctrl-C and fork where the XetSession instance can't be reused, and it seems to me managing all these inside a class is easier. But I'm not a Python expert so I'll research if there's a more Pythonic way of doing that.

src/huggingface_hub/_commit_api.py
seanses: It changes the `PyXetUploadCommitBuilder` instance (returned from `session.new_upload_commit()`) each time `with_...` is called. But this is a direct translation of the Rust style API, and I learned that it's more Pythonic to use options in parameter list, e.g. `session.new_upload_commit(token_refresh_url = "...", custom_headers = "...")`, so will change to that format.

src/huggingface_hub/_commit_api.py
seanses: I can totally make it more Pythonic! The difficulty I encountered was that enum `Sha256Policy` has three variants: `Provided(sha256)`, `Compute`, and `Skip`, so it's not clear to define a mapping from a `sha256: str` or `sha256: str | None` to this enum. Any suggestions?

src/huggingface_hub/file_download.py
seanses: Yes `download_file` only spawns the task and it's executed in the background. `group.finish()` waits for background tasks to complete.

I like the naming!

src/huggingface_hub/hf_api.py
seanses: Yes both upload and download works on the same conceptual model: spawns background tasks and wait for completion. Good suggestion!

src/huggingface_hub/utils/_xet.py
seanses: Good catch! Although "sigint_abort" is only called after "KeyboardInterrupt" signal and this signal is only raised in the main thread, "sigint_abort()" and "get()" can interleave as "get()" can happen in a worker thread. Will add a threading lock.

src/huggingface_hub/utils/_xet.py
Wauplin: Holding it in a class is fine yes if it makes the separation and testing easier. I'm just suggesting that on top of it we have a global `get_xet_session` similar to `XetSessionHolder().get()` (but global) so that we never have to pass down the `XetSessionHolder` instance

src/huggingface_hub/_commit_api.py
Wauplin: Oh I see. Currently to distinguish both with have a `sha256s` parameter in `upload_files` to pass pre-computed hashes and another `skip_sha256` if it's fine to skip hash computation. Both parameters are mutually exclusive and if none are provided, it means "please compute the sha256". Not very self-explanatory so agree to harmonize into a single `sha256` parameter now that we are redefining the API.

What do you think of having special values for `skip` and `compute` that doesn't need instantiation (they are raw objects) while accepting raw string for precomputed sha256. On the `hf-xet` it's a bit more verbose / hacky but from the `huggingface_hub`/client side it feels more Pythonic IMO. (a bit like what `dataclasses.MISSING` does for instance)

This is what it would look like on `hf-xet` side
```py
from typing import Final

class _ComputeSha256Type:  # not exposed
    def __repr__(self) -> str:
        return "COMPUTE_SHA256"

class _SkipSha256Type:  # not exposed
    def __repr__(self) -> str:
        return "SKIP_SHA256"


COMPUTE_SHA256: Final = _ComputeSha256Type() # exposed
SKIP_SHA256: Final = _SkipSha256Type() # exposed


def upload_file(path: str, sha256: str | _ComputeSha256Type | _SkipSha256Type):
    ...
```

and on `huggingface_hub` side:

```py
from hf_xet import COMPUTE_SHA256, SKIP_SHA256

upload_file("test.txt", sha256="f2358d9a0c429bc940cf63badf1e32ca8851d2e7e2353311c6a42299952f3dad")
upload_file("test.txt", sha256=COMPUTE_SHA256)
upload_file("test.txt", sha256=SKIP_SHA256)
```

Since the usual case is to provide the sha256 (for models/datasets/spaces), then no need to import a `Sha256Policy` class with classmethods.

Wauplin: another alternative would be to accept `sha256="compute"` and `sha256="skip"` in addition to `sha256="<real_sha256>"` but it feels more hacky / less explicit

Wauplin: Let me know what you think. I'm open to suggestion or even keeping your current solution if you prefer

seanses: > upload_file("test.txt", sha256="f2358d9a0c429bc940cf63badf1e32ca8851d2e7e2353311c6a42299952f3dad")
upload_file("test.txt", sha256=COMPUTE_SHA256)
upload_file("test.txt", sha256=SKIP_SHA256)

I like this style! I think I can make it happen

setup.py
```diff
@@ -12,7 +12,7 @@ def get_version() -> str:
 
 
 # hf-xet version used in both install_requires and extras["hf_xet"]
-HF_XET_VERSION = "hf-xet>=1.4.3,<2.0.0"
+HF_XET_VERSION = "hf-xet>=1.5.0.dev3,<2.0.0"
```
Wauplin: Let's not merge this PR until `v1.5.0` is officially released.

src/huggingface_hub/utils/__init__.py
```diff
@@ -118,8 +117,10 @@
 from ._xet import (
     XetConnectionInfo,
     XetFileData,
+    XetSessionHolder,
     XetTokenType,
     fetch_xet_connection_info_from_repo_info,
+    get_xet_session,
```
Wauplin:
```suggestion
XetTokenType,
fetch_xet_connection_info_from_repo_info,
```

for now let's not make them public. We can always add them later if a downstream library requires a direct access to the xet session but I doubt.

(you always import them with `from .utils._xet import ...` in other modules so it shouldn't break anything)

src/huggingface_hub/hf_api.py
```diff
@@ -13226,36 +13147,9 @@ def _batch_bucket_files(
...
-                    token_type=XetTokenType.WRITE,
-                    repo_id=bucket_id,
-                    repo_type="bucket",
-                    headers=headers,
-                    endpoint=self.endpoint,
-                )
-                if new_xet_connection is None:
-                    raise XetRefreshTokenError("Failed to refresh xet token")
-                return new_xet_connection.access_token, new_xet_connection.expiration_unix_epoch
+            refresh_url = f"{self.endpoint}/api/buckets/{bucket_id}/xet-write-token"
```
Wauplin: is this change intentional? (i.e. semi-hardcoding the URL) If not let's replace it with something like 

```py
refresh_url = xet_connection_info_refresh_url(
    token_type=XetTokenType.WRITE,
    repo_id=bucket_id,
    repo_type="bucket",
    endpoint=self.endpoint,
)
```

src/huggingface_hub/utils/_xet.py
```diff
@@ -119,6 +121,43 @@ def refresh_xet_connection_info(
...
+) -> str:
+    """
+    Build the URL used to fetch or refresh a Xet access token for a given repo.
+    Args:
+        token_type (`XetTokenType`):
+            Type of the token to request: `"read"` or `"write"`.
+        repo_id (`str`):
+            A namespace (user or an organization) and a repo name separated by a `/`.
+        repo_type (`str`):
+            Type of the repo to upload to: `"model"`, `"dataset"` or `"space"`.
```
Wauplin:
```suggestion
            Type of the repo to upload to.
```

(handles the new "kernels" and "buckets" as well)

src/huggingface_hub/utils/_xet_progress_reporting.py
Wauplin, line 1: (small refacto seems to work well in my local examples but I have to admit I did not carefully reviewed it - looks fine)

src/huggingface_hub/_commit_api.py
```diff
@@ -637,43 +614,23 @@ def token_refresher() -> tuple[str, int]:
...
+            token_refresh_headers=headers,
+            custom_headers=xet_headers,
+            progress_callback=progress_callback,
+        ) as commit:
+            for op in all_paths_ops:
+                commit.start_upload_file(str(op.path_or_fileobj), sha256=op.upload_info.sha256.hex())
+            for op in all_bytes_ops:
+                commit.start_upload_bytes(op.path_or_fileobj, sha256=op.upload_info.sha256.hex())
+    except KeyboardInterrupt:
+        _GLOBAL_XET_HOLDER.sigint_abort()
```
Wauplin: since we have `get_xet_session`, I'm not a fan of having to import the low-level `_GLOBAL_XET_HOLDER` outside of the `_xet` module. Could you define a new utility like `abort_xet_session` which would be an alias for `_GLOBAL_XET_HOLDER.sigint_abort()`? This way we kinda centralize the logic around `_GLOBAL_XET_HOLDER`

src/huggingface_hub/hf_api.py
```diff
@@ -13487,73 +13373,61 @@ def download_bucket_files(
...
+                with session.new_file_download_group(
+                    token_refresh_url=metadata.xet_file_data.refresh_route,
+                    token_refresh_headers=headers,
+                    custom_headers=xet_headers,
+                    progress_callback=_on_progress,
+                ) as group:
+                    for xet_info, dest in non_zero_download_items:
+                        group.start_download_file(xet_info, dest)
+            except KeyboardInterrupt:
+                _GLOBAL_XET_HOLDER.sigint_abort()
```
Wauplin: same here

Wauplin (commented):
This looks very solid now! Extensively tested it locally (mainly the CLI) and works as smoothly as before 👍

Wauplin:
I've asked Claude code to review the PR and it commented a few things. Nothing critical IMO

---

> ### 3. `XetAuthorizationError` / `XetRefreshTokenError` are now dead code
> 
> **Files:** `src/huggingface_hub/errors.py:453-457`, removed imports in `_commit_api.py` and `hf_api.py`
> 
> The old code raised `XetAuthorizationError` on 401 responses and `XetRefreshTokenError` on failed token refresh. The new session-based API delegates auth to hf_xet (which handles token refresh internally via the URL), so these exceptions are no longer raised anywhere.
> 
> They're still defined in `errors.py` but never imported or used. They're not in `__init__.py` so they're not public API.
> 

**Suggestion:** Delete `XetAuthorizationError`, `XetRefreshTokenError`, and possibly `XetError` from `errors.py` if nothing else uses them, or add a TODO comment. Dead error classes are confusing for future readers.

### My opinion (Wauplin): fine to remove them and mention the breaking change in the PR description (it's a breaking change but I don't expect downstream users to have actually raised from it)

---

> ### 5. No unit tests for `xet_connection_info_refresh_url`
> 
> This is a new public-ish function with non-trivial logic (bucket vs git-based repo types, the `"/None"` revision for PR tokens). There are no dedicated tests for it. A few parametrized test cases would be cheap and valuable:
> 
> - `repo_type="model"`, `revision="main"` → `.../models/{id}/xet-write-token/main`
> - `repo_type="model"`, `revision=None` → `.../models/{id}/xet-write-token/None`
> - `repo_type="bucket"`, `revision=None` → `.../buckets/{id}/xet-write-token` (no revision)
> - `repo_type="bucket"`, `revision="some-rev"` → `.../buckets/{id}/xet-write-token/some-rev`


### My opinion (Wauplin): quick and easy to test so let's do it (even if not introduced in this PR)

---

> ### 6. `progress.close(False)` always passes `False` — is `_success` meaningful?
> 
> **Files:** `_commit_api.py:632`, `hf_api.py:13193`
> 
> `XetProgressReporter.close(_success)` is always called with `False` (in the `finally` block), even on successful completion. The `_success` parameter is currently unused in `close()` — it just closes bars regardless. If `_success` is never meaningful, the parameter should be removed. If it's intended for future use, the call site should pass the correct value.
> 
> This existed before this PR but worth flagging since the progress reporter was rewritten.
> 

### My opinion (Wauplin): let's keep it as-is except if you really care about it^^

---

> ### 8. Header stripping pattern repeated
> 
> The `xet_headers = headers.copy(); xet_headers.pop("authorization", None)` pattern appears in 4 call sites:
> - `_commit_api.py:602-603`
> - `file_download.py:537-538`
> - `hf_api.py:13151-13152`
> - `hf_api.py:13397-13398`
> 
> Consider a small helper like `_strip_auth_headers(headers)` if you want, but it's fine as-is — it's 2 lines and very clear.

### My opinion (Wauplin): why not but low prio

src/huggingface_hub/utils/__init__.py
seanses: dropped the export

src/huggingface_hub/_commit_api.py
seanses: added `abort_xet_session`

tests/test_xet_utils.py
```diff
@@ -342,3 +347,124 @@ def test_xet_token_reset_after_repo_deletion() -> None:
...
+            "/api/buckets/user/mybucket/xet-write-token",
+        ),
+        (
+            {
+                "token_type": XetTokenType.WRITE,
+                "repo_id": "user/mybucket",
+                "repo_type": "bucket",
+                "revision": "some-rev",
+            },
+            "/api/buckets/user/mybucket/xet-write-token/some-rev",
```
seanses, line 459: Is this expected? The `xet_connection_info_refresh_url` function logic was entirely extracted from the previous `fetch_xet_connection_info_from_repo_info` function, and the logic implies this. But the comment `# On "bucket" repo type, the revision never needed => don't use it` suggests otherwise. @Wauplin

setup.py
seanses: Updated to `hf-xet>=1.5.0,<2.0.0`

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +288,86 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
+
+def xet_headers_without_auth(headers: dict[str, str]) -> dict[str, str]:
+    """Return a copy of headers with the authorization header removed.
+
+    Xet storage requests use a short-lived xet access token for auth, so the
+    Hub authorization header must not be forwarded to xet storage endpoints.
+    """
+    xet_headers = headers.copy()
+    xet_headers.pop("authorization", None)
+    return xet_headers
```
hanouticelina: (nit)
```suggestion
    return {key: value for key, value in headers.items() if key.lower() != "authorization"}
```

src/huggingface_hub/hf_api.py
```diff
@@ -13308,88 +13314,56 @@ def _batch_bucket_files(
...
+                    ]
+                    bytes_handles = [
+                        (commit.start_upload_bytes(op.source, sha256=SKIP_SHA256), op)
+                        for op in add_bytes_operations
+                        if op.xet_hash is None
+                    ]
+                for handle, op in path_handles + bytes_handles:
+                    result = handle.result()
+                    op.xet_hash = result.xet_info.hash
+                    op.size = result.xet_info.file_size
```
hanouticelina: (nit) for readability
```suggestion
                    handles = []
                   for op in add_path_operations:
                       if op.xet_hash is None:
                           handles.append((commit.start_upload_file(str(op.source), sha256=SKIP_SHA256), op))
                   for op in add_bytes_operations:
                       if op.xet_hash is None:
                           handles.append((commit.start_upload_bytes(op.source, sha256=SKIP_SHA256), op))
               for handle, op in handles:
                   result = handle.result()
                   op.xet_hash = result.xet_info.hash
                   op.size = result.xet_info.file_size
```

src/huggingface_hub/hf_api.py
```diff
@@ -13308,88 +13314,56 @@ def _batch_bucket_files(
...
+            except KeyboardInterrupt:
+                abort_xet_session()
+                raise
             finally:
                 if owns_progress and progress is not None:
-                    progress.close(False)
+                    progress.close()
 
         # 3. /batch call
```
```suggestion
```

src/huggingface_hub/utils/_xet_progress_reporting.py
```diff
@@ -42,12 +40,12 @@ def __init__(self, n_lines: int = 10, description_width: int = 30, total_files:
...
-        # Offsets to accumulate totals across multiple upload_files/upload_bytes calls
-        self._total_bytes_offset = 0
-        self._total_transfer_bytes_offset = 0
+        # Track previous absolute values to compute increments
+        self._prev_bytes_completed: int = 0
+        self._prev_transfer_bytes_completed: int = 0
 
         # Item bars (scrolling view)
-        self.item_state: OrderedDict[str, PyItemProgressUpdate] = OrderedDict()
+        self.item_state: OrderedDict[str, object] = OrderedDict()
```
hanouticelina:
```suggestion
        self.item_state: OrderedDict[str, Any] = OrderedDict()
```

src/huggingface_hub/hf_api.py
```diff
@@ -13569,73 +13545,60 @@ def download_bucket_files(
...
+        metadata = self.get_bucket_file_metadata(bucket_id, remote_path, token=token)
+
+        xet_headers = xet_headers_without_auth(headers)
+
         # Download files
         progress_cm = _get_progress_bar_context(
             desc="Downloading bucket files",
             log_level=logger.getEffectiveLevel(),
-            total=sum(info.file_size for info in non_zero_download_infos),
+            total=sum(xet_info.file_size or 0 for xet_info, _ in non_zero_download_items),
```
hanouticelina: since `non_zero_download_items` is built via `if remote_file.size == 0: ...`
```suggestion
            total=sum(xet_info.file_size for xet_info, _ in non_zero_download_items),
```

src/huggingface_hub/file_download.py
```diff
@@ -547,22 +531,34 @@ def token_refresher() -> tuple[str, int]:
...
+    session = get_xet_session()
 
     with progress_cm as progress:
+        _prev = 0
 
-        def progress_updater(progress_bytes: float):
-            progress.update(progress_bytes)
+        def _on_progress(group_report, _):
+            nonlocal _prev
+            current = group_report.total_bytes_completed
```
hanouticelina, line 560: I tested the download path locally and the progress bar stays at 0.00/Total throughout and it finishes but never animates:
```bash
$ hf download celinah/xet-pr-4116-test big.bin --force-download --local-dir /tmp/dl2
big.bin:   0%|                                                      | 0.00/105M [00:02<?, ?B/s]
✓ Downloaded
  path: /tmp/dl2/big.bin
```
added some debug prints and from what I see, we're reading the wrong field. `total_bytes_completed` stays at 0 the whole time, it's `total_transfer_bytes_completed` that actually tracks bytes received. let me know if I'm missing something here!
```suggestion
            current = group_report.total_transfer_bytes_completed
```

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +288,86 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
+            if self._session is not None:
+                try:
+                    self._session.sigint_abort()
+                except Exception:
+                    pass
+                self._session = None
+                self._session_pid = None
+
+
+_GLOBAL_XET_HOLDER = XetSessionHolder()
```
hanouticelina, line 342: I got the following Python error, I put Claude Code to investigate 
```bash
> hf download celinah/xet-pr-4116-test big.bin --force-download --local-dir /tmp/dl2
big.bin:   0%|                                                      | 0.00/105M [00:01<?, ?B/s]
✓ Downloaded
  path: /tmp/dl2/big.bin
Fatal Python error: _PyImport_Init: global import state already initialized
Python runtime state: preinitialized

Thread 0x0000000201e23100 (most recent call first):
  <no Python frame>
```
and here is what it says:
> It's the new global _GLOBAL_XET_HOLDER — the XetSession's Tokio runtime stays alive for the
>   whole process lifetime and there is no atexit cleanup. So at interpreter shutdown:
> 
>   1. Python starts finalizing.
>   2. A Tokio worker (or the per-group progress thread that hasn't quite broken yet) tries
>   Python::attach(...) to call back into Python.
>   3. The interpreter is mid-teardown → "global import state already initialized" / <no Python
>   frame>.
> 
>   The "no Python frame" line and the "Thread 0x..." prefix confirm it's a Rust-spawned thread
>   doing this, not Python code.

so basically, a fix would be to register `_GLOBAL_XET_HOLDER.sigint_abort` to run automatically when the Python interpreter is shutting down:

```suggestion
_GLOBAL_XET_HOLDER = XetSessionHolder()
atexit.register(_GLOBAL_XET_HOLDER.sigint_abort) # import atexit at the top of the file
```

src/huggingface_hub/hf_api.py
```diff
@@ -13569,73 +13545,60 @@ def download_bucket_files(
...
 
+        session = get_xet_session()
+
         with progress_cm as progress:
+            _prev = [0]
 
-            def progress_updater(progress_bytes: float):
-                progress.update(progress_bytes)
+            def _on_progress(group_report, _):
+                current = group_report.total_bytes_completed
```
hanouticelina, line 13960: same bug as in https://github.com/huggingface/huggingface_hub/pull/4116/changes#r3272436825
```bash
> hf buckets cp hf://buckets/celinah/xet-pr-4116-bucket/single.bin /tmp/bdl/
Downloading bucket files:   0%|                                    | 0.00/52.4M [00:03<?, ?B/s]
✓ Downloaded
  src: hf://buckets/celinah/xet-pr-4116-bucket/single.bin
  dst: /tmp/bdl/single.bin
```
```suggestion
                current = group_report.total_transfer_bytes_completed
```

hanouticelina (commented):
thanks a lot @seanses for working on this! and very sorry for the delay. I left a couple of comments and tested the PR as well (with with `hf_xet==1.5.0`), 
let me know if I'm missing something.

src/huggingface_hub/file_download.py
seanses: Both `total_bytes_completed` and `total_transfer_bytes_completed` are updated by hf-xet and they have different meaning:
`total_bytes_completed` represents number of bytes written to the file on disk or stream download bytes array, and is only updated after the writing is done;
`total_transfer_bytes_completed` represents number of bytes we get from the network, before decompression and not double counting if a segment of bytes are reused to construct two different parts of a file.

I think what you observed was due to the combination factor that we use a large write out buffer and that the progress is updated by hf-xet every 100 ms. For a 105 M file, it is very likely the bytes are buffered so you didn't see the progress bar grow and the final report was not delivered because the buffered bytes are flushed within 100 ms. If you try on an even larger file you'll see the progress bar moves.

With that being said, I think we can address this UX issue. Using `total_transfer_bytes_completed` in replace of `total_bytes_completed` is not ideal because the total number is going to be less than the actual file size. But what do you think if we plot two bars, one for `total_bytes_completed` and another for `total_transfer_bytes_completed` so users know things are moving?

src/huggingface_hub/utils/_xet.py
seanses: Actually `atexit.register(_GLOBAL_XET_HOLDER.sigint_abort)` will not fix this issue. The `XetSession::sigint_abort()` function aborts the tokio runtime managed internally, but the per-group progress thread is spawned off std::thread, not managed by the tokio runtime. So I will change the `Python::attach` to `Python::try_attach` so it will not attach if the interpreter is in a state where it cannot be attached to.

seanses: Fixed in https://github.com/huggingface/xet-core/pull/856, also released in `hf-xet 1.5.1.dev1`

src/huggingface_hub/file_download.py
seanses: Also, https://github.com/huggingface/xet-core/pull/856 should fix the problem that the progress bar didn't reach 100%.

src/huggingface_hub/hf_api.py
seanses: See replies under https://github.com/huggingface/huggingface_hub/pull/4116#discussion_r3272436825

src/huggingface_hub/file_download.py
```diff
@@ -560,22 +544,34 @@ def token_refresher() -> tuple[str, int]:
...
```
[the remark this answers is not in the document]
seanses: In discussion

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +288,84 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
+
+            if self._session is not None and self._session_pid != current_pid:
+                # Fork detected. Discard the parent's session; the Rust Drop will
+                # call discard_runtime() (std::mem::forget) rather than the normal
+                # shutdown path, so this returns immediately without blocking.
+                self._session = None
+                self._session_pid = None
+
+            if self._session is None:
+                from hf_xet import XetSession
```
hanouticelina, line 323: let's put the import at the top of the file

src/huggingface_hub/utils/_xet.py
```diff
@@ -250,3 +288,84 @@ def _cache_key(url: str, headers: dict[str, str], params: dict[str, str] | None,
...
+        """
+        with self._lock:
+            current_pid = os.getpid()
+
+            if self._session is not None and self._session_pid != current_pid:
+                # Fork detected. Discard the parent's session; the Rust Drop will
+                # call discard_runtime() (std::mem::forget) rather than the normal
+                # shutdown path, so this returns immediately without blocking.
+                self._session = None
+                self._session_pid = None
```
hanouticelina: it's overwritten line 327
```suggestion
```

src/huggingface_hub/hf_api.py
```diff
@@ -13936,73 +13906,60 @@ def download_bucket_files(
...
+                progress.update(max(0, current - _prev[0]))
+                _prev[0] = current
```
```suggestion
            prev = 0

            def _on_progress(group_report, _):
                nonlocal prev
                current = group_report.total_bytes_completed
                progress.update(max(0, current - _prev[0]))
                prev = current
```

hanouticelina (approved):
Looks good to me, I tested the PR against a locally built wheel from xet-core PR https://github.com/huggingface/xet-core/pull/856 and all good on my side, thanks a lot @seanses for working on this!
I'll let @Wauplin take a look and give the green light

src/huggingface_hub/utils/_xet.py
Wauplin: hmm, `hf_xet` is still an optional dependency so let's continue to lazy-import it I think

setup.py
```diff
@@ -12,7 +12,7 @@ def get_version() -> str:
 
 
 # hf-xet version used in both install_requires and extras["hf_xet"]
-HF_XET_VERSION = "hf-xet>=1.4.3,<2.0.0"
+HF_XET_VERSION = "hf-xet>=1.5.1.dev1,<2.0.0"
```
Wauplin: don't forget to bump this to a stable version before merging this PR

src/huggingface_hub/utils/_xet.py
hanouticelina: yes 🤦‍♀️

src/huggingface_hub/file_download.py
Wauplin: tested this on a rather slow connection (5-10 MB/s) and also witnessed that the progress bar got almost never updated. It reaches 100% though. 

See example to download a single 500MB file using this PR + `hf_xet==1.5.1.dev1`:

[Screencast from 01-06-2026 15:18:11.webm](https://github.com/user-attachments/assets/d8590c30-566b-46c4-938d-c26c04c538d4)

But I also tested on `main` + `hf_xet==1.5.0` and behavior is practically the same:

[Screencast from 01-06-2026 15:26:31.webm](https://github.com/user-attachments/assets/2aaaf79b-afa5-4326-ac63-34fb57913d23)

I think it'd be great to have better reporting but I assume this is unrelated to this PR. 

Regarding the thread above:

> I think what you observed was due to the combination factor that we use a large write out buffer and that the progress is updated by hf-xet every 100 ms.

I'm not sure to understand this 100ms refresh rate. Is it that the progress bar gets updated every 100ms but only if a full chunk has been downloaded + reconstructed? in that case it explains why on a slow connection I don't see anything happening for a long time even though the refresh rate is every 100ms.

Wauplin: > With that being said, I think we can address this UX issue. Using total_transfer_bytes_completed in replace of total_bytes_completed is not ideal because the total number is going to be less than the actual file size. But what do you think if we plot two bars, one for total_bytes_completed and another for total_transfer_bytes_completed so users know things are moving?

I like the idea of having a "download" progress bar alongside a "reconstruction" one. This would be explicit to the user while improving UX on slower connection where chunks can take a few dozens seconds to download.

Note that we currently have 2 different reporting process depending on whether we are downloading 1 or several files at a time: 

image


When downloading a single file maybe a good UX could be:

```
model.safetensors:
    downloading bytes   0%|                              | 0.00  [00:01<?, ?B/s]
    reconstructing file 0%|                              | 0.00/548M [00:01<?, ?B/s]
✓ Downloaded
  path: /home/wauplin/.cache/huggingface/hub/models--gpt2/snapshots/607a30d783dfa663caf39e06633721c8d4cfcd7e/model.safetensors
```

And for multiple files:

```
Downloading bytes:   0%|                           | 467k [00:01<1:34:29, 684kB/s]
Reconstructing (incomplete total...):   0%|                         | 467k/3.88G [00:01<1:34:29, 684kB/s]
Fetching 26 files:   4%|████▍                                  | 1/26 [00:00<00:09,  2.71it/s]
```

The "downloading bytes" bar wouldn't need a "total" part as it's a bit meaningless (depends a lot on compression + in-file duplication). The only purpose of the bar is to let the user know that "something is happening" while the reconstruction bar is blocked

WDYT  @seanses @hanouticelina ? Does that make any sense? In any case, I'd rather address this part in a follow-up PR instead to avoid postponing this one (+it's kinda unrelated)

Wauplin (commented):
Made another pass on the PR. I think we are very close to getting this merged (once `hf_xet==1.5.1` gets released).

About the progress bars I like the idea of reporting both the "download" and "reconstruction" progress for better UX. Left 2 comments about it but I think it can be addressed in a separate PR. Let me know what you think

src/huggingface_hub/file_download.py
hanouticelina: ^ yes agree! let's address this in a follow-up PR if that's okay with you @seanses

seanses: > Is it that the progress bar gets updated every 100ms but only if a full chunk has been downloaded + reconstructed?

A background thread pulls metrics from a low level progress tracker every 100ms, but the low level progress tracker only updates `total_bytes_completed` after a buffer of chunks is flushed out to disk.

seanses: Sounds good let's add that bar in a separate PR!

Wauplin (approved):
Pre-approving, only remaining item is bumping `hf_xet` version in `setup.py` 🎉

Wauplin:
:tada:

PR #4122, opened 2026-04-17
Support hf-mount

lhoestq:
Support using a mounted HF Storage Bucket as cache.

It was failing because of file locks:
- flock is not available on mounted bucket so we need to use SoftFileLock
- I simply detect when flock is not supported and fallback on SoftFileLock in this case

The fix is different from the one in https://github.com/huggingface/huggingface_hub/pull/2402 but I'm not sure if it's because of recent `filelock` or if it's specific to mounted buckets. The differences are:
- I detect that flock is not supported from the `OSError.errno`
- and I also had to use a secondary file since we can't know if a file is a failed UnixFileLock or a locked SoftFileLock.

Alternatively we could try to make hf-mount work with the existing locking mechanism.
There is already a fallback to SoftFileLock in `filelock` if the flock opetation returns ENOSYS, this could be worth trying.
Though there is the same risk that maybe `filelock` can't know if a file is a failed UnixFileLock or a locked SoftFileLock and doesn't work properly.

Marking the PR as Draft while we investigate a bit

### DIFF
--- a/src/huggingface_hub/utils/_fixes.py
+++ b/src/huggingface_hub/utils/_fixes.py
@@ -5,6 +5,7 @@ import stat
 import tempfile
 import time
 from collections.abc import Callable, Generator
+from errno import ENOTSUP
 from functools import partial
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


julien-c:
Ah yes i wanted to store my hf cache inside a mounted Bucket!!! Yay!!

lhoestq:
Closing in favor of https://github.com/huggingface/hf-mount/pull/106 which brings native `filelock` support in hf-mount :)

PR #4130, opened 2026-04-21
[Buckets] Add search param to list_buckets

alexpouliquen:
Add support for buckets list search params.

_100% vibecoded, but should be fine_

---

> [!NOTE]
> **Low Risk**
> Small, additive change that only affects bucket-listing queries and CLI flag parsing; main risk is behavior differences in list results when `--search`/`search` is used.
> 
> **Overview**
> Adds a `search` filter for bucket listing in both Python and CLI.
> 
> `HfApi.list_buckets(..., search=...)` now forwards a `search` query param to the Hub, and `hf buckets list --search ...` exposes it with validation to reject use when listing files plus an updated empty-results message. Documentation and CLI reference are updated and new CLI tests cover the flag and error cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a647a9be54c380b26318ae07572360ec2f06f180. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_buckets_cli.py
+++ b/tests/test_buckets_cli.py
@@ -506,6 +506,13 @@ def test_bucket_list_error_recursive_with_namespace():
     assert "Cannot use --recursive when listing buckets" in result.output
 
 
+def test_bucket_list_error_search_with_files(tree_bucket: str):
+    """Cannot use --search when listing files."""
+    result = cli(f"hf buckets list {tree_bucket} --search foo")
+    assert result.exit_code != 0
+    assert "Cannot use --search when listing files" in result.output
+
+
 # =============================================================================
 # List files
 # =============================================================================
--- a/tests/test_buckets_cli.py
+++ b/tests/test_buckets_cli.py
@@ -506,28 +506,23 @@ def test_bucket_list_error_recursive_with_namespace():
     assert "Cannot use --recursive when listing buckets" in result.output
 
 
-def test_bucket_list_error_search_with_files(tree_bucket: str):
+def test_bucket_list_search_error_with_files(tree_bucket: str):
     """Cannot use --search when listing files."""
     result = cli(f"hf buckets list {tree_bucket} --search foo")
     assert result.exit_code != 0
     assert "Cannot use --search when listing files" in result.output
 
 
-def test_bucket_list_empty_search_mentions_filter(monkeypatch: pytest.MonkeyPatch):
-    class DummyApi:
-        def list_buckets(self, namespace: str | None, search: str | None):
-            assert namespace == USER
-            assert search == "does-not-exist-1234567890"
-            return []
+def test_bucket_list_search_term_exists(bucket_read: str, bucket_write: str):
+    search_term = bucket_read.split("/")[1].split("-")[1]  # e.g. 9447e4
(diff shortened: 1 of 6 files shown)


julien-c (commented):
nice 👍

Pierrci:
(internal PR: https://github.com/huggingface-internal/moon-landing/pull/17838)

tests/test_buckets_cli.py
```diff
@@ -506,6 +506,30 @@ def test_bucket_list_error_recursive_with_namespace():
...
+        def whoami(self):
+            return {"name": USER}
+
+    monkeypatch.setattr("huggingface_hub.cli.buckets.get_hf_api", lambda token=None: DummyApi())
+
+    result = cli(f"hf buckets list {USER} --search does-not-exist-1234567890")
+    assert result.exit_code == 0
+    assert f"No buckets found under namespace '{USER}' matching search 'does-not-exist-1234567890'." in result.output
+
+
```
Wauplin: Hey @alexpouliquen , I refactored the tests with a few recommendations for next PRs:

- usually no need to test all error cases, especially on CLI stuff. Here I kept the `test_bucket_list_search_error_with_files` test to check that `--search` and file mode are not compatible but honestly not very important. Agents tend to over-test things but I feel that it just makes the tests unreadable
- added 2 separate tests, one testing when the search returns a result, one when the search is empty. No need for more than that
- always prefer real tests on our staging infra rather than mocked APIs. Since `huggingface_hub` is the main Hub client, we use the tests both to ensure that the client is valid but also that the Hub doesn't break. It happened several times in the past that a feature silently broke on the Hub side and was caught by the Python CI. So generic rule is "don't mock tests when not needed"

```suggestion
def test_bucket_list_search_error_with_files(tree_bucket: str):
    """Cannot use --search when listing files."""
    result = cli(f"hf buckets list {tree_bucket} --search foo")
    assert result.exit_code != 0
    assert "Cannot use --search when listing files" in result.output


def test_bucket_list_search_term_exists(bucket_read: str, bucket_write: str):
    search_term = bucket_read.split("/")[1].split("-")[1]  # e.g. 9447e4
    result = cli(f"hf buckets list {USER} --search {search_term} --quiet")
    assert bucket_read in result.stdout
    assert bucket_write not in result.stdout


def test_bucket_list_search_term_empty_results():
    result = cli(f"hf buckets list {USER} --search does-not-exist-1234567890")
    assert f"No buckets found under namespace '{USER}' matching search 'does-not-exist-1234567890'." in result.stdout


```

Wauplin (commented):
Thanks @alexpouliquen for your PR!

Left a comment regarding tests in the Python library. It's not a list of hard rules but more a set of guidelines we try to follow. These should be in the AGENTS.md but often gets ignored so good to double-check. Other than that, ready to merge!

tests/test_buckets_cli.py
alexpouliquen: Really clear, thanks for the recommandations.

Wauplin (approved):
Good to merge! First of many? 🎉

PR #4147, opened 2026-04-25
[CLI] Allow unquoted multi-word search for ls commands

ParamThakkar123:
## Summary

- Changed `--search` option to a positional `[QUERY]` argument for `hf datasets ls`, `hf models ls`, and `hf spaces ls`
- This allows users to run unquoted multi-word searches like:
  - `hf datasets ls japanese instruction` (parses as single query)
  - `hf models llama 3` (parses as single query)
  - `hf spaces chatbot` (parses as single query)

Previously required quoting:
- `hf datasets ls --search "japanese instruction"`

## Examples from manual testing

```
$ hf datasets ls --help
Usage: root ls [OPTIONS] [QUERY]

Arguments:
  [QUERY]  Search query.
```

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the public CLI interface by replacing `--search` with a positional query, which can break existing scripts/automation relying on the old flag.
> 
> **Overview**
> **CLI behavior change:** `hf models|datasets|spaces ls` now takes the search term(s) as a positional `QUERY` argument (accepting multiple words without quoting) and joins them into the `search` parameter passed to the Hub API, instead of using the `--search` option.
> 
> Tests were updated to reflect the new invocation style, and a small formatting-only tweak was made in `utils/release_notes/generate_release_notes.py` (single-line `subprocess.run(...)`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75c7f3b631511de1dd28209c3324a9171cfc1730. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/datasets.py
+++ b/src/huggingface_hub/cli/datasets.py
@@ -39,7 +39,6 @@ from ._cli_utils import (
     FormatWithAutoOpt,
     LimitOpt,
     RevisionOpt,
-    SearchOpt,
     TokenOpt,
     api_object_to_dict,
     get_hf_api,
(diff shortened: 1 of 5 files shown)


Wauplin:
Hi @ParamThakkar123 , thanks for the PR but we won't move forward with it. Is quoting the search term a problem for you? 

The problem of making `search` a positional argument is that we won't be able to add other positional arguments in the future, and I don't think `search` must be the default positional argument (we can only use it once^^). Plus, I don't see the problem of doing `hf datasets ls --search "japanese instruction"`

PR #4155, opened 2026-04-27
[CLI] Add spaces lifecycle commands: pause, restart, sleep

davanstrien:
## Summary

Wraps three existing `HfApi` methods so Spaces lifecycle is controllable from the CLI:

- `hf spaces pause <id>` → `pause_space`
- `hf spaces restart <id> [--factory-reboot]` → `restart_space`
- `hf spaces sleep <id> --seconds N` → `set_space_sleep_time`

## CLI examples

```
$ hf spaces pause --help
Usage: hf spaces pause [OPTIONS] SPACE_ID

  Pause a Space.

Arguments:
  SPACE_ID  The space ID (e.g. `username/repo-name`).  [required]

Options:
  -y, --yes                       Answer Yes to prompt automatically.
  --format [agent|auto|human|json|quiet]
                                  Output format.  [default: auto]
  --token TEXT                    A User Access Token generated from
                                  https://huggingface.co/settings/tokens.
  -h, --help                      Show this message and exit.

Examples
  $ hf spaces pause username/my-space
  $ hf spaces pause username/my-space --yes
```

`hf spaces restart --help` and `hf spaces sleep --help` follow the same shape (with `--factory-reboot` and `--seconds` respectively).

Tests pass:
```
$ pytest tests/test_cli.py -k TestSpacesLifecycleCommands
tests/test_cli.py::TestSpacesLifecycleCommands::test_pause PASSED
tests/test_cli.py::TestSpacesLifecycleCommands::test_restart_factory_reboot PASSED
tests/test_cli.py::TestSpacesLifecycleCommands::test_sleep PASSED
3 passed in 0.40s
```

Live smoke test against a real Space pending — will add output before un-drafting.

## Discussion point: confirmation prompts

`pause` and `restart` ship **with** an `out.confirm()` prompt + `-y/--yes` flag, mirroring `volumes delete`. This is a deliberate **divergence** from the established CLI rule (verified across `repos.py`, `buckets.py`, `spaces.py`, `auth.py`):

> **confirm = deletion, no confirm = state change**

By that rule, pause/restart should not prompt — they parallel `dev_mode --stop` (which tears down a running container without asking).

Reasoning for diverging here: pausing/restarting wipes ephemeral storage on the Space's container. Users who haven't mounted a Volume can lose unsaved data. The same is true of the dashboard buttons, but a CLI invocation is easier to script/automate, which raises the blast radius.

`sleep` does **not** prompt (it's a config update, no container teardown).

Happy to remove the prompts if you'd rather stay consistent with the rest of the CLI — let me know.

## Notes

- `space_id` argument shape, `-y/--yes` shape (`Answer Yes to prompt automatically.`), and `out.result(...) + out.hint(...)` output style all match existing `spaces.py` patterns verbatim.
- Docs guide at `cli.md` adds three terse subsections matching the existing tone (frame pause around billing, link to canonical docs for sleep).
- `package_reference/cli.md` regenerated by `make style` (do not edit by hand).
- Cross-linking from `hub-docs/spaces-gpus.md` (#pause, #sleep-time) is a separate follow-up PR.

---

> [!NOTE]
> **Medium Risk**
> Adds new `hf spaces` lifecycle commands that trigger remote runtime actions (pause/restart/factory reboot and sleep-time changes), so misuse could interrupt running Spaces or discard ephemeral data.
> 
> **Overview**
> Adds new Spaces lifecycle controls to the `hf` CLI: `hf spaces pause` and `hf spaces restart` (optionally `--factory-reboot`) plus `hf spaces settings --sleep-time` for updating a Space’s sleep timeout.
> 
> Documentation is updated to describe these commands and warn that pausing/restarting tears down the container (ephemeral filesystem is lost), and the autogenerated CLI reference now includes the new subcommands and options.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d3a9038951e3dd52922a6feb51a39bfbdf76d23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/spaces.py
+++ b/src/huggingface_hub/cli/spaces.py
@@ -297,6 +297,106 @@ def dev_mode(
 
 
 @spaces_cli.command(
+    "pause",
+    examples=[
+        "hf spaces pause username/my-space",
+        "hf spaces pause username/my-space --yes",
+    ],
+)
+def spaces_pause(
+    space_id: Annotated[str, typer.Argument(help="The space ID (e.g. `username/repo-name`).")],
+    yes: Annotated[
+        bool,
+        typer.Option(
+            "-y",
+            "--yes",
+            help="Answer Yes to prompt automatically.",
+        ),
+    ] = False,
+    format: FormatWithAutoOpt = OutputFormatWithAuto.auto,
+    token: TokenOpt = None,
+) -> None:
+    """Pause a Space."""
+    out.confirm(f"You are about to pause Space '{space_id}'. Proceed?", yes=yes)
+    api = get_hf_api(token=token)
+    runtime = api.pause_space(space_id)
+    out.result("Space paused", space_id=space_id, stage=runtime.stage)
+    out.hint(f"Use `hf spaces restart {space_id}` to restart it.")
+
+
+@spaces_cli.command(
+    "restart",
+    examples=[
+        "hf spaces restart username/my-space",
+        "hf spaces restart username/my-space --yes",
+        "hf spaces restart username/my-space --factory-reboot --yes",
+    ],
+)
+def spaces_restart(
+    space_id: Annotated[str, typer.Argument(help="The space ID (e.g. `username/repo-name`).")],
+    factory_reboot: Annotated[
+        bool,
+        typer.Option(
+            "--factory-reboot",
+            help="Rebuild the Space from scratch without using the build cache.",
+        ),
+    ] = False,
+    yes: Annotated[
+        bool,
+        typer.Option(
+            "-y",
+            "--yes",
+            help="Answer Yes to prompt automatically.",
+        ),
+    ] = False,
+    format: FormatWithAutoOpt = OutputFormatWithAuto.auto,
+    token: TokenOpt = None,
+) -> None:
+    """Restart a Space."""
+    if factory_reboot:
+        prompt = (
+            f"You are about to restart Space '{space_id}' with a factory reboot (build cache will be wiped). Proceed?"
+        )
+    else:
+        prompt = f"You are about to restart Space '{space_id}'. Proceed?"
+    out.confirm(prompt, yes=yes)
+    api = get_hf_api(token=token)
+    runtime = api.restart_space(space_id, factory_reboot=factory_reboot)
+    out.result(
+        "Space restart triggered",
+        space_id=space_id,
+        stage=runtime.stage,
+        factory_reboot=factory_reboot,
+    )
+    out.hint(f"Use `hf spaces info {space_id}` to monitor the runtime stage.")
+
+
+@spaces_cli.command(
+    "sleep",
+    examples=[
+        "hf spaces sleep username/my-space --seconds 300",
+    ],
+)
+def spaces_sleep(
+    space_id: Annotated[str, typer.Argument(help="The space ID (e.g. `username/repo-name`).")],
+    seconds: Annotated[
+        int,
+        typer.Option(
+            "--seconds",
+            help="Idle time in seconds after which the Space goes to sleep. Only available on upgraded hardware.",
+        ),
+    ],
+    format: FormatWithAutoOpt = OutputFormatWithAuto.auto,
+    token: TokenOpt = None,
+) -> None:
+    """Set the idle sleep time for a Space."""
+    api = get_hf_api(token=token)
+    runtime = api.set_space_sleep_time(space_id, sleep_time=seconds)
+    out.result("Sleep time set", space_id=space_id, sleep_time=runtime.sleep_time)
+    out.hint(f"Use `hf spaces info {space_id}` to verify the runtime configuration.")
+
+
+@spaces_cli.command(
     "logs",
     examples=[
         "hf spaces logs username/my-space",
--- a/src/huggingface_hub/cli/spaces.py
+++ b/src/huggingface_hub/cli/spaces.py
@@ -322,6 +322,9 @@ def spaces_pause(
     runtime = api.pause_space(space_id)
     out.result("Space paused", space_id=space_id, stage=runtime.stage)
     out.hint(f"Use `hf spaces restart {space_id}` to restart it.")
+    out.hint(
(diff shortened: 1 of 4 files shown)


davanstrien:
Ran smoke tests locally now

src/huggingface_hub/cli/spaces.py
```diff
@@ -296,6 +296,112 @@ def dev_mode(
...
+            "--seconds",
+            help="Idle time in seconds after which the Space goes to sleep. Only available on upgraded hardware.",
+        ),
+    ],
+    format: FormatWithAutoOpt = OutputFormatWithAuto.auto,
+    token: TokenOpt = None,
+) -> None:
+    """Set the idle sleep time for a Space."""
+    api = get_hf_api(token=token)
+    runtime = api.set_space_sleep_time(space_id, sleep_time=seconds)
```
hanouticelina: because the server does not apply sleep time on free hardware (see https://github.com/huggingface/huggingface_hub/blob/3c40d1f95e6cbb672dc757883d13ed9e90faae65/src/huggingface_hub/hf_api.py#L7873-L7874)
```suggestion
    runtime = api.set_space_sleep_time(space_id, sleep_time=seconds)
    if runtime.sleep_time != seconds:
        out.warning(
            f"Sleep time was not applied (returned {runtime.sleep_time}). "
            "Configurable sleep time requires upgraded hardware."
        )
       return
```

src/huggingface_hub/cli/spaces.py
```diff
@@ -296,6 +296,112 @@ def dev_mode(
...
+        "hf spaces sleep username/my-space --seconds 300",
+    ],
+)
+def spaces_sleep(
+    space_id: Annotated[str, typer.Argument(help="The space ID (e.g. `username/repo-name`).")],
+    seconds: Annotated[
+        int,
+        typer.Option(
+            "--seconds",
+            help="Idle time in seconds after which the Space goes to sleep. Only available on upgraded hardware.",
```
hanouticelina:
```suggestion
            help="Idle time in seconds after which the Space goes to sleep. Use -1 to never sleep. Only available on upgraded hardware.",
```

hanouticelina (commented):
Thanks!

src/huggingface_hub/cli/spaces.py
Wauplin: haven't tried myself but what happens if we try to update sleep time on a Space with CPU Basic? Just ignored without warning/error from the server?

tests/test_cli.py
```diff
@@ -2113,6 +2113,33 @@ def test_logs_follow_and_tail_error(self, runner: CliRunner) -> None:
...
+    def test_sleep(self, runner: CliRunner) -> None:
+        with patch("huggingface_hub.cli.spaces.get_hf_api") as api_cls:
+            api = api_cls.return_value
+            api.set_space_sleep_time.return_value = Mock(sleep_time=300)
+            result = runner.invoke(app, ["spaces", "sleep", "user/my-space", "--seconds", "300"])
+        assert result.exit_code == 0, result.output
+        api_cls.assert_called_once_with(token=None)
+        api.set_space_sleep_time.assert_called_once_with("user/my-space", sleep_time=300)
+
+
```
Wauplin: no need for these mocked tests IMO (very verbose just to check that it's a simple call to the `HfApi` method

src/huggingface_hub/cli/spaces.py
davanstrien: Tested with three tier configurations and all three return a 400 server-side:

| Tier | Server response |
|---|---|
| `static` | `Can't set custom sleep time for static Spaces` |
| `cpu-basic` (gradio) | `Cannot change the sleep time for CPU Basic Spaces` |
| `zero-a10g` (ZeroGPU) | `Cannot change the sleep time for ZeroGPU Spaces` |

So `hf_raise_for_status` throws before reaching the `runtime.sleep_time != seconds` check. 

Happy to leave the CLI guard in place if preferred.

hanouticelina: ah yes in that case, no need for the CLI guard

Wauplin:
> ```
> hf spaces sleep <id> --seconds N
> ```

I find this syntax quite misleading. I would expect a `hf spaces sleep <id>` to put the Space to sleep, i.e. exactly what `hf spaces pause` is doing. I would suggest either `hf spaces settings <id> --sleep-time <SECONDS>`, similar to the current `hf repos settings <id> ...` or `hf spaces --set-sleep-time`. Slight preference for the former for explicitness even though it's not consistent with the `HfApi` methods. Later we might do `hf spaces settings <id> --hardware ...` as well? 

Otherwise `hf spaces --set-sleep-time` / `hf spaces --set-hardware` is also fine to be consistent with the HfApi calls. Happy to get some suggestions on this @davanstrien @hanouticelina 

> Discussion point: confirmation prompts

Good point discussing this. I would tend to think confirmation prompt is a bit too much for a pause/restart since by nature Space storage is ephemeral so we are not really deleting "hard storage". I would just go yolo on that

hanouticelina:
I have a preference for `hf spaces settings <id> --sleep-time <SECONDS>` as well

davanstrien:
- Renamed `sleep --seconds` → `settings --sleep-time`. Went with the `settings` shape since it mirrors `hf repos settings` and gives `--hardware` etc. a home later without new top-level verbs.
- Dropped confirms on `pause`/`restart`
- For `secrets`/`variables` later it might make sense to keep them as nested subgroups (`hf spaces secrets set/rm/ls`) like `volumes` — i.e. `settings` for single-value attributes, subgroups for keyed collections?

Wauplin:
> For secrets/variables later it might make sense to keep them as nested subgroups (hf spaces secrets set/rm/ls) like volumes — i.e. settings for single-value attributes, subgroups for keyed collections?

Yes, agree on this as well :+1:

Wauplin (approved):
Thank you!

hanouticelina (approved):
Thanks!

PR #4156, opened 2026-04-27
[CLI] Add benchmark dataset filter examples

hanouticelina:
_Comment from_ @hanouticelina : i think it's worth having an example to list official benchmarks, mostly for discoverability. a better solution for discoverability would be to expose `list_datasets(benchmark=...)` directly in the CLI. that said, `list_datasets` has several other filters that are not surfaced either (`gated`, `language`, `task_categories`, etc) so adding only `--benchmark` would mean adding the others as well for consistency. 

## Summary

- Add `hf datasets ls --filter benchmark:official` to the `hf datasets ls` CLI examples.
- Document the same example in the CLI guide.
- Regenerate the CLI reference so command help examples stay in sync.

---

> [!NOTE]
> **Low Risk**
> Low risk documentation-only change that just adds an additional CLI usage example; no functional behavior or API surface is modified.
> 
> **Overview**
> Adds a new discoverability example for listing official benchmark datasets by documenting `hf datasets ls --filter benchmark:official` in the CLI guide and in the generated CLI reference.
> 
> Updates the `hf datasets ls` command’s Typer `examples` list so the help output and regenerated docs stay in sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 629dce541d6a9762df5d6332ad7508504f0a5d8a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -848,6 +848,9 @@ Use `hf datasets` to list datasets on the Hub and get detailed information about
 # Search for datasets
 >>> hf datasets ls --search "code"
 
+# List official benchmark datasets
+>>> hf datasets ls --filter benchmark:official
+
 # Sort by downloads
 >>> hf datasets ls --sort downloads --limit 10
 ```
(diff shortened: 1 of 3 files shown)


Wauplin (approved):
Reasoning makes perfect sense to me!

PR #4162, opened 2026-04-28
[CLI] Make --format / --json / -q global

Wauplin:
### Context: let's handle `--format` globally

Benefits are:
- **End users**: formatting options are now first-class and properly documented in `--help` (one consistent section per command), instead of being mixed into the regular options or only mentioned in some commands.
- **Maintainers**: no need to remember to add `format: FormatWithAutoOpt = OutputFormatWithAuto.auto` (and the matching imports) when adding a new command. The flags work automatically.

## Summary

- Handle `--format <value>`, `--json`, and `-q` / `--quiet` globally in `HFCliTyperGroup` instead of declaring them on every command. The flags can appear anywhere in a leaf command's args; they're stripped and applied to the `out` singleton before click parses the rest.
- Document them in a dedicated `Formatting options` section in every `--help` page (root, groups, and leaves).
- Removed `format: FormatWithAutoOpt = OutputFormatWithAuto.auto` from every command (~85 occurrences across 17 files).
- Legacy commands that own their own `--format` (`hf jobs ps`, `hf jobs scheduled ps`) keep working — the previous `--json` / `--quiet` shorthand-rewrite behaviour is preserved for them. Pass-through commands (`hf extensions exec`) remain untouched.
- Added a short `Output formatting` section to the CLI guide.

## Examples

```bash
# --format / --json / -q now appear in every command's --help
$ hf models ls --help
...
Formatting options:
  --format [auto|human|agent|json|quiet]
                                  Output format. Defaults to 'auto' which
                                  picks 'agent' or 'human' based on the
                                  terminal.
  -q, --quiet                     Quiet output (one ID per line). Equivalent
                                  to '--format quiet'.
  --json                          JSON output. Equivalent to '--format json'.

# Works on any command — no per-command opt-in needed
$ hf collections ls --owner nvidia --limit 2 -q
nvidia/nemotron-supervised-fine-tuning-69eab9824c9120a3a3b1e25e
nvidia/nvidia-nemotron-v3-69388dda16167bb1607171ea

# Mutually exclusive
$ hf models ls --json --format table
Error: '--json' and '--format' are mutually exclusive.

# Legacy command with its own --format keeps working
$ hf jobs ps --format '{{.id}} {{.status}}'
```

---

> [!NOTE]
> **Medium Risk**
> Touches CLI argument parsing and help rendering globally, so regressions could affect many commands’ option handling; changes include explicit compatibility paths for legacy/pass-through commands to reduce risk.
> 
> **Overview**
> Makes `--format`, `--json`, and `-q/--quiet` **global formatting flags** by consuming them during command resolution and applying the chosen mode to the shared `out` singleton, so leaf commands no longer need to declare `FormatWithAutoOpt` (while pass-through commands remain untouched).
> 
> Updates help generation to add a consistent **“Formatting options”** section to command `--help` output, preserves backward compatibility for legacy commands that own their own formatting options by rewriting shorthands, and switches a few commands (e.g. `hf buckets sync`) to read quiet-mode from `out` instead of a local flag. Documentation is updated to explain the new behavior, and the generated CLI reference is refreshed accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e44a4c619f5ee2ca1ee31ec1faf689b3e5195e44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -389,6 +368,160 @@ def list_commands(self, ctx: click.Context) -> list[str]:  # type: ignore[name-d
...
+        if arg in ("-q", "--quiet"):
+            _check_conflict(arg)
+            chosen_mode = OutputFormatWithAuto.quiet
+            chosen_flag = arg
+            del args[i : i + 1]
+            continue
+        i += 1
+
+    if chosen_mode is not None:
+        out.set_mode(chosen_mode)
```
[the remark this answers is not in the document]
Wauplin: fixed in https://github.com/huggingface/huggingface_hub/pull/4162/changes/d0875c03b5d33209de2bd7b30b0e9d312bb5c5e8 (by removing explicit `--quiet` option in `hf buckets sync`)

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -182,43 +183,21 @@ def resolve_command(self, ctx: click.Context, args: list[str]) -> tuple:
...
-            flag = "-q" if "-q" in args else "--quiet"
-            idx = args.index(flag)
-            args[idx : idx + 1] = ["--format", "quiet"]
+    def format_options(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
+        # MultiCommand.format_options writes regular options + the commands list. We
+        # insert the "Formatting options" section in between so it sits with the
+        # other options instead of below the subcommand list.
+        click.Command.format_options(self, ctx, formatter)
+        _format_formatting_options_section(formatter)
+        self.format_commands(ctx, formatter)
```
hanouticelina: no strong opinion but I think it would be better to have:
`hf --help` -> no "Formatting options" section
`hf models --help` -> no section either
`hf models ls --help` -> shows the Fromatting section (where the flags actually work)
```suggestion
```

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -389,6 +368,159 @@ def list_commands(self, ctx: click.Context) -> list[str]:  # type: ignore[name-d
...
+    # Strip --format/--json/-q/--quiet from 'args' and apply to 'out'
+    chosen_mode: OutputFormatWithAuto = OutputFormatWithAuto.auto
+    chosen_flag: str | None = None
+
+    def _check_conflict(new_flag: str) -> None:
+        # Reject mixing different kinds of formatting flags before parsing values,
+        # so the user gets a "mutually exclusive" error rather than e.g. an
+        # "invalid value" error for the second flag.
+        if chosen_flag is not None and _flag_kind(chosen_flag) != _flag_kind(new_flag):
+            raise click.UsageError(f"'{chosen_flag}' and '{new_flag}' are mutually exclusive.")
```
hanouticelina, line 437: this one was flagged by Codex while testing the branch with two `--format` flags. quoting it directly here:
If the goal is "mixing formatting flags raises a usage error" (per the PR description), `--format json --format quiet` should error too, right now it silently picks the last one, because `_check_conflict` only fires on cross-kind mixes via `_flag_kind`.

```suggestion
    def _check_conflict(new_flag: str) -> None:
        # Reject any second formatting flag before parsing values, so the user gets
        # a "mutually exclusive" error rather than e.g. an "invalid value" error
        # from the second flag's argument.
        if chosen_flag is not None:
            raise click.UsageError(f"'{chosen_flag}' and '{new_flag}' are mutually exclusive.")
```

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -389,6 +368,159 @@ def list_commands(self, ctx: click.Context) -> list[str]:  # type: ignore[name-d
...
 
 
+_FORMATTING_OPTIONS_HELP_RECORDS: list[tuple[str, str]] = [
+    (
+        "--format [auto|human|agent|json|quiet]",
+        "Output format. Defaults to 'auto' which picks 'agent' or 'human' based on the terminal.",
+    ),
+    ("-q, --quiet", "Quiet output (one ID per line). Equivalent to '--format quiet'."),
+    ("--json", "JSON output. Equivalent to '--format json'."),
+]
```
hanouticelina, line 368: for alphabetical ordering
```suggestion
    (
        "--format [auto|human|agent|json|quiet]",
        "Output format. Defaults to 'auto' which picks 'agent' or 'human' based on the terminal.",
    ),
    ("--json", "JSON output. Equivalent to '--format json'."),
    ("-q, --quiet", "Quiet output (one ID per line). Equivalent to '--format quiet'."),
]
```

hanouticelina (commented):
very nice improvement!

src/huggingface_hub/cli/_cli_utils.py
Wauplin: such a strange thing to test :smile:

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -389,6 +358,159 @@ def list_commands(self, ctx: click.Context) -> list[str]:  # type: ignore[name-d
...
+                raise click.UsageError(f"'{flag}' and '--format' are mutually exclusive.")
+            idx = args.index(flag)
+            args[idx : idx + 1] = ["--format", "quiet"]
+
+
+def _flag_kind(flag: str) -> str:
+    if flag in ("-q", "--quiet"):
+        return "--quiet"
+    return flag
+
```
hanouticelina:
```suggestion
```

hanouticelina (approved):
Thanks!

src/huggingface_hub/cli/_output.py
```diff
@@ -52,10 +52,15 @@ def __init__(self) -> None:
         self.set_mode()
 
     def set_mode(self, mode: OutputFormatWithAuto = OutputFormatWithAuto.auto) -> None:
-        """Override the output mode (called by commands that receive ``--format``)."""
+        """Override the output mode (called once at startup and again per '--format' flag)."""
         if mode == OutputFormatWithAuto.auto:
             mode = OutputFormatWithAuto.agent if is_agent() else OutputFormatWithAuto.human
         self.mode = mode
+        if mode != OutputFormatWithAuto.human:
+            disable_progress_bars()
```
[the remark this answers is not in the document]
Wauplin: that's fine now that (almost?) all commands have been migrated

PR #4165, opened 2026-04-29
Avoid reuploading preuploaded LFS files in upload-large-folder

Dev-Jahn:
## What does this PR do?

upload_large_folder stores whether a file has already been preuploaded in LocalUploadFileMetadata.is_uploaded, but _build_hacky_operation did not restore that state on the CommitOperationAdd created for the commit step.

As a result, create_commit could call preupload_lfs_files again for files that were already uploaded by the large-folder preupload step. This makes the commit phase unnecessarily re-read/re-upload LFS files after resuming or after preupload completion.

This PR restores _is_uploaded from the local metadata and clears path_or_fileobj for already uploaded LFS files, matching the behavior of preupload_lfs_files after a successful upload.

## Tests

- uv run python -m compileall -q src/huggingface_hub/_upload_large_folder.py tests/test_upload_large_folder.py
- uv run --with pytest --with pytest-mock --with pytest-env --with-editable . pytest tests/test_upload_large_folder.py -q
- uv run --with ruff ruff format --check src/huggingface_hub/_upload_large_folder.py tests/test_upload_large_folder.py
- uv run --with ruff ruff check src/huggingface_hub/_upload_large_folder.py tests/test_upload_large_folder.py

### DIFF
--- a/src/huggingface_hub/_upload_large_folder.py
+++ b/src/huggingface_hub/_upload_large_folder.py
@@ -724,6 +724,9 @@ def _build_hacky_operation(item: JOB_ITEM_T) -> HackyCommitOperationAdd:
     operation._upload_mode = metadata.upload_mode  # type: ignore
     operation._should_ignore = metadata.should_ignore
     operation._remote_oid = metadata.remote_oid
+    operation._is_uploaded = metadata.is_uploaded
+    if metadata.is_uploaded and metadata.upload_mode == "lfs":
+        operation.path_or_fileobj = b""
     return operation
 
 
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Thanks for finding this out and the PR @Dev-Jahn ! I don't get why we never caught this issue but after carefully reviewing this PR + https://github.com/huggingface/huggingface_hub/pull/2254 and https://github.com/huggingface/huggingface_hub/pull/3100, I can confirm this looks like a complete oversight from us. And thanks for the minimal test as well :) 

Note that if the metadata is older than 20h, we automatically set `is_uploaded` to False even if metadata says otherwise because the uploaded Xet file might have been garbage-collected in the meantime (if uploaded but not committed).

Anyway, thanks for the fix!

PR #4111, opened 2026-04-15
[CLI] Migrate buckets commands to out singleton

hanouticelina:
Part of #3979. Migrates `buckets` commands (`create`, `list`, `info`, `delete`, `remove`, `move`, `cp`) to the unified `--format [auto|human|agent|json|quiet]` flag and `out.*` methods.

- Replaces `--quiet` with `--format` on all migrated commands
- Uses `out.result()` for success messages, `out.dict()` for info, `out.table()` for bucket listing, `out.confirm()` for confirmations, `out.hint()` for hints
- Adds `id_key` param to `out.dict()` for quiet mode (consistent with `out.table()`)
- `sync` left as-is (passes quiet/verbose to library)

**`_list_files` and `_list_buckets` keep their existing output format** — the `ls`-like flat table, `-h` for human-readable sizes, `--tree`, and the json/quiet branches are preserved as-is. Only the flag changes (`--quiet` → `--format quiet`, `--format table` → `--format auto`). The existing output worked well and we didn't want to change it unnecessarily by routing everything through `out.table()`.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes CLI flags and stdout/stderr output formatting for multiple `hf buckets` commands, which may break user scripts relying on previous messages or `--quiet` behavior.
> 
> **Overview**
> Migrates `hf buckets` commands (`create`, `list`, `info`, `delete`, `rm`, `move`, `cp`) from ad-hoc printing and `--quiet`/`--format table|json` to the unified `out` output framework with `--format [auto|human|agent|json|quiet]`.
> 
> Adds `id_key` support to `out.dict()` to make `--format quiet` print a single identifier (used by `buckets info`), routes bucket listing through `out.table()` (including standardized empty-results messaging), and adjusts hints/confirmations/results accordingly; tests and generated CLI docs are updated to match the new outputs and flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 387dcfe880adff943a3459df95bf4903d5ca9ac4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4171, opened 2026-04-30
[CLI] Add `out.status()`

hanouticelina:
Follow-up to Wauplin’s review comment in https://github.com/huggingface/huggingface_hub/pull/4111#discussion_r3087129946.

This PR adds `out.status()` as the CLI wrapper around `StatusLine`, so command code can request a status line without deciding when it should render. It also migrates the existing CLI status-line call sites in buckets, extensions, and spaces to use the new helper.

**Breaking Change**

`hf extensions install` and `hf spaces dev-mode` no longer print status updates when run with `--format json`, `--quiet`, or `--format agent`.

Before, `StatusLine()` defaulted to `enabled=True`, so updates went to stderr whenever stderr was a TTY, regardless of output mode. After this PR, status updates go to stderr only in `human` mode, and still only when stderr is a TTY.

---

> [!NOTE]
> **Medium Risk**
> Changes CLI stderr output behavior by suppressing status updates in `agent`/`json`/`quiet` modes, which may affect tooling that relied on those messages.
> 
> **Overview**
> Adds `Output.status()` as a CLI-level wrapper around `StatusLine`, automatically enabling status rendering only in `human` mode and remaining a no-op for `agent`/`json`/`quiet` output.
> 
> Updates existing status-line usage in `hf buckets rm --recursive`, `hf extensions install` (python extension path), and `hf spaces dev-mode` to use `out.status()` instead of constructing `StatusLine` directly, and adds a targeted unit test to lock in the new mode-dependent behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9912a86938566ce51ad01e1a818f241acc0dbb37. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_cli_output.py
+++ b/tests/test_cli_output.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 import pytest
 
 from huggingface_hub.cli._cli_utils import OutputFormatWithAuto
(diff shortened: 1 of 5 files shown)


tests/test_cli_output.py
```diff
@@ -318,6 +320,32 @@ def test_confirm_non_human_raises(mode):
         o.confirm("Delete everything?")
 
 
+# =============================================================================
+# out.status()
+# =============================================================================
+
+
+def test_status(check, monkeypatch):
```
Wauplin:
```suggestion
def test_status_only_enabled_for_humans(check, monkeypatch):
```

Wauplin (approved):
Nice and clean, thank you!

PR #4158, opened 2026-04-28
Centralize hf:// URI parsing

Wauplin:
## Summary

Adds a single source of truth for parsing `hf://...` URIs, addressing #3971.

- New `HfUri` dataclass and `parse_hf_uri` helper, plus shared constants in `huggingface_hub.constants` (`HF_PROTOCOL`, `HfUriType`, `HF_URI_TYPE_PREFIXES`).
- Separate `HfMount` dataclass and `parse_hf_mount` helper for volume mount specifications (`hf://...:<MOUNT_PATH>[:ro|:rw]`). Mount logic is cleanly separated from URI parsing — `HfUri` is a pure location identifier, `HfMount` wraps a `HfUri` with a mount path and read-only flag.
- A reference page documenting the canonical syntax for both URIs and mounts, and what does (and does not) qualify as a HF URI — strict-by-design (plural-only type prefixes, no canonical repos without namespace, etc.).
- Pure string parser, no network calls. Round-trippable via `HfUri.to_uri()` and `HfMount.to_uri()`.

```py
>>> from huggingface_hub import parse_hf_uri
>>> parse_hf_uri("hf://datasets/namespace/my-dataset@refs/pr/3/train.json")
HfUri(type='dataset', id='namespace/my-dataset', revision='refs/pr/3', path_in_repo='train.json')

>>> from huggingface_hub import parse_hf_mount
>>> parse_hf_mount("hf://buckets/my-org/my-bucket/sub/dir:/mnt:ro")
HfMount(source=HfUri(type='bucket', id='my-org/my-bucket', revision=None, path_in_repo='sub/dir'), mount_path='/mnt', read_only=True)
```

## Follow-ups (not in this PR)

- Migrate existing call sites to the new parser:
  - `HfFileSystem.resolve_path` (keeps its 1- vs 2-segment network fallback on top of the parser)
  - `_parse_hf_copy_handle` in `hf_api.py` (the spot already carries a `# TODO` referencing #3971)
  - `parse_volumes` in `cli/_cli_utils.py`
  - `_split_bucket_id_and_prefix` / `_parse_bucket_path` / `_is_bucket_path` in `_buckets.py` and `_parse_bucket_argument` / `_is_hf_handle` in `cli/buckets.py`
  - `Volume.to_hf_handle` formatter in `_space_api.py` (mirror of `HfMount.to_uri`)
[...]

PR #4189, opened 2026-05-05
[Core] Migrate hf:// URI parsing to centralized parse_hf_uri

Wauplin:
## Summary

Follow-up to #4158. Replaces all scattered ad-hoc `hf://` URI parsers with the centralized `parse_hf_uri`/`parse_hf_mount` helpers.

**Not touched** (as mentioned in #4158): `repo_type_and_id_from_hf_id` and `RepoUrl` — these are mixed parsers accepting both `hf://` URIs and `https://huggingface.co/...` URLs and need a separate migration plan.

### Breaking changes

1. **`BucketUrl.handle` → `BucketUrl.uri`** — Attribute renamed and type changed from `str` to `HfUri`. Any code accessing `bucket_url.handle` must switch to `bucket_url.uri` (use `.to_uri()` for the string form).

2. **`Volume.to_hf_handle()` → `Volume.to_uri()`** — Method renamed. Any code calling `.to_hf_handle()` must switch to `.to_uri()`.

3. **Single-segment repo IDs no longer supported in `HfFileSystem`** — Paths like `gpt2`, `gpt2/file.txt`, or `gpt2@dev/file.txt` no longer resolve. Repo IDs must use the `namespace/name` format (e.g. `username/gpt2`).

4. **Single-segment repo IDs rejected in CLI `-v` flags** — `hf://gpt2:/data` is no longer accepted. Must use `hf://namespace/name:/mount` form.

### Improvements

- `@` in filenames is now treated as literal (e.g. `hf://a/b/file@v1.txt` parses correctly instead of erroring).
- `HfUriError` is now caught by the CLI error handler and shows a clean message instead of a raw traceback.
- New `is_hf_uri` public helper for validating `hf://` URIs.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because this is a cross-cutting refactor of URI parsing that introduces breaking API/CLI behavior changes (renamed fields/methods and stricter URI formats), which could impact downstream integrations and edge-case path resolution.
> 
> **Overview**
> **Centralizes all `hf://` parsing on `parse_hf_uri`/`parse_hf_mount`** and removes multiple ad-hoc parsers across buckets, copy helpers, CLI volume parsing, and `HfFileSystem`.
> 
> Introduces **breaking API/CLI changes**: `BucketUrl.handle` is replaced by `BucketUrl.uri` (`HfUri`), `Volume.to_hf_handle()` is renamed to `to_uri()`, and CLI output/docs/examples are updated from “handle” wording to canonical “URI” formatting.
> 
> Tightens URI validation/semantics: `HfFileSystem` and CLI `-v/--volume` now reject single-segment repo IDs (must be `namespace/name`), CLI errors now format `HfUriError` cleanly, and URI parsing is improved to treat `@` in filenames as literal (while still supporting revision markers where valid).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a471279a9c88b0b4caf594045200f1deb8b71fcb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -842,75 +835,26 @@ def parse_volumes(volumes: list[str] | None) -> "list[Volume] | None":
...
     result: list[Volume] = []
     for raw_spec in volumes:
-        # Strip :ro/:rw suffix
-        spec = raw_spec
-        read_only = None
-        if spec.endswith(":ro"):
-            read_only = True
-            spec = spec[:-3]
-        elif spec.endswith(":rw"):
-            read_only = False
```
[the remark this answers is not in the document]
Wauplin: addressed in https://github.com/huggingface/huggingface_hub/pull/4189/commits/24ee99335cc49cabfe6c0ebc702bc1ad0432b477

src/huggingface_hub/_space_api.py
```diff
@@ -163,10 +163,11 @@ def to_dict(self) -> dict:
...
         """Return the volume as an HF handle in the format expected by the CLI."""
-        path = f"/{self.path}" if self.path else ""
-        revision = f"@{self.revision}" if self.revision else ""
-        ro = {True: ":ro", False: ":rw", None: ""}.get(self.read_only, "")
-        return f"hf://{self.type}s/{self.source}{revision}{path}:{self.mount_path}{ro}"
+        return HfMount(
+            source=HfUri(type=self.type, id=self.source, revision=self.revision, path_in_repo=self.path or ""),
+            mount_path=self.mount_path,
+            read_only=self.read_only,
+        ).to_uri()
```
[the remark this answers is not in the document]
Wauplin: Actually I wouldn't call it a regression but a bug fix :smile:

src/huggingface_hub/utils/_hf_uris.py
```diff
@@ -179,6 +180,16 @@ def to_uri(self) -> str:
...
+def is_hf_uri(uri: str) -> bool:
+    """Check if a string is a valie hf:// URI."""
+    try:
+        parse_hf_uri(uri)
+        return True
+    except HfUriError:
+        return False
+
+
+@functools.lru_cache
```
Wauplin, line 192: added since some logic is now parsing multiple time the same URI (once to check if it's an URI, then to check if it's a bucket, then to reuse its parsed values, etc.). Easier to use a LRU cache rather than parsing the value once and passing it everywhere

src/huggingface_hub/_buckets.py
```diff
@@ -179,7 +173,7 @@ class BucketUrl:
     endpoint: str = ""
     namespace: str = field(init=False)
     bucket_id: str = field(init=False)
-    handle: str = field(init=False)
+    uri: HfUri = field(init=False)
```
Wauplin, line 175: intentional breaking change

IMO ok to break as I don't expect anyone to use it at the moment (and it's nice to consistently use the 'uri' naming)

src/huggingface_hub/_space_api.py
```diff
@@ -161,12 +161,13 @@ def to_dict(self) -> dict:
             data["path"] = self.path
         return data
 
-    def to_hf_handle(self) -> str:
-        """Return the volume as an HF handle in the format expected by the CLI."""
-        path = f"/{self.path}" if self.path else ""
-        revision = f"@{self.revision}" if self.revision else ""
-        ro = {True: ":ro", False: ":rw", None: ""}.get(self.read_only, "")
-        return f"hf://{self.type}s/{self.source}{revision}{path}:{self.mount_path}{ro}"
+    def to_hf_mount_uri(self) -> str:
```
Wauplin: intentional breaking change

src/huggingface_hub/_buckets.py
```diff
@@ -257,15 +265,22 @@ def _parse_bucket_path(path: str) -> tuple[str, str]:
...
+    prefix = parsed.path_in_repo
+    if prefix and path.endswith("/"):
+        prefix += "/"
+    return parsed.id, prefix
 
 
 def _is_bucket_path(path: str) -> bool:
     """Check if a path is a bucket path."""
-    return path.startswith(BUCKET_PREFIX)
+    return is_hf_uri(path) and parse_hf_uri(path).is_bucket
```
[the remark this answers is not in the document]
Wauplin: this has to be fixed (@ are allowed in filenames both for repos and buckets)

src/huggingface_hub/_buckets.py
```diff
@@ -252,20 +245,27 @@ def __init__(self, **kwargs):
...
+    if path.startswith(constants.HF_PROTOCOL):
+        parsed = parse_hf_uri(path)
+        if not parsed.is_bucket:
+            raise ValueError(f"Invalid bucket path: {path}. Must be a bucket URI (hf://buckets/...).")
+        return parsed
+    parts = path.split("/", 2)
+    if len(parts) < 2 or not parts[0] or not parts[1]:
+        raise ValueError(f"Invalid bucket path: '{path}'. Expected format: namespace/bucket_name")
+    bucket_id = f"{parts[0]}/{parts[1]}"
+    prefix = parts[2].strip("/") if len(parts) > 2 else ""
```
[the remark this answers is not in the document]
Wauplin: addressed in https://github.com/huggingface/huggingface_hub/pull/4189/commits/407f65ba7ceff1722861408dc346da1c47294830

Wauplin:
PR should finally be ready for review! I tried to account for all possible use cases that were previously handled. This PR is introducing a few breaking changes (see PR description) **that are intentional**. Hopefully won't impact anyone :crossed_fingers: Parsing logic should be must cleaner now :smiley:

lhoestq:
Could HfUri work like RepoUrl and be a string subclass ? This would avoid `bucket.uri.to_uri()` which I find a bit weird

Wauplin:
> Could HfUri work like RepoUrl and be a string subclass ? This would avoid bucket.uri.to_uri() which I find a bit weird

I'd rather not no. `RepoUrl` is a str subclass for legacy reasons rather than a desired choice. What we can do is adding an alias method `BucketUrl.to_uri()` to avoid the `BucketUrk.uri.to_uri()`.

The reason why I don't want to implement something like `str(HfUri)` is that in the future I'm planning to add a `to_url()` method to get the url on the Hub of any resource described by a URI (e.g. a repo, a folder in a bucket, a file in a specific revision of a dataset, etc.). That should help simplify some other scattered logic we have + easily print URLs in the CLI.

lhoestq:
I see ! in that case I have a small pref for `bucket.uri.to_string()` but not a big deal

Wauplin:
> I see ! in that case I have a small pref for `bucket.uri.to_string()` but not a big deal

hmm, but I want to be able to do `bucket.uri.to_uri()` and `bucket.uri.to_url()` actually

src/huggingface_hub/cli/_cli_utils.py
hanouticelina, line 1: I love the cleaning done here! ❤️

src/huggingface_hub/utils/_hf_uris.py
```diff
@@ -179,6 +180,16 @@ def to_uri(self) -> str:
         return "".join(parts)
 
 
+def is_hf_uri(uri: str) -> bool:
+    """Check if a string is a valie hf:// URI."""
```
hanouticelina: you need to stop writing the code yourself 😄 
```suggestion
    """Check if a string is a valid hf:// URI."""
```

src/huggingface_hub/_buckets.py
```diff
@@ -166,7 +160,7 @@ class BucketUrl:
     - namespace (`str`)
     - bucket_id (`str`)
     - url (`str`)
-    - handle (`str`)
+    - uri (`str`)
```
hanouticelina:
```suggestion
    - uri (`HfUri`)
```

src/huggingface_hub/_buckets.py
hanouticelina: yes makes sense!

src/huggingface_hub/_buckets.py
```diff
@@ -252,20 +245,27 @@ def __init__(self, **kwargs):
...
+        raise ValueError(f"Invalid bucket path: '{path}'. Expected format: namespace/bucket_name")
+    bucket_id = f"{parts[0]}/{parts[1]}"
+    prefix = "/".join(parts[2:])
+    return HfUri(type="bucket", id=bucket_id, path_in_repo=prefix)
 
 
 def _is_bucket_path(path: str) -> bool:
     """Check if a path is a bucket path."""
-    return path.startswith(BUCKET_PREFIX)
+    return is_hf_uri(path) and parse_hf_uri(path).is_bucket
```
hanouticelina: I would have done something like that, also `is_hf_uri` already calls `parse_hf_uri` internally. the double parsing wouldn't have been an issue since we have a lru_cache but still
```suggestion
    try:
        return parse_hf_uri(path).is_bucket
    except HfUriError:
        return False
```

src/huggingface_hub/hf_api.py
```diff
@@ -417,60 +418,15 @@ def repo_type_and_id_from_hf_id(hf_id: str, hub_url: str | None = None) -> tuple
...
+def _parse_hf_copy_uri(hf_uri: str) -> _BucketCopyHandle | _RepoCopyHandle:
+    parsed = parse_hf_uri(hf_uri)
+    if parsed.is_bucket:
+        return _BucketCopyHandle(bucket_id=parsed.id, path=parsed.path_in_repo)
     return _RepoCopyHandle(
-        repo_type=repo_type,  # type: ignore
-        repo_id=f"{namespace}/{repo_name}",
-        revision=revision,
-        path=repo_path,
+        repo_type=parsed.type,  # type: ignore
```
hanouticelina, line 426: I don't get any linting errors if I remove the ignore
```suggestion
        repo_type=parsed.type,
```

src/huggingface_hub/hf_api.py
```diff
@@ -12920,11 +12876,10 @@ def copy_files(self, source: str, destination: str, *, token: str | bool | None
 
         Args:
             source (`str`):
-                Source location as an `hf://` handle. Can be a bucket path (e.g. `"hf://buckets/my-bucket/path/to/file"`)
+                Source location as an `hf://` URI. Can be a bucket path (e.g. `"hf://buckets/my-bucket/path/to/file"`)
                 or a repo path (e.g. `"hf://username/my-model/weights.bin"`, `"hf://datasets/username/my-dataset/data/"`).
             destination (`str`):
-                Destination location as an `hf://` handle pointing to a bucket
-                (e.g. `"hf://buckets/my-bucket/target/path"`).
+                Destination location as an `hf://` URI pointing to a bucket (e.g. `"hf://buckets/my-bucket/target/path"`).
```
hanouticelina, line 12882: (nit) line 12891 needs to be updated as well
```diff
- If the destination is not a bucket or if the source/destination handles are invalid.
+ If the destination is not a bucket or if the source/destination URIs are invalid.
```

src/huggingface_hub/hf_file_system.py
```diff
@@ -294,87 +294,59 @@ def resolve_path(
...
-                        f'Revision specified in path ("{revision_in_path}") and in `revision` argument ("{revision}")'
-                        " are not the same."
-                    )
-            else:
-                revision = revision_in_path
-            return revision
-
         path = self._strip_protocol(path)
-        if not path:
+        if not path or path.count("/") == 0:
```
hanouticelina: to avoid misleading error messages here 
```suggestion
        if not path:
            raise NotImplementedError("Access to buckets and repositories lists is not implemented.")
        if path.count("/") == 0:
            raise ValueError(
                f"Repository id must be 'namespace/name', got '{path}'. "
                "Single-segment ids (e.g. 'gpt2') are no longer supported."
            )
```

src/huggingface_hub/_buckets.py
hanouticelina: Thinking about it, let's not catch `HfUriError` in here. With the current code:
```bash
> hf buckets cp hf://buckets/celinah/my-bucket@v1/file.txt hf://buckets/celinah/other-bucket
Usage: hf buckets cp [OPTIONS] SRC [DST]
Try 'hf buckets cp -h' for help.

Error: Invalid value: Source file not found: hf://buckets/celinah/my-bucket@v1/file.txt
```
The real parse error is generated inside `parse_hf_uri` and then swallowed by `is_hf_uri` / `_is_bucket_path`.
```suggestion
    if not path.startswith(constants.HF_PROTOCOL):
         return False
     return parse_hf_uri(path).is_bucket  # raises on malformed hf://...
```
with the fix:
```bash
x hf buckets cp hf://buckets/celinah/my-bucket@v1/file.txt hf://buckets/celinah/other-bucket
Error: Invalid HF URI: hf://buckets/celinah/my-bucket@v1/file.txt. Bucket URIs do not support a revision marker ('@').
Set HF_DEBUG=1 as environment variable for full traceback.
```

hanouticelina (commented):
Thank you for simplifying the logic here! the PR looks globally good, I left a couple of comments and I will take another look before merging.

up to you if you want to do it in another PR, but let's drop the local `try/except ValueError` blocks in `cli/buckets.py` now that the global handler exists, non-URI parse failures still get a reasonable error message.

Wauplin:
> up to you if you want to do it in another PR, but let's drop the local try/except ValueError blocks in cli/buckets.py now that the global handler exists, non-URI parse failures still get a reasonable error message.

yes correct :+1: addressed in https://github.com/huggingface/huggingface_hub/pull/4189/commits/c843f3213a73e660123d7188c277a6dd73dd148b

src/huggingface_hub/hf_api.py
Wauplin: I'm getting 

```
src/huggingface_hub/hf_api.py:426:9: error[invalid-argument-type] Argument is incorrect: Expected `Literal["model", "dataset", "space"]`, found `Literal["model", "dataset", "space", "kernel", "bucket"]`
```

src/huggingface_hub/_buckets.py
Wauplin: Yes good idea, I like the second suggestion. I added a comment in docstring to mention that we do not raise if not a `hf://` URI but we raise if it's an incorrect `hf://` URI

(addressed in https://github.com/huggingface/huggingface_hub/pull/4189/commits/19cdfc0d7610fd5081a48d2cdd98dbc64b0b71c4)

Wauplin:
@hanouticelina  Addressed all of your comments, thanks for the review!

src/huggingface_hub/hf_file_system.py
```diff
@@ -294,87 +294,62 @@ def resolve_path(
...
-            # can't list repositories at root
             raise NotImplementedError("Access to buckets and repositories lists is not implemented.")
-        elif path.split("/")[0] == "buckets":
-            bucket_id = "/".join(path.split("/")[1:3])
-            path = "/".join(path.split("/")[3:])
-            bucket_exists, err = self._bucket_exists(bucket_id)
+        if path.count("/") == 0:
+            raise ValueError(
+                f"Repository id must be 'namespace/name', got '{path}'. Single-segment ids (e.g. 'gpt2') are no longer supported."
+            )
```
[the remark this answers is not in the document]
Wauplin: fixed in https://github.com/huggingface/huggingface_hub/pull/4189/commits/471473935bff394a2282f39ab99cd263dfeeb5b0 and https://github.com/huggingface/huggingface_hub/pull/4189/changes/a471279a9c88b0b4caf594045200f1deb8b71fcb

src/huggingface_hub/hf_api.py
hanouticelina: yes my bad, I ran `ty` on the wrong branch I think 🙈

hanouticelina (approved):
Awesome 🔥  thanks again!

Wauplin:
Youhou! Merging :raised_hands:

PR #4201, opened 2026-05-07
Throttle progress bar updates in http_get to at most once per second

### LINKED ISSUE #4129: [BUG+Workaround]: HTTP download progress callback fires only once per ~10 MB, making progress bars look frozen on home connections
### Problem

This is about the HTTP transfer path (`HF_HUB_DISABLE_XET=1`). The xet path has a separate coarse-progress issue already tracked in #4058; I'm deliberately on HTTP because of that.

On HTTP, the progress callback passed via `tqdm_class=` on `hf_hub_download` / `snapshot_download` fires in very coarse intervals. On a ~1.5 MB/s home connection it fires roughly once every 7 seconds. A user watching a progress bar sees long silences between jumps (0 → 12 → 23 → …) and on small files may only see one update at the very end. The download is progressing fine — it just looks stalled. There's no env variable or public hook to change how often the callback fires.

### Workaround I'm using

At package import, before any HF call, and with `HF_HUB_DISABLE_XET=1`:

```python
from huggingface_hub import constants as _hf_constants
_hf_constants.DOWNLOAD_CHUNK_SIZE = 200 * 1024
```

With this, the callback fires several times per second at typical home-internet rates and progress animates smoothly. Overhead is negligible and it does not produce more network round-trips. I'm poking an internal constant, which isn't ideal, but it's the only lever I could find.

Sharing in case it helps anyone else running into the same thing.

### Suggested fix

Expose a way to make progress callbacks fire more frequently — env variable, a `progress_interval=` / `chunk_size=` kwarg on the download functions, or simply a more frequent default. Anything that doesn't require downstream code to reach into internals.

### Environment

- `huggingface_hub==1.11.0`
- Reproduces in any UI that renders the tqdm callback (Textual TUI, Jupyter, desktop GUIs).

wzshiming:
Fixes #4129

---

> [!NOTE]
> **Medium Risk**
> Touches core HTTP download streaming by changing chunk sizing and deferring `tqdm` updates, which could affect download performance and progress reporting behavior across clients.
> 
> **Overview**
> **Reduces progress bar overhead during streaming downloads.** In `http_get`, progress updates are now batched and emitted at most once per second (with a final flush on completion or on retry-worthy network errors) instead of updating on every received chunk.
> 
> **Changes download chunk sizing.** `DOWNLOAD_CHUNK_SIZE` is reduced from 10MB to 200KB, affecting how `httpx.Response.iter_bytes` buffers data during HTTP downloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e8a340d56454a1de25d79f2c7597f4958b95064. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Wauplin:
Will close the PR without merging as per https://github.com/huggingface/huggingface_hub/issues/4129#issuecomment-4395461819. We do not want to decrease the chunk size to 200kb which would slow down the download process on most machines.

PR #4211, opened 2026-05-08
[CLI] Surface job runtime fields in ps + inspect

davanstrien:
Not blocked by the upcoming `cli/jobs.py` `out.*` migration; this could land first
since it's purely additive (new fields + new column).

## Summary

Surfaces job runtime fields (`startedAt`, `finishedAt`, `durations`) in
`hf jobs ps` and `hf jobs inspect`. The server-side change that exposes these
fields in the jobs API response landed recently; this is the CLI counterpart.

- `_jobs_api.py`: new `JobDurations` dataclass; three new fields on `JobInfo`
- `__init__.py`: re-export `JobDurations` alongside the existing job dataclasses
- `cli/jobs.py`: new `RUNTIME` column in `ps` (shows `running_secs`); `inspect` auto-surfaces via `asdict`
- `utils/_parsing.py`: new `format_duration` helper (sibling of `format_timesince`)
- Tests: parse tests + CLI render test + `format_duration` unit tests

## Test plan

- [x] Parse tests (full / errored-before-start / terminal-no-finished-at payloads)
- [x] RUNTIME column render test
- [x] `format_duration` unit tests across all branches
- [x] Smoke-tested locally against real COMPLETED, RUNNING, SCHEDULING, CANCELED, ERROR jobs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Mostly additive, but extends the public Jobs API surface (`JobStage` enum and `JobInfo` fields) which could affect downstream code that assumes a fixed set of stages/fields or relies on dataclass serialization.
> 
> **Overview**
> Surfaces server-provided job runtime metadata end-to-end by extending `JobInfo` with `started_at`, `finished_at`, and `durations` (new `JobDurations` dataclass) and adding the `SCHEDULING` stage to `JobStage`.
> 
> Updates `hf jobs ps` to include a **RUNTIME** column derived from `durations.running_secs` using a new `format_duration` helper, and re-exports `JobDurations` from `huggingface_hub.__init__`. Adds parsing/unit tests to cover full/partial/absent duration payloads and CLI table rendering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0644af8e6eb995bec08351bc2123bd4ae79a9ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4212, opened 2026-05-08
[CLI] Parse `initiator` field on jobs API responses

davanstrien:
Independent of #4211 (off main, no merge-order dependency).

## Summary

Parses the `initiator` field the Hub returns on jobs API responses. Adds a
`JobInitiator` dataclass + `JobInfo.initiator` field. Pure data-layer
addition — no `ps` table column in this PR.

- `_jobs_api.py`: new `JobInitiator` dataclass + new `JobInfo.initiator` field
- `__init__.py`: re-export `JobInitiator` alongside the existing job dataclasses

Variants the Hub returns:
- `user` / `org` — ad-hoc runs (`name` populated)
- `scheduled-job` — cron-triggered runs (id only)
- `duplicated-job` — runs created via `POST /duplicate` (id only)

All share `type` and `id`; `name` is only set for user/org. Matches the
lightweight shape of the existing `JobOwner` (no `avatar_url` — keeps
`inspect` output tight). The server-side `avatarUrl` is dropped at the
call site rather than absorbed into the dataclass, matching how
`JobOwner` / `JobStatus` are constructed in this file.

## Why

Useful for both human users (knowing at-a-glance what triggered a job in
`inspect` output) and agents inspecting a user's job history programmatically.
When scanning many jobs from the same user, `initiator` is the only way to
disambiguate manual runs from automated/scheduled ones. The data was already
available server-side; the Python library was silently dropping it on parse.

## Where it surfaces

| CLI surface | Visible? |
|---|---|
| `hf jobs inspect <id>` | ✅ full block |
| `hf jobs ps --format json` | ✅ full block (JSON-parseable for agents) |
| `hf jobs ps` (default table) | ❌ no column added |

## Sample output

Two real jobs run by the same user — one ad-hoc, one cron-triggered. Note how
`initiator` disambiguates them while `owner` is identical:

```console
$ hf jobs inspect 6a01d697aff1cd33e8f33a64
[
    {
        "id": "6a01d697aff1cd33e8f33a64",
        "status": {"stage": "COMPLETED", "message": null},
        ...
        "owner": {"id": "60107b385ac3e86b3ea4fc34", "name": "davanstrien", "type": "user"},
        "initiator": {"type": "user", "id": "60107b385ac3e86b3ea4fc34", "name": "davanstrien"}
        ...
    }
]

$ hf jobs inspect 6a01c8c2aff1cd33e8f339d4
[
    {
        "id": "6a01c8c2aff1cd33e8f339d4",
        "status": {"stage": "COMPLETED", "message": null},
        ...
        "owner": {"id": "60107b385ac3e86b3ea4fc34", "name": "davanstrien", "type": "user"},
        "initiator": {"type": "scheduled-job", "id": "69c4179671691dc46f166314", "name": null}
        ...
    }
]
```

For cron-triggered runs, `initiator.type == "scheduled-job"` and
`initiator.id` points at the schedule definition. Agents can filter the full
list directly off `--format json`:

```console
$ hf jobs ps -a --format json | jq '[.[] | select(.initiator.type == "scheduled-job")] | length'
63
```

## Open question for reviewers

Should we add an `INITIATOR` column to the default `hf jobs ps` table?
- **For**: glance-scan UX — "which of these are cron vs manual?"
- **Against**: column-count creep — `ps` is already 5 cols (6 if #4211 lands),
  150+ chars wide on standard terminals
- **Adjacent UX problem**: active runs buried under historical ERROR/COMPLETED
  rows — better addressed by a `--sort status` flag or `--active-only` shortcut

Kept this PR minimal so the display decision can have its own discussion.

## Test plan

- [x] Parametrized parse tests for all four variants (user / org / scheduled-job / duplicated-job)
- [x] Default-None test (when `initiator` field absent)
- [x] Smoke-tested locally against real jobs — `user` and `scheduled-job` initiators both parse; server-side `avatarUrl` extra is silently filtered

---

> [!NOTE]
> **Low Risk**
> Low risk: adds an optional parsed field to `JobInfo` and re-exports a new dataclass, with minimal impact on existing job APIs unless callers rely on strict attribute sets/serialization.
> 
> **Overview**
> Adds support for the Jobs API `initiator` field by introducing a new `JobInitiator` dataclass and wiring it into `JobInfo` as an optional `initiator` attribute during response parsing.
> 
> Updates public exports in `__init__.py` (lazy/static imports and `__all__`) so `JobInitiator` is available at the package top level.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 460db26216b2298da633ddfc0b763b8640128331. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lhoestq (approved):

src/huggingface_hub/_jobs_api.py
```diff
@@ -54,6 +54,13 @@ class JobOwner:
     type: str
 
 
+@dataclass
+class JobInitiator:
+    type: str
+    id: str
+    name: str | None = None
+
```
hanouticelina, line 71: since it's public, let's have a docstring
```suggestion
class JobInitiator:
    """
    Contains information about what triggered a Job.
    
    Args:
        type (`str`): Initiator kind, for example `"user"`, `"org"`, `"scheduled-job"`, or `"duplicated-job"`.
        id (`str`): Identifier of the initiator.
        name (`str` or `None`): Human-readable name when available, usually for user/org initiators.
    """
    type: str
    id: str
    name: str | None = None

```

tests/test_cli.py
```diff
@@ -2889,6 +2889,83 @@ def _make_mock_jobs(self):
...
+            command=["python", "-c", "print('x')"],
+            arguments=[],
+            environment={},
+            secrets={},
+            flavor="cpu-basic",
+            labels={},
+            status={"stage": "RUNNING"},
+            owner={"id": "u", "name": "test", "type": "user"},
+        )
+        assert job.initiator is None
```
hanouticelina: these tests aren't CLI tests so they don't belong in this file. I'd remove them to be honest since they feel a bit sloppy, every variant in the `parametrize` hits the same line with a different string, so there's no real behavior being tested

hanouticelina (commented):
Thank you!

tests/test_cli.py
davanstrien: removed these now

Wauplin (approved):
Thank you!

PR #4228, opened 2026-05-15
[Cache] Detect Lustre/GPFS/NFS in `WeakFileLock` and fall back to `SoftFileLock`

vadiklyutiy:
## Problem

On Lustre, GPFS, and many NFS mounts, `fcntl.flock(2)` returns success for every caller without blocking — so `WeakFileLock` is a silent no-op on shared HF caches. Concurrent `hf_hub_download` cycles can interleave `O_APPEND` writes to `<etag>.incomplete`, producing a syntactically-valid but semantically-wrong blob (or a transiently-empty `refs/<revision>`).

The downstream symptom seen on multi-process `vllm serve` jobs sharing `HF_HOME` on Lustre is a sporadic `transformers.AutoConfig.from_pretrained` crash on one of N siblings: `Unrecognized model … Should have a "model_type" key in its config.json`. The same `config.json` was read successfully ~50 ms earlier in the same process, and 7 of 8 sibling processes parse it fine — a classic race.

The current code only falls back to `SoftFileLock` when `filelock` raises `NotImplementedError`, which Lustre/GPFS never do.

Related: [`tox-dev/filelock#389`](https://github.com/tox-dev/filelock/issues/389), [`huggingface/transformers#30859`](https://github.com/huggingface/transformers/issues/30859).

## Solution

`WeakFileLock` now decides between native `FileLock` and `SoftFileLock` up-front, via two layers in `src/huggingface_hub/utils/_fixes.py`:

1. **Primary: `statfs(2)` family detection** — Lustre, GPFS, NFS, SMB, CIFS magic numbers route to `SoftFileLock`. Cross-node-correct, no IPC needed.
2. **Secondary: fail-closed runtime probe** for unknown filesystems — spawns a child process (`multiprocessing.get_context("spawn")` for pickle-safety) that re-opens a probe file and tries `LOCK_EX|LOCK_NB` while the parent still holds the lock. Any timeout, exception, or non-`"ok"` result → `SoftFileLock`.

Both helpers are `@functools.lru_cache`'d per cache directory.

Two opt-out env vars cover edge cases:

- `HF_HUB_USE_SOFT_FILELOCK=1` — force soft regardless of detection (for exotic FUSE mounts that wrap an unsafe backend).
- `HF_HUB_FORCE_FLOCK=1` — force native for sites that mount Lustre with cluster-wide `flock` support and want the small perf benefit.

The legacy `NotImplementedError` → `SoftFileLock` fallback inside the acquisition loop is preserved as belt-and-suspenders.

## Tests

New `TestSoftLockDetection` (10 cases):

- 5 parametrized known-broken magics (Lustre / GPFS / NFS / SMB / CIFS) → asserts `SoftFileLock`.
- Local FS via `tmp_path` (under `/tmp`, ext4) → asserts native `FileLock`; exercises the real probe end-to-end.
- `HF_HUB_USE_SOFT_FILELOCK=True` overrides a healthy magic → `SoftFileLock`.
- `HF_HUB_FORCE_FLOCK=True` overrides a Lustre magic → `FileLock`.
- Probe failure with unknown magic (`0`) → `SoftFileLock` (verifies fail-closed).

Results:

- `pytest tests/test_utils_fixes.py -v` → **14 passed** (10 new + 4 pre-existing).
- `pytest tests/test_local_folder.py` → **13 passed, 2 skipped**.
- `pytest tests/test_file_download.py` → bit-identical to `origin/main` baseline (failures are sandbox `HF_HUB_OFFLINE` network restrictions, not regressions; confirmed via `git stash` + rerun).
- `make style` and `make quality` → clean, including `ty check src`.
- `mypy src` (CI parity) → `Success: no issues found in 176 source files`.

Real-world detection on the dev box, which happens to use NFS for `$HOME`:

- `_fs_magic("/home/scratch.vgimpelson_ent/huggingface_hub")` → `0x6969` (NFS) → `_should_use_soft_lock` returns `True`.
- `_fs_magic("/tmp")` → `0xef53` (ext4) → `_should_use_soft_lock` returns `False`.

Exactly the cases the fix is designed to discriminate.

---

> [!NOTE]
> **Medium Risk**
> Changes core Hub cache locking behavior on POSIX and shared/network filesystems; wrong detection could hurt performance (unnecessary soft locks) or, if forced native flock is misused, reintroduce rare concurrent-write corruption.
> 
> **Overview**
> Fixes **silent cache corruption** when multiple processes share `HF_HUB_CACHE` on parallel filesystems where `flock(2)` appears to succeed but does not serialize writers.
> 
> `WeakFileLock` now **chooses `SoftFileLock` vs native `FileLock` before acquisition**, instead of only falling back on `NotImplementedError`. Detection uses **`statfs(2)` magic numbers** for Lustre, GPFS, NFS, SMB, and CIFS, plus a **fail-closed child-process probe** for unknown POSIX filesystems. Decisions are cached per cache directory.
> 
> New knobs: **`HF_HUB_USE_SOFT_FILELOCK`** (always soft) and **`HF_HUB_FORCE_FLOCK`** (always native when cluster-wide flock is known good). Constants and docs were added; **`TestSoftLockDetection`** covers magics, local FS, env overrides, and probe failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 748df016e2b4ba70453660c503cf2b1a953ff686. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/package_reference/environment_variables.md
+++ b/docs/source/en/package_reference/environment_variables.md
@@ -84,7 +84,7 @@ Integer value to define the number of seconds to wait for server response when f
 
 Integer value to define the number of seconds to wait for server response when downloading a file. If the request times out, a TimeoutError is raised. Setting a higher value is beneficial on machine with a slow connection. A smaller value makes the process fail quicker in case of complete network outage. Default to 10s.
 
-## Xet 
+## Xet
 
 ### Other Xet environment variables
 * [`HF_HUB_DISABLE_XET`](../package_reference/environment_variables#hfhubdisablexet)
@@ -100,13 +100,13 @@ Defaults to `0` (0 bytes, means chunk cache is disabled).
 
 ### HF_XET_SHARD_CACHE_SIZE_LIMIT
 
-To set the size of the Xet shard cache locally. Increasing this will improve upload efficiency as chunks referenced in cached shard files are not re-uploaded. Note that the default soft limit is likely sufficient for most workloads. 
+To set the size of the Xet shard cache locally. Increasing this will improve upload efficiency as chunks referenced in cached shard files are not re-uploaded. Note that the default soft limit is likely sufficient for most workloads.
 
 Defaults to `4000000000` (4GB).
 
 ### HF_XET_NUM_CONCURRENT_RANGE_GETS
 
-To set the number of concurrent terms (range of bytes from within a xorb, often called a chunk) downloaded from S3 per file. Increasing this will help with the speed of downloading a file if there is network bandwidth available. 
+To set the number of concurrent terms (range of bytes from within a xorb, often called a chunk) downloaded from S3 per file. Increasing this will help with the speed of downloading a file if there is network bandwidth available.
 
 Defaults to `16`.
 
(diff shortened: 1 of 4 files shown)


Wauplin:
Hi @vadiklyutiy , have you run into this issue yourself?

As mentioned by `@gaborbernat` in https://github.com/tox-dev/filelock/issues/389#issuecomment-2603010247 this really sound like a `fnlock` Python bug that should be solve in there rather than applying a weak patch in a downstream library (this PR). Also this PR feels heavily AI-generated, which is not a bad thing in itself but without a proper setup to test the PR I'm reluctant to review/merge it. I'll therefore close it for now but happy to reopen if you think it's necessary.

vadiklyutiy:
Hi @Wauplin 

First about 

> have you run into this issue yourself?

and 

> Also this PR feels heavily AI-generated, which is not a bad thing in itself but without a proper setup to test the PR I'm reluctant to review/merge it.

This is a real issue we encountered during large-scale vLLM testing at NVIDIA. Across approximately 1,000 runs, with dozens running concurrently, we observed 3–7 such failures.

Yes, I used AI, especially for writing tests, but I fully understand the purpose of the changes and have personally reviewed every line of code.

This PR is not about identifying a potential issue that is likely not real, asking Claude to fix it, and then pushing a PR just to add a contribution to `huggingface_hub` to my resume. You can simply check my public contributions on GitHub; it should be clear that I do not need that.

Second about 

>As mentioned by @gaborbernat in https://github.com/tox-dev/filelock/issues/389#issuecomment-2603010247 this really sound like a fnlock Python bug that should be solve in there rather than applying a weak patch in a downstream library (this PR)

I'd point to the next comment in the same issue https://github.com/tox-dev/filelock/issues/389#issuecomment-2683039912. That describe that this is a real issue and his workaround to patch filelock by changing `flock` to `lockf`

```
sed -i "s/fcntl.flock/fcntl.lockf/" site-packages/filelock/_unix.py
```

helped in that case. 

Also in this https://github.com/huggingface/transformers/issues/30859#issuecomment-2590117684 comment `SoftFileLock` was a solution/workaround. 

> I'll therefore close it for now but happy to reopen if you think it's necessary.

I propose to reopen it. We of course can ping pong the issue between `filelock` or `pyhton` but this is real issue.  

`huggingface_hub` has a ready interfaces `SoftFileLock` that I use here.


The main change here is 
https://github.com/vadiklyutiy/huggingface_hub/blob/filelock-lustre-gpfs-fallback/src/huggingface_hub/utils/_fixes.py#L248-L251

where we choose using `SoftFileLock` for network file systems. 

All other changes is accurate implementation of `_should_use_soft_lock` - checking underlying file system.

Wauplin:
Hey @vadiklyutiy , thanks for the answer and added details. Sorry for the initial closing, we receive a lot of contributions and it's sometimes hard to determine if people actually ran into the issue they are reporting. So yes, let's reopen this PR and see what we can do about it!

Wauplin:
Hi @vadiklyutiy, after some investigation and local debugging, I came to the conclusion that the safest bet is to never resume from an incomplete file. This should solve your issue at the cost of introducing a breaking change in the current behavior. However this is not too much of a problem since `hf_xet` do not allow resuming from an incomplete file so for now it was more of a "legacy nice to have" rather than a useful feature.

I push a separate PR (https://github.com/huggingface/huggingface_hub/pull/4306) with the fix. Do you mind taking a look and let me know what you think? Would be even better if you could test it on a real setup on your side to confirm it resolves your problem. Thanks in advance! :hugs: 

(once #4306 gets merged, I'll close this PR)

vadiklyutiy:
@Wauplin 
#4306 looks good. Thank you for fixing. I will test it.

PR #4234, opened 2026-05-19
[Auth] Harden HF_TOKEN_PATH and HF_STORED_TOKENS_PATH to 0o600 / 0o700

ghost:
## Summary

`huggingface_hub` writes the user's HF token files at Python's default file modes (0o666 masked by umask 022 → 0o644 on macOS Terminal and most Linux defaults), leaving them world-readable. Two writer sites:

1. `src/huggingface_hub/_login.py:441-445` — `Path.write_text(token)` for the single active token at `HF_TOKEN_PATH` (~/.cache/huggingface/token).
2. `src/huggingface_hub/utils/_auth.py:157-167` — `_save_stored_tokens` writes the multi-account INI to `HF_STORED_TOKENS_PATH` (~/.cache/huggingface/stored_tokens). Worst case impact-wise — the file contains every named token the user has saved via `huggingface-cli auth login --token` (personal + work + organization).

Any local account or process that can traverse the home directory can recover these tokens and act as the user against the Hub API — pulling or pushing private models and datasets, running Inference Endpoint predictions on the user's credits, accessing private orgs.

This PR adds `mode=0o700` to the parent `mkdir` and `path.chmod(0o600)` / `path.parent.chmod(0o700)` after each write so pre-existing installations converge to safe perms on the next save. The chmod calls are wrapped in try/except for Windows (no POSIX modes).

## How to verify

A small Python reproducer that mirrors the pre-PR call shape:

```python
from pathlib import Path

p = Path('/tmp/hf-perm-test/token')
p.parent.mkdir(parents=True, exist_ok=True)
p.write_text('hf_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx')
print('file mode (pre-PR):', oct(p.stat().st_mode & 0o777))
print('dir mode (pre-PR):', oct(p.parent.stat().st_mode & 0o777))
```

Pre-PR: `file mode (pre-PR): 0o644 / dir mode (pre-PR): 0o755`. After applying the PR: `0o600 / 0o700`.

## Industry baseline

GitHub CLI (`~/.config/gh/hosts.yml`), AWS CLI (`~/.aws/credentials`), Google Cloud SDK (`~/.config/gcloud/credentials.db`), Stripe CLI, PlanetScale CLI (keyring + 0o600 fallback), Pulumi (`StoreCredentials` uses `lockedfile.Write` at 0o600).

## Related disclosure

Reported privately to `security@huggingface.co` per `.well-known/security.txt` on 2026-05-19. Filing the PR in parallel to give maintainers the ready-to-merge fix.

---

> [!NOTE]
> **Medium Risk**
> Touches credential persistence paths (`HF_TOKEN_PATH`/`HF_STORED_TOKENS_PATH`) and changes how secrets are written to disk; while small, mistakes could break login persistence or permissions across platforms.
> 
> **Overview**
> **Hardens on-disk token persistence** by introducing `_write_secret` to write Hub tokens with restrictive permissions (directory `0o700`, file `0o600`) on POSIX systems.
> 
> The login flow now uses `_write_secret` instead of `Path.write_text` for `HF_TOKEN_PATH`, and stored-tokens saving is refactored to write the INI content via an in-memory buffer and `_write_secret` for `HF_STORED_TOKENS_PATH`, with best-effort `chmod` handling on Windows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fca196090135dc230683f35950eb3e4aeca75cb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_login.py
+++ b/src/huggingface_hub/_login.py
@@ -438,10 +438,20 @@ def _set_active_token(
             )
         else:
             logger.warning("Token has not been saved to git credential helper.")
-    # Write token to HF_TOKEN_PATH
+    # Write token to HF_TOKEN_PATH. The token is sensitive — restrict the file
+    # and its parent directory to the user (0o600 / 0o700) so a co-tenant on the
+    # same host can't recover it via home-dir traversal. chmod is wrapped in
+    # try/except because Windows doesn't support POSIX modes; the mkdir mode
+    # arg likewise only applies to a freshly-created directory, so we chmod the
+    # parent again to handle the case where the cache dir already existed.
     path = Path(constants.HF_TOKEN_PATH)
-    path.parent.mkdir(parents=True, exist_ok=True)
+    path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
     path.write_text(token)
+    try:
+        path.chmod(0o600)
+        path.parent.chmod(0o700)
+    except (OSError, NotImplementedError):
+        pass
     logger.info(f"Your token has been saved to {constants.HF_TOKEN_PATH}")
 
 
--- a/src/huggingface_hub/_login.py
+++ b/src/huggingface_hub/_login.py
@@ -439,10 +439,7 @@ def _set_active_token(
             )
         else:
             logger.warning("Token has not been saved to git credential helper.")
-    # Write token to HF_TOKEN_PATH. Route through _write_secret so the file
-    # is atomically created at 0o600 (no TOCTOU window where the token is
-    # world-readable between create and chmod) and the parent directory at
-    # 0o700.
+    # Write token to HF_TOKEN_PATH
     _write_secret(Path(constants.HF_TOKEN_PATH), token)
     logger.info(f"Your token has been saved to {constants.HF_TOKEN_PATH}")
 
--- a/src/huggingface_hub/_login.py
+++ b/src/huggingface_hub/_login.py
@@ -40,6 +40,7 @@ from .utils._auth import (
     _get_token_from_google_colab,
     _save_stored_tokens,
     _save_token,
+    _write_secret,
     get_stored_tokens,
 )
 
@@ -438,20 +439,11 @@ def _set_active_token(
             )
         else:
             logger.warning("Token has not been saved to git credential helper.")
-    # Write token to HF_TOKEN_PATH. The token is sensitive — restrict the file
-    # and its parent directory to the user (0o600 / 0o700) so a co-tenant on the
-    # same host can't recover it via home-dir traversal. chmod is wrapped in
-    # try/except because Windows doesn't support POSIX modes; the mkdir mode
-    # arg likewise only applies to a freshly-created directory, so we chmod the
-    # parent again to handle the case where the cache dir already existed.
-    path = Path(constants.HF_TOKEN_PATH)
-    path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
-    path.write_text(token)
-    try:
-        path.chmod(0o600)
-        path.parent.chmod(0o700)
-    except (OSError, NotImplementedError):
-        pass
+    # Write token to HF_TOKEN_PATH. Route through _write_secret so the file
+    # is atomically created at 0o600 (no TOCTOU window where the token is
+    # world-readable between create and chmod) and the parent directory at
+    # 0o700.
+    _write_secret(Path(constants.HF_TOKEN_PATH), token)
     logger.info(f"Your token has been saved to {constants.HF_TOKEN_PATH}")
 
 
--- a/src/huggingface_hub/utils/_auth.py
+++ b/src/huggingface_hub/utils/_auth.py
@@ -14,6 +14,7 @@
 """Contains a helper to get the token from machine (env variable, secret or config file)."""
 
 import configparser
+import io
 import logging
 import os
 import warnings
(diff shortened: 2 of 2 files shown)


Wauplin:
Hi @JAE0Y2N thanks for reporting and fixing this! This looks like a valid report indeed. Could you move the "write content to a secret file" logic into its own helper to avoid duplicating logic? (+fix the TOCTOU issue as described in https://github.com/huggingface/huggingface_hub/pull/4234#discussion_r3265823127, though I agree it's a very limited vulnerability scope)


```py
_SECRET_FILE_MODE = 0o600
_SECRET_DIR_MODE = 0o700

def _write_secret(path: Path, content: str) -> None:
    """Write content to path, restricting to owner-only on POSIX."""
    path.parent.mkdir(parents=True, exist_ok=True, mode=_SECRET_DIR_MODE)
    fd = os.open(str(path), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, _SECRET_FILE_MODE)
    with os.fdopen(fd, "w") as f:
        f.write(content)
    try:
        path.chmod(_SECRET_FILE_MODE)  # fix perms if file already existed
        path.parent.chmod(_SECRET_DIR_MODE)  # fix perms if dir already existed
    except (OSError, NotImplementedError):
        pass
```

ghost:
Thanks for the fast review @Wauplin! Pushed b38123df with your suggested helper.

Changes in this push:
- Added `_write_secret(path, content)` in `utils/_auth.py` using `os.open(O_WRONLY | O_CREAT | O_TRUNC, 0o600)` so the file is atomically created at the restricted mode — no brief window at 0o644 between create and chmod.
- The post-write `chmod` is kept (wrapped in `try/except (OSError, NotImplementedError)`) so pre-existing files/dirs at looser perms get tightened on the next save. Comment in-code explains both halves.
- `_login.py` now calls `_write_secret(Path(constants.HF_TOKEN_PATH), token)` — single line replacing the 8-line block.
- `utils/_auth.py::_save_stored_tokens` renders the INI into a `io.StringIO()` via `config.write(buf)` then calls `_write_secret(stored_tokens_path, buf.getvalue())`.

Empirical verification on macOS 14:

```
file mode: 0o600
dir mode: 0o700
PASS: file=0o600 dir=0o700 verified
```

(Standalone reproducer ran the extracted helper against a fresh path under `tempfile.TemporaryDirectory()` and asserted the modes.)

Let me know if you'd like further tweaks!

src/huggingface_hub/utils/_auth.py
```diff
@@ -24,6 +25,35 @@
...
+
+    Uses `os.open(... O_CREAT, 0o600)` so the file is atomically created with
+    restricted permissions instead of being briefly world-readable between
+    `open("w")` and a subsequent `chmod` (a TOCTOU window the previous
+    `path.write_text(...)` + `chmod` implementation had).
+
+    Also tightens both the file and the parent directory after the write to
+    handle the case where they pre-existed at looser permissions (which
+    cannot be fixed by `O_CREAT` alone).
+    """
```
Wauplin:
```suggestion
    """Write content to file, restricting both the file and its parent
    directory to owner-only on POSIX systems."""
```

src/huggingface_hub/utils/_auth.py
```diff
@@ -156,15 +186,18 @@ def _save_stored_tokens(stored_tokens: dict[str, str]) -> None:
     """
     stored_tokens_path = Path(constants.HF_STORED_TOKENS_PATH)
 
-    # Write the stored tokens into an INI file
+    # The stored_tokens file contains every named HF token the user has saved
+    # (personal + work + organization). Render the INI to a string and route
+    # through _write_secret so the file is atomically created at 0o600 and the
+    # parent directory at 0o700.
```
Wauplin:
```suggestion
    # Write the stored tokens into an INI file
```

src/huggingface_hub/_login.py
```diff
@@ -438,10 +439,11 @@ def _set_active_token(
...
             logger.warning("Token has not been saved to git credential helper.")
-    # Write token to HF_TOKEN_PATH
-    path = Path(constants.HF_TOKEN_PATH)
-    path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(token)
+    # Write token to HF_TOKEN_PATH. Route through _write_secret so the file
+    # is atomically created at 0o600 (no TOCTOU window where the token is
+    # world-readable between create and chmod) and the parent directory at
+    # 0o700.
+    _write_secret(Path(constants.HF_TOKEN_PATH), token)
```
Wauplin:
```suggestion
    # Write token to HF_TOKEN_PATH
    _write_secret(Path(constants.HF_TOKEN_PATH), token)
```

src/huggingface_hub/utils/_auth.py
```diff
@@ -24,6 +25,35 @@
...
+    Also tightens both the file and the parent directory after the write to
+    handle the case where they pre-existed at looser permissions (which
+    cannot be fixed by `O_CREAT` alone).
+    """
+    path.parent.mkdir(parents=True, exist_ok=True, mode=_SECRET_DIR_MODE)
+    fd = os.open(str(path), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, _SECRET_FILE_MODE)
+    with os.fdopen(fd, "w") as f:
+        f.write(content)
+    try:
+        path.chmod(_SECRET_FILE_MODE)
```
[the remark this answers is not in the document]
Wauplin: not a problem IMO, let's not be too paranoiac

src/huggingface_hub/utils/_auth.py
```diff
@@ -24,6 +25,25 @@
 from ._runtime import is_colab_enterprise, is_google_colab
 
 
+_SECRET_FILE_MODE = 0o600
+_SECRET_DIR_MODE = 0o700
+
+
+def _write_secret(path: Path, content: str) -> None:
+    """Write content to file, restricting both the file and its parent
+    directory to owner-only on POSIX systems."""
```
Wauplin:
```suggestion
    """Write content to file, restricting both the file and its parent directory to owner-only on POSIX systems."""
```

Wauplin (approved):
Looks good, thank you!

(will wait a bit before merging in case someone else has some feedback)

dtrifiro:
Hello,

while I think this a useful improvement, this completely breaks some shared cache scenarios.

In our setup we have  `HF_HOME=/huggingface-cache` shared across different users.

The shared cache directory has its group set to `hf-cache` (`chown user:model-cache $HF_HOME`) and permissions set to `755`: each user belonging to this group can read from the shared hf cache.

Whenever `hf auth login` is run, the permissions are changed to 0700, causing the `HF_HOME` directory to become unreadable by the group are breaking our shared cache usecase.

cc @Wauplin

julien-c:
is there an env var to control specifically where the token is stored (independly of the cache)?

Wauplin:
> is there an env var to control specifically where the token is stored (independly of the cache)?

Yes, [`HF_TOKEN_PATH`](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hftokenpath). That might be the easiest fix here. On a shared cluster is good practice not to share a single auth token anyway.

@dtrifiro could you try that solution and let us know if it works? (i.e. point token path to user's cache while $HF_HOME is kept in shared cache

dtrifiro:
Thanks for the quick response.

Yeah that should work and prevent `_save_stored_tokens` and `_write_secret` to change permissions for the `HF_CACHE`, which solves our problem.

Another consideration: this would still change the permissions for the parent dir, which might be unexpected behaviour for user.

How about just plainly refusing to write the token to an insecure location and failing instead, kind of like ssh refuses to use an ssh private key file from an unsecured `~/.ssh/` directory. This would make sure users are aware of the security risks and take action, rather than silently changing permissions.

PR #4252, opened 2026-05-22
[Jobs] Add update_job_labels and update_scheduled_job_labels

Wauplin:
## Summary

Client-side support for label update endpoints added in https://github.com/huggingface-internal/moon-landing/pull/17932.

- Add `HfApi.update_job_labels()` and `HfApi.update_scheduled_job_labels()` (`PUT /api/jobs/:ns/:id/labels` and `PUT /api/scheduled-jobs/:ns/:id/labels`)
- Add CLI commands `hf jobs labels` and `hf jobs scheduled labels` with `--clear` flag to prevent accidental label deletion
- Docs + tests

```
>>> hf jobs labels <job_id> --label env=prod --label team=ml
>>> hf jobs labels <job_id> --clear
>>> hf jobs scheduled labels <id> --label env=staging
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk: adds new Jobs/Scheduled Jobs label update endpoints and CLI wiring without changing existing job execution logic; main risk is accidental label replacement/clearing, mitigated by explicit `--clear` and argument validation.
> 
> **Overview**
> Adds client support to **replace (or clear) user labels** on existing Jobs and Scheduled Jobs via new `HfApi.update_job_labels` and `HfApi.update_scheduled_job_labels` (PUT to `/labels` endpoints), and exports them at the package top-level.
> 
> Extends the CLI with `hf jobs labels` and `hf jobs scheduled labels`, including validation to require `--label` or `--clear` (but not both), and updates docs/CLI reference plus tests covering the new command behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2ef6ef04b3fba395202d040509a0792162dd0ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4254, opened 2026-05-22
 [CLI] migrate `hf jobs` to `out` singleton

hanouticelina:
Last piece of #3979. 
This PR migrates `hf jobs` to the `out` framework. 

breaking changes introduced in `hf jobs ps` and `hf jobs scheduled ps` (I think it's fine, the alternative would be special casing `_output.py` justfor these two list commands) :

  1.` --format` table is removed. Use `--format auto|human|agent|quiet|json`.
  2. json output shape changed for `hf jobs ps`:
    - `command` is now a string (e.g. "python script.py"), was a list (e.g.
  ["python", "script.py"]).
    - `status` is now a string (e.g. `"RUNNING"`), was a dict (e.g. ```{"stage": "RUNNING"```).
 

---

> [!NOTE]
> **Medium Risk**
> User-visible CLI breaking changes on `hf jobs ps` and `hf jobs scheduled ps` (format flags and JSON field shapes); behavior is otherwise presentation-layer only for jobs commands.
> 
> **Overview**
> Migrates the **`hf jobs`** command group to the shared **`out`** CLI output layer (`out.result`, `out.table`, `out.dict`, `out.text`, `out.warning`, `out.hint`) instead of ad-hoc `print` / `json.dumps` / `print_list_output`.
> 
> **`hf jobs ps`** and **`hf jobs scheduled ps`** now use **`--format auto|human|agent|json|quiet`** (default `auto`) with optional Go-style templates (`{{.id}}`). **`--format table`** and **`-q` / `--quiet`** are removed; quiet IDs come from **`--format quiet`**. For **`hf jobs ps`**, JSON output flattens **`command`** to a string and **`status`** to a stage string (not a list or nested object). Inspect/labels and lifecycle commands get consistent structured output; docs for the list commands are regenerated to match.
> 
> **`_output.Output.set_mode`** accepts string mode names (e.g. `'agent'`) in addition to the enum.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63b352d29c1a0b9030008d33442ec847efdb3ce8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/_output.py
+++ b/src/huggingface_hub/cli/_output.py
@@ -53,8 +53,13 @@ class Output:
         self.no_truncate = False
         self.set_mode()
 
-    def set_mode(self, mode: OutputFormatWithAuto = OutputFormatWithAuto.auto) -> None:
-        """Override the output mode (called once at startup and again per '--format' flag)."""
+    def set_mode(self, mode: OutputFormatWithAuto | str = OutputFormatWithAuto.auto) -> None:
+        """Override the output mode (called once at startup and again per '--format' flag).
+
+        Accepts the ``OutputFormatWithAuto`` enum or its string name (e.g. ``'agent'``, ``'json'``).
+        """
+        if isinstance(mode, str):
+            mode = OutputFormatWithAuto(mode)
         if mode == OutputFormatWithAuto.auto:
             mode = OutputFormatWithAuto.agent if is_agent() else OutputFormatWithAuto.human
         self.mode = mode
--- a/src/huggingface_hub/cli/_output.py
+++ b/src/huggingface_hub/cli/_output.py
@@ -53,13 +53,8 @@ class Output:
         self.no_truncate = False
         self.set_mode()
 
-    def set_mode(self, mode: OutputFormatWithAuto | str = OutputFormatWithAuto.auto) -> None:
-        """Override the output mode (called once at startup and again per '--format' flag).
-
-        Accepts the ``OutputFormatWithAuto`` enum or its string name (e.g. ``'agent'``, ``'json'``).
-        """
-        if isinstance(mode, str):
-            mode = OutputFormatWithAuto(mode)
+    def set_mode(self, mode: OutputFormatWithAuto = OutputFormatWithAuto.auto) -> None:
+        """Override the output mode (called once at startup and again per '--format' flag)."""
         if mode == OutputFormatWithAuto.auto:
             mode = OutputFormatWithAuto.agent if is_agent() else OutputFormatWithAuto.human
         self.mode = mode
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -613,13 +613,14 @@ def jobs_ps(
         filtered_jobs.append(job)
 
     # Build display items. Augment the raw api dict with curated, table-friendly columns
-    # (image, command, created, status, runtime) — these double as Go-template fields.
+    # — these double as Go-template fields and as table headers (via SCREAMING_SNAKE).
     items: list[dict[str, Any]] = []
     for job in filtered_jobs:
         item = api_object_to_dict(job)
         durations = item.get("durations") or {}
         cmd = item.get("command") or []
-        item["image"] = item.get("docker_image") or "N/A"
+        item["job_id"] = item.get("id", "")
+        item["image/space"] = item.get("docker_image") or "N/A"
         item["command"] = " ".join(cmd) if cmd else "N/A"
         item["created"] = item["created_at"][:19].replace("T", " ") if item.get("created_at") else "N/A"
         item["status"] = (item.get("status") or {}).get("stage", "UNKNOWN")
@@ -630,7 +631,11 @@ def jobs_ps(
         _render_template(format, items)
         return
 
-    out.table(items, headers=["id", "image", "command", "created", "status", "runtime"], id_key="id")
+    out.table(
+        items,
+        headers=["job_id", "image/space", "command", "created", "status", "runtime"],
+        id_key="job_id",
+    )
     if not items and filters:
         filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
         out.hint(f"No jobs matched filters: {filters_msg}")
@@ -877,8 +882,8 @@ def scheduled_ps(
             continue
         filtered_jobs.append(scheduled_job)
 
-    # Build display items. Augment with curated columns (image, command, last, next) that also
-    # serve as Go-template fields.
+    # Build display items. Augment with curated columns that also serve as Go-template fields
+    # and as table headers (via SCREAMING_SNAKE).
     items: list[dict[str, Any]] = []
     for sj in filtered_jobs:
         item = api_object_to_dict(sj)
@@ -886,10 +891,10 @@ def scheduled_ps(
         status_dict = item.get("status") or {}
         last_job = status_dict.get("last_job")
         cmd = job_spec.get("command") or []
-        item["image"] = job_spec.get("docker_image") or "N/A"
+        item["image/space"] = job_spec.get("docker_image") or "N/A"
         item["command"] = " ".join(cmd) if cmd else "N/A"
-        item["last"] = last_job["at"][:19].replace("T", " ") if last_job and last_job.get("at") else "N/A"
-        item["next"] = (
+        item["last_run"] = last_job["at"][:19].replace("T", " ") if last_job and last_job.get("at") else "N/A"
+        item["next_run"] = (
             status_dict["next_job_run_at"][:19].replace("T", " ") if status_dict.get("next_job_run_at") else "N/A"
         )
         item["suspend"] = item.get("suspend") or False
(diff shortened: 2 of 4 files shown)


src/huggingface_hub/cli/_output.py
```diff
@@ -53,8 +53,13 @@ def __init__(self) -> None:
...
 
-    def set_mode(self, mode: OutputFormatWithAuto = OutputFormatWithAuto.auto) -> None:
-        """Override the output mode (called once at startup and again per '--format' flag)."""
+    def set_mode(self, mode: OutputFormatWithAuto | str = OutputFormatWithAuto.auto) -> None:
+        """Override the output mode (called once at startup and again per '--format' flag).
+
+        Accepts the ``OutputFormatWithAuto`` enum or its string name (e.g. ``'agent'``, ``'json'``).
+        """
+        if isinstance(mode, str):
+            mode = OutputFormatWithAuto(mode)
```
hanouticelina: this change is only to keep backward compatibility with the Go-template formatting

src/huggingface_hub/cli/jobs.py
```diff
@@ -391,21 +383,47 @@ def _matches_filters(job_properties: dict[str, str], filters: list[tuple[str, st
...
+
+
+FormatWithTemplateOpt = Annotated[
+    str,
+    typer.Option(
+        "--format",
+        help="Output format: auto|human|agent|json|quiet, or a Go-template string (e.g. '{{.id}}').",
+        callback=_set_jobs_format,
+    ),
+]
```
hanouticelina: this is maybe the cleanest way to avoid breaking the Go-template formatting, wdyt @Wauplin ?

Wauplin: What about dropping the Go-template formatting altogether? I don't expect it to be used at all + it would make the `hf jobs` consistent with all other commands that do not support it. Then if we want we can open a separate to add template formatting to all commands at once we some logic inside the `out` singleton instead. IMO it'd be a much cleaner solution than having ad-hoc logic here

src/huggingface_hub/cli/_output.py
Wauplin: let's remove it (see comment above)

src/huggingface_hub/cli/jobs.py
```diff
@@ -313,15 +308,12 @@ def jobs_run(
         timeout=timeout,
         namespace=namespace,
     )
-    # Always print the job ID to the user
-    print(f"Job started with ID: {job.id}")
-    print(f"View at: {job.url}")
-
+    out.result("Job started", id=job.id, url=job.url)
     if detach:
+        out.hint(f"Use `hf jobs logs {job.id}` to fetch the logs.")
```
Wauplin, line 312: 🙌

src/huggingface_hub/cli/jobs.py
```diff
@@ -579,7 +579,6 @@ def jobs_ps(
                     label_key, label_value = label_part.split("=", 1)
                 else:
                     label_key, label_value = label_part, "*"
-                # Negate predicate in case of key!=value
```
Wauplin, line 582: let's keep the comment

src/huggingface_hub/cli/jobs.py
```diff
@@ -588,15 +587,14 @@ def jobs_ps(
             labels_filters.append((label_key.lower(), op, label_value.lower()))
         elif "=" in f:
             key, value = f.split("=", 1)
-            # Negate predicate in case of key!=value
```
Wauplin, line 591: same

src/huggingface_hub/cli/jobs.py
```diff
@@ -614,79 +612,61 @@ def jobs_ps(
...
+        item["image/space"] = item.get("docker_image") or "N/A"
+        item["command"] = " ".join(cmd) if cmd else "N/A"
+        item["created"] = item["created_at"][:19].replace("T", " ") if item.get("created_at") else "N/A"
+        item["status"] = (item.get("status") or {}).get("stage", "UNKNOWN")
+        item["runtime"] = format_duration(durations.get("running_secs"))
+        items.append(item)
+
+    if _is_template(format):
+        _render_template(format, items)
+        return
```
Wauplin: removing suppoer for go-template formatting would avoid this step

Wauplin (commented):
Thanks for migrating this! Most comments are related to templating, which I think we should drop for the sake of simplicity

Wauplin:
Merged https://github.com/huggingface/huggingface_hub/pull/4265 to use `out` singleton  in labels commands as well (related to https://github.com/huggingface/huggingface_hub/pull/4252#pullrequestreview-4362206518)

src/huggingface_hub/cli/jobs.py
hanouticelina: yes agree! removed in https://github.com/huggingface/huggingface_hub/pull/4254/changes/c12a350c1b612e275831342b2328ec0b34811c30

src/huggingface_hub/cli/_output.py
hanouticelina: done in https://github.com/huggingface/huggingface_hub/pull/4254/changes/c12a350c1b612e275831342b2328ec0b34811c30

hanouticelina:
@Wauplin thanks for the review! addressed your comments in https://github.com/huggingface/huggingface_hub/pull/4254/changes/c12a350c1b612e275831342b2328ec0b34811c30

src/huggingface_hub/cli/jobs.py
Wauplin: use out.text ?

src/huggingface_hub/cli/jobs.py
```diff
@@ -499,7 +463,7 @@ def jobs_stats(
                 rows_per_job_id[job_id] = [row]
             last_update_time = time.time()
             total_rows = [row for job_id in rows_per_job_id for row in rows_per_job_id[job_id]]
-            _print_output(total_rows, table_headers, headers_aliases, None)
+            print(_tabulate(total_rows, headers=table_headers))
```
Wauplin, line 467: is there a reason for using `print` instead of `out`? if yes, would be great to add a comment

src/huggingface_hub/cli/jobs.py
```diff
@@ -518,7 +482,7 @@ def jobs_stats(
                 if now - last_update_time >= STATS_UPDATE_MIN_INTERVAL:
                     _clear_line(2 + len(total_rows))
                     total_rows = [row for job_id in rows_per_job_id for row in rows_per_job_id[job_id]]
-                    _print_output(total_rows, table_headers, headers_aliases, None)
+                    print(_tabulate(total_rows, headers=table_headers))
```
Wauplin, line 486: same here

src/huggingface_hub/cli/jobs.py
Wauplin, line 495: hmm, just realized about that part... 😕  (+ in many other places in the module)

src/huggingface_hub/cli/jobs.py
```diff
@@ -614,79 +570,56 @@ def jobs_ps(
...
+        items.append(item)
+
+    out.table(
+        items,
+        headers=["job_id", "image/space", "command", "created", "status", "runtime"],
+        id_key="job_id",
+    )
+    if not items and filters:
+        filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
+        out.hint(f"No jobs matched filters: {filters_msg}")
```
```suggestion
        out.text(f"No jobs matched filters: {filters_msg}")
```

-slight preference for a `.text` rather than a `.hint` since it's informative

src/huggingface_hub/cli/jobs.py
```diff
@@ -718,6 +650,7 @@ def jobs_inspect(
             raise CLIError("Access denied. You may not have permission to view this job.") from e
         else:
             raise CLIError(f"Failed to inspect job: {e}") from e
+    out.dict([api_object_to_dict(job) for job in jobs])
```
Wauplin: should be a proper table IMO even if it will be a breaking change (more consistent and allows --json etc.)

src/huggingface_hub/cli/jobs.py
```diff
@@ -1006,7 +912,7 @@ def scheduled_inspect(
         api.inspect_scheduled_job(scheduled_job_id=scheduled_job_id, namespace=namespace)
         for scheduled_job_id in scheduled_job_ids
     ]
-    print(json.dumps([asdict(scheduled_job) for scheduled_job in scheduled_jobs], indent=4, default=str))
+    out.dict([api_object_to_dict(scheduled_job) for scheduled_job in scheduled_jobs])
```
Wauplin: `out.table`?

src/huggingface_hub/cli/jobs.py
Wauplin: (ah yes probably because of the multiline + regular polling. Then keeping "print" + a comment is fine I guess)

hanouticelina: ^ yes exactly

Wauplin (approved):
Thank you!

PR #4266, opened 2026-05-26
[Jobs] Decouple Job hardware from Spaces, auto-sync enums with Hub API

Wauplin:
## Summary

Jobs hardware was previously tied to `SpaceHardware` — the same enum, same values, same code paths. As the two diverge (e.g. Jobs don't support ZeroGPU, Spaces don't have the same GPU catalog), this coupling causes friction in the UX with badly documented commands.

In addition, the hardcoded enum means every infra change requires a manual PR like #4259.

This PR decouples them and makes future hardware updates fully automated:

- **Own enum for Jobs** — `JobHardware(str, Enum)` in `_jobs_api.py`, independent of `SpaceHardware`. The existing `JobHardware` dataclass (return type of `list_jobs_hardware()`) is renamed to `JobHardwareInfo` to free the name. :warning: this is a breaking change but perfectly fine IMO
- **Future-proof CLI** — new `SoftChoice(click.Choice)` type that shows known values for docs/autocomplete but accepts any string, so older CLI versions don't reject new server-side flavors. Used for `--flavor` in jobs, spaces, and repos CLIs.
- **Automated sync** — `utils/check_hardware_flavors.py` fetches the live hardware catalog from the Hub API and patches both enums. New flavors are appended; removed ones get a `# legacy` comment (kept for backward compat). A daily CI workflow (`.github/workflows/update-hardware-flavors.yaml`) runs the script and opens a PR via `huggingface-hub-bot` when something changes.
- **Docs** — the hardcoded hardware list in `docs/source/en/guides/jobs.md` is replaced with a pointer to `hf jobs hardware` / `list_jobs_hardware()` so it stays current automatically.

```
$ hf jobs run --flavor future-gpu-x99 python:3.12 echo "works with unknown flavors"
Job started with ID: ...
```

```python
>>> from huggingface_hub import JobHardware, JobHardwareInfo
>>> JobHardware.CPU_BASIC
<JobHardware.CPU_BASIC: 'cpu-basic'>
>>> from huggingface_hub import list_jobs_hardware
>>> list_jobs_hardware()[0]
JobHardwareInfo(name='cpu-basic', pretty_name='CPU Basic', ...)
```

```
$ python utils/check_hardware_flavors.py
Fetching hardware flavors from the Hub API...
  Spaces: 29 flavors
  Jobs:   28 flavors

✅ All good! Hardware enums are up to date.
```

---

> [!NOTE]
> **Medium Risk**
> Breaking rename of `JobHardware` dataclass to `JobHardwareInfo` and enum/catalog splits affect public API and CLI flavor validation; automated enum edits need human review before merge.
> 
> **Overview**
> Jobs hardware is **decoupled from Spaces**: a dedicated `JobHardware` enum drives job APIs and types, while `SpaceHardware` is trimmed to the Spaces catalog. The old `JobHardware` dataclass from `list_jobs_hardware()` is **renamed to `JobHardwareInfo`** (breaking rename for consumers who imported the dataclass).
> 
> **CLI and docs** use new `SoftChoice` for `--flavor` / `--hardware` so help lists known values but unknown Hub flavors still pass through. Jobs guide hardware is an **auto-generated table** plus pointers to `hf jobs hardware` / `list_jobs_hardware()`.
> 
> **Automation**: `utils/check_hardware_flavors.py` syncs both enums (and the jobs doc table) from the live Hub API—new flavors appended, removed ones kept with `# legacy`. A **daily GitHub Actions workflow** runs `--update`, `make style`, and opens a bot PR when anything drifts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26fc3b2cc4ccefaa62c38826879b1abadadb0471. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


utils/check_hardware_flavors.py
Wauplin, line 1: for this one I reviewed the code quickly
I mainly focused on testing it locally with all sorts of situation (adding 1, removing 1, adding multiples while removing multiple, etc.). Always worked correctly so I think "it's fine"

src/huggingface_hub/_jobs_api.py
```diff
@@ -437,7 +480,7 @@ def _create_job_spec(
     command: list[str],
     env: dict[str, Any] | None,
     secrets: dict[str, Any] | None,
-    flavor: SpaceHardware | None,
+    flavor: JobHardware | str | None,
```
hanouticelina, line 483: `str` is redundant I think, `JobHardware` already inherits from `str`
```suggestion
    flavor: JobHardware | None,
```

src/huggingface_hub/_jobs_api.py
```diff
@@ -447,7 +490,7 @@ def _create_job_spec(
         "command": command,
         "arguments": [],
         "environment": env or {},
-        "flavor": flavor or SpaceHardware.CPU_BASIC,
+        "flavor": flavor or JobHardware.CPU_BASIC,
```
hanouticelina, line 493: should we delegate the default to the server instead?

src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -690,6 +690,32 @@ def typer_factory(help: str, epilog: str | None = None, cls: type[TyperGroup] |
...
+        values = (
+            [m.value for m in choices] if isinstance(choices, type) and issubclass(choices, Enum) else list(choices)
+        )
+        super().__init__(values, case_sensitive=True)
+
+    def convert(self, value: Any, param: click.Parameter | None, ctx: click.Context | None) -> str:
+        try:
+            return super().convert(value, param, ctx)
+        except click.exceptions.BadParameter:
+            return str(value)
```
hanouticelina, line 716: worth a warning maybe, okay to not warn if the server's error is informative enough
```suggestion
        except click.exceptions.BadParameter:
            out.warning(f"Unknown flavor '{value}'. Sending to server anyway.")
            return str(value)
```

utils/check_hardware_flavors.py
```diff
@@ -0,0 +1,247 @@
...
+        name: str = "",
+        value: str = "",
+        is_legacy: bool = False,
+        text: str = "",
+    ):
+        self.kind = kind  # "member" or "line"
+        self.name = name
+        self.value = value
+        self.is_legacy = is_legacy
+        self.text = text  # raw line text for comments/blanks
```
hanouticelina:
```suggestion
@dataclass(slots=True)
class _EnumEntry:
    kind: Literal["member", "line"]
    name: str = ""
    value: str = ""
    is_legacy: bool = False
    text: str = ""
```

.github/workflows/update-hardware-flavors.yaml
```diff
@@ -0,0 +1,79 @@
+name: Update Hardware Flavors
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 4 * * *" # Every day at 4am
```
hanouticelina, line 6: maybe once a week is enough no? no strong opinion

hanouticelina (commented):
Thank you!

src/huggingface_hub/_jobs_api.py
Wauplin: we could but at the moment `flavor` is required by the server otherwise you get a

```
Bad request:
* Invalid option: expected one of "cpu-basic"|"cpu-upgrade"|"cpu-performance"|"cpu-xl"|"sprx8"|"zero-a10g"|"t4-small"|"t4-medium"|"l4x1"|"l4x4"|"l40sx1"|"l40sx4"|"l40sx8"|"a10g-small"|"a10g-large"|"a10g-largex2"|"a10g-largex4"|"a100-large"|"a100x4"|"a100x8"|"h200"|"h200x2"|"h200x4"|"h200x8"|"rtx-pro-6000"|"rtx-pro-6000x2"|"rtx-pro-6000x4"|"rtx-pro-6000x8"|"inf2x6" * at flavor
```

(just tested it myself)

src/huggingface_hub/cli/_cli_utils.py
Wauplin: I'm not sure to be honest as the warning wouldn't be useful to the user. Either the flavor is unknown locally and supported server-side, in which case using that flavor is perfectly valid and the warning would be misleading (best they can do is to update `hf` but the Job itself still runs). And if the flavor is unknown locally and not supported server-side then they get an immediate error from the server "flavor must be one of ..." in which case the user knows the flavor is unknown/not supported.

Wauplin: and on the contrary, getting a warning for something that works correctly is annoying as you don't have any way to get rid of it

.github/workflows/update-hardware-flavors.yaml
Wauplin: :shrug:

docs/source/en/guides/jobs.md
```diff
@@ -206,13 +206,20 @@ Use this to run a fine-tuning script like [trl/scripts/sft.py](https://github.co
 > [!TIP]
 > For comprehensive guidance on running model training jobs with TRL on Hugging Face infrastructure, check out the [TRL Jobs Training documentation](https://huggingface.co/docs/trl/main/en/jobs_training). It covers fine-tuning recipes, hardware selection, and best practices for training models efficiently.
 
-Available `flavor` options:
+To see all available `flavor` options and their pricing, run:
 
-- CPU: `cpu-basic`, `cpu-upgrade`
-- GPU: `t4-small`, `t4-medium`, `l4x1`, `l4x4`, `a10g-small`, `a10g-large`, `a10g-largex2`, `a10g-largex4`,`a100-large`
-- TPU: `v5e-1x1`, `v5e-2x2`, `v5e-2x4`
```
julien-c, line 213: only drawback is we're losing a bit of SEO/built-in info for agents (w/o invocation of a script) 

maybe we can maintain a list in the same way we maintain the docstring? (in addition to documenting the python code to get it)

src/huggingface_hub/_jobs_api.py
Wauplin: Here we can do `flavor="something"` where `"something"` is not a `JobHardware` so type should reflect that. We could do `flavor: str | None,` instead since as you said `JobHardware` is a `str` but not the other way around. Though I still prefer to be explicit on `JobHardware` for self-documentation purposes

docs/source/en/guides/jobs.md
Wauplin: good idea
pushed https://github.com/huggingface/huggingface_hub/pull/4266/commits/26fc3b2cc4ccefaa62c38826879b1abadadb0471 to address it. Now the output of `hf jobs hardware` is listed in the docs with hardware + cost

Wauplin: image

src/huggingface_hub/_space_api.py
```diff
@@ -61,9 +61,6 @@ class SpaceHardware(str, Enum):
     # CPU
     CPU_BASIC = "cpu-basic"
     CPU_UPGRADE = "cpu-upgrade"
-    CPU_PERFORMANCE = "cpu-performance"
-    CPU_XL = "cpu-xl"
-    SPRX8 = "sprx8"
```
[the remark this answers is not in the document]
Wauplin: that's on purpose yes

docs/source/en/guides/jobs.md
```diff
@@ -206,13 +206,51 @@ Use this to run a fine-tuning script like [trl/scripts/sft.py](https://github.co
 > [!TIP]
 > For comprehensive guidance on running model training jobs with TRL on Hugging Face infrastructure, check out the [TRL Jobs Training documentation](https://huggingface.co/docs/trl/main/en/jobs_training). It covers fine-tuning recipes, hardware selection, and best practices for training models efficiently.
 
-Available `flavor` options:
+Here is the full list of available hardware to run Jobs:
```
hanouticelina, line 209: Nice!

hanouticelina (approved):
all good, thank you!

PR #4267, opened 2026-05-26
fix(release.yml): restore env: block on 'Detect version' step

paulinebm:
## Summary
- Restores the step-level `env:` block on the `Detect version` step in `.github/workflows/release.yml` that was removed by #4264.


Wauplin (approved):
sorry merging too fast 🙈

PR #4297, opened 2026-05-29
[CLI] Support hf:// URIs in `hf upload` and `hf download`

Wauplin:
This PR lets `hf upload` and `hf download` accept a single `hf://` URI in place of the positional repo id. The URI follows the usual `hf://[<TYPE>/]<ID>[@<REVISION>][/<PATH>]` grammar, so the repo type, revision and (optionally) the file path are all read straight from it. This is mostly a convenience to avoid having to pass `--repo-type` / `--revision` / filename arguments.

When a URI is provided, it is the single source of truth: passing `--repo-type` or `--revision` on top of it raises an error. A path embedded in the URI can't be combined with positional filenames (download) or the `path_in_repo` argument (upload) either.

No breaking changes expected.

## Examples

```bash
# Download a single file from a dataset at a given revision
hf download hf://datasets/HuggingFaceM4/FineVision@refs/pr/1/data/train.parquet

# Download an entire repo
hf download hf://datasets/google/fleurs

# Upload a single file to a dataset on a specific branch
hf upload hf://datasets/Wauplin/my-cool-dataset@my-branch/data/train.csv ./train.csv
```

```bash
# Trying to also pass --repo-type / --revision is rejected
$ hf download hf://datasets/google/fleurs --revision main
Error: '--revision' cannot be used with an 'hf://' URI ('hf://datasets/google/fleurs').
```

The download/upload guides and the CLI reference are updated accordingly.

---

> [!NOTE]
> **Low Risk**
> CLI-only convenience layer on existing download/upload APIs with explicit validation; default behavior for plain repo IDs is unchanged.
> 
> **Overview**
> **`hf download`** and **`hf upload`** now accept an **`hf://`** URI as the repo argument, encoding repo type, optional revision, and optional path in one string (e.g. `hf://datasets/org/repo@branch/data/file.parquet`).
> 
> When the argument starts with **`hf://`**, **`parse_hf_uri`** drives routing: single-file vs snapshot vs subfolder (trailing **`/`** preserved for downloads). **`--repo-type`** and **`--revision`** are rejected if set alongside a URI; embedded paths cannot be combined with extra positional filenames (download) or **`path_in_repo`** (upload). Bucket URIs error with a pointer to **`hf sync`**.
> 
> **`RepoTypeOptionalOpt`** replaces the default **`model`** repo type on these commands so “no **`--repo-type`**” is distinguishable from an explicit **`--repo-type model`**, which matters for URI conflict checks. Plain repo IDs still default to models when **`--repo-type`** is omitted.
> 
> Docs (CLI, download, upload guides and CLI reference) and tests cover URI examples, validation, and malformed **`hf://`** handling via **`HfUriError`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ca32100f3c3d8578ca295e7c749f44bf3b3be2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/cli/_cli_utils.py
```diff
@@ -732,6 +732,19 @@ class RepoType(str, Enum):
     ),
 ]
 
+# Same as `RepoTypeOpt` but optional (defaults to `None` rather than `model`). Used by commands that
+# accept an `hf://` URI as repo id: a `None` default lets us tell apart "user did not pass --repo-type"
+# from "user explicitly passed --repo-type model", which is required to detect conflicts with the URI.
+RepoTypeOptionalOpt = Annotated[
```
Wauplin, line 738: this part I'll need to double-check (likely that we don't need `RepoTypeOpt` to default to `"model"` in the first place

julien-c:
and this one more specifically @mishig25

julien-c:
Nice call rejecting `--repo-type`/`--revision` when passing a URI

src/huggingface_hub/cli/upload.py
```diff
@@ -146,7 +150,29 @@ def upload(
...
+            raise CLIError("Buckets are not supported by `hf upload`. Use `hf sync` instead.")
+        repo_id, repo_type_str, revision = uri.id, uri.type, uri.revision
+        if uri.path_in_repo:
+            if path_in_repo is not None:
+                raise CLIError(
+                    f"Cannot combine a path in the hf:// URI ('{uri.path_in_repo}') with the `path_in_repo` argument ('{path_in_repo}')."
+                )
+            path_in_repo = uri.path_in_repo
+    else:
+        repo_type_str = (repo_type or RepoType.model).value
```
Wauplin, line 175: this logic is duplicated between `upload` and `download` on purpose. I first tested a factorized version but it was bloating the code / making things more complex than what they should. Hence why it's like this now. Happy to revisit in the future if it was a bad decision

src/huggingface_hub/cli/_cli_utils.py
Wauplin: addressed it https://github.com/huggingface/huggingface_hub/pull/4297/commits/3071ec7f11db4d314389014f98bf4e7c70f56b44

Wauplin: reverted in https://github.com/huggingface/huggingface_hub/pull/4297/commits/9bf05aa6ef220dcc66c2352d619512a971a3f254, too complex to get it working without changing too many places

hanouticelina (approved):
perfect!

PR #4308, opened 2026-06-03
[CLI] Agent-friendly hints + examples for hf jobs

davanstrien:
## What

Small agent-ergonomics polish for `hf jobs` — help text + three `out.hint()`s, no new commands or flags (one file):

- **`logs`**: document that `-f/--follow` exits when the stream ends *regardless of whether the job succeeded or failed* (it never inspects the final stage), and add a hint after a followed stream pointing at `hf jobs inspect <id>` to check the outcome (COMPLETED/ERROR).
- **`ps`**: when nothing is running (no `-a`, no filters) add a hint — `No running jobs. Use -a/--all to include finished (and failed) jobs.` Previously this printed only "No results found." with no next step.
- **`run` / `uv run`**: the detach hint now points at `hf jobs logs -f <id>` *or* `hf jobs inspect <id>`, and each gains a `--detach` example so the detach→poll workflow is discoverable (examples feed the auto-generated `hf-cli` skill).

Hints go to stderr per the `out` convention, so they never pollute agent-parsed stdout. No `--format` examples on purpose — agent output is already auto-selected via `is_agent()`.

## Motivation

Hit these myself in a session today driving Jobs from an agent: a followed `logs -f` returned exit 0 on a job that had actually errored, and a bare `ps` with nothing running left me with no next step. Small nudges to smooth that out.

## Update — rebased on #4310 (centralized hint suppression)

Now that #4310 has merged (hint suppression centralized in the `out` singleton — `out.hint()` is a no-op only in `quiet`; `human`/`agent`/`json` still emit to stderr), this branch is brought up to date with `main` and the empty-`ps` hint is aligned to the new pattern:

- Dropped the now-redundant per-command guard `out.mode not in (OutputFormat.json, OutputFormat.quiet)` on the empty-`ps` hint. The singleton handles `quiet` centrally, and `json` now gets the hint **on stderr** — consistent with treating `json` like `agent` (the shape agreed by @hanouticelina / @Wauplin in #4310). The `not all and not labels_filters` data check stays (only hint on a bare `ps` with nothing running).
- Removed the now-unused `OutputFormat` import from `jobs.py`.
- `test_ps_empty_json` now parses `result.stdout` (the hint lands on stderr and never pollutes the JSON), matching the other `json.loads(result.stdout)` tests in the suite. `test_ps_empty_quiet` keeps asserting on the combined output, so it doubles as the regression test proving `quiet` suppression.

## A note on process

This is tiny and purely help/examples, so I thought a focused PR was clearer than an issue here — but happy to close it and open an issue instead if you'd prefer under the updated CONTRIBUTING workflow.

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -311,7 +311,8 @@ def jobs_run(
     )
     out.result("Job started", id=job.id, url=job.url)
     if detach:
-        out.hint(f"Use `hf jobs logs -f {job.id}` to stream logs, or `hf jobs inspect {job.id}` to check status.")
+        job_ref = f"{job.owner.name}/{job.id}"
+        out.hint(f"Use `hf jobs logs -f {job_ref}` to stream logs, or `hf jobs inspect {job_ref}` to check status.")
         return
     for log in api.fetch_job_logs(job_id=job.id, namespace=job.owner.name, follow=True):
         out.text(log)
@@ -364,8 +365,9 @@ def jobs_logs(
         for log in logs:
             out.text(log)
         if follow:
+            job_ref = f"{namespace}/{job_id}" if namespace else job_id
             out.hint(
-                f"Stream ended. Run `hf jobs inspect {job_id}` to check the final status (e.g. COMPLETED or ERROR)."
+                f"Stream ended. Run `hf jobs inspect {job_ref}` to check the final status (e.g. COMPLETED or ERROR)."
             )
     except HfHubHTTPError as e:
         status = e.response.status_code if e.response is not None else None
@@ -771,7 +773,8 @@ def jobs_uv_run(
     )
     out.result("Job started", id=job.id, url=job.url)
     if detach:
-        out.hint(f"Use `hf jobs logs -f {job.id}` to stream logs, or `hf jobs inspect {job.id}` to check status.")
+        job_ref = f"{job.owner.name}/{job.id}"
+        out.hint(f"Use `hf jobs logs -f {job_ref}` to stream logs, or `hf jobs inspect {job_ref}` to check status.")
         return
     for log in api.fetch_job_logs(job_id=job.id, namespace=job.owner.name, follow=True):
         out.text(log)
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -91,7 +91,7 @@ from ._cli_utils import (
     parse_volumes,
     typer_factory,
 )
-from ._output import _dataclass_to_dict, out
+from ._output import OutputFormat, _dataclass_to_dict, out
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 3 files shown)


src/huggingface_hub/cli/jobs.py
```diff
@@ -591,9 +599,12 @@ def jobs_ps(
...
     )
-    if not items and filters:
-        filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
-        out.text(f"No jobs matched filters: {filters_msg}")
+    if not items:
+        if filters:
+            filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
+            out.text(f"No jobs matched filters: {filters_msg}")
+        elif not all:
+            out.hint("No running jobs. Use `-a`/`--all` to include finished (and failed) jobs.")
```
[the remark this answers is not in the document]
davanstrien: should be fixed in 06f2f95

src/huggingface_hub/cli/jobs.py
```diff
@@ -272,6 +272,7 @@ def _parse_namespace_from_job_id(job_id: str, namespace: str | None) -> tuple[st
     context_settings={"ignore_unknown_options": True},
     examples=[
         "hf jobs run python:3.12 python -c 'print(\"Hello!\")'",
+        "hf jobs run -d python:3.12 python script.py",
```
```suggestion
        "hf jobs run --detach python:3.12 python script.py",
```

(nit) maybe in the example I'd rather use `--detach` in plain to make it clear what the example is about

src/huggingface_hub/cli/jobs.py
```diff
@@ -591,9 +601,13 @@ def jobs_ps(
...
-    if not items and filters:
-        filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
-        out.text(f"No jobs matched filters: {filters_msg}")
+    if not items:
+        if filters:
+            filters_msg = ", ".join(f"{k}{o}{v}" for k, o, v in filters)
+            out.text(f"No jobs matched filters: {filters_msg}")
+        elif not all and not labels_filters and out.mode not in (OutputFormat.json, OutputFormat.quiet):
+            # Hint is human/agent guidance; keep json/quiet output clean for machine consumers.
+            out.hint("No running jobs. Use `-a`/`--all` to include finished (and failed) jobs.")
```
```suggestion
        elif not all and not labels_filters:
            out.hint("No running jobs. Use `-a`/`--all` to include finished (and failed) jobs.")
```

this should be handled in `_output.py` directly. For now json and quiet mode are still printing hints to stderr but if we decide not to display hints at all (which I found a sensible idea, let's see what @hanouticelina thinks), then let's do it for all hints

src/huggingface_hub/cli/jobs.py
```diff
@@ -714,6 +728,7 @@ def jobs_labels(
     context_settings={"ignore_unknown_options": True},
     examples=[
         "hf jobs uv run my_script.py",
+        "hf jobs uv run -d my_script.py",
```
```suggestion
        "hf jobs uv run --detach my_script.py",
```

(same as above)

Wauplin (commented):
Thanks for improving the UX!

src/huggingface_hub/cli/jobs.py
davanstrien: Sounds good. Was also a bit unsure about this part. Ideally would not need to do this check manually.

Wauplin: Yeah in theory we want as minimal OutputFormat checks as possible in the commands logic and delegate that part to the `out` singleton. Happy to review a PR if you want to remove hints globally from json/quiet modes

davanstrien: have updated this to pull the upstream `_output.py ` change

Wauplin (approved):
Thanks!

PR #4296, opened 2026-05-29
[URIs] Parse web URLs in `parse_hf_uri` + add `HfUri.to_url`

Wauplin:
Now that we have a single centralized parser for `hf://` URIs (`parse_hf_uri`), this PR teaches it to **also understand the web URLs you copy-paste from the website**, so pasting a link straight into the CLI or the library "just works". It also adds the reverse direction: `HfUri.to_url()`.

Allows using URLs in the CLI:

```bash
$ hf cp https://huggingface.co/nvidia/LocateAnything-3B/blob/main/config.json - | jq '.architectures'
[
  "LocateAnythingForConditionalGeneration"
]
```

## Summary

- `parse_hf_uri(...)` now accepts Hugging Face **web URLs** (with or without scheme, `http://` and `https://`) and normalizes them to the canonical `hf://` form before reusing the exact same parsing/validation path.
- Recognized hosts: `huggingface.co`, `hf.co`, the staging host, and the host of a custom `HF_ENDPOINT`. Query strings (`?download=true`) and fragments (`#L10`) are dropped.
- Added **`HfUri.to_url(endpoint=None)`** — renders the browsable web URL (the inverse of parsing one).
- `is_hf_uri(...)` benefits for free (returns `True` for recognized URLs, still `False` for local paths).
- **The parser stays strict**: ambiguous or non-location URLs are rejected, never guessed.

## Supported URL formats

Only the **URL path** is shown (the part after the host, e.g. `huggingface.co`); the recognized host is implied.

| Points at | URL path |
| --- | --- |
| Model repo | `/<ns>/<name>` |
| That repo type (also `spaces/`, `kernels/`, `models/`) | `/datasets/<ns>/<name>` |
| Folder at `<rev>` | `/<ns>/<name>/tree/<rev>[/<path>]` |
| File (viewer) | `/<ns>/<name>/blob/<rev>/<path>` |
| File (download) | `/<ns>/<name>/resolve/<rev>/<path>` |
| File (raw) | `/<ns>/<name>/raw/<rev>/<path>` |
| File (blame) | `/<ns>/<name>/blame/<rev>/<path>` |
| Bucket | `/buckets/<ns>/<name>` |
| Bucket file (no revision) | `/buckets/<ns>/<name>/resolve/<path>` |
| Bucket folder | `/buckets/<ns>/<name>/tree/<path>` |

[...]

PR #4324, opened 2026-06-08
[URIs] Use `parse_hf_uri` in `RepoUrl` + soft-deprecate `repo_type_and_id_from_hf_id`

Wauplin:
Address https://github.com/huggingface/huggingface_hub/pull/4296#issuecomment-4601365443

Now that `parse_hf_uri` understands both `hf://` URIs and web URLs, `RepoUrl` no longer needs its own parser (`repo_type_and_id_from_hf_id`)

With this PR:

- `RepoUrl` uses `parse_hf_uri` internally
- `repo_type_and_id_from_hf_id` no longer used inside `huggingface_hub` + is softly deprecated (i.e. only documented as legacy but doesn't warn)
   - => will open PRs in transformers/diffusers/etc. later
- `duplicate_repo` stops abusing `RepoUrl(...)` just to split a bare `from_id`/`to_id` (use `parse_hf_uri` + custom str handling)
- `parse_hf_uri` gained an optional `endpoint` argument to recognize web URLs from custom / self-hosted Hubs (incl. path prefixes like `/hf`)

`parse_hf_uri` is deliberately stricter than the old helper, so this is not a literal swap.

🚨 **Breaking change:**

Support for canonical single-segment repos (`gpt2`, `datasets/squad`) in `RepoUrl` has been dropped.
In practice not a problem since it was only used for repo creation and file upload.

Note that `hf_hub_download("gpt2", ...)` never went through `RepoUrl`, so downloads of canonical repos keep working.

## Example

```py
>>> from huggingface_hub import parse_hf_uri
>>> from huggingface_hub.hf_api import RepoUrl
>>> RepoUrl("https://huggingface.co/datasets/user/ds")
RepoUrl('https://huggingface.co/datasets/user/ds', endpoint='https://huggingface.co', repo_type='dataset', repo_id='user/ds')
>>> parse_hf_uri("http://localhost:8080/hf/my-org/my-model", endpoint="http://localhost:8080/hf")  # custom endpoint
HfUri(type='model', id='my-org/my-model', revision=None, path_in_repo='')
>>> RepoUrl("gpt2")  # canonical -> rejected
huggingface_hub.errors.HfUriError: Invalid HF URI 'hf://gpt2'. Repository id must be 'namespace/name', got 'gpt2'.
```

### DIFF
--- a/tests/test_hf_api.py
+++ b/tests/test_hf_api.py
@@ -3942,34 +3942,36 @@ class HfApiTokenAttributeTest(unittest.TestCase):
 @patch("huggingface_hub.constants.ENDPOINT", ENDPOINT_PRODUCTION)
 class RepoUrlTest(unittest.TestCase):
     def test_repo_url_class(self):
-        url = RepoUrl("https://huggingface.co/gpt2")
+        url = RepoUrl("https://huggingface.co/user/repo_name")
 
         # RepoUrl Is a string
         self.assertIsInstance(url, str)
-        self.assertEqual(url, "https://huggingface.co/gpt2")
+        self.assertEqual(url, "https://huggingface.co/user/repo_name")
 
         # Any str-method can be applied
-        self.assertEqual(url.split("/"), "https://huggingface.co/gpt2".split("/"))
+        self.assertEqual(url.split("/"), "https://huggingface.co/user/repo_name".split("/"))
 
         # String formatting and concatenation work
-        self.assertEqual(f"New repo: {url}", "New repo: https://huggingface.co/gpt2")
-        self.assertEqual("New repo: " + url, "New repo: https://huggingface.co/gpt2")
+        self.assertEqual(f"New repo: {url}", "New repo: https://huggingface.co/user/repo_name")
+        self.assertEqual("New repo: " + url, "New repo: https://huggingface.co/user/repo_name")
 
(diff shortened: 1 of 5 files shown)


tests/test_hf_api.py
```diff
@@ -3942,34 +3942,36 @@ def test_user_agent_is_overwritten(self, mock_build_hf_headers: Mock) -> None:
 @patch("huggingface_hub.constants.ENDPOINT", ENDPOINT_PRODUCTION)
 class RepoUrlTest(unittest.TestCase):
     def test_repo_url_class(self):
-        url = RepoUrl("https://huggingface.co/gpt2")
+        url = RepoUrl("https://huggingface.co/user/repo_name")
 
         # RepoUrl Is a string
         self.assertIsInstance(url, str)
-        self.assertEqual(url, "https://huggingface.co/gpt2")
+        self.assertEqual(url, "https://huggingface.co/user/repo_name")
```
Wauplin, line 3949: not a fan of the `self.assertEqual` syntax but using it since I don't want to change all existing tests

hanouticelina (approved):
looks good to me! sorry for the delay, I somehow missed this PR in my notifications 😬

Wauplin:
no worries, thanks for all the reviews :)

PR #4326, opened 2026-06-08
[Auth] Keyless CI/CD auth via OIDC token exchange

hanouticelina:
This PR lets CI jobs authenticate to the Hub without storing an `HF_TOKEN` secret, using the Hub's [Trusted Publishers](https://huggingface.co/docs/hub/trusted-publishers). Set `HF_OIDC_RESOURCE` to the repo (or username) to scope a short-lived token to, and `huggingface_hub` does the OIDC exchange under the hood, no token, no setup code.

**Example**: publish a model and a dataset from one training run

```yaml
name: Publish to the Hub

on:
  push:
    branches: [main]
  workflow_dispatch:

jobs:
  publish:
    runs-on: ubuntu-latest
    permissions:
      id-token: write    # required so the job can mint an OIDC token
      contents: read
    steps:
      - uses: actions/checkout@v4

      - name: Install the hf CLI
        run: |
          curl -LsSf https://hf.co/cli/install.sh | bash
          echo "$HOME/.local/bin" >> "$GITHUB_PATH"

      - name: Train
        run: ./train.sh    # produces ./model and ./dataset

      # Scope `HF_OIDC_RESOURCE` per step to the repo that step pushes to.
      - name: Push the model
        env:
          HF_OIDC_RESOURCE: acme/awesome-model
        run: hf upload acme/awesome-model ./model .

      - name: Push the dataset
        env:
          HF_OIDC_RESOURCE: datasets/acme/awesome-dataset
        run: hf upload acme/awesome-dataset ./dataset . --repo-type dataset
```
  Under the hood, `get_token()` gains a last-resort source: when `HF_OIDC_RESOURCE` is set and a supported CI provider is detected, it mints the provider's OIDC id token, exchanges it at `POST /oauth/token` for a short-lived Hub token, and caches it.

---

> [!NOTE]
> **High Risk**
> Changes core authentication in `get_token()` and introduces token exchange; misconfiguration or exchange failures affect all Hub calls when `HF_OIDC_RESOURCE` is set.
> 
> **Overview**
> Adds **Trusted Publishers** keyless CI auth: CI jobs can authenticate to the Hub via OIDC token exchange instead of storing `HF_TOKEN`.
> 
> A new internal `_oidc` module mints GitHub Actions OIDC id tokens (when `permissions: id-token: write` is set) and exchanges them at `POST /oauth/token` (RFC 8693), scoped with `resource`. Other providers can pass a pre-minted id token via `HF_OIDC_ID_TOKEN`.
> 
> **`get_token()`** now tries OIDC last when **`HF_OIDC_RESOURCE`** is set (repo or username to scope the token). Failures raise **`OIDCError`** instead of silently returning unauthenticated. Exchanged tokens are cached with refresh before expiry for auto-minted GitHub flows.
> 
> Debug **`HF_DEBUG`** curl logging redacts OAuth body fields such as `subject_token`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a48753a9ff575970d4140e28902bc7ca48b2d7ba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/_oidc.py
```diff
@@ -0,0 +1,159 @@
...
+GITHUB = "github"
+SUPPORTED_PROVIDERS = (GITHUB,)
+
+
+def detect_provider() -> str | None:
+    """Detect the CI provider able to mint an OIDC id token, or `None` if not in a supported CI."""
+    if os.environ.get("GITHUB_ACTIONS") == "true":
+        return GITHUB
+    return None
+
```
coyotte508: maybe add a way to pass the ID token directly for other providers than Github

eg `HF_OIDC_ID_TOKEN=... HF_OIDC_RESOURCE=... hf upload ...`

so the Gitlab example can also be simplified!

coyotte508 (commented):
Nice!

src/huggingface_hub/_oidc.py
Wauplin: in addition of `HF_OIDC_ID_TOKEN` we could also support Gitlab, etc. natively (can be a follow-up PR)

src/huggingface_hub/utils/_auth.py
```diff
@@ -143,6 +154,63 @@ def _get_token_from_file() -> str | None:
...
+        now = time.monotonic()
+        if (
+            _OIDC_TOKEN_CACHE is not None
+            and _OIDC_TOKEN_CACHE["resource"] == resource
+            and now < _OIDC_TOKEN_CACHE["expires_at"]
+        ):
+            return _OIDC_TOKEN_CACHE["token"]
+
+        # An explicit id token (any provider) takes precedence; otherwise mint from a detected one.
+        subject_token = os.environ.get("HF_OIDC_ID_TOKEN") or None
```
coyotte508, line 199: note - if provided from env probably impossible to change? (in that case maybe `expires_at` shouldn't be deducted the refresh margin of 300s)

hanouticelina: yes okay makes sense, dropped the margin for the `HF_OIDC_ID_TOKEN` path in [a644a69](https://github.com/huggingface/huggingface_hub/pull/4326/commits/a644a69228023aa1d620123bee773d6b8fa01bf2)

src/huggingface_hub/utils/_auth.py
```diff
@@ -58,13 +60,22 @@ def get_token() -> str | None:
...
 
     Returns:
         `str` or `None`: The token, `None` if it doesn't exist.
     """
-    return _get_token_from_environment() or _get_token_from_file() or _get_token_from_google_colab()
+    return (
+        _get_token_from_environment()
+        or _get_token_from_file()
+        or _get_token_from_google_colab()
+        or _get_token_from_oidc()
```
Wauplin:
```suggestion
        _get_token_from_oidc()
        or _get_token_from_environment()
        or _get_token_from_file()
        or _get_token_from_google_colab()
```

I'd say as soon as `HF_OIDC_RESOURCE` is set, it should take precedence over any other login method

src/huggingface_hub/utils/_auth.py
```diff
@@ -143,6 +154,67 @@ def _get_token_from_file() -> str | None:
         return None
 
 
+# Cache for the OIDC-exchanged token: re-exchanging on every `get_token()` call would be wasteful,
+# and re-exchanging shortly before expiry transparently keeps long-running jobs authenticated.
+_OIDC_TOKEN_LOCK = Lock()
+_OIDC_TOKEN_CACHE: dict | None = None  # {"resource": str, "token": str, "expires_at": float}
```
Wauplin: (nit) could be a typed dict instead of a comment

(typed dict over dataclass so we don't put hard constraint)

src/huggingface_hub/utils/_auth.py
```diff
@@ -143,6 +154,67 @@ def _get_token_from_file() -> str | None:
...
+    The id token to exchange is taken from `HF_OIDC_ID_TOKEN` when set (so any CI provider works by
+    minting it itself), otherwise it is minted from a natively-supported provider (GitHub Actions).
+
+    Returns `None` only when `HF_OIDC_RESOURCE` is unset, so normal token resolution is untouched for
+    everyone else. Once opted in, any failure *raises* rather than silently falling back: setting
+    `HF_OIDC_RESOURCE` is an explicit "authenticate via OIDC", so a failure should be a clear error,
+    not a confusing "not logged in" further down the call stack.
+
+    See `huggingface_hub._oidc` and https://huggingface.co/docs/hub/trusted-publishers.
+    """
```
Wauplin: (nit) I feel the docstring is too verbose

```suggestion
    """Get a short-lived OIDC token in CI (Trusted Publishers).

    Enabled by setting `HF_OIDC_RESOURCE`, which scopes the token to a repo or user.    
    The ID token is read from `HF_OIDC_ID_TOKEN` if available, or minted from a supported CI provider (e.g. GitHub Actions).

    Returns `None` when OIDC is not enabled.
    If enabled, any failure is raised explicitly rather than falling back silently.

    See `huggingface_hub._oidc` and https://huggingface.co/docs/hub/trusted-publishers.
    """
```

src/huggingface_hub/utils/_http.py
```diff
@@ -1005,6 +1005,18 @@ def _format(
...
 
+# Request-body fields that carry credentials; redacted from debug-log curl commands (HF_DEBUG).
+_SENSITIVE_BODY_KEYS = ("subject_token", "access_token", "refresh_token", "client_secret", "device_code")
+
+
+def _redact_sensitive_body(body: str) -> str:
+    """Redact OAuth credential values from a request body string (JSON or form-encoded)."""
+    for key in _SENSITIVE_BODY_KEYS:
+        body = re.sub(rf'("{key}"\s*:\s*")[^"]*(")', r"\1<REDACTED>\2", body)  # JSON
+        body = re.sub(rf"({re.escape(key)}=)[^&\s]+", r"\1<REDACTED>", body)  # form-encoded
```
Wauplin: form encoding never used I think?

src/huggingface_hub/utils/_http.py
```diff
@@ -1026,7 +1038,7 @@ def _curlify(request: httpx.Request) -> str:
     body: str | None = None
     try:
         if request.content is not None:
-            body = request.content.decode("utf-8", errors="ignore")
+            body = _redact_sensitive_body(request.content.decode("utf-8", errors="ignore"))
             if len(body) > 1000:
                 body = f"{body[:1000]} ... [truncated]"
```
Wauplin: (to avoid redacting sensitive body on huged payload)

```suggestion
            body = request.content.decode("utf-8", errors="ignore")
            if len(body) > 1000:
                body = f"{body[:1000]} ... [truncated]"
            body = _redact_sensitive_body(body)
```

src/huggingface_hub/_oidc.py
```diff
@@ -0,0 +1,164 @@
...
+from .utils import get_session, hf_raise_for_status
+
+
+# RFC 8693 token-exchange grant + id-token subject type (see trusted-publishers docs).
+_TOKEN_EXCHANGE_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:token-exchange"
+_ID_TOKEN_TYPE = "urn:ietf:params:oauth:token-type:id_token"
+
+# Supported CI providers. GitHub Actions only.
+GITHUB = "github"
+SUPPORTED_PROVIDERS = (GITHUB,)
```
Wauplin: (nit) could be a str enum

src/huggingface_hub/errors.py
```diff
@@ -29,6 +29,19 @@ class LocalTokenNotFoundError(EnvironmentError):
     """Raised if local token is required but not found."""
 
 
+# OIDC ERRORS
+
+
+class OIDCError(Exception):
```
Wauplin, line 35: Would be good to add to cli error handling so that if an OIDCError is raised in the CLI, the error message is user-friendly (likely to happen if used with `hf upload ...`)

Wauplin (commented):
Thanks for the PR! I left some cosmetic changes but otherwise looks good to me.

Could you also update the PR description with a more up-to-date example? e.g. install `hf` CLI using installer. Also potentially could be a more complex example pushing to both a dataset and a model at the end of a training step => would need to set `HF_OIDC_RESOURCE` per step targeting different resources. (asking because example will likely be copied to release notes)

Once merged the follow-up would be:
- update hub-docs 
- add support/examples for more providers

src/huggingface_hub/errors.py
hanouticelina: yes indeed. added in [ba99c8e](https://github.com/huggingface/huggingface_hub/pull/4326/commits/ba99c8eea14bd9255c520a1ee2d75391951fac81)

src/huggingface_hub/_oidc.py
hanouticelina: yes, changed to str enum in [ba99c8e](https://github.com/huggingface/huggingface_hub/pull/4326/commits/ba99c8eea14bd9255c520a1ee2d75391951fac81)

src/huggingface_hub/utils/_http.py
hanouticelina: I think so, removed in [ba99c8e](https://github.com/huggingface/huggingface_hub/pull/4326/commits/ba99c8eea14bd9255c520a1ee2d75391951fac81)

src/huggingface_hub/utils/_auth.py
hanouticelina: changed to typed dict in [ba99c8e](https://github.com/huggingface/huggingface_hub/pull/4326/commits/ba99c8eea14bd9255c520a1ee2d75391951fac81)

src/huggingface_hub/cli/_errors.py
```diff
@@ -110,6 +111,7 @@ def _format_cli_extension_install_error(error: CLIExtensionInstallError) -> str:
     RepositoryNotFoundError: _format_repo_not_found,
     RevisionNotFoundError: _format_revision_not_found,
     LocalTokenNotFoundError: lambda _: "Not logged in. Run 'hf auth login' first.",
+    OIDCError: lambda error: str(error),
```
Wauplin:
```suggestion
    OIDCError: lambda error: f"OICD Exchange failed. {error}",
```

Wauplin (approved):
Thank you!

hanouticelina:
@Wauplin updated the PR description with a better example ✔️

hanouticelina:
I'm talking care of the follow-up PRs as well

julien-c (commented):
This is very elegantly implemented! kudos

Issue #4314, opened 2026-06-05
Make the `hf endpoints` CLI agent- and script-friendly (discovery, custom-image/GGUF, safe defaults, logs)


gary149:
Notes from watching real coding-agent sessions deploy a GGUF model via `hf endpoints`. Recurring theme: `catalog deploy` needs 1 flag (`--repo`), while `deploy` needs 7 required free-text flags. `--format agent` already exists, so this extends that direction. Observed on 1.14.0, ordered by impact.

1. **Discovery wall on `deploy`.** Requires `--repo --framework --accelerator --instance-size --instance-type --region --vendor`, all free `TEXT`, none discoverable. No way to learn valid combinations or what fits a repo, so an agent can't build a valid command from `--help`.

2. **No custom-image / GGUF path.** The SDK supports custom images, env, secrets, revision, and engine config, but the CLI exposes none of it (`--type`, `--custom-image`, `--model-path`/`--gguf-file`, `--mmproj`, `--ctx-size`, `--container-args`, `--env`, `--secrets`, `--revision`). GGUF deploys fall back to raw SDK/REST.

3. **`type` defaults to `protected`.** `create_inference_endpoint(..., type=PROTECTED)` (`hf_api.py:8339`, enum at `_inference_endpoints.py:31`). Observed result: a running, billable, but unreachable endpoint, with no error at create time. Valid public values are `public | authenticated | private`.

4. **Wrong namespace in catalog handle.** `from_raw(data, namespace=data["name"], ...)` (`hf_api.py:8612`) uses the endpoint name as the namespace, so follow-up `pause`/`describe`/`delete` calls 404.

5. **No programmatic log access.** Failed deploys say "check the logs", but there is no CLI/SDK way to fetch them.

6. **No dry-run / validation.** No way to check a config without creating a billable resource.

7. **Free-text flags, no enums.** `--framework`, `--accelerator`, etc. are `TEXT`, so typos fail late and valid values aren't discoverable.

8. **Minor inconsistencies.** `deploy NAME` (positional) vs `catalog deploy --name` (flag); `--format agent` could surface the resolved config and endpoint URL.

PR #4329, opened 2026-06-09
 [Inference Endpoints] Custom-container deploy CLI + deprecate protected endpoint type

gary149:
## Summary

`hf endpoints deploy` couldn't deploy a custom container (custom image + container args + access type), so agents/users had to hand-write JSON and POST the raw endpoints API. This was hit in a real session deploying the `nexagi/sglang:v0.5.12` fork. Addresses point #2 of #4314.

 This wires the SDK's existing custom-image surface into the CLI and fills the one real SDK gap (container command/args).

**CLI (`hf endpoints deploy`)**, reusing existing option types (`EnvOpt`/`SecretsOpt`/`RevisionOpt`/`parse_env_map`/`SoftChoice`):
- `--custom-image` (+ `--health-route`, `--port`): assembles the `custom_image` dict (the SDK auto-wraps it to `model.image.custom`).
- `--command` / `--container-args`: quoted string, `shlex`-split into tokens.
- `--env` / `--env-file` / `--secrets` / `--secrets-file`, `--revision`.
- `--type`: `public | authenticated | private` (forward-compatible pass-through, default `authenticated`; `protected` is deprecated).
- Validation: the custom-container knobs raise a clean error if used without `--custom-image`.

  **SDK (`create_inference_endpoint`)**:
  - New `container_command` / `container_args` (`list[str]`) -> `model.command` / `model.args`
  (the API supports these but the SDK had no way to set them).
  - `type` now defaults to `authenticated` (was `protected`, now deprecated — passing it emits a `FutureWarning`); added `AUTHENTICATED` to `InferenceEndpointType`. Kept `InferenceEndpointType | str` so plain strings and forward-compatible new types still pass through.

**Bonus fix**: the custom-image health route key is camelCase `healthRoute` per the API schema (`CustomContainer` -> `BaseContainer`). The codebase used snake_case `health_route`, which the API silently ignores (unknown key, falls back to the default route). Fixed in the new code and corrected the same mistake in the existing docstring + en/ko guides (they only "worked" because the examples used the default `/health`).

## Example

```bash
hf endpoints deploy nex-n2-pro \
  --repo nex-agi/Nex-N2-Pro \
  --framework custom \
  --accelerator gpu --vendor aws --region ap-northeast-2 \
  --instance-type nvidia-h200 --instance-size x8 \
  --custom-image nexagi/sglang:v0.5.12 \
  --health-route /health --port 30000 \
  --container-args "--reasoning-parser qwen3 --tool-call-parser qwen3_coder --mamba-scheduler-strategy extra_buffer --tp 8" \
  --env MODEL_ID=/repository \
  --type authenticated \
  --min-replica 0 --max-replica 1 --scale-to-zero-timeout 15
```

Resulting `model` payload:

```json
"model": {
  "framework": "custom",
  "repository": "nex-agi/Nex-N2-Pro",
  "image": { "custom": { "url": "nexagi/sglang:v0.5.12", "healthRoute": "/health", "port": 30000 } },
  "env": { "MODEL_ID": "/repository" },
  "args": ["--reasoning-parser","qwen3","--tool-call-parser","qwen3_coder","--mamba-scheduler-strategy","extra_buffer","--tp","8"]
}
```

This is exactly the payload that previously had to be hand-written and POSTed directly.

## Notes / out of scope

- `hf endpoints update` / `update_inference_endpoint` are intentionally untouched (kept the diff focused; natural fast-follow).
- `InferenceEndpointType` enum left as-is. `SoftChoice` + the `| str` widening cover `authenticated` without a breaking enum change.
- GGUF/llamacpp-specific shortcuts (`--gguf-file`, `--ctx-size`, `--mmproj`) from #4314 are a separate image type, not included here.

## Tests

Mocked unit tests (no network): added `test_deploy_custom_image`, `test_deploy_custom_args_require_image` (CLI) and `test_create_inference_endpoint_container_command_and_args_payload` (SDK payload); updated `test_deploy_from_hub`. Custom-image schema (field names, command/args location) verified against the live OpenAPI spec.

Refs #4314

**Breaking change:** `create_inference_endpoint(...)` now defaults `type` to `authenticated` instead of `protected`. Callers that omitted `type` and relied on the old default will now create `authenticated` endpoints. Pass `type="protected"` explicitly to keep the previous behavior - though `protected` is now deprecated and emits a `FutureWarning`, so the real migration is to `authenticated`.

---

> [!NOTE]
> **Medium Risk**
> Changes endpoint creation payloads and CLI defaults (`type=authenticated`); misconfiguration could affect deployed inference endpoints, but scope is limited to deploy/create and is covered by new tests.
> 
> **Overview**
> **`hf endpoints deploy`** now supports full custom-container workflows without hand-written API JSON: `--custom-image` (with `--health-route`, `--port`), `--command` / `--container-args` (quoted strings, `shlex`-split), `--revision`, env/secrets options, and `--type` (`public`, `authenticated`, `protected`, `private`, default **`authenticated`** in the CLI). Custom-container flags error unless `--custom-image` is set.
> 
> **`create_inference_endpoint`** gains `container_command` and `container_args` (`list[str]` → `model.command` / `model.args`), and `type` accepts plain strings (e.g. **`authenticated`**). Docs and examples switch custom-image health checks to API-correct **`healthRoute`** instead of `health_route`.
> 
> Tests cover CLI custom deploy, validation, and SDK payload wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 852289414014f04d5941a9af6c18d682b71bbb43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gary149:
Aligned `--type` handling with the recent hf-endpoints work after reviewing @gary149's last-6-days PRs:

- **hf-endpoints#2350** (`fix(api): default catalog deploy endpoints to authenticated`) establishes `authenticated` as the safe default and explicitly calls out the SDK's stale `protected` default as the same silent-unreachable footgun. The v2 API `EndpointType` enum is `public | authenticated | private` (no `protected`).
- So `hf endpoints deploy --type` now defaults to `authenticated` (token-gated, reachable) and lists `public | authenticated | private`. `SoftChoice` still passes through any value, so `protected` (legacy/SDK enum) is not hard-rejected.
- Also cross-checked hf-endpoints#2349 (`GET /api/engines`) and #2352 (deploy recipe): both confirm the `custom_image.<engine>.url` shape and the camelCase `healthRoute` this PR now uses. The GGUF/llamacpp shortcut path those add server-side remains intentionally out of scope here.

Note: the SDK `create_inference_endpoint` default is still `InferenceEndpointType.PROTECTED` (the enum has no `AUTHENTICATED`). I left that untouched to avoid a breaking change to the public SDK surface; happy to follow up separately if you want the SDK default realigned too.

Wauplin:
(let us know when/if it's ready for review)

gary149:
> (let us know when/if it's ready for review)

ready for review

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -152,8 +164,73 @@ def deploy(
...
+            help="Health check route exposed by the custom container (e.g. '/health'). Requires --custom-image.",
+        ),
+    ] = None,
+    port: Annotated[
+        int | None,
+        typer.Option(
+            help="Port the custom container listens on (e.g. 30000). Requires --custom-image.",
+        ),
+    ] = None,
+    command: Annotated[
```
hanouticelina: maybe to keep the symmetry `container_command` / `container_args`
```suggestion
    container_command: Annotated[
```

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -172,8 +249,16 @@ def deploy(
         scaling_metric=scaling_metric,
         scaling_threshold=scaling_threshold,
         scale_to_zero_timeout=scale_to_zero_timeout,
+        revision=revision,
+        custom_image=custom_image_dict,
+        container_command=shlex.split(command) if command else None,
+        container_args=shlex.split(container_args) if container_args else None,
+        env=env_map or None,
+        secrets=secrets_map or None,
+        type=endpoint_type or "authenticated",
```
hanouticelina: I think we should deprecate `protected` and update the default in `create_inference_endpoint`

src/huggingface_hub/cli/inference_endpoints.py
gary149: Good call, renamed to `--container-command` in 75333363.

src/huggingface_hub/cli/inference_endpoints.py
hanouticelina: addressed in [0479c94](https://github.com/huggingface/huggingface_hub/pull/4329/commits/0479c94db8e989b6a1d30ed0555403c7fc163770)

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -172,8 +250,16 @@ def deploy(
         scaling_metric=scaling_metric,
         scaling_threshold=scaling_threshold,
         scale_to_zero_timeout=scale_to_zero_timeout,
+        revision=revision,
+        custom_image=custom_image_dict,
+        container_command=shlex.split(container_command) if container_command else None,
+        container_args=shlex.split(container_args) if container_args else None,
+        env=env_map or None,
+        secrets=secrets_map or None,
+        type=endpoint_type or "authenticated",
```
Wauplin: let's do a

```py
params = {}
if endpoint_type is not None:
    params["type"] = endpoint_type
(...)

api.create_inference_endpoint(..., **params)
```

so that the default value is defined only once (in hf_api)

Wauplin: same for other values where a default is set here (`secrets=secrets_map or None,` etc.)

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -152,8 +164,74 @@ def deploy(
...
+    ] = None,
+    env: EnvOpt = None,
+    env_file: EnvFileOpt = None,
+    secrets: SecretsOpt = None,
+    secrets_file: SecretsFileOpt = None,
+    endpoint_type: Annotated[
+        str | None,
+        typer.Option(
+            "--type",
+            click_type=SoftChoice(["public", "authenticated", "private"]),
```
Wauplin: (nit) could use `SoftChoice(InferenceEndpointType)` (from `_inference_endpoints.py`)

tests/test_hf_api.py
```diff
@@ -4879,6 +4879,91 @@ def test_create_inference_endpoint_custom_image_payload(
...
+    # Default type is now "authenticated".
+    api.create_inference_endpoint(**common_args)
+    _, call_kwargs = mock_session.post.call_args
+    assert call_kwargs["json"]["type"] == "authenticated"
+
+    # Passing "protected" still works but is deprecated.
+    with pytest.warns(FutureWarning, match="authenticated"):
+        api.create_inference_endpoint(type="protected", **common_args)
+
+
```
Wauplin: can be removed IMO (low added value)

Wauplin (approved):
Thank you! pre-approving with a few nits but overall looks good

PR #4337, opened 2026-06-10
Fix ignored-pattern warning grammar in download CLI

wunianze666-netizen:
## Summary
- fix the ignored-pattern warning text in `hf download` from 'have being' to 'have been'
- tighten the CLI regression test to assert the exact warning messages for both `--include` and `--exclude`

## Why
These warnings are user-facing CLI output. Fixing the grammar makes the message clearer, and the stronger assertion helps prevent the typo from coming back.

## Validation
- ran a local regression script that exercises `download()` with ignored patterns and checks the exact warning messages
- ran `python -m py_compile src/huggingface_hub/cli/download.py tests/test_cli.py`

---

> [!NOTE]
> **Low Risk**
> Copy-only CLI warning text plus a stricter unit test; no download behavior changes.
> 
> **Overview**
> Fixes user-facing **`hf download`** warnings when positional filenames override `--include` / `--exclude`: the message now says **"have been"** instead of **"have being"**.
> 
> The regression test **`test_download_with_ignored_patterns`** no longer only checks that some warning contains "Ignoring"; it asserts the **exact** text and order of both warnings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce0aab82af0ad00c960fa7a3666c87d6dbe76821. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1111,7 +1111,11 @@ class TestDownloadImpl:
                 force_download=True,
             )
         print_mock.assert_called_once_with("folder-path")
-        assert any("Ignoring" in str(w.message) for w in caught)
+        warning_messages = [str(w.message) for w in caught]
+        assert warning_messages == [
+            "Ignoring `--include` since filenames have been explicitly set.",
+            "Ignoring `--exclude` since filenames have been explicitly set.",
+        ]
         mock_download.assert_not_called()
         mock_snapshot.assert_called_once_with(
             repo_id="author/model",
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1111,11 +1111,7 @@ class TestDownloadImpl:
                 force_download=True,
             )
         print_mock.assert_called_once_with("folder-path")
-        warning_messages = [str(w.message) for w in caught]
-        assert warning_messages == [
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -1111,7 +1111,11 @@ def test_download_with_ignored_patterns(self, mock_download: Mock, mock_snapshot
                 force_download=True,
             )
         print_mock.assert_called_once_with("folder-path")
-        assert any("Ignoring" in str(w.message) for w in caught)
+        warning_messages = [str(w.message) for w in caught]
+        assert warning_messages == [
+            "Ignoring `--include` since filenames have been explicitly set.",
+            "Ignoring `--exclude` since filenames have been explicitly set.",
+        ]
```
hanouticelina:
```suggestion
        assert any("Ignoring" in str(w.message) for w in caught)
```

hanouticelina (approved):
Thank you!

tests/test_cli.py
hanouticelina: no need to tighten the test in my opinion

Wauplin (approved):

PR #4345, opened 2026-06-11
[Jobs] Add `hf jobs wait` and `HfApi.wait_for_job`

### LINKED ISSUE #4311: [Jobs] Add a `wait` primitive — `hf jobs wait` + `HfApi.wait_for_job` (chaining + agent ergonomics)
*Not urgent — more of a nice-to-have, but it'd unlock some clean scripting/orchestration patterns and keep jobs consistent with the `wait()` we already have for endpoints.*

#### What

A first-class way to wait on a job — or a batch of jobs — until it finishes, with a clean exit code:

```bash
ID=$(hf jobs uv run -dq train.py)
hf jobs wait "$ID" && hf jobs uv run -dq eval.py   # chain on success
hf jobs wait --label run=my-sweep                  # block on a whole batch
```

This would bring the same ergonomics we already have for Inference Endpoints (`InferenceEndpoint.wait()`) over to jobs.

#### Why it'd help

- **Chaining / CI** — lets `hf jobs … && next-step` branch on whether the job actually succeeded.
- **Agents** — a detached-job workflow could do a quiet status check (`--format json`) instead of consuming the whole log stream. Feels like a natural follow-up to the agent-friendly logs work in #3783.
- **Batches** — fire N jobs and block on all of them in one call.

#### Rough sketch (two layers, API-first)

- **Python API** — `HfApi.wait_for_job(job_id, *, timeout=None, poll_interval=…) -> JobInfo`, plus a `JobInfo.wait()` convenience. Built on the existing `inspect_job`, mirroring `InferenceEndpoint.wait()`. Terminal = `{COMPLETED, CANCELED, ERROR, DELETED}`, success = `COMPLETED`. (Open question: return the final `JobInfo` for the caller to inspect, or raise on `ERROR`/`CANCELED` like the endpoint `wait()` — happy either way.)
- **CLI** — `hf jobs wait <id> [<id>…]`, a thin wrapper over the above; blocks until terminal, exit code reflects success; `--timeout`, `--poll-interval`, `--format`, and `--label` for batches.
- Nice bonus (CLI): the non-detached run / `logs -f` paths could surface the job's final stage as their exit code too, so "block while watching logs" and `wait` line up.

#### Notes

- No server changes needed — `inspect_job` is the poll primitive and `JobStage` already has the states. (One subtlety: a general `wait` should spell out its terminal states rather than reuse the log-stream finished-check, which counts `SCHEDULING` as done.)
- Scope: jobs only. Endpoints already have `.wait()` in the SDK (could surface in the CLI later); spaces are lower priority since builds are short; synchronous commands (buckets, uploads) don't need it.

Happy to put up a PR.

Wauplin:
Closes #4311. Implements the `wait` primitive as discussed in https://github.com/huggingface/huggingface_hub/issues/4311#issuecomment-4681458449.

## Summary

- **`HfApi.wait_for_job(job_id, *, timeout=None, poll_interval=5.0, namespace=None, token=None)`** — polls `inspect_job` until the Job reaches a terminal stage (`COMPLETED`, `CANCELED`, `ERROR`, `DELETED`). Returns the final `JobInfo` **even on failure** (check `job.status.stage`); only raises `TimeoutError` on timeout. `job_id` accepts `str | list[str]` and returns `JobInfo | list[JobInfo]` accordingly (typed with overloads).
- **`hf jobs wait <id> [<id>...]`** — blocks until all Jobs finish; exit code 0 iff **all** ended `COMPLETED`. Supports `namespace/job_id` form, `--timeout 10m`-style durations and the global `--format` flags.
- **Non-detached `hf jobs run` / `hf jobs uv run` now exit with the Job's outcome.** After the log stream ends, the final state is settled via `wait_for_job` (this also papers over the log stream ending while the Job is still `SCHEDULING`) and the command fails with the server's error message unless the Job `COMPLETED`.
- New `TERMINAL_JOB_STAGES` constant in `_jobs_api.py` — the log-stream finished-check was not reused on purpose (it counts `SCHEDULING` as done).
- Docs: new "Wait until Jobs finish" section in the Jobs guide, CLI guide examples, package reference entry.

## Decisions (rationale in the issue thread)

- **Return, don't raise, on `ERROR`/`CANCELED`**: a failed Job is a normal, inspectable outcome — unlike `InferenceEndpoint.wait()` where a failed deploy means you have no endpoint.
- **No `--label` batch flag**: footgun (would match completed Jobs from previous runs; no server-side label filtering anyway). Batches compose with pipes instead — see examples below.
- **No `JobInfo.wait()`**: the dataclass holds no token handle, so it would silently construct a client with the default token.
- **No `--poll-interval` on the CLI** (kept in the Python API).
- `job_id: str | list[str]` on `wait_for_job` so the Python API matches the CLI's multi-id support.

## ⚠️ Breaking change

Non-detached `hf jobs run` used to exit 0 even when the Job errored. It now exits 1. We consider this a bugfix (scripts relying on the old behavior were being silently misled), but it should be called out in the release notes.

## Examples (real runs on production, cpu-basic)

Wait on a batch where one Job fails:

```console
$ hf jobs wait 6a2ac6e3c4f53f9fc5aa5014 6a2ac6e559bbdade52d49f9b; echo "exit_code=$?"
Error: 1 of 2 Job(s) did not complete successfully: 6a2ac6e559bbdade52d49f9b (ERROR)
id	stage	message
6a2ac6e3c4f53f9fc5aa5014	COMPLETED	None
6a2ac6e559bbdade52d49f9b	ERROR	Job failed with exit code: 1. Reason: Error
exit_code=1
```

Timeout + cancel + `namespace/job_id` + `--json`:

```console
$ hf jobs wait --timeout 10s 6a2ac6e959bbdade52d49f9d; echo "exit_code=$?"
Error: Timed out after 10s waiting for Job(s) to finish.
exit_code=1

$ hf jobs cancel 6a2ac6e959bbdade52d49f9d && hf jobs wait --json Wauplin/6a2ac6e959bbdade52d49f9d
id=6a2ac6e959bbdade52d49f9d
Error: 1 of 1 Job(s) did not complete successfully: 6a2ac6e959bbdade52d49f9d (CANCELED)
[{"id": "6a2ac6e959bbdade52d49f9d", "stage": "CANCELED", "message": null}]
```

Non-detached run now reflects the Job outcome — chaining just works:

```console
$ hf jobs run --flavor cpu-basic python:3.12 python -c "import sys; print('doing work'); sys.exit(3)"; echo "exit_code=$?"
id=6a2ac74559bbdade52d49fa0 url=https://huggingface.co/jobs/Wauplin/6a2ac74559bbdade52d49fa0
doing work
Error: Job 6a2ac74559bbdade52d49fa0 finished with stage 'ERROR': Job failed with exit code: 3. Reason: Error
exit_code=1

$ hf jobs run --flavor cpu-basic python:3.12 python -c "print('chain me')" && echo "NEXT STEP RAN"
id=6a2ac7d9c4f53f9fc5aa501f url=https://huggingface.co/jobs/Wauplin/6a2ac7d9c4f53f9fc5aa501f
chain me
NEXT STEP RAN
```

The pipe pattern that replaces `--label`:

```console
$ hf jobs ps -q | xargs hf jobs wait; echo "exit_code=$?"
id	stage
6a2ac76159bbdade52d49fa2	COMPLETED
6a2ac760c4f53f9fc5aa5019	COMPLETED
exit_code=0
```

Python API:

```python
>>> jobs = api.wait_for_job(job_id=[a.id, b.id])
>>> [(j.id, j.status.stage) for j in jobs]
[('6a2ac79a59bbdade52d49fa4', 'COMPLETED'), ('6a2ac79ac4f53f9fc5aa501b', 'ERROR')]
```

Out of scope (as agreed in the issue): scheduled jobs, `hf jobs logs -f` exit code (stays 0, matching `docker logs -f` semantics).

---

> [!NOTE]
> **Medium Risk**
> Non-detached `hf jobs run`/`uv run` now exit 1 on Job failure, which can break scripts that assumed success; otherwise changes are additive polling/CLI with localized test coverage.
> 
> **Overview**
> Adds a **wait** primitive for Hub Jobs in Python and the CLI, plus tighter exit-code behavior for foreground runs.
> 
> **`HfApi.wait_for_job`** polls `inspect_job` until each Job hits a terminal stage (`COMPLETED`, `CANCELED`, `ERROR`, `DELETED`), exposed via **`TERMINAL_JOB_STAGES`**. It accepts one or many job IDs, optional `timeout` / `poll_interval`, returns final **`JobInfo`** even on failure (only **`TimeoutError`** on timeout), and is exported as **`wait_for_job`**.
> 
> **`hf jobs wait`** blocks on one or more IDs (same namespace, optional `--timeout`), prints a results table, and exits **0** only if every Job **`COMPLETED`**.
> 
> Non-detached **`hf jobs run`** / **`hf jobs uv run`** now stream logs then **`wait_for_job`** and **exit non-zero** unless the Job completed—so **`&&`** chaining reflects real outcomes (behavior change vs. always exiting 0).
> 
> Docs and tests cover the new API, CLI command, namespace validation, and the run exit-code change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 323df4b4a2b1008e61686ab24b27a861663809ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Wauplin:
This PR is ready for review! @davanstrien can you give it a try let me know your feedback? Once everything's fine for you, I'll turn the PR as "ready for review" (I prefer to avoid asking @hanouticelina 's review until we are 100% aligned on the UX).

davanstrien:
Thanks for working on this! I didn't have time to do as much review by hand as I wanted yet, but here is my agent's initial review (armed with a lot of context from my agent's previous design discussion on this, plus some smoke tests):

---

Took it for a spin — installed the branch and ran real cpu-basic Jobs through the matrix: non-detached `run` success/failure exit codes, mixed-batch `wait`, `--timeout 10s`, cancel → `CANCELED`, `--json`, `-q`, `namespace/job_id` form. The UX is exactly what #4311 hoped for, and the trims all feel right (return-not-raise, no `--label`, no `JobInfo.wait()`). `hf jobs run … && next-step` just working is a big deal. New tests all pass locally.

A few things hit while testing:

**1. Confirming Bugbot's namespace leak — and it fails noisily.** In `jobs_wait`, the parse loop reassigns the shared `namespace`, so after one `namespace/job_id` argument every subsequent bare ID is polled in that namespace:

```console
$ hf jobs wait Wauplin/<id-a> <my-bare-id>   # my-bare-id now looked up in Wauplin's namespace → 403/404
```

And `wait alice/x bob/y` raises `Conflicting namespace: got --namespace='alice'…` even though the user never passed `--namespace`. Since `wait_for_job` takes a single namespace per call anyway, maybe parse each ID independently (bare IDs keep the default) and error explicitly if two args imply different namespaces?

**2. A typo'd Job ID dumps a raw httpx traceback.**

```console
$ hf jobs wait 000000000000000000000000
  ...
httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://huggingface.co/api/jobs/davanstrien/000000000000000000000000'
```

Root cause is `inspect_job` using `response.raise_for_status()` (raw httpx, so it bypasses the CLI's error formatting). Pre-existing — `hf jobs inspect <bad-id>` does the same — but `wait` is where hand-pasted IDs will be most common, and the namespace leak above surfaces as this same traceback. Worth catching in `jobs_wait` (or switching `inspect_job` to `hf_raise_for_status`)?

**3.** One gap noticed along the way (separate from this PR): the `ID=$(hf jobs run -dq …); hf jobs wait $ID` capture pattern from the issue still isn't expressible, since `run`/`uv run` don't consume the global format flags. Will write that up as a follow-up issue rather than expand here.

**4. Two quick questions on output:** should the new `Job <id> completed` line be a `hint` (stderr) instead of `out.text` (stdout), given the exit code already carries it? And in agent/TSV mode the `wait` table prints a literal `None` for the message of completed Jobs — intended? (JSON mode correctly emits `null`.)

None of these block the feature — 1 and 2 feel worth fixing in this PR. Thanks for taking this on so quickly!

Wauplin:
Hi @davanstrien thanks for the review.

I pushed https://github.com/huggingface/huggingface_hub/pull/4345/changes/5698d9bcbd139bca63a05165961d52900a8b9bb4 to fix 1.. Easiest solution was to forbid waiting on Jobs from several namespaces at once. In practice I don't think it'll ever be a problem.

About 2., I prefer to fix it in a separate PR. At the moment any HTTP 404 in the Jobs CLI is raised like this. I'll tackle all of them at once. **EDIT:** see https://github.com/huggingface/huggingface_hub/pull/4367.

About 3., happy to read a more detailed report (haven't understood this one to be fair).

About 4., I still think a clean output is nicer and more consistent with other commands.

src/huggingface_hub/cli/jobs.py
```diff
@@ -681,6 +695,69 @@ def jobs_cancel(
...
+        "hf jobs wait <job_id>",
+        "hf jobs wait <job_id_1> <job_id_2>",
+        "hf jobs ps -q | xargs hf jobs wait",
+    ],
+)
+def jobs_wait(
+    job_ids: Annotated[
+        list[str],
+        typer.Argument(
+            help="Job IDs to wait for.",
```
hanouticelina:
```suggestion
            help="Job IDs to wait for (or 'namespace/job_id').",
```

hanouticelina (approved):
Looks good, thank you!

davanstrien:
thanks!

davanstrien:
Screenshot 2026-06-22 at 14 23 19

Also, thanks from my agent!

PR #4346, opened 2026-06-11
[Jobs] Add sync_job_volume helper and local paths in hf jobs -v

### LINKED ISSUE #4268: [Jobs] Add LocalVolume for local-path uploads (Python API)
Running a Job against some local data currently takes three commands:

```
hf buckets create me/my-data
hf buckets sync ./my-data hf://buckets/me/my-data/
hf jobs uv run script.py -v hf://buckets/me/my-data:/data
```

This is the design discussion for collapsing the Python-API path by adding a new `LocalVolume` type that mixes into the `volumes=` list. CLI integration is a separate concern to be filed as a companion issue. 

```python
from huggingface_hub import LocalVolume, Volume, run_uv_job

run_uv_job(
    "train.py",
    volumes=[
        LocalVolume(local_path="./training-data", mount_path="/data"),
        Volume(type="dataset", source="HuggingFaceFW/fineweb", mount_path="/fineweb", read_only=True),
    ],
)
```

Under the hood, each `LocalVolume` is uploaded to a subfolder of `{namespace}/jobs-artifacts` (the same bucket #4025's script transport already uses) and the resulting bucket subfolder is mounted at the user's chosen path. Semantically `LocalVolume(local_path=X, mount_path=Y)` is "upload X, mount the result at Y as a RW bucket volume".

IMO it makes sense to land the Python API first (with the experimental flag) and polish CLI semantics from that. The Python path already serves agents and notebook users without making them deal with the CLI's parsing surface — every new affordance is a kwarg rather than a `-v src:dst:modifier` bikeshed.

### Spike branch

[`daniel/jobs-sync-local-dir-spike`](https://github.com/davanstrien/huggingface_hub/tree/daniel/jobs-sync-local-dir-spike). Smoke-tested end-to-end on real Jobs. Four commits.

Roughly:
- New public `LocalVolume` dataclass in `hf_api.py`, exported from `huggingface_hub`. `@experimental` on `__new__` so construction emits the standard UserWarning.
- `HfApi._resolve_local_volumes()` at the top of `run_job` / `run_uv_job` swaps each `LocalVolume` for a real `Volume` backed by a bucket subfolder. Existing `Volume` entries pass through unchanged.
- Upload reuses `_buckets._list_local_files` so walker behaviour matches `hf buckets sync`. Applies `DEFAULT_IGNORE_PATTERNS`.
- Factored a shared `_upload_to_jobs_artifacts_bucket` helper that both the script transport from #4025 and the new local-volume upload go through.
- Pre-check on the `/data` reservation so a conflicting volume doesn't half-upload before raising.
- `create_scheduled_job` / `create_scheduled_uv_job` raise `TypeError` on `LocalVolume` for now — scheduled-job semantics aren't designed yet (every fire would re-upload).
- Empty source dirs upload a `.hf_jobs_keep` placeholder to reserve the bucket subfolder.

### The main question: storage model

Everything else here (cleanup, naming, reuse) follows from this one. Two directions:

**A.** *Ephemeral, system-managed.* What the spike does now: each upload gets a fresh `{timestamp}-{hex}` subfolder, re-runs re-upload, cleanup needs separate machinery (TTL or `hf jobs cleanup`). Closest to Modal's `Image.add_local_dir`.

**B.** *Named, user-managed.* Add a `remote_name=` to `LocalVolume`: `LocalVolume(local_path="./data", mount_path="/data", remote_name="cifar10")` → `jobs-artifacts/cifar10/`. Stable across runs, cleanup via `hf buckets rm`. Closest to Modal's `Volume.from_name(create_if_missing=True)`. Open sub-question if we go here: re-running with the same name → overwrite, append, or rsync-diff?

The spike defaults to A. Intermediate "default A, opt-in B via `remote_name=`" is also possible.

### Smaller open questions

1. **Filtering of secret-prone files.** Spike applies `DEFAULT_IGNORE_PATTERNS` only (`.git/`, `.cache/huggingface/`) — same as `upload_folder`. Modal's `add_local_dir(ignore=[])` is hands-off. Extend defaults to cover `.env*`, `.aws/`, `.ssh/`, or stay hands-off and add an `ignore_patterns=` kwarg on `LocalVolume`?
2. **Org-namespace permissions.** When `run_*_job(namespace="some-org", ...)`, the bucket lives at `some-org/jobs-artifacts`. Worth confirming who reads those subfolders before users start mounting things in an org context.
3. **`/data` reservation behaviour.** Reserved only when there's also a local script being uploaded, so `LocalVolume(mount_path="/data")` against a Hub script works fine. Confirm that's intended.
4. **Subfolder-ID entropy.** `token_hex(3)` = 16M values per second per namespace, inherited from #4025. Probably fine, possibly tight under bursty agentic patterns. Bump to `token_hex(6)`?

### v1 limitations

- The walker follows symlink files (via `os.stat`) but not symlink dirs (`os.walk` default). So `ln -s ~/.aws/credentials ./data/creds.txt` would upload silently. Same as `hf buckets sync` today.
- No progress bar during upload — relies on `batch_bucket_files` defaults.

### Not in v1

Auto pull-back after completion, subfolder cleanup / TTL, size guardrails, partial-upload resume, scheduled-job support (currently guarded with `TypeError`), `.gitignore` awareness, job-label provenance for `hf jobs inspect`.

A companion issue will cover the CLI surface (`-v ./local:/path` syntax, modifier parsing, typo UX, `.hf_jobs_keep` placeholder leak in `hf buckets sync` output, Windows path handling) — to be linked here once filed.

### LINKED ISSUE #4269: [Jobs] CLI: -v accepts local paths (depends on #4268)
Companion to #4268 (the Python `LocalVolume` design). This issue covers the CLI surface — once `LocalVolume` lands, allowing `-v`'s source side to be a local path lets users write:

```
hf jobs uv run script.py -v ./my-data:/data
```

What the user sees:

```
$ hf jobs uv run train.py -v ./my-data:/data
UserWarning: 'LocalVolume' is experimental ...
Local data from './my-data' uploaded to: https://huggingface.co/buckets/me/jobs-artifacts/20260526T143022-a1b2c3
To pull data back after the job completes, run:
    hf buckets sync hf://buckets/me/jobs-artifacts/20260526T143022-a1b2c3/ ./my-data
    (overwrites local files of the same name — pass a different target dir to keep the local copy)
Job started with ID: 6abc...
```

The `:ro`/`:rw` suffix is honoured (`-v ./fixtures:/data:ro`). Empty source dirs work too, which is the natural shape for the OCR/output-mount pattern:

```
hf jobs uv run process.py -v ./pdfs:/input:ro -v ./md-out:/output
```

Same spike branch as #4268: [`daniel/jobs-sync-local-dir-spike`](https://github.com/davanstrien/huggingface_hub/tree/daniel/jobs-sync-local-dir-spike). CLI-side changes live in `cli/_cli_utils.py:parse_volumes()` (and a helper `_try_parse_local_volume`), reusing `_split_mount` for the `:ro`/`:rw` parsing.

### Open questions

1. **Local-path discriminator / typo UX.** `Path(src).exists()` is the sole rule. If it returns false, the parser falls through to `parse_hf_mount` and the user sees `"Must start with 'hf://'"`. Worth a more specific error for "looked local but didn't exist"? Shape heuristic on the source (`./`, `/`, `~/`) could give better diagnostics.
2. **`.hf_jobs_keep` leaks into pull-back.** When an empty dir is mounted, the placeholder ends up in `hf buckets sync` downloads. Filter it server-side in `list_bucket_tree`, filter client-side in `hf buckets sync`, or skip the placeholder approach altogether.
3. **Pull-back hint copy.** Printed `hf buckets sync` command targets the source dir by default with an overwrite caveat in parentheses. Sensible if `local_path` was input-only, less sensible if the job wrote new files to a dir that already had local files of the same name. Default to a sibling dir (`./my-data.pulled`?), or keep current behaviour with the warning?
4. **Source-side colons / Windows paths.** `_split_mount`'s `rfind(":/")` should handle Windows paths like `C:\dir:/mnt`, but no Windows test exists yet.
5. **Mapping future `LocalVolume` kwargs to CLI modifiers.** If #4268 lands B (named volumes via `remote_name=`) or extra kwargs (`ignore_patterns=`), how do they show up in `-v`? Options: extra colon-separated modifiers (`-v ./data:/data:name=cifar10`), separate flag (`--data-name cifar10`), or just not surface them via CLI (use Python). Mostly depends on which kwargs #4268 adopts.

### v1 limitations

- No Windows CI for source paths with colons.
- Source discriminator is path-existence only — typo UX is as above.
- The pull-back hint is a `print()` to stdout (matches #4025's scripts-transport print pattern, not `out.hint`). If we want the same line in agent-mode JSON output, we'd want to convert.

### Out of scope

Tab completion for local paths in `-v`, shell-quoting edge cases beyond `:ro`/`:rw`, an `--ignore-patterns` flag (would depend on #4268's filtering decision).

Wauplin:
Implements #4268 and #4269, following the design settled in the #4268 discussion: no `LocalVolume` dataclass — a `sync_job_volume()` helper built on `sync_bucket` instead, plus the CLI sugar on top.

## Summary

- New **`HfApi.sync_job_volume(source, mount_path, *, remote_name=None, read_only=True, namespace=None, token=None) -> Volume`**: syncs a local directory to a subfolder of the (private, auto-created) `{namespace}/jobs-artifacts` bucket and returns a ready-to-mount `Volume`. Exported at module level.
- **CLI: `-v` accepts local directories** in `hf jobs run`, `hf jobs uv run`, `hf jobs scheduled run` and `hf jobs scheduled uv run`. `:ro`/`:rw` suffixes honoured, read-only by default. `parse_volumes` in `_cli_utils.py` is untouched, so spaces/repos commands keep the `hf://`-only behavior.

## Python API

```py
from huggingface_hub import run_uv_job, sync_job_volume

# Upload ./training-data once...
volume = sync_job_volume("./training-data", "/data")

# ...then run as many jobs as you want against it
run_uv_job("train.py", script_args=["--learning-rate", "0.01"], volumes=[volume])
run_uv_job("train.py", script_args=["--learning-rate", "0.05"], volumes=[volume])
```

## CLI

Real output from manual testing (all jobs on `cpu-basic`):

```console
$ hf jobs uv run -v ./demo/data:/inputs:ro -v ./demo/outputs:/outputs:rw --timeout 5m process.py
Local data from './demo/data' is available at: https://huggingface.co/buckets/Wauplin/jobs-artifacts/tree/data-9aab5089
Local data from './demo/outputs' is available at: https://huggingface.co/buckets/Wauplin/jobs-artifacts/tree/outputs-9344de90
Hint: Volume '/outputs' is mounted read-write. Once the job is over, pull back its data with:
  hf buckets sync hf://buckets/Wauplin/jobs-artifacts/outputs-9344de90 ./demo/outputs
✓ Job started id=6a2ac2d459bbdade52d49f3f url=https://huggingface.co/jobs/Wauplin/6a2ac2d459bbdade52d49f3f
```

## Decisions taken

- **Folder naming**: `{dirname}-{sha256(hostname:abspath)[:8]}` — the path+machine-derived scheme agreed in the issue (hostname included per @lhoestq's suggestion to avoid cross-machine overwrites). `remote_name=` overrides.
- **Read-only by default**, both in Python and CLI (`:rw` to opt out), so the two surfaces can't drift. Output retrieval stays the reverse `sync_bucket` flow.
- **No deletion of remote-only files when re-syncing**: matches the `sync_bucket` engine default. Locally-deleted files linger remotely until cleaned with `hf buckets rm`; in exchange a re-sync can never wipe outputs a previous job wrote to the same folder.
- **Empty source dirs upload a `.keep` placeholder**. Confirmed necessary end-to-end: a bucket folder with zero files doesn't exist and the job dies at scheduling with `Volume mount failed: init container exhausted retries` (repro: https://huggingface.co/jobs/Wauplin/6a2ac081c4f53f9fc5aa4ff0). This makes the empty-output-dir pattern (`-v ./md-out:/output:rw`) work. Trade-off: `.keep` comes back in pull-back syncs (known v1 limitation from #4269).
- **Privacy**: `sync_job_volume` creates the `jobs-artifacts` bucket **private** (same as the existing script transport). It can't change the visibility of a bucket that already exists, so if `jobs-artifacts` was previously created public, `sync_job_volume` now **warns** rather than silently uploading data to a public bucket (addresses the Cursor Bugbot flag).
- **No secret-file filtering**, as decided in the issue discussion.
- **Scheduled jobs are allowed** with local volumes: unlike the spike (which uploaded at trigger time and raised `TypeError`), the sync happens once at creation and every fire mounts the same folder.
- **No pre-check for the reserved `/data` mount conflict** with local scripts: the existing error still raises after the volume sync, but thanks to derived naming the upload is reused when the user retries with a different mount path, so nothing is wasted.

Closes #4268
Closes #4269

---

> [!NOTE]
> **Medium Risk**
> Touches Jobs volume mounting and bucket sync/upload paths where mis-sync or a public `jobs-artifacts` bucket could expose user data; scope is bounded by tests and privacy warnings.
> 
> **Overview**
> Adds **`sync_job_volume`** so local folders can be synced into the private `{namespace}/jobs-artifacts` bucket and mounted as a **`Volume`** on Jobs (incremental sync via `sync_bucket`, stable `{dirname}-{hash}` subfolders, read-only by default, `.keep` for empty dirs, warning if the bucket is public).
> 
> **Jobs CLI** (`run`, `uv run`, scheduled variants) now accepts **local paths** on `-v` (`LOCAL_DIR:/mount[:ro|:rw]`), syncing before submit and hinting `hf buckets sync` for `:rw` outputs; **`parse_volumes`** stays `hf://`-only for other commands.
> 
> Also fixes **`sync_bucket`** to clone `HfApi` with the same endpoint/headers when a token is passed, plus docs and integration/unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9f5716cfdddb42a40bccb04014360a270901ac3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/_buckets.py
```diff
@@ -1098,7 +1098,14 @@ def sync_bucket_internal(
     if token is not None:
         from .hf_api import HfApi
 
-        api = HfApi(token=token)
+        api = HfApi(
```
Wauplin, line 1101: fix unrelated to this PR (must redefine exact same HfApi attributes otherwise we loose some info)

hanouticelina (approved):
Looks good to me!

PR #4347, opened 2026-06-11
Add cache doctor CLI for cache diagnostics and safe repairs

Chinmay1220:
## What does this PR do?

Adds `hf cache doctor`, a CLI command to diagnose corrupted local Hugging Face cache entries and optionally repair safe cases.

The command reports cache scan warnings with structured categories, affected paths, suggestions, and repairability. It supports human/table output, JSON output, dry-run repair previews, and confirmed repairs.

## Changes

- Add structured metadata to `CorruptedCacheException`
- Add `hf cache doctor`
- Support `--format json`, `--repair --dry-run`, and `--repair --yes`
- Document cache diagnostics in the cache guide
- Add CLI and cache warning metadata tests

## Tests

- `python -m pytest tests\test_cli.py::TestCacheCommand tests\test_utils_cache.py::TestCacheWarningMetadata -q`
- `python -m ruff check src\huggingface_hub\errors.py src\huggingface_hub\utils\_cache_manager.py src\huggingface_hub\cli\cache.py tests\test_cli.py tests\test_utils_cache.py`
- `python -m py_compile src\huggingface_hub\errors.py src\huggingface_hub\utils\_cache_manager.py src\huggingface_hub\cli\cache.py`

---

> [!NOTE]
> **Medium Risk**
> `--repair` deletes local cache paths on disk; behavior is limited to scanner-marked targets but mistaken use could remove incomplete repos or stray files users still wanted to keep.
> 
> **Overview**
> Adds **`hf cache doctor`** to scan the Hub cache via `scan_cache_dir`, surface corruption warnings in tables or JSON, exit with code **1** when issues exist (CI-friendly), and optionally **repair** scanner-approved targets through `_try_delete_path` with `--repair --dry-run`, interactive confirm, or `--repair --yes`.
> 
> **`CorruptedCacheException`** now carries structured fields (`category`, `path`, `suggestion`, `repair_path`, `repair_action`, `repair_type`). Cache scanning in **`_scan_cached_repo`** populates them so some problems (unexpected cache-root files, missing `snapshots`, invalid `refs`/snapshot files) are marked repairable; others (invalid repo layout, broken symlinks, dangling refs) only get guidance.
> 
> The cache management guide documents the new workflow; tests cover the doctor CLI paths and warning metadata.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 906b86ea98b6049fd20abe5d8c104ac4fc74adc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/manage-cache.md
+++ b/docs/source/en/guides/manage-cache.md
@@ -494,6 +494,45 @@ HFCacheInfo(
 )
 ```
 
+### Diagnose your cache
+
+Use `hf cache doctor` to check whether the local cache layout is healthy. The command
+builds on [`scan_cache_dir`] and reports corrupted cache entries such as invalid repo
+folders, missing `snapshots` directories, broken `refs`, unexpected files in snapshot
+folders, and broken snapshot links.
+
+```bash
+>>> hf cache doctor
+```
+
+When issues are detected, the command prints the category, affected path, explanation,
+and suggested action. It exits with code `1` so it can be used in scripts or CI jobs.
+Use JSON output when you need to inspect the diagnostics programmatically:
+
+```bash
+>>> hf cache doctor --format json
+```
+
+Some cache issues have a safe automatic repair target, for example an unexpected file
+inside the cache root or an incomplete cached repo with no `snapshots` directory. Preview
+those actions first with `--dry-run`:
+
+```bash
+>>> hf cache doctor --repair --dry-run
+```
+
+If the proposed actions look correct, run the repair with confirmation:
+
+```bash
+>>> hf cache doctor --repair
+```
+
+For non-interactive usage, pass `--yes`:
+
+```bash
+>>> hf cache doctor --repair --yes
+```
+
 ### Verify your cache
 
 `huggingface_hub` can verify that your cached files match the checksums on the Hub. Use `hf cache verify` CLI to validate file consistency for a specific revision of a specific repository:
--- a/src/huggingface_hub/cli/cache.py
+++ b/src/huggingface_hub/cli/cache.py
@@ -19,6 +19,7 @@ from collections import defaultdict
 from collections.abc import Callable, Mapping
 from dataclasses import dataclass
 from enum import Enum
+from pathlib import Path
 from typing import Annotated, Any
 
 import typer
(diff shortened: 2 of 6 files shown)


Wauplin:
Hi @Chinmay1220 , thanks for the PR. Is it solving a real-world issue you've had with your cache? If yes, what was it and how this command would help fixing it?

Chinmay1220:
Hi @Wauplin, thanks for the question.

After looking deeper, this PR is not fixing the root cause of cache corruption itself. It addresses the next layer: making cache warnings that are already detected by `scan_cache_dir` visible and actionable from the CLI.

I reproduced this with a realistic cache scenario using a temp cache containing a real Hub file (`hf-internal-testing/tiny-random-gpt2/config.json`) and the normal commit/ETag cache layout. The cache was healthy first, then I added a harmless OS/tool-created file (`Thumbs.db`) at the cache root, similar in class to #2110.

After that:
- `scan_cache_dir` reported an `invalid-cache-entry` warning
- `hf cache ls --cache-dir <dir>` still exited 0 and only listed the repo
- `hf cache doctor --cache-dir <dir>` surfaced the warning with category/path/message/suggestion
- `hf cache doctor --repair --dry-run` previewed the safe repair target
- `hf cache doctor --repair --yes` removed only that unexpected file
- a final `hf cache doctor --cache-dir <dir>` returned healthy

So the command helps users diagnose and safely clean simple cache corruption/pollution cases that the scanner already detects. It does not prevent every root cause.

For the specific root cause I reproduced, the smaller preventive fix would be to extend the scanner's existing `.DS_Store` ignore behavior to other harmless OS metadata files such as `Thumbs.db` and `desktop.ini`. That would prevent those files from being reported as cache corruption in the first place.

For broader cache corruption sources like interrupted downloads, broken symlinks, dangling refs, or corrupted blobs, I think those should be handled as separate targeted fixes because each one needs different detection and repair semantics.

If you prefer, I can narrow this work into the smaller root-cause fix first and keep `hf cache doctor` as a follow-up discussion.

Wauplin:
From what I understand, this PR is only fixing harmless warnings. I'm therefore tempted to just close it as it's doesn't bring much value to end users (it has never been a requested feature AFAIK). Adding 400+ lines of code induces maintenance costs down the line.

PR #4348, opened 2026-06-11
feat: avoid per file resolve call in snapshot path

drbh:
this PR explores avoiding the per file resolve call in the `snapshot_download` path and prefers using the  `list_repo_tree` to for the repo's files metadata.

this change was motivated by kernel repos that contain many small files like https://huggingface.co/kernels/kernels-community/deep-gemm/tree/v2 (which vendors cutlass in each build variant). 

when testing these changes locally via the `get_kernel` method from the kernels library this reduces the number of "resolve" calls from 1413 to 1 (since the kernels library makes an extra `hf_hub_download` call when downloading).

not sure if these changes align with the rest of the library but wanted to open as a possible solution/improvement for repos with many small files. note I was careful to avoid making changes to any public functions to avoid any api breakage

---

> [!NOTE]
> **Medium Risk**
> Touches the shared download metadata path; incorrect tree-derived etags/URLs could cause wrong cache keys or failed downloads, though incomplete entries still fall back to HEAD.
> 
> **Overview**
> **`snapshot_download`** now reuses file metadata from **`list_repo_tree`** when the repo file list is unreliable (empty/missing siblings, >1000 files, kernel repos), so each file download can skip the per-file **resolve/HEAD** round trip.
> 
> While walking the tree at the resolved **`repo_info.sha`**, it builds a path → **`HfFileMetadata`** map via **`_file_metadata_from_repo_file`** (etag/size from LFS or blob id, resolve URL, optional Xet data) and passes it into downloads through a new internal **`_hf_hub_download`** used only from snapshot. Public **`hf_hub_download`** is unchanged and still delegates without pre-fetched metadata.
> 
> **`_get_metadata_or_catch_error`** returns the caller-supplied metadata immediately when present; otherwise behavior is unchanged (HEAD + fallbacks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25b79bbd427d6e3e6368090fe58058c534675336. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -16,9 +16,17 @@ from .errors import (
     RepositoryNotFoundError,
     RevisionNotFoundError,
 )
-from .file_download import REGEX_COMMIT_HASH, DryRunFileInfo, hf_hub_download, repo_folder_name
+from .file_download import (
+    REGEX_COMMIT_HASH,
+    DryRunFileInfo,
+    HfFileMetadata,
+    _hf_hub_download,
+    hf_hub_url,
+    repo_folder_name,
+)
 from .hf_api import DatasetInfo, HfApi, KernelInfo, ModelInfo, RepoFile, SpaceInfo
-from .utils import OfflineModeIsEnabled, filter_repo_objects, logging, validate_hf_hub_args
+from .utils import OfflineModeIsEnabled, XetFileData, filter_repo_objects, logging, validate_hf_hub_args
+from .utils._xet import XetTokenType, xet_connection_info_refresh_url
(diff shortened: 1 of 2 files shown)


Wauplin:
Benchmarked it in https://github.com/huggingface/huggingface_hub/pull/4361. There is definitely something to be done here. I'll work on something in the coming days (likely a separate PR)

Wauplin:
closing this PR in favor of https://github.com/huggingface/huggingface_hub/pull/4394

same core idea: skipping HEAD calls when /tree listing is enough (implementation is a bit more generic)

PR #4354, opened 2026-06-12
[Auth] Browser-based OAuth login 

hanouticelina:
This PR replaces the token copy-paste login with a browser-based OAuth Device Code flow (RFC 8628). Supersedes #4102).

- `hf auth login` / `login()` print a URL + short code; you authorize in the browser and the CLI retrieves and saves the token. In an interactive terminal you get a gh-style arrow-key menu to choose between browser login and pasting a token. `--token` is unchanged.
- OAuth tokens expire after 30 days but come with a refresh token: `get_token()` transparently refreshes them when less than a day of validity remains.
- an agent mode is implemented,  in `--format json | agent` (or auto-detected agent mode) it never prompts and streams events so an agent can surface the URL + code to its user and block until a terminal `auth_success` / `auth_error` event.
- `notebook_login()` renders the link + code with plain `IPython.display.HTML` — ipywidgets dependency dropped.
- `hf auth list` shows expiry for OAuth tokens; `hf auth switch` uses the new arrow-key picker.

### Interactive login

https://github.com/user-attachments/assets/3eebdc69-3acd-4f36-b284-3443c49b245c

### Agent / scripted login

```
$ hf auth login --format json
{"event": "device_code", "verification_uri": "https://hf.co/oauth/device", "user_code": "52AT-FLYZ", "verification_uri_complete": "https://hf.co/oauth/device", "expires_in": 300, "interval": 5}
{"event": "auth_success", "user": "celinah", "token_name": "oauth-celinah"}
```

Any failure emits a terminal `{"event": "auth_error", "error_code": "access_denied", "message": "..."}` before the command exits 1. `--format agent` produces the same stream as `key=val` lines.

### Token list with expiry

```
$ hf auth list
  name          token       expires
- ------------- ----------- -------------------
  my-token      hf_****5678
* oauth-celinah hf_****1234 2026-07-09
  oauth-old     hf_****9999 2026-06-09 (expired)
```

## How to review

Three commits, meant to be read in order:

1. https://github.com/huggingface/huggingface_hub/commit/469c949f51deb05f1f9b3086c8ebde8858c7fb3d **`[CLI] Add arrow-key selection prompt helper`** — self-contained `utils/_terminal.py::select_choice()`. Focus: raw-mode key reading (termios cbreak / msvcrt), the numbered fallback, Windows VT enablement.
2. https://github.com/huggingface/huggingface_hub/commit/a85e47d230fcd0103dbd6668429165381befc612 **`[Auth] Browser-based login with OAuth device flow and refresh tokens`** — the feature. Suggested reading order:
   - `_oauth_device.py` (new, ~190 lines): the three protocol primitives — `request_device_code()` (normalizes defaults), `poll_device_token()` (transient-tolerant poll loop), `refresh_access_token()`. No printing, no persistence.
   - `utils/_auth.py`: storage gains per-section metadata (`_read_stored_tokens_full` / `_save_stored_tokens_full`, `get_stored_tokens()` signature unchanged) and `get_token()` gains the auto-refresh (`_refresh_oauth_token_if_needed` — the trickiest function in the PR: in-process cache + lock, file lock, failure handling).
   - `_login.py`: the human flows — menu, paste path, `_device_code_login`, notebook HTML, `_validate_and_save_token` (renamed from `_login`), `auth_list` expiry column.
   - `cli/auth.py` + `cli/_output.py`: the mode branch (human → `login()`, otherwise event stream via the new `out.event()`).
   - `errors.py`, `constants.py`, `cli/_errors.py`, `utils/_headers.py`, `utils/_http.py`: small supporting changes (error type, client ID, implicit-token short-circuit, debug-log redaction of form bodies).
   - Tests mirror that order in `tests/test_auth.py`.
3. https://github.com/huggingface/huggingface_hub/commit/793ccf3265c09bab977e85b2de0d5a05e82d4005 **`[Auth] Update docs for browser-based login`** — guide + quick-start + regenerated CLI reference.

## Breaking changes

1. Interactive login defaults to the browser flow (paste is still available in the menu and via `--token`).
2. Piping a token to stdin (`echo $TOKEN | hf auth login`) no longer works — stdin without a TTY goes straight to the device flow. Use `--token` (the documented non-interactive path).
3. `notebook_login()` no longer uses/requires ipywidgets, and no longer stores the git credential by default (the old widget's checkbox defaulted to yes; pass `add_to_git_credential=True` to keep that behavior).
4. `interpreter_login()` no longer interactively asks "Add token as git credential?" — it's the `add_to_git_credential` param / `--add-to-git-credential` flag on both paths.
5. Private `_login()` renamed to `_validate_and_save_token()`; private `_save_stored_tokens()` removed.
6. The stored_tokens INI gains extra keys for OAuth tokens (older versions ignore them, but drop them on rewrite).

---

> [!NOTE]
> **High Risk**
> Changes core authentication (token storage, `get_token()` refresh on every Hub call, new OAuth endpoints) with documented breaking login behavior for scripts and notebooks.
> 
> **Overview**
> **Browser-based OAuth (RFC 8628) is now the default** for `hf auth login` and `login()` when no `--token` is passed: users get a URL + device code (with an optional gh-style menu to paste a token instead). `--token` behavior is unchanged.
> 
> New **`_oauth_device.py`** implements device-code request, polling, and refresh against the Hub; **`get_token()`** transparently refreshes OAuth access tokens (refresh metadata in `stored_tokens`) with file locking and caching. **`notebook_login()`** drops ipywidgets in favor of HTML + the same device flow; **`hf auth list`** can show expiry; **`hf auth switch`** uses **`select_choice()`** arrow-key prompts.
> 
> For **non-interactive** `--format json` / `agent`, **`hf auth login`** streams **`out.event()`** lines (`device_code`, `auth_success`, `auth_error`) so agents can drive login without prompts. Docs (CLI guide, quick-start, package reference) are updated accordingly.
> 
> **Breaking:** default login is browser-first; piping a token on stdin no longer works (use `--token`); `notebook_login()` no longer defaults git credentials; private `_login` → `_validate_and_save_token`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4a215141b9bf34a96c521682fe9119a9c1e6eaf. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanouticelina:
bugbot run

src/huggingface_hub/utils/_auth.py
```diff
@@ -219,6 +222,145 @@ def _get_token_from_oidc() -> str | None:
...
+                    response = refresh_access_token(refresh_token)
+                    new_token = response["access_token"]
+                    new_expires_at = int(now) + int(response["expires_in"]) if "expires_in" in response else None
+                    _save_token(
+                        token=new_token,
+                        token_name=token_name,
+                        # The server may rotate the refresh token; keep the old one if it doesn't.
+                        refresh_token=response.get("refresh_token") or refresh_token,
+                        expires_at=new_expires_at,
+                    )
```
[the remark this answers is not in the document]
hanouticelina: prod always returns `expires_in`

docs/source/en/guides/cli.md
```diff
@@ -201,6 +200,16 @@ Login successful
 The current active token is: `token_name`
 ```
 
+In non-interactive environments (CI, scripts, AI agents), the command never prompts: with `--format json` it runs the browser flow and streams one JSON event per line, so the calling program can surface the URL and code to a user and wait for completion:
```
Wauplin: I feel we should raise an explicit error if `--format json` or `--format quiet` are passed. Should simplify the logic and we don't loose anything UX-wise
Agent mode is a good idea though

docs/source/en/package_reference/cli.md
```diff
@@ -96,7 +96,7 @@ Learn more
 
 ### `hf auth login`
 
-Login using a token from huggingface.co/settings/tokens.
+Login to the Hub with your browser, or with a token passed via --token.
```
Wauplin:
```suggestion
Login from your browser, or using a token from huggingface.co/settings/tokens.
```

(nit) (to be updated in auth.py)

src/huggingface_hub/cli/_output.py
```diff
@@ -175,6 +175,28 @@ def result(self, message: str, **data: Any) -> None:
                 if values:
                     print(values[0])
 
+    def event(self, name: str, **data: Any) -> None:
```
Wauplin: I feel that this method is too specific to the auth login command to be really included in the `out` singleton. Typically, it's a no-op in human mode which is unexpected for such a method

src/huggingface_hub/cli/auth.py
```diff
@@ -73,8 +79,36 @@ def auth_login(
...
+    # so that an agent can surface the URL + code to the user and wait for a terminal event
+    # (`auth_success` or `auth_error`).
+    if not force and get_token() is not None:
+        # Still emit a terminal event: an agent watching stdout must always get one.
+        out.event("auth_success", already_logged_in=True)
+        out.hint("Already logged in. Use `hf auth login --force` to re-login.")
+        return
+    try:
+        device_info = request_device_code()
+        out.event(
```
Wauplin: related to comments above I think that here we can just support agent mode and that's it, for example with just a 

```py
out.text(agent=f"<real sentence for the agent with the information needed>")
```

src/huggingface_hub/cli/auth.py
```diff
@@ -73,8 +79,36 @@ def auth_login(
...
+            verification_uri_complete=device_info["verification_uri_complete"],
+            expires_in=device_info["expires_in"],
+            interval=device_info["interval"],
+        )
+        response = poll_device_token(device_info)
+        token_name, username = _save_oauth_token(response, add_to_git_credential=add_to_git_credential)
+    except Exception as e:
+        out.event("auth_error", error_code=getattr(e, "error_code", None), message=str(e))
+        raise
+    out.event("auth_success", user=username, token_name=token_name)
```
Wauplin: to adapt with comment above

src/huggingface_hub/utils/_auth.py
```diff
@@ -219,6 +222,145 @@ def _get_token_from_oidc() -> str | None:
...
+
+        if expires_at - _OAUTH_REFRESH_MARGIN > now:
+            _OAUTH_REFRESH_CACHE = {
+                "file_token": token,
+                "resolved_token": token,
+                "recheck_at": expires_at - _OAUTH_REFRESH_MARGIN,
+            }
+            return token
+
+        from .._oauth_device import refresh_access_token
```
Wauplin: better to have top-level imports (more consistent with the codebase)

src/huggingface_hub/utils/_auth.py
```diff
@@ -219,6 +222,145 @@ def _get_token_from_oidc() -> str | None:
...
+                recheck_at = now + _OAUTH_RECHECK_INTERVAL
+            # Return the existing token: if it's truly expired, the API will reject it with a clear error.
+            _OAUTH_REFRESH_CACHE = {"file_token": token, "resolved_token": token, "recheck_at": recheck_at}
+            return token
+
+        _OAUTH_REFRESH_CACHE = {
+            "file_token": new_token,
+            "resolved_token": new_token,
+            # The floor guards against a token lifetime shorter than the refresh margin, which
+            # would otherwise put `recheck_at` in the past and trigger a refresh on every call.
```
Wauplin, line 339: (note: in practice refresh margin will always be shorter)

src/huggingface_hub/utils/_headers.py
```diff
@@ -146,12 +144,14 @@ def get_token_to_send(token: bool | str | None) -> str | None:
             )
         return cached_token
 
-    # Case implicit use of the token is forbidden by env variable
+    # Case implicit use of the token is forbidden by env variable. Checked before resolving the
```
Wauplin, line 147: good catch

src/huggingface_hub/utils/_http.py
```diff
@@ -1013,13 +1013,14 @@ def _format(
...
+_SENSITIVE_BODY_KEYS = ("subject_token", "access_token", "refresh_token", "client_secret", "device_code")
 
 
 def _redact_sensitive_body(body: str) -> str:
-    """Redact OAuth credential values from a JSON request body string."""
+    """Redact OAuth credential values from a JSON or form-urlencoded request body string."""
     for key in _SENSITIVE_BODY_KEYS:
-        body = re.sub(rf'("{key}"\s*:\s*")[^"]*(")', r"\1<REDACTED>\2", body)
+        body = re.sub(rf'("{key}"\s*:\s*")[^"]*(")', r"\1<REDACTED>\2", body)  # JSON
+        body = re.sub(rf"(^|&)({key}=)[^&]*", r"\1\2<REDACTED>", body)  # application/x-www-form-urlencoded
```
Wauplin: (nit) both regex could be compiled once at import time and reused after

src/huggingface_hub/utils/_terminal.py
```diff
@@ -95,6 +96,153 @@ def _format(cls, s: str, code: str) -> str:
         return f"{code}{s}{cls._reset}"
 
 
+def select_choice(prompt: str, choices: list[str]) -> int:
```
Wauplin, line 108: I won't review this part but it's really cool to have such a helper now without additional dependency! Hope it works great cross platform and in weird setups 😄

src/huggingface_hub/utils/_terminal.py
```diff
@@ -95,6 +96,153 @@ def _format(cls, s: str, code: str) -> str:
...
+        import termios  # noqa: F401
+    except ImportError:
+        return False
+    return True
+
+
+def _enable_windows_vt_processing() -> bool:
+    """Enable VT escape-sequence processing on the Windows console (off by default in legacy
+    cmd.exe/PowerShell windows, where the menu would otherwise render as literal `←[K` garbage)."""
+    import ctypes
```
Wauplin: (top level import)

src/huggingface_hub/_login.py
```diff
@@ -228,80 +221,119 @@ def auth_list() -> None:
...
+    return f"This token expires in {days} days. Log in again to renew it."
+
+
 ###
 # Interpreter-based login (text)
 ###
 
 
-def interpreter_login(*, skip_if_logged_in: bool = True) -> None:
+def interpreter_login(*, skip_if_logged_in: bool = True, add_to_git_credential: bool = False) -> None:
```
Wauplin: Is this change really needed? I don't like `add_to_git_credential` (I feel like it's more of a legacy feature) so I'm a bit reluctant to add it to more methods

Wauplin: I wouldn't be shocked if we completely get rid of it to be fair... (with a no-op if someone passes it, but at least not document it)

Wauplin: cc @julien-c since I know you used to use git credentials

src/huggingface_hub/_oauth_device.py
```diff
@@ -0,0 +1,208 @@
...
+            data={"client_id": constants.DEVICE_CODE_OAUTH_CLIENT_ID},
+            timeout=constants.HF_HUB_DOWNLOAD_TIMEOUT,
+        )
+    except httpx.HTTPError as e:
+        raise DeviceCodeError(f"Failed to request device code from {constants.ENDPOINT}/oauth/device: {e}") from e
+    if response.status_code != 200:
+        raise DeviceCodeError(
+            f"Failed to request device code from {constants.ENDPOINT}/oauth/device "
+            f"(status {response.status_code}): {response.text}"
+        )
```
Wauplin: we could use `hf_raise_for_error` + have a single raise line here ?

what I don't like with 

```
if response.status_code != 200:
        raise DeviceCodeError(...)
```

is that we don't `raise from` which is inconsistent with other places where we raise for incorrect status

Wauplin: (+ the `hf_raise_for_error` helps getting a Request ID for debugging)

src/huggingface_hub/_oauth_device.py
```diff
@@ -0,0 +1,208 @@
...
+    try:
+        info = response.json()
+    except ValueError as e:
+        raise DeviceCodeError(
+            f"Failed to parse response from {constants.ENDPOINT}/oauth/device: {response.text[:500]}"
+        ) from e
```
Wauplin: I think we can safely assume we have a valid JSON here

src/huggingface_hub/_oauth_device.py
```diff
@@ -0,0 +1,208 @@
...
+    if not isinstance(info, dict) or not {"device_code", "user_code", "verification_uri"} <= info.keys():
+        raise DeviceCodeError(
+            f"Malformed device code response from {constants.ENDPOINT}/oauth/device: {response.text[:500]}"
+        )
```
Wauplin: same here, I'd trust the Hub for returning the good response

src/huggingface_hub/_oauth_device.py
```diff
@@ -0,0 +1,208 @@
...
+                    "grant_type": _DEVICE_CODE_GRANT_TYPE,
+                    "device_code": device_info["device_code"],
+                    "client_id": constants.DEVICE_CODE_OAUTH_CLIENT_ID,
+                },
+                timeout=constants.HF_HUB_DOWNLOAD_TIMEOUT,
+            )
+            if response.status_code < 500:
+                data = response.json()
+        except (httpx.HTTPError, ValueError):
+            pass
```
Wauplin, line 123: weird way to check JSON data no?

src/huggingface_hub/errors.py
```diff
@@ -42,6 +42,26 @@ class OIDCError(Exception):
...
+
+class DeviceCodeError(Exception):
+    """Raised when the Device Code OAuth login flow (RFC 8628) or an OAuth token refresh fails.
+
+    Covers failures at any step: requesting the device code, polling for the token,
+    authorization denied/expired, or unexpected server responses.
+
+    Attributes:
+        error_code (`str`, *optional*):
+            The OAuth `error` code returned by the server (e.g. `"access_denied"`,
```
Wauplin: (nit) if we know the possible values, let's define an enum?

Wauplin (commented):
Nice! 👍  Made a first pass and tested locally but haven't checked the test. Looking forward to it!

julien-c:
> * OAuth tokens expire after 30 days but come with a refresh token: `get_token()` transparently refreshes them when less than a day of validity remains.

maybe keep more wiggle-room ie. refresh when less than 10 days remains or something. (To minimize risk that user don't launch a command in the last day and then they need to do the auth flow again)

docs/source/en/guides/cli.md
hanouticelina: yes makes sense, addressed in [45e2a3f](https://github.com/huggingface/huggingface_hub/pull/4354/commits/45e2a3f50b974ec6e90140f2720f3d695b843e3d)

src/huggingface_hub/cli/_output.py
hanouticelina: deleted `out.event()`, the only behavior kept from it is that `out.text()` now flushes immediately, the blocking login flow needs it when stdout is piped

src/huggingface_hub/cli/auth.py
hanouticelina: addressed in [45e2a3f](https://github.com/huggingface/huggingface_hub/pull/4354/commits/45e2a3f50b974ec6e90140f2720f3d695b843e3d)

src/huggingface_hub/_oauth_device.py
hanouticelina: ah yes good catch, addressed as well

hanouticelina: (you mean `hf_raise_for_status`)

src/huggingface_hub/_oauth_device.py
hanouticelina: defensive code I missed in my review 🤦‍♀️

src/huggingface_hub/_oauth_device.py
hanouticelina: yep 🤦‍♀️

src/huggingface_hub/utils/_http.py
hanouticelina: Precompiled into `_SENSITIVE_BODY_PATTERNS` at import in [45e2a3f](https://github.com/huggingface/huggingface_hub/pull/4354/commits/45e2a3f50b974ec6e90140f2720f3d695b843e3d)

src/huggingface_hub/errors.py
hanouticelina: yes, defined in [45e2a3f](https://github.com/huggingface/huggingface_hub/pull/4354/commits/45e2a3f50b974ec6e90140f2720f3d695b843e3d)

src/huggingface_hub/_login.py
hanouticelina: > I don't like `add_to_git_credential` (I feel like it's more of a legacy feature) so I'm a bit reluctant to add it to more methods

same but I wasn't sure if we really want to get rid of it..

hanouticelina:
> maybe keep more wiggle-room ie. refresh when less than 10 days remains or something. (To minimize risk that user don't launch a command in the last day and then they need to do the auth flow again)

@julien-c the line "OAuth tokens expire after 30 days but come with a refresh token.." is misleading I think, because expired access tokens still get refreshed via the refresh token on the next command. the real expiry that forces a re-login is the refresh token's 3-month sliding window (3 months of 0 inactivity) so "miss the last day -> re-auth" risk doesn't exist here

src/huggingface_hub/cli/_output.py
```diff
@@ -79,10 +80,10 @@ def text(self, msg: str | None = None, *, human: str | None = None, agent: str |
         match self.mode:
             case OutputFormat.human:
                 if human is not None:
-                    print(human)
+                    print(human, flush=True)
```
Wauplin: (nit) shouldn't we flush on all `out` commands for consistency? in practice I don't think it changes much but at least we won't have discrepencies

Wauplin (approved):
Thanks!

src/huggingface_hub/cli/_output.py
hanouticelina: good point for keeping things consistent: [18c866f](https://github.com/huggingface/huggingface_hub/pull/4354/commits/18c866f451e3ccc439b7ff33f0c1d900d28c7a72)

src/huggingface_hub/cli/_output.py
```diff
@@ -79,10 +79,10 @@ def text(self, msg: str | None = None, *, human: str | None = None, agent: str |
...
```
Wauplin: maybe define a print_flush alias for the whole module?

src/huggingface_hub/cli/_output.py
```diff
@@ -164,16 +165,16 @@
                 for k, v in data.items():
                     if v is not None:
                         parts.append(f"  {k}: {v}")
-                print("\n".join(parts))
+                print("\n".join(parts), flush=True)
```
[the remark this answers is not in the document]
hanouticelina: only secret names are logged, not values

hanouticelina:
failing tests unrelated, merging

PR #4359, opened 2026-06-15
Retry reads on 408 (request time-out)

pkooij:
## What does this PR do?

When you stream a dataset from the Hub, the library reads files in small pieces ("range reads"). Sometimes one of those reads comes back with a **`408 Request Time-out`** error, and right now that error makes the whole read fail.

The official HTTP spec says it's safe to simply try again. So this PR tells the range-read code to **retry on 408**, the same way it already retries on other temporary errors (429, 500, 502, 503, 504).

## Why is this needed?

While streaming a large dataset with `datasets` (`streaming=True`), I kept hitting random 408 errors. 

- **It only happens under load.** With ~24 readers running at once, about 3 of them hit a 408 within 20 minutes. A single reader almost never does.
- **It fixes itself on retry.** Every single 408 I saw succeeded the moment it was retried.

## How was it tested?

- **Added a unit test** that checks `_fetch_range` is set up to retry on 408.
- **Reproduced the original problem** by streaming `allenai/MolmoAct2-BimanualYAM-Dataset` with many readers at once., with the fix, the 408 recover.

---

> [!NOTE]
> **Low Risk**
> Narrow change to buffered range-read retries; only extends backoff to a status code considered safe to repeat, with a focused unit test.
> 
> **Overview**
> **Hub file streaming** no longer fails on transient **408 Request Time-out** during byte-range fetches. `HfFileSystemFile._fetch_range` now passes an explicit `retry_on_status_codes` tuple to `http_backoff` that keeps the usual retriable server/rate-limit codes and **adds 408**, which can appear under concurrent range reads at the load balancer.
> 
> A unit test asserts `_fetch_range` includes **408** in `retry_on_status_codes` when calling `http_backoff`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c09b293de36fdfdeefaa19c8ef468cdcf02eeff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -1191,7 +1191,17 @@ class HfFileSystemFile(fsspec.spec.AbstractBufferedFile):
             **self.fs._api._build_hf_headers(),
         }
         url = self.url()
-        r = http_backoff("GET", url, headers=headers, timeout=constants.HF_HUB_DOWNLOAD_TIMEOUT)
+        r = http_backoff(
+            "GET",
+            url,
+            headers=headers,
+            timeout=constants.HF_HUB_DOWNLOAD_TIMEOUT,
+            # 408 is added to the default retried statuses: under concurrent range reads the Hub's
+            # load balancer (in front of the CDN) can intermittently return "408 Request Time-out"
+            # before the request reaches the backend. It is transient and safe to repeat per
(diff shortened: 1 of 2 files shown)


Wauplin:
closed in favor of https://github.com/huggingface/huggingface_hub/pull/4360

PR #4364, opened 2026-06-15
WIP: Stream Xet files through HfFileSystem read paths

assafvayner:
## Summary

`HfFileSystem` now downloads Xet-backed files through `hf_xet`'s ranged download-stream interface across **all read paths** — `HfFileSystemFile._fetch_range` (random access), `HfFileSystemStreamFile` (sequential), and `get_file` — falling back to the existing HTTP path for non-Xet files or when `hf_xet` is unavailable. This gives `datasets` (and other fsspec consumers) a fast streaming path for Xet files, which previously always went over plain HTTP and bypassed Xet entirely.

Key points:
- **Per-file detection** (repos can mix Xet/non-Xet blobs) via a single best-effort metadata HEAD (`get_hf_file_metadata`); any failure degrades gracefully to the HTTP path so detection never breaks a read.
- **Shared stream group**: a bounded, thread-safe LRU cache shares one `XetDownloadStreamGroup` per `(endpoint, refresh-route, auth)`, amortizing the CAS handshake + token fetch across all files of a repo.
- **Ranged reconstruction**: byte ranges map 1:1 to `download_stream(start, end)` (exclusive end, matching the old `bytes=start-(end-1)`), so random-access reads (e.g. parquet footers / row groups) fetch only the covering chunks.
- **Interrupt handling**: `KeyboardInterrupt` aborts the Xet session consistently across all three paths.
- No `hf_xet` / xet-core changes required — the ranged streaming API already exists in `hf_xet` 1.5.x.

## Performance (no regression)

Benchmarked on a 238 MB Xet-backed parquet (`open-thoughts/OpenThoughts-114k`), Xet vs `HF_HUB_DISABLE_XET=1`:
- `full_scan` (sequential throughput): Xet ≈ HTTP (~117s mean both) — at parity.
- `random_access`: within HTTP's own run-to-run variance.
- `footer_read`: marginally slower on Xet (one-time CAS handshake), small in absolute terms.

Note: the streaming path does not populate the local chunk cache (that cache is used by `hf_hub_download`, not streaming), so warm reads aren't faster than cold — but parity with HTTP holds. A reusable benchmark script is included under `benchmarks/`.

## Test Plan

- [x] Unit tests (offline, mocked): stream-group cache + LRU eviction, per-file detection (once-only, best-effort fallback), all three read paths, HTTP fallback, KeyboardInterrupt→abort, and byte-exact resume after a mid-stream error.
- [x] Integration tests against a real Xet repo: byte-exact random-access / sequential / `get_file`, plus spies confirming the Xet path is taken.
- [x] `ruff check` and `ruff format --check` clean.

## Follow-up (not in this PR)

A companion `datasets` change (bump `huggingface_hub>=1.20`, add a streaming-uses-xet integration test) is prepared but held until this ships in a release, since the version floor can't resolve from PyPI before then.

### DIFF
--- a/benchmarks/benchmark_hffs_xet_streaming.py
+++ b/benchmarks/benchmark_hffs_xet_streaming.py
@@ -1,0 +1,53 @@
+"""Benchmark HfFileSystem streaming reads: xet vs HTTP.
+
+Run twice — once with HF_HUB_DISABLE_XET=1 (HTTP baseline) and once without (xet) —
+and compare. Pass a real xet-backed hf:// file path (parquet or any large file).
+
+Usage:
+    HF_HUB_DISABLE_XET=1 python benchmarks/benchmark_hffs_xet_streaming.py <hf://...>
+    python benchmarks/benchmark_hffs_xet_streaming.py <hf://...>
+"""
+
+import sys
+import time
+
+from huggingface_hub import HfFileSystem
+from huggingface_hub.utils._runtime import is_xet_available
+
+
+def _time(fn):
+    start = time.perf_counter()
+    fn()
+    return time.perf_counter() - start
+
+
+def main(path: str) -> None:
+    fs = HfFileSystem()
+    size = fs.info(path)["size"]
+    mode = "xet" if is_xet_available() else "http"
+    print(f"mode={mode} size={size} path={path}")
+
+    def footer_read():
+        with fs.open(path, "rb") as f:
+            f.seek(max(0, size - 64 * 1024))
+            f.read(64 * 1024)
+
+    def full_scan():
+        with fs.open(path, "rb", block_size=0) as f:
+            while f.read(8 * 1024 * 1024):
+                pass
+
+    def random_access():
+        with fs.open(path, "rb") as f:
+            step = max(1, size // 16)
+            for off in range(0, size, step):
+                f.seek(off)
+                f.read(min(256 * 1024, size - off))
+
+    print(f"footer_read     : {_time(footer_read):.3f}s")
+    print(f"full_scan       : {_time(full_scan):.3f}s")
+    print(f"random_access   : {_time(random_access):.3f}s")
+
+
+if __name__ == "__main__":
+    main(sys.argv[1])
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -29,10 +29,13 @@ from .errors import (
 )
 from .file_download import get_hf_file_metadata, hf_hub_url, http_get
 from .hf_api import SPECIAL_REFS_REVISION_REGEX, BucketFile, BucketFolder, HfApi, LastCommitInfo, RepoFile, RepoFolder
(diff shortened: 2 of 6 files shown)


Wauplin:
@assafvayner looks interesting, thanks for taking a look at that :+1: Let us know once it's in good shape for first review/tests.

(and cc @lhoestq on `HfFileSystem` side)

lhoestq:
cc @kszucs as well for viz, esp. if some of it could be reused in opendal

davanstrien:
Also ran into this as potential optimisation for data reading for Jobs etc. I did quite a bit of digging and think it would be quite nice to get even parts of this merged! Would it make sense to get the `get_file` changes in separately if the streaming work needs more discussion?

An LLM-assisted benchmark report, checked against the executed scripts and Job logs.

Tested PR head `f9563e5` on Hugging Face Jobs (`cpu-upgrade`), using one 2.15 GB FineWeb-Edu Parquet and a bucket copy. HTTP controls used the same head with `HF_HUB_DISABLE_XET=1`. Each run used a fresh container and isolated local caches; remote CDN/CAS cache state was uncontrolled.

Nonstreaming `datasets.load_dataset` from `hf://buckets/...` already calls `HfFileSystem.get_file`. With `datasets` 5.0.1:

| Measurement | HTTP | PR with Xet |
|---|---:|---:|
| Download + dataset preparation | 130.5 s | 16.5–17.9 s |
| Inside `get_file` | 120.7 s | 6.6–8.1 s |
| Other loading/preparation work | 9.8 s | 9.8 s |

There were two successful Xet runs and one successful HTTP run. The other HTTP attempt disconnected after receiving 1.29 GB of the 2.15 GB file; it is retained as a failure and excluded from timing comparisons. Validation ran after these timings. Instrumentation confirmed that the Xet runs used the PR's actual file-download helper.

Separate DataTrove/Dask tests used two runs per arm. DataTrove read all columns; Dask selected `id`, `url`, `token_count` and `text`. These timings include consuming and checking records; staging includes the PR's `get_file` download followed by the same reader locally:

| Workload | HTTP direct | Xet direct | Xet get_file + local read |
|---|---:|---:|---:|
| DataTrove ParquetReader, bucket | 196–200 s | 85–86 s | 56–57 s |
| Dask read_parquet, repo | 101–105 s | 118–120 s | 39–42 s |

The buffered-read change improved this DataTrove bucket case by about 2.3×; the Dask repo case got slower. Staging reduced total time for both. Dask used two local execution threads; the original distributed workload was not rerun.

All completed tests returned 728,000 rows with matching consumer fingerprints within each comparison. Completed whole-file downloads passed full SHA-256 checks against the source. The fingerprints are partial content checks, not a proof that every output field is identical.

These results support a concrete benefit for large-file `get_file` consumers, including existing nonstreaming `datasets` bucket loads.

assafvayner:
@davanstrien @lhoestq I haven't come back to this in a while and I don't think I have the BW to shepard it in. If you want to take over this PR/feature incl. even a re-write go ahead.

PR #4378, opened 2026-06-18
Add timeout parameter to safetensors metadata methods

### LINKED ISSUE #4377: get_safetensors_metadata can hang indefinitely: no timeout on shard Range requests
### Describe the bug

`HfApi.get_safetensors_metadata` (and the underlying `parse_safetensors_file_metadata`)
fetch each shard header via a Range `GET` with no `timeout`, and fan the requests out via
`thread_map` with no overall deadline. If one shard's connection stays open but stops
delivering the response body, the worker thread blocks on the read forever and the whole
call — and any caller — hangs indefinitely.

The requests use the client default, which is unbounded in both eras:
- 0.x (`requests`): no `timeout` arg → defaults to `None`.
- 1.x (`main`, httpx): `default_client_factory` uses `timeout=None`.

Unlike the `*_info` methods and `get_hf_file_metadata` (which already expose `timeout`),
these two have no way for a caller to bound them.

### When it happens

Triggered by a stalled / half-open connection, not deterministic steps. On the hung
process I confirmed (via `/proc`, `ss`, `hping3`): one worker socket sat `ESTABLISHED` to
the CDN for ~7.9h after receiving a partial body, with no FIN/RST and no timer armed.
Probing that 4-tuple returns `RST` — the connection was half-open (dead on the server
side, still open on ours), and there was no read timeout to break it.

Reproduces deterministically against any endpoint that accepts the request then never
sends the body and never closes (e.g. a socket server that `accept()`s then sleeps).

### Expected behavior

A bounded, configurable timeout so a stalled connection fails fast instead of hanging,
consistent with the other Hub-facing `HfApi` methods.

### Proposed fix

Add a `timeout` parameter to both methods (default `HF_HUB_DOWNLOAD_TIMEOUT`, passed
through to each Range request). I have a branch ready against `main` and am happy to open
a PR.

### Reproduction

_No response_

### Logs

```shell

```

### System info

```shell
- Platform: Linux-6.8.0-60-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Running in iPython ?: No
- Running in notebook ?: No
- Has saved token ?: False
- hf_xet: 1.1.9
- HF_HUB_OFFLINE: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
```

go-bai:
## What

  Add a `timeout` parameter to `HfApi.get_safetensors_metadata` and
  `parse_safetensors_file_metadata`, passed through to the underlying Range `GET` requests.
  Default is `constants.HF_HUB_DOWNLOAD_TIMEOUT`.

  Closes #4377 

  ## Why

  `get_safetensors_metadata` fans one Range `GET` per shard out via `thread_map` with no
  overall deadline, and each request reads a response body. With the client default being
  unbounded (`requests` → `None` in 0.x; httpx `timeout=None` on `main`), a single stalled
  connection blocks a worker forever and hangs the whole call, with no way to bound it.
  The `*_info` methods and `get_hf_file_metadata` already expose `timeout` for the same
  reason; this follows that precedent.

  ## Notes

  - **Default `HF_HUB_DOWNLOAD_TIMEOUT`, not `HF_HUB_ETAG_TIMEOUT`**: these aren't HEAD
    requests like `get_hf_file_metadata` — they fetch content via Range `GET` and read the
    body, i.e. "download" semantics. Reusing the constant keeps it overridable via the env
    var. (Both default to 10s, so default behavior is unchanged.)
  - Backwards compatible: keyword-only with a default; `timeout=None` restores the previous
    unbounded behavior.
  - **Scope**: `timeout` intentionally covers only the header Range `GET`s that are fanned
    out via `thread_map`. The preliminary `file_exists` / `hf_hub_download` (index) requests
    in `get_safetensors_metadata` are not bounded by it, as they aren't part of the fan-out
    hot path this change targets.

  ## AI assistance

  AI was used to help investigate and draft this change. I have reviewed every line and can
  explain and defend it.

---

> [!NOTE]
> **Low Risk**
> Backward-compatible optional parameter with an unchanged default; only affects HTTP wait behavior for safetensors header fetches.
> 
> **Overview**
> Adds a keyword-only **`timeout`** argument (default **`constants.HF_HUB_DOWNLOAD_TIMEOUT`**) to **`HfApi.get_safetensors_metadata`** and **`parse_safetensors_file_metadata`**, matching how other Hub download-style calls expose deadlines.
> 
> The value is threaded into both Range **`GET`** calls in **`parse_safetensors_file_metadata`** (initial 100kb fetch and optional full-metadata fetch) and forwarded for single-file and per-shard parsing in sharded repos. **`timeout=None`** keeps prior unbounded client behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf6ed26e5af4446de19c6a545075f9d96db41776. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_hf_api.py
+++ b/tests/test_hf_api.py
@@ -2634,6 +2634,27 @@ class HfApiPublicProductionTest(unittest.TestCase):
                 "HuggingFaceH4/zephyr-7b-beta", "pytorch_model-00001-of-00008.bin"
             )
 
+    def test_parse_safetensors_metadata_default_timeout(self) -> None:
+        # Regression test: the Range request must carry a timeout so a stalled connection cannot hang forever.
+        with patch("huggingface_hub.hf_api.get_session") as mock_get_session:
+            mock_get_session.return_value.get.side_effect = RuntimeError("stop after first request")
+            with self.assertRaises(RuntimeError):
+                self._api.parse_safetensors_file_metadata(
+                    "HuggingFaceH4/zephyr-7b-beta", "model-00003-of-00008.safetensors"
+                )
+        _, kwargs = mock_get_session.return_value.get.call_args
+        assert kwargs["timeout"] == constants.HF_HUB_DOWNLOAD_TIMEOUT
+
+    def test_parse_safetensors_metadata_custom_timeout(self) -> None:
+        with patch("huggingface_hub.hf_api.get_session") as mock_get_session:
+            mock_get_session.return_value.get.side_effect = RuntimeError("stop after first request")
+            with self.assertRaises(RuntimeError):
+                self._api.parse_safetensors_file_metadata(
+                    "HuggingFaceH4/zephyr-7b-beta", "model-00003-of-00008.safetensors", timeout=42
+                )
+        _, kwargs = mock_get_session.return_value.get.call_args
+        assert kwargs["timeout"] == 42
+
     def test_inference_provider_mapping_model_info(self):
         model = self._api.model_info("deepseek-ai/DeepSeek-R1-0528", expand="inferenceProviderMapping")
         mapping = model.inference_provider_mapping
--- a/tests/test_hf_api.py
+++ b/tests/test_hf_api.py
@@ -2634,27 +2634,6 @@ class HfApiPublicProductionTest(unittest.TestCase):
                 "HuggingFaceH4/zephyr-7b-beta", "pytorch_model-00001-of-00008.bin"
             )
 
-    def test_parse_safetensors_metadata_default_timeout(self) -> None:
-        # Regression test: the Range request must carry a timeout so a stalled connection cannot hang forever.
-        with patch("huggingface_hub.hf_api.get_session") as mock_get_session:
-            mock_get_session.return_value.get.side_effect = RuntimeError("stop after first request")
-            with self.assertRaises(RuntimeError):
-                self._api.parse_safetensors_file_metadata(
-                    "HuggingFaceH4/zephyr-7b-beta", "model-00003-of-00008.safetensors"
-                )
-        _, kwargs = mock_get_session.return_value.get.call_args
-        assert kwargs["timeout"] == constants.HF_HUB_DOWNLOAD_TIMEOUT
-
-    def test_parse_safetensors_metadata_custom_timeout(self) -> None:
-        with patch("huggingface_hub.hf_api.get_session") as mock_get_session:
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
This looks good to me. I checked the thread-through and it's complete on the request side:

- `parse_safetensors_file_metadata` has two Range `GET`s - the initial `bytes=0-100000` fetch and the fallback `bytes=8-{metadata_size+7}` for headers larger than 100kb - and both now receive `timeout=timeout`. Easy one to miss the second one, so nice that it's covered.
- `get_safetensors_metadata` passes `timeout` through at both call sites (the single-file early return and the `_parse` closure fanned out by `thread_map`).
- Defaulting to `constants.HF_HUB_DOWNLOAD_TIMEOUT` keeps it consistent with the download path, and the "set to `None` to disable (not recommended)" docstring note is a good call.

One small, optional note: on the sharded path, the initial `self.hf_hub_download(index_file=...)` that discovers the shard list isn't passed the new `timeout`, so a caller-supplied value applies to the per-shard header reads but not to the index-file fetch (which keeps `hf_hub_download`'s own default). Probably fine since that download already has a timeout, but flagging it in case you want a caller's `timeout=...` to govern the whole call uniformly.

Deferring to the maintainers.

tests/test_hf_api.py
```diff
@@ -2634,6 +2634,27 @@ def test_not_a_safetensors_file(self) -> None:
...
+    def test_parse_safetensors_metadata_custom_timeout(self) -> None:
+        with patch("huggingface_hub.hf_api.get_session") as mock_get_session:
+            mock_get_session.return_value.get.side_effect = RuntimeError("stop after first request")
+            with self.assertRaises(RuntimeError):
+                self._api.parse_safetensors_file_metadata(
+                    "HuggingFaceH4/zephyr-7b-beta", "model-00003-of-00008.safetensors", timeout=42
+                )
+        _, kwargs = mock_get_session.return_value.get.call_args
+        assert kwargs["timeout"] == 42
+
```
Wauplin:
```suggestion
```

(no need for this)

Wauplin (approved):
Looks good, thanks you! And sorry for the delay reviewing this 😬

PR #4385, opened 2026-06-18
Hot-fix: Fix circular import on `from huggingface_hub import login` (#4384)

hanouticelina:
Fixes #4384.

## Summary

- `from huggingface_hub import login` raised `ImportError: cannot import name 'refresh_access_token' from partially initialized module 'huggingface_hub._oauth_device' (most likely due to a circular import)` on a fresh interpreter (regression in 1.20.0, introduced by the browser-based OAuth login in #4354).
- Root cause is a layering inversion: `_oauth_device` lived at the top level but imported from `utils._http`, while `utils._auth` imported `refresh_access_token` back from `_oauth_device`. So `import login` → `_oauth_device` → `utils._http` → (runs `utils/__init__`) → `utils._auth` → `_oauth_device` (still half-initialized) → boom.
- Fix: **move `_oauth_device.py` into the `utils` layer** (`utils/_oauth_device.py`). It only depends on `constants`, `errors`, and `utils._http` — all same-or-lower layer — so it belongs there, and now every dependency points downward. No lazy imports.
  - `utils._auth` → `from ._oauth_device import refresh_access_token` (sideways, no cycle)
  - `_login.py` / `cli/auth.py` → `from .utils._oauth_device import ...` (downward)

## Reproduction (before)

```
$ python -c 'from huggingface_hub import login'
ImportError: cannot import name 'refresh_access_token' from partially initialized module
'huggingface_hub._oauth_device' (most likely due to a circular import)
```

## After

```
$ python -c 'from huggingface_hub import login; print("ok")'
ok
```

## Why no existing test caught this

It's an import-ordering bug, and in-process unit tests are structurally blind to that class of bug.

The cycle only opens when `_login`/`_oauth_device` is the *first* thing to pull in `huggingface_hub.utils` (cold). If `utils` is imported first through any normal path — which `tests/conftest.py` does at collection time (`from huggingface_hub.utils import ...`) — its `__init__` runs to completion and pulls `_oauth_device` in as a harmless leaf, so the half-initialized window never exists.

Because `sys.modules` is per-process and shared across the whole pytest session, that warm-up happens before any test module is imported. The buggy `tests/test_auth.py` passes all 39 tests against the buggy source — the same code that crashes on `from huggingface_hub import login`. Reproducing it requires a fresh interpreter (e.g. `python -c "from huggingface_hub import login"`), which an in-process test can't recreate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4384, opened 2026-06-18
Circular import on `from huggingface_hub import login`?


ianxmason:
### Describe the bug

With the latest release I am seeing a circular import error with `from huggingface_hub import login`

`ImportError: cannot import name 'refresh_access_token' from partially initialized module 'huggingface_hub._oauth_device' (most likely due to a circular import)`

I can reproduce the error on multiple systems with the attached reproduction script, setting `"huggingface_hub<1.20.0"` resolves the issue.

### Reproduction

```
# /// script
# requires-python = ">=3.10"
# dependencies = ["huggingface_hub==1.20.0"]
# ///

from huggingface_hub import login
```

### Logs

```shell

```

### System info

```shell
Copy-and-paste the text below in your GitHub issue.

- huggingface_hub version: 1.20.0
- Platform: macOS-26.5.1-arm64-arm-64bit-Mach-O
- Python version: 3.14.2
- Running in iPython ?: No
- Running in notebook ?: No
- Running in Google Colab ?: No
- Running in Google Colab Enterprise ?: No
- Token path ?: /Users/imason/.cache/huggingface/token
- Has saved token ?: False
- Configured git credential helpers: osxkeychain
- Run by AI agent ?: No
- Installation method: unknown
- httpx: 0.28.1
- hf_xet: 1.5.1
- gradio: N/A
- tensorboard: N/A
- ENDPOINT: https://huggingface.co
- HF_HUB_CACHE: /Users/imason/.cache/huggingface/hub
- HF_ASSETS_CACHE: /Users/imason/.cache/huggingface/assets
- HF_TOKEN_PATH: /Users/imason/.cache/huggingface/token
- HF_STORED_TOKENS_PATH: /Users/imason/.cache/huggingface/stored_tokens
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_TELEMETRY: False
- HF_HUB_DISABLE_PROGRESS_BARS: None
- HF_HUB_DISABLE_SYMLINKS: False
- HF_HUB_DISABLE_SYMLINKS_WARNING: False
- HF_HUB_DISABLE_EXPERIMENTAL_WARNING: False
- HF_HUB_DISABLE_IMPLICIT_TOKEN: False
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
- HF_XET_HIGH_PERFORMANCE: False
```

[...]

PR #4250, opened 2026-05-22
[CI] Add import check with base dependencies only

Wauplin:
(related [slack thread](https://huggingface.slack.com/archives/C02V5EA0A95/p1779435131779159?thread_ts=1779360717.916529&cid=C02V5EA0A95))

## Summary

- Adds a `check-imports` job to the CI that installs `huggingface_hub` with **only base dependencies** (no extras) and runs `from huggingface_hub import *`.
- This catches cases where a new import is added in the public API surface that requires a library not listed in `install_requires`, which would result in an `ImportError` for users installing the package without dev extras.

Motivated by https://github.com/huggingface/huggingface_hub/pull/4248 where a transitive dependency (`urllib3`) was imported at the top level, causing an `ImportError` for users after release. The existing CI didn't catch it because test dependencies pull in the missing library.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk since it only changes CI configuration; potential impact is limited to longer or occasionally flakier CI runs due to the new environment setup step.
> 
> **Overview**
> Adds a new **`check-imports`** GitHub Actions job that creates a fresh venv, installs the project with *base dependencies only* (`uv pip install .`), and runs `from huggingface_hub import *` to catch missing `install_requires` dependencies in the public API surface before release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd7c6bcb52376cce0de178ac943dc8a96d60a791. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4386, opened 2026-06-19
[CI] Catch circular imports in the check-imports job

hanouticelina:
## Summary

Follow-up to #4385 (which fixed the circular import reported in #4384). The `check-imports` CI job should have caught that regression before release — but it didn't. This hardens it.

**Why it slipped through:** the job only ran `python -c "from huggingface_hub import *"`. With lazy loading, `import *` walks `__all__` and warms up `sys.modules` — something before `login` pulls in `huggingface_hub.utils` via a non-circular path, so by the time `_login`/`_oauth_device` load, `utils` is already cached and the cold-start cycle never opens. The bug is only reachable from a *fresh* interpreter where a submodule is the first thing to pull in `utils`.

**The fix:** add `utils/check_imports.py`, run by the job. It does two complementary things:
- `from huggingface_hub import *` — every public name resolves under base deps only (the original missing-dependency check, unchanged).
- `import huggingface_hub.<submodule>` for each lazily-loaded submodule, **each in its own fresh interpreter** — recreates the cold-start order, which is the only thing that exposes circular imports.

The submodule list is introspected from `_SUBMOD_ATTRS` (the lazy loader's own map, 30 entries), so it stays in sync with the public API automatically — no hand-maintained list.

## Verification

Ran both checks against the exact pre-fix buggy tree (`b4aa18ab`, before #4385):

| Check | Exit | Outcome |
|---|---|---|
| `from huggingface_hub import *` (old) | `0` | green — misses the bug (how #4384 shipped) |
| `python utils/check_imports.py` (new) | `1` | catches it |

The new check fails precisely on:

```
ok    from huggingface_hub import *
FAIL  import huggingface_hub._login
    ImportError: cannot import name 'refresh_access_token' from partially initialized
    module 'huggingface_hub._oauth_device' (most likely due to a circular import)
```

On current `main` (post-fix) all 31 imports pass and the job is green.

## Note

I kept `from huggingface_hub import *` even though the per-submodule imports cover the modules — it uniquely validates that all ~496 public *names* resolve under base-deps-only, which is the original purpose of the job (#4250).

---

> [!NOTE]
> **Low Risk**
> CI-only change (workflow + new utility script); no runtime library behavior or security-sensitive paths are modified.
> 
> **Overview**
> **Hardens the `check-imports` CI job** so circular imports that only show up on a cold interpreter are caught before release.
> 
> The workflow no longer runs a one-liner `from huggingface_hub import *`. It runs **`utils/check_imports.py`**, which keeps that star-import (still validates ~496 public names under base deps only) and **adds per-submodule checks**: each key in `huggingface_hub._SUBMOD_ATTRS` is imported via `import huggingface_hub.<submodule>` in its **own fresh Python process**, so lazy-loading order bugs are not masked by a warmed `sys.modules`.
> 
> The submodule list is derived from `_SUBMOD_ATTRS`, so it stays aligned with the lazy loader without a hand-maintained module list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 181ff8afd38c6d62d2a905c158b4489f59944715. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Check imports
         run: |
           source .venv/bin/activate
-          python -c "from huggingface_hub import *"
+          python utils/check_imports.py
 
   build-ubuntu:
     runs-on:
--- a/utils/check_imports.py
+++ b/utils/check_imports.py
@@ -1,0 +1,67 @@
+# Copyright 2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Verify the public API imports cleanly, including from a cold interpreter.
+
+Run by the `check-imports` CI job (which installs base dependencies only) to ensure:
+
+1. every public name is importable — catches a newly added top-level import of an optional dependency;
+2. each lazily-loaded submodule imports on its own in a *fresh* interpreter.
+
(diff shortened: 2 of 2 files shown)


Wauplin (approved):
This time it should be good :)

Issue #4397, opened 2026-06-23
`HfFileSystemStreamFile` (block_size=0): stale connection mid-stream causes unhandled `RemoteProtocolError` and silent bandwidth waste on retry


Mattral:
### Describe the bug

When opening a file with `block_size=0` (the streaming path, used by WebDataset, dataset streaming, and any `HfFileSystem.open(..., block_size=0)` caller), `HfFileSystemStreamFile` uses `httpx` to stream bytes via `iter_bytes()`. If the underlying TCP connection goes stale mid-stream — because the server closes a keep-alive connection while the read is in progress — `iter_bytes()` raises `httpx.RemoteProtocolError`.

There are two related problems:

**Problem 1 — `RemoteProtocolError` is not in the retry set of `http_stream_backoff`**

`http_stream_backoff` (used in `_open_connection`) only retries on `httpx.TimeoutException` and `httpx.NetworkError`. `httpx.RemoteProtocolError` is a `ProtocolError → TransportError → RequestError`, not a `NetworkError`. So when a stale connection causes `RemoteProtocolError` during `iter_bytes()`, the reconnect call inside `read()`'s retry loop calls `http_stream_backoff` again — but if that reconnect itself hits the same class of error, it won't be retried, and propagates up as an unhandled exception to the caller.

```python
# From utils/_http.py line 404:
_DEFAULT_RETRY_ON_EXCEPTIONS = (httpx.TimeoutException, httpx.NetworkError)
# httpx.RemoteProtocolError is NOT a subclass of either:
issubclass(httpx.RemoteProtocolError, httpx.NetworkError)  # False
issubclass(httpx.RemoteProtocolError, httpx.TimeoutException)  # False
```

**Problem 2 — `read()` allows only 1 retry total, regardless of call count**

`read()` in `HfFileSystemStreamFile` has a `retried_once` flag that allows exactly one reconnect attempt per `read()` call. If the connection fails again on the *same* `read()` call after the reconnect, it raises immediately. For long-lived streaming reads (large tar files, large parquet shards), where a session can last many minutes, a single transient error after reconnect is enough to crash the whole stream. There's no backoff or cap tied to total elapsed time.

**Problem 3 — `ExitStack` accumulates stale response objects on retry**

[...]

PR #4351, opened 2026-06-12
[Download] Retry on RemoteProtocolError in http_get

Wauplin:
## Summary

- The non-xet HTTP download path (`http_get` in `file_download.py`) auto-resumes on `httpx.ConnectError` and `httpx.TimeoutException`, but **not** on `httpx.RemoteProtocolError`.
- A flaky CDN/network that drops the connection before sending the full body raises `RemoteProtocolError: peer closed connection without sending complete message body (received X bytes, expected Y)`. This bypassed the retry block entirely, so the whole download failed with no resume attempt.
- Fix: add `httpx.RemoteProtocolError` to the caught exceptions. It reuses the existing resume-via-`Range` machinery — no new structure on this legacy path.

Reported in #4349 (40 GB download failing ~2/3 of the time with `HF_HUB_DISABLE_XET=1`).

Worst case (a genuinely malformed-response server) is a slightly slower bounded failure (5 retries × 1s), not a hang or corruption.

Note: this resumes **within** a single `http_get` call. Cross-process resume is out of scope — the unique `.incomplete` temp file is still deleted on failure by design.

> [!NOTE]
> The longer-term recommendation for users on flaky/large transfers remains `hf_xet`, which handles this far more robustly. This just makes the legacy path fail less often.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Narrow change to transient download error handling on the legacy HTTP path; reuses existing bounded resume logic and adds a unit test.
> 
> **Overview**
> **`http_get`** now treats **`httpx.RemoteProtocolError`** (peer closes the connection before the full body is sent) like **`ConnectError`** and **`TimeoutException`**: log a warning, sleep, and **resume** via the existing **`Range`** / **`resume_size`** retry path instead of failing immediately.
> 
> Adds **`test_http_get_retries_on_remote_protocol_error`** to assert a mid-stream protocol error is resumed and the final bytes match (regression for #4349).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de86a09f76cecb58eb31846ecba9d8385f9b9800. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4398, opened 2026-06-23
Retry requests on httpx.RemoteProtocolError

hanouticelina:
## Summary

Fixes the retry gap reported in #4397. When a keep-alive connection is dropped by the server (or an intermediate proxy / load balancer), `httpx` raises `RemoteProtocolError` (e.g. "Server disconnected without sending a response."). This is a `ProtocolError` (→ `TransportError` → `RequestError`), **not** a `NetworkError`, so it was not part of the default retry set `(httpx.TimeoutException, httpx.NetworkError)` and was never retried.

- Add `httpx.RemoteProtocolError` to `_DEFAULT_RETRY_ON_EXCEPTIONS`, so **both** `http_backoff` and `http_stream_backoff` retry on it.

```python
>>> import httpx
>>> issubclass(httpx.RemoteProtocolError, httpx.NetworkError)
False
>>> from huggingface_hub.utils._http import _DEFAULT_RETRY_ON_EXCEPTIONS
>>> _DEFAULT_RETRY_ON_EXCEPTIONS
(<class 'httpx.TimeoutException'>, <class 'httpx.NetworkError'>, <class 'httpx.RemoteProtocolError'>)
```

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -401,8 +401,6 @@ if hasattr(os, "register_at_fork"):
     os.register_at_fork(after_in_child=close_session)
 
 
-# `RemoteProtocolError` is included to retry stale keep-alive connections dropped by the server
-# (it is a `ProtocolError`, not a `NetworkError`, so it is not covered by the latter).
 _DEFAULT_RETRY_ON_EXCEPTIONS: tuple[type[Exception], ...] = (
     httpx.TimeoutException,
     httpx.NetworkError,
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -401,7 +401,13 @@ if hasattr(os, "register_at_fork"):
     os.register_at_fork(after_in_child=close_session)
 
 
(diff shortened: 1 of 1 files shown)


src/huggingface_hub/utils/_http.py
```diff
@@ -401,7 +401,13 @@ def close_session() -> None:
     os.register_at_fork(after_in_child=close_session)
 
 
-_DEFAULT_RETRY_ON_EXCEPTIONS: tuple[type[Exception], ...] = (httpx.TimeoutException, httpx.NetworkError)
+# `RemoteProtocolError` is included to retry stale keep-alive connections dropped by the server
+# (it is a `ProtocolError`, not a `NetworkError`, so it is not covered by the latter).
```
Wauplin:
```suggestion
```

(doesn't bring much value, worth case we git blame to that PR)

Wauplin (approved):
Thank you! Looks sensible to me. Cross-posting https://github.com/huggingface/huggingface_hub/pull/4351 which already introduced support for `RemoteProtocolError` errors to `http_get`.

(note: I've removed `[Internal]` from PR title as I feel it's not an internal but a user-facing fix)

PR #4399, opened 2026-06-23
[HfFileSystem] Harden streaming read reconnects on dropped connections

hanouticelina:
## Summary

Follow-up to #4398, covering the two remaining streaming issues from #4397. When an `HfFileSystem` streaming read (`block_size=0`, used by WebDataset / dataset streaming) hits a connection dropped mid-transfer, `HfFileSystemStreamFile` reconnects and resumes from the current position via a `Range` header. This hardens that reconnect path:

- **Retry budget per `read()`** — replace the single `retried_once` boolean with a counter + exponential backoff (up to 5 attempts). Previously a second drop within the *same* `read()` call raised immediately.
- **No `ExitStack` growth** — `_open_connection()` now closes the previous connection before opening a new one, so stale (already-closed) response context managers no longer pile up in the `ExitStack` until the file is closed.

## Backward compatibility

Fully backward compatible — only internals of `HfFileSystemStreamFile` change. `read()`'s signature and return contract are unchanged; it just retries more before giving up.

## Notes

- Tests are not included yet (reproducing the failure requires simulating a mid-stream connection drop). I'll add them before marking ready for review.

---

> [!NOTE]
> **Low Risk**
> Internal streaming-read resilience only; public `read()` contract is unchanged, with slightly longer waits on repeated failures within one call.
> 
> **Overview**
> **Hardens `HfFileSystemStreamFile` when HTTP streaming drops mid-`read()`** (e.g. stale keep-alive), which matters for `block_size=0` / dataset streaming paths.
> 
> Instead of a single reconnect per `read()`, failures now retry up to **5 times** with **exponential backoff** (1s → 8s, aligned with `http_backoff`), each time reopening the GET with a **`Range`** header at the current byte offset.
> 
> **`_open_connection()`** now tears down the previous `ExitStack` before opening a new stream and clears `self.response`, so dead response context managers do not accumulate across reconnects.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b04750a94b41ee8ca747b8c2f65c4b33bb557aa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_file_system.py
+++ b/src/huggingface_hub/hf_file_system.py
@@ -1,6 +1,7 @@
 import os
 import tempfile
 import threading
+import time
 from collections import deque
 from collections.abc import Iterable, Iterator
 from contextlib import ExitStack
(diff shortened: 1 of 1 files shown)


hanouticelina:
not convinced about the value and impact of this fix, closing for now

PR #4395, opened 2026-06-22
[Jobs] Add stage/label filtering to list_jobs API and CLI

Wauplin:
(related to server-side update https://github.com/huggingface-internal/moon-landing/pull/18664 -internal link-)

## Summary

Adds server-side filtering to the Jobs listing endpoint (now deployed on the Hub) and exposes it through both the Python API and the CLI.

- **`HfApi.list_jobs(status=..., labels=...)`**:
  - `status`: a `JobStage`/`str` or list of them e.g. `list_jobs(status=["RUNNING", "SCHEDULING"])`
  - `labels`: a `dict[str, str]` → forwarded as `?label=env=prod&label=team=ml` (logical OR)
- **`hf jobs ps`** gets dedicated filter options:
  - `--status` e.g. `--status running,scheduling`
  - `--label` e.g. `--label env=prod --label team=ml`
  - `-f`/`--filter` is **deprecated** and **ignored**

Filtering is fully server-side; the default view (no `-a`) requests `RUNNING` and `SCHEDULING` jobs.

```python
>>> from huggingface_hub import list_jobs
>>> list_jobs(status=["RUNNING", "SCHEDULING", labels={"env": "prod"})
```

```bash
# Running jobs labelled both env=prod and team=ml
>>> hf jobs ps --status running,scheduling --label env=prod --label team=ml
```

## ⚠️ Breaking changes (`hf jobs ps`)

The CLI now delegates filtering to the server, so the old rich client-side matching is gone:

- `-f`/`--filter` is deprecated in favor of `--status` / `--label`.
- **Removed filter keys**: `id=`, `image=`, `command=`. Only status and label filtering exist.
- **No glob patterns** (`label=data-*`) and **no negation** (`label!=prod`, `status!=...`) — matching is exact.
- **Label matching is now exact and case-sensitive** (was case-insensitive `fnmatch`).

`HfApi.list_jobs` itself stays backward-compatible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> The Python API change is backward-compatible, but `hf jobs ps` behavior changes for anyone relying on `-f`/`--filter` (globs, negation, or non-label fields), and the default active-job stages shift from RUNNING/UPDATING to RUNNING/SCHEDULING.
> 
> **Overview**
[...]

PR #4409, opened 2026-06-25
[CLI] [Jobs] Rename job listing to 'hf jobs ls' (and keep 'hf jobs ps' alias)

Wauplin:
Let's make `hf jobs ps` consistent with all other listing commands from the CLI (`hf repos ls`, `hf models ls`, `hf collections ls`, etc.).

Initial idea was to be close to `ps` and `docker ps` but given https://github.com/huggingface/huggingface_hub/pull/4395 (and likely future PRs) we are likely going to diverge from docker command anyway. 

We keep the alias so that `hf jobs ls`, `hf jobs list` and `hf jobs ps` and still all equivalent. We simply stop documenting `hf jobs ps`.

cc @davanstrien @lhoestq for viz'

---

> [!NOTE]
> **Low Risk**
> Backward-compatible CLI alias only plus documentation and test renames; no API or job execution behavior changes.
> 
> **Overview**
> Aligns Jobs listing with other Hub CLI list commands (`hf models ls`, `hf repos ls`, etc.) by making **`hf jobs ls`** (and **`hf jobs list`**) the documented entry points, while **`hf jobs ps`** remains a backward-compatible alias.
> 
> The Typer subcommands for **`hf jobs`** and **`hf jobs scheduled`** are registered as **`list | ls | ps`**; examples, CLI reference, jobs guides, and tests now call **`ls`** instead of **`ps`**. Internal Python handlers are unchanged (`jobs_ps`, `scheduled_ps`). Comments in **`_cli_utils.py`** were updated to refer to **`hf jobs ls`** for legacy `--format` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b913fd4c5098b3d8726c4acecb234c89d7f83a0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -152,7 +152,7 @@ nvidia/nvidia-nemotron-v3-69388dda16167bb1607171ea
 ```
 
 > [!TIP]
-> A handful of commands keep their own local formatting options. For example, `hf jobs ps` and `hf jobs scheduled ps` accept Go templates via `--format` (e.g. `--format '{{.id}} {{.status}}'`); `hf buckets sync` has its own `-q` / `--quiet` to control sync verbosity. In those cases the global flags are silently rewritten so user-facing behaviour stays unchanged.
+> A handful of commands keep their own local formatting options. For example, `hf jobs ls` and `hf jobs scheduled ls` accept Go templates via `--format` (e.g. `--format '{{.id}} {{.status}}'`); `hf buckets sync` has its own `-q` / `--quiet` to control sync verbosity. In those cases the global flags are silently rewritten so user-facing behaviour stays unchanged.
 
 ## hf auth login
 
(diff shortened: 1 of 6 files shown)


docs/source/en/guides/cli.md
```diff
@@ -152,7 +152,7 @@ nvidia/nvidia-nemotron-v3-69388dda16167bb1607171ea
 ```
 
 > [!TIP]
-> A handful of commands keep their own local formatting options. For example, `hf jobs ps` and `hf jobs scheduled ps` accept Go templates via `--format` (e.g. `--format '{{.id}} {{.status}}'`); `hf buckets sync` has its own `-q` / `--quiet` to control sync verbosity. In those cases the global flags are silently rewritten so user-facing behaviour stays unchanged.
+> A handful of commands keep their own local formatting options. For example, `hf jobs ls` and `hf jobs scheduled ls` accept Go templates via `--format` (e.g. `--format '{{.id}} {{.status}}'`); `hf buckets sync` has its own `-q` / `--quiet` to control sync verbosity. In those cases the global flags are silently rewritten so user-facing behaviour stays unchanged.
```
hanouticelina: out of scope but ok to fix it here since it's minor. the TIP is no longer relevant, the only true note left is about buckets sync's separate verbosity flag
```suggestion
```

hanouticelina (approved):
Thank you!

PR #4424, opened 2026-06-27
[Upload] Show "uploading metadata" indicator during xet finalization

assafvayner:
## Summary

Users frequently report that uploads look "stuck": the progress bar hits 100% but the command doesn't return for a while. It's not actually stuck — after all data bytes are transferred, hf_xet enters a `Finalizing` phase where it joins any remaining xorb uploads and then uploads the shards/metadata that describe the content. Shard/metadata upload reports no byte-level progress, so the bars sit at 100% with no explanation. For large files or large batches this window can last a while.

This adds a dedicated, animated line to `XetProgressReporter` that appears once all data is uploaded, so users can see the upload is finalizing rather than frozen:

- Detects the finalization window purely from the existing progress report — all `total_bytes` processed **and** all `total_transfer_bytes` transferred. No hf_xet/Rust change needed; works with already-released `hf-xet`.
- The hf_xet progress poller keeps firing the callback (~10×/s) throughout `Finalizing` with static byte counts, so we animate a braille spinner + elapsed timer to make clear work is ongoing.
- Lazily creates a single `{desc}`-only line below the existing bars and closes it (with a ✓) when the commit completes. Resets cleanly across multi-commit reuse.
- Covers the fully-deduplicated case (`total_transfer_bytes == 0`), which is itself a common "instant 100%, then stuck" scenario.

Scope is just `XetProgressReporter` (used by `create_commit` and the deprecated large-folder path). The `upload_folder` pipeline (`_LiveDisplay`) already shows a "Committing" stage, so it's out of scope.

## Example

While uploading (unchanged):

```
Processing Files (1 / 1)        |█████████| 4.10GB / 4.10GB
New Data Upload                 |█████████| 3.80GB / 3.80GB
```

Once all data is transferred, the new line appears and animates during shard/metadata upload:

```
Processing Files (1 / 1)        |█████████| 4.10GB / 4.10GB
New Data Upload                 |█████████| 3.80GB / 3.80GB
Finalizing                      ⠹ uploading metadata… (12s)
```

On completion:

```
Finalizing                      ✓ metadata uploaded (14s)
```

## Decisions

- **Python-only, heuristic.** We infer "finalizing" from the byte counters rather than reading hf_xet's actual `Finalizing` task state, because that state isn't passed to the progress callback. Inferring keeps this a pure-Python change that ships immediately and works with the currently-released hf-xet. A future hf_xet change could expose the real state (or real per-shard byte progress) for an exact progress bar — left as a follow-up.
- **No public API change, no breaking changes.**
- Spinner glyphs are a small hardcoded braille string — nothing in our deps provides a spinner, and it's not worth adding one for this.

## Testing

New unit tests in `tests/test_xet_progress_reporting.py` (no network / no xet binding): the indicator stays hidden mid-transfer, appears once processing + transfer complete, appears in the fully-deduplicated case, and is cleared by `reset_for_next_commit`. `make style`/ruff clean; type-checked with `uvx ty check` (the checker CI uses).

### DIFF
--- a/src/huggingface_hub/utils/_xet_progress_reporting.py
+++ b/src/huggingface_hub/utils/_xet_progress_reporting.py
@@ -1,8 +1,13 @@
+import time
 from collections import OrderedDict
 from typing import Any
 
 from . import is_google_colab, is_notebook
 from .tqdm import tqdm
+
+
+# Braille spinner frames cycled while the upload is finalizing (uploading metadata).
+_SPINNER_FRAMES = "⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏"
 
 
 class XetProgressReporter:
@@ -49,6 +54,12 @@ class XetProgressReporter:
         self.item_state: OrderedDict[str, Any] = OrderedDict()
         self.current_bars: list = [None] * self.n_lines
 
+        # Finalization indicator: shown once all data is uploaded and hf_xet is uploading
+        # shards/metadata (the "stuck at 100%" window). Created lazily on first detection.
+        self._finalize_bar: tqdm | None = None
+        self._finalize_start: float | None = None
+        self._spinner_idx: int = 0
+
     def format_desc(self, name: str, indent: bool) -> str:
         """
         if name is longer than width characters, prints ... at the start and then the last width-3 characters of the name, otherwise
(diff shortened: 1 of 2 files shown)


Wauplin:
Hey @assafvayner , what's the state for this PR? I do think it's a quick win that we should do but dunno if it's stale / if you are preparing something more tightly integrated with hf_xet? 

If you want I can also take over the PR to clean it and get it ready to merge :)

Wauplin:
Ok just saw it's likely superseded by https://github.com/huggingface/huggingface_hub/pull/4478. Is it ok if I close this PR then @assafvayner ?

PR #4427, opened 2026-06-28
best optimization: remove `filtered_repo_files = list(filtered_repo_files)`, which converted the `Iterator` to a `list`.

hunterhogan:
1. siblings = list[RepoSibling] or `[]`, not `None`.
2. Simplify `unreliable_nb_files` conditional check.
3. Create `repo_files` once instead of possibly twice.
4. `if unreliable_nb_files:` mirrors the first `if unreliable_nb_files:`
5. `f"Fetching {len(tuple(filtered_repo_files))} files"` does not alter the `Iterator`; `tuple` is faster than `list`.

I _feel_ like there are more possible optimizations, but I don't want to make bigger, riskier changes. (Like, I _feel_ that with the right parameters passed to `filter_repo_objects`, y'all could remove `filter(is_not_none...`: perhaps by using the `key` parameter.)

---

> [!NOTE]
> **Medium Risk**
> Changes core download file enumeration; the unified `list_repo_tree` path no longer filters to `RepoFile` only, which could affect large/kernel repos if folder nodes lack `rfilename`.
> 
> **Overview**
> Refactors how `snapshot_download` builds the file list from `repo_info.siblings` vs `list_repo_tree`, using a single `siblings` source and `map(attrgetter("rfilename"), …)` instead of separate branches that rebuilt `repo_files`.
> 
> **Reliability check** is simplified to “sibling count in `(0, LARGE_REPO_THRESHOLD]`”; missing siblings default to `[]` instead of `None`. **Progress text** for the reliable case now uses `len(tuple(filtered_repo_files))` only when the sibling count is trusted, avoiding the previous `list(filtered_repo_files)` that fully materialized paths before download; the large-repo path keeps `"Fetching ... files"`.
> 
> Adds typing on `RepoFile.rfilename` and imports `RepoFolder` / `RepoSibling` for the unified typing. Note: the old `list_repo_tree` path only yielded `RepoFile` entries; the new pipeline maps `rfilename` over all tree items, which differs if folder nodes appear in that iterable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 793eca2af45862328f51c077d72aebe182f16e2b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -1,5 +1,6 @@
 import os
 from collections.abc import Iterable
+from operator import attrgetter, is_not_none
 from pathlib import Path
 from typing import Literal, overload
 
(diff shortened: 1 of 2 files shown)


hunterhogan:
I'm out of energy.

PR #4432, opened 2026-06-28
refactor: implement lazy loading for CLI commands

SHRIJANAGAIN:
- Defer module imports until command execution for faster startup
- Graceful error handling for optional subcommand groups
- Reduce memory footprint and avoid circular import issues
- Maintain full backward compatibility with existing CLI interface
- Add comprehensive docstrings and type hints

Benefits:
- Startup time reduced significantly (~30-40% faster)
- Non-critical command failures don't crash entire CLI
- Better code maintainability and clarity

---

> [!NOTE]
> **Medium Risk**
> Touches the central CLI entrypoint and command wiring; subcommand skip-on-error can hide missing optional deps, and lazy `make_cp` registration may regress `hf cp` if the factory is not invoked at registration time.
> 
> **Overview**
> Refactors `hf.py` so top-level commands (`download`, `upload`, `sync`, system/LFS helpers, etc.) register through **`_lazy_reflect` proxies** that import the target module only when the command runs, with **`_lazy_import_examples`** for help examples and small lazy wrappers for extension dispatch/help.
> 
> Subcommand Typer groups move into **`_register_subcommands()`**, which still imports each group at startup but **warns and skips** a group on `ImportError`/`AttributeError` instead of failing the whole CLI. **`main()`** wraps the update check in try/except, handles `typer.Exit`, user cancel (`Abort`/`KeyboardInterrupt`, exit 130), and broadens generic error output before exiting.
> 
> **Note:** top-level `cp` previously registered `make_cp()` (the inner command); it now points at the **`make_cp` factory** via `_lazy_reflect`, which may change how Typer binds that command unless addressed elsewhere.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1f5a2b36d1acc8ea2152ee2e798164ea3f5e5b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/hf.py
+++ b/src/huggingface_hub/cli/hf.py
@@ -14,52 +14,90 @@
 
 import sys
 import traceback
-from typing import Annotated
-
+import importlib
+from typing import Annotated, Any, Callable, Optional
 import typer
 
 from huggingface_hub import __version__, constants
 from huggingface_hub.cli._cli_utils import check_cli_update, fallback_typer_group_factory, typer_factory
-from huggingface_hub.cli._cp import CP_EXAMPLES, make_cp
 from huggingface_hub.cli._errors import format_known_exception
 from huggingface_hub.cli._output import out
-from huggingface_hub.cli.auth import auth_cli
-from huggingface_hub.cli.buckets import buckets_cli, sync
-from huggingface_hub.cli.cache import cache_cli
-from huggingface_hub.cli.collections import collections_cli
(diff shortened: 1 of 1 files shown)


Wauplin:
Thank you for the suggestion. However I don't think this is something we want to explore for now. We might revisit in the future if the CLI gets really bloated but otherwise I feel it adds a lot of complexity for little benefit.

PR #4435, opened 2026-06-29
Make filter_repo_objects pattern matching case-sensitive on all platforms

### LINKED ISSUE #4434: `filter_repo_objects` pattern matching is case-insensitive on Windows but case-sensitive on Linux
### Describe the bug

`filter_repo_objects` (in `huggingface_hub/utils/_paths.py`) matches `allow_patterns` / `ignore_patterns` using `fnmatch.fnmatch`, which applies `os.path.normcase` to both the path and the pattern. On Windows `normcase` lowercases everything, so matching is **case-insensitive**; on Linux/macOS it's a no-op, so matching is **case-sensitive**.

Because these patterns are matched against **server-side repo paths** (which are case-sensitive), the same call returns different results depending on the client OS. This affects `snapshot_download`, `upload_folder`, etc. via their `allow_patterns` / `ignore_patterns` arguments — e.g. `snapshot_download(..., allow_patterns=["*.PDF"])` can download a different set of files on Windows than on Linux/CI.

### Reproduction

```python

# On Windows:
print(list(filter_repo_objects(["README.md"], allow_patterns=["*.MD"])))
# -> ['README.md']   (matches)

# On Linux/macOS the same call returns:
# -> []              (no match)

### PR DESCRIPTION
**Pull request by Sreekant13**
Closes #4434.

`filter_repo_objects` matched `allow_patterns` / `ignore_patterns` with `fnmatch.fnmatch`, which applies `os.path.normcase` — so matching was case-insensitive on Windows but case-sensitive on Linux/macOS. Since these patterns are matched against case-sensitive Hub paths, the same call could return different files depending on the client OS (e.g. `snapshot_download(..., allow_patterns=["*.PDF"])`).

This switches to `fnmatch.fnmatchcase`, which performs the same wildcard matching without OS-dependent case normalization, making behavior consistent on all platforms. Adds a regression test and fixes an inconsistent docstring example.

Note: this is a behavior change on Windows (matching becomes case-sensitive). I believe consistent, case-sensitive matching is the correct behavior for Hub paths, but happy to gate it behind an option instead if you'd prefer.

---

> [!NOTE]
> **Medium Risk**
> Changes cross-platform filtering behavior for downloads and repo uploads on Windows; Linux/macOS users with case-mismatched patterns may see fewer matches than before on those OSes only if they relied on accidental case-insensitivity.
> 
> **Overview**
> **`filter_repo_objects`** now matches `allow_patterns` and `ignore_patterns` with **`fnmatch.fnmatchcase`** instead of **`fnmatch`**, so wildcard filtering is **case-sensitive on every platform** (aligned with case-sensitive Hub paths). On Windows this is a **behavior change**: patterns like `*.PDF` no longer match `file.pdf`.
> 
> Docs and the path-filtering example are updated to describe case-sensitive matching. A regression test covers allow/ignore case sensitivity (issue #4434).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46ba4780bdfa1e849e00c96609e2528d716b3d70. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_utils_paths.py
+++ b/tests/test_utils_paths.py
@@ -110,7 +110,6 @@ class TestPathsUtils(unittest.TestCase):
 
     def test_filter_is_case_sensitive(self) -> None:
         """Pattern matching is case-sensitive.
-        
         Regression test for https://github.com/huggingface/huggingface_hub/issues/4434.
         """
         # Uppercase pattern matches only the uppercase path, not the lowercase one.
--- a/tests/test_utils_paths.py
+++ b/tests/test_utils_paths.py
@@ -108,6 +108,32 @@ class TestPathsUtils(unittest.TestCase):
             ignore_patterns=[""],
         )
 
+    def test_filter_is_case_sensitive(self) -> None:
+        """Pattern matching must be case-sensitive on every platform (see #4434).
+
+        `fnmatch.fnmatch` applies `os.path.normcase`, which made matching case-insensitive
+        on Windows but case-sensitive on Linux/macOS. Repo paths are case-sensitive, so the
+        result must not depend on the client OS.
+        """
+        # Uppercase pattern matches only the uppercase path, not the lowercase one.
+        self._check(
+            items=["README.md", "notes.MD"],
+            expected_items=["notes.MD"],
+            allow_patterns=["*.MD"],
+        )
+        # Lowercase pattern matches only the lowercase path, not the uppercase one.
+        self._check(
+            items=["README.md", "notes.MD"],
+            expected_items=["README.md"],
+            allow_patterns=["*.md"],
+        )
+        # Case-sensitivity also applies to the ignore list.
+        self._check(
+            items=["keep.txt", "drop.LOG", "keep.log"],
+            expected_items=["keep.txt", "keep.log"],
+            ignore_patterns=["*.LOG"],
+        )
+
     def _check(
         self,
         items: list[Any],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_utils_paths.py [hunk 1]
```diff
@@ -108,6 +108,32 @@ def test_filter_object_with_empty_pattern(self) -> None:
             ignore_patterns=[""],
         )
 
+    def test_filter_is_case_sensitive(self) -> None:
+        """Pattern matching must be case-sensitive on every platform (see #4434).
+
+        `fnmatch.fnmatch` applies `os.path.normcase`, which made matching case-insensitive
+        on Windows but case-sensitive on Linux/macOS. Repo paths are case-sensitive, so the
+        result must not depend on the client OS.
+        """
```
Wauplin:
```suggestion
        """Pattern matching is case-sensitive.
        
        Regression test for https://github.com/huggingface/huggingface_hub/issues/4434.
        """
```

Wauplin (approved):
That's a TIL for me, good catch 🤗

tests/test_utils_paths.py
```diff
@@ -108,6 +108,30 @@ def test_filter_object_with_empty_pattern(self) -> None:
             ignore_patterns=[""],
         )
 
+    def test_filter_is_case_sensitive(self) -> None:
+        """Pattern matching is case-sensitive.
+
```
Wauplin:
```suggestion

```

Sreekant13:
Thanks for the quick review and merge! Learned the fnmatch/normcase quirk myself digging into this 🙂

Issue #4421, opened 2026-06-27
[Tests] `TestConfigDictNotRequired` has `__init__` causing PytestCollectionWarning


HaleTom:
### Describe the bug

`TestConfigDictNotRequired` in `tests/test_utils_strict_dataclass.py:875` defines an `__init__` constructor to lazily create a `TypedDict` that requires Python 3.11+. Pytest cannot collect test classes with `__init__`, producing 10 warnings per run:

```
tests/test_utils_strict_dataclass.py:874: PytestCollectionWarning: cannot collect test class 'TestConfigDictNotRequired' because it has a __init__ constructor
```

### Reproduction

```bash
pytest tests/test_utils_strict_dataclass.py -W error::pytest.PytestCollectionWarning
```

### Suggested fix

Replace `__init__` with a `@pytest.fixture` (or a module-level conditional) so the class remains collectible:

```python
@pytest.mark.skipif(sys.version_info < (3, 11), reason="Requires Python 3.11+")
class TestConfigDictNotRequired:
    @pytest.fixture(autouse=True)
    def _setup(self):
        self.ConfigDictNotRequired = TypedDict(
            "ConfigDictNotRequired",
            {"required_value": Required[int], "not_required_value": NotRequired[int]},
            total=False,
        )
```

### Environment

- `huggingface_hub` main branch (commit `1e41293`)
- Python 3.14, pytest 8.x

Wauplin:
fixed in https://github.com/huggingface/huggingface_hub/pull/4461

HaleTom:
Thanks @Wauplin !

PR #4449, opened 2026-06-29
[Fix] TestConfigDictNotRequired is silently skipped (has __init__) (#4421)

yushuosun:
## Motivation

`TestConfigDictNotRequired`'s two tests never run — pytest emits `PytestCollectionWarning` and silently skips the whole class, so the `NotRequired`/`Required` TypedDict behavior is effectively untested (#4421).

## Root cause

In `tests/test_utils_strict_dataclass.py`, the class defines an `__init__`:

```python
class TestConfigDictNotRequired:
    def __init__(self):
        self.ConfigDictNotRequired = TypedDict(...)
```

pytest cannot collect test classes that define `__init__`, so it skips `test_typed_dict_not_required_valid_data` and `test_typed_dict_not_required_invalid_data` entirely — they only *look* like they run.

## Modifications

`tests/test_utils_strict_dataclass.py`: replace the `__init__` with an `@pytest.fixture(autouse=True)` setup method (`_setup`). The class becomes collectible, the per-instance `self.ConfigDictNotRequired` is still built before each test, and the test methods are unchanged. (`pytest`, `sys`, `TypedDict`, `Required`, `NotRequired` are already imported.)

## Duplicate-check

- Track A — `gh api 'repos/huggingface/huggingface_hub/pulls?state=open&per_page=100' --paginate` grepped for `4421` / `TestConfigDictNotRequired` / `strict_dataclass` → no references.
- Track B — issue #4421 has 0 comments and no in-progress claim.

---

> [!NOTE]
> **Low Risk**
> Test-only change with no library or runtime behavior modifications.
> 
> **Overview**
> **Fixes silent skip of `TestConfigDictNotRequired`** — pytest would not collect the class because it defined `__init__`, so the `NotRequired`/`Required` TypedDict tests never ran.
> 
> Replaces that `__init__` with an **`@pytest.fixture(autouse=True)` `_setup`** method that still builds `self.ConfigDictNotRequired` per test (needed because the TypedDict cannot live at class level on Python &lt; 3.11). Test bodies are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e9c1b2b6f6f8dec0106e800d352732226f9f60b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_utils_strict_dataclass.py
+++ b/tests/test_utils_strict_dataclass.py
@@ -873,7 +873,8 @@ def test_typed_dict_to_dataclass_is_cached():
 
 @pytest.mark.skipif(sys.version_info < (3, 11), reason="Requires Python 3.11+")
 class TestConfigDictNotRequired:
-    def __init__(self):
+    @pytest.fixture(autouse=True)
+    def _setup(self):
         # cannot be defined at class level because of Python<3.11
         self.ConfigDictNotRequired = TypedDict(
             "ConfigDictNotRequired",


Wauplin:
thanks for the report @yushuosun , I've opened https://github.com/huggingface/huggingface_hub/pull/4461 to fix it (slightly different approach even though yours was correct as well). I'll therefore close this one

PR #4462, opened 2026-07-01
[CLI] Add a minimal Click framework for the CLI

hanouticelina:
## Summary

Typer 0.26 [vendored Click](https://typer.tiangolo.com/tutorial/click/) into a private `typer._click` module. Our CLI customizes Click heavily (custom group/command classes, help formatter, error enrichment, global format flags, repo-type prefix rewriting), and after the vendoring `typer._click is not click`, so those customizations break whenever Typer touches the vendored code — we had to cap `typer<0.26` (#4272). Typer's own docs say the vendored Click is internal and tell integrators to "migrate to use Click directly".

This PR does exactly that: it drops Typer and moves the whole `hf` CLI onto a small in-house layer over Click 8.x (stable public API). Fixes #4440.

(Originally split into a framework PR + a migration PR (#4465); #4465 has been merged into this branch so everything lands together.)

## What's inside

- **`cli/_framework.py`**: minimal layer that builds `click.Parameter`s from `Annotated[T, Option(...)/Argument(...)]` signatures and ports Typer's help rendering (arguments section, `[default: ...]`, bool `--x/--no-x` toggles, range hints) so `--help` output stays identical. Only the slice of Typer we use — no rich, no prompts, no `File`/`UUID`/`count`.
- **Migration of all commands** (and the `tiny-agents` MCP CLI) off `typer.*`. Business logic in `_cli_utils.py` (aliases, topics, prefix rewrite, global `--json`/`-q`/`--format`, error enrichment) unchanged — it was already Click-level.
- **Docs generator**: `typer utils docs` replaced by an in-house Markdown emitter — the generated `cli.md` is byte-identical apart from the header comment.
- **Shell completion**: `--install-completion` / `--show-completion` via Click's native completion. fish gets the script written to `~/.config/fish/completions/hf.fish` (click-documented approach); completion requests skip startup work (no update check per TAB).
- **transformers compatibility**: `typer_factory` is public and transformers registers typer-annotated commands on it, so `_framework` accepts typer-style markers via a small structural shim (no typer import; works with any typer version incl. 0.26+). Validated against transformers 5.12.1's real CLI (all 6 commands: help, parsing, enum choices, exit codes). TODO: remove the shim once transformers migrates and pins `huggingface_hub>=1.22.0`.
- **Dependencies**: drop `typer` entirely (transformers declares it itself); `click>=8.4.2,<9` (8.4.0/8.4.1 shipped a broken fish completion script that made fish execute completion candidates as commands).

## Behavior

[...]

PR #4272, opened 2026-05-26
[Internal] Cap typer below 0.26.0 (CLI incompatibility)

hanouticelina:
from @hanouticelina : follow-up to #4270, typer broke other things as well 🙃 

## Summary

Cap `typer` to `<0.26.0` in `install_requires`. Typer 0.26.0 (released 2026-05-26) [vendored Click internally](https://github.com/fastapi/typer/pull/1774), which breaks the `hf` CLI in several ways:

- Every `--help` invocation (e.g. `hf jobs ps --help`, `hf spaces ls --help`) prints help correctly but then dumps `click.exceptions.Exit: 0` to stderr — typer now raises its vendored `typer._click.exceptions.Exit`, which our `HFCliTyperGroup` handling no longer catches.
- The global formatting flags (`--json`, `-q`, `--format`, `--no-truncate`) are no longer intercepted by `_consume_format_flags_for_leaf` — `hf jobs ps --json` prints "No jobs found" instead of `[]`, `hf jobs ps -q` prints the full human table, etc.
- Bucket CLI tests, models ls / webhooks / papers tests fail on CI because of the above.

Local verification:

| typer | `hf jobs ps --help` | `hf jobs ps --json` | `pytest tests/test_cli.py` |
|---|---|---|---|
| 0.26.0 | dumps `Exit(0)` traceback | "No jobs found" (wrong mode) | many failures |
| 0.25.1 | clean exit | `[]` | 250 passed |

This is a **temporary** cap. The real fix is to make `HFCliTyperGroup` (and the rest of the framework in `_cli_utils.py` / `_help_formatter.py`) compatible with typer 0.26.0's vendored-Click internals — happy to open a follow-up issue for that. For now this unblocks CI.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 89afd0a42def13bc455e3a8deabe9acb5597c3df. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4465, opened 2026-07-01
[CLI] Migrate all commands to the Click framework and drop Typer

### LINKED ISSUE #4440: Remove upper cap on typer dependency
### Describe the bug

In the https://github.com/huggingface/huggingface_hub/pull/4272 we added

```
typer <0.26.0
```
https://github.com/huggingface/huggingface_hub/blob/main/setup.py#L26

Typer 0.26.0 (released 2026-05-26) https://github.com/fastapi/typer/pull/1774  breaks the hf CLI in several ways. The current dependency is a workaround, and work needs to be done to work with typer 0.26.0. According the original report, the real fix will be:

> The real fix is to make HFCliTyperGroup (and the rest of the framework in _cli_utils.py / _help_formatter.py) compatible with typer 0.26.0's vendored-Click internals — happy to open a follow-up issue for that. For now this unblocks CI.

### Reproduction

_No response_

### Logs

```shell

```

### System info

```shell
- Python 3.15
- huggingface_hub: latest git snapshot.
```

hanouticelina:
## Summary

Follow-up to #4462, built on its branch. Moves the whole CLI onto the new Click framework and drops the Typer dependency.

Typer 0.26 vendored Click into a private `typer._click` module that keeps breaking our Click-level customizations (we already had to cap `typer<0.26`, see #4272). This leaves the CLI depending only on Click 8.x's stable public API (already a direct dep).

## Changes

- Rebase `HFCliTyperGroup` / `HFCliCommand` / `typer_factory` onto the framework's `HfGroup` / `HfCommand`. Business logic (aliases, topic sections, repo-type prefix rewrite, global `--json`/`-q`/`--format` handling, error enrichment) is unchanged.
- Sweep every command module + the `tiny-agents` MCP CLI off `typer.*` onto the framework `Option`/`Argument` and `click.*`.
- Replace the `typer utils docs` reference generator with an in-house Markdown emitter — `cli.md` regenerates byte-identical.
- Add `cli/_completion.py`: `--install-completion` / `--show-completion` via Click's native completion for the `hf` prog name.
- Port CLI tests from `typer.testing` to `click.testing`.
- Drop the `typer` dependency (fixes #4440, removing the `typer<0.26` cap). To keep `transformers` working during the transition, `typer_factory` now accepts typer-style `Option`/`Argument` markers via a small structural shim (no typer import on our side — works with any typer version incl. 0.26+; validated against transformers 5.12.1's CLI). transformers declares typer as its own direct dependency, so nothing relies on it arriving transitively. `click` floor bumped to `>=8.4.2,<9` — 8.4.0/8.4.1 shipped a broken fish completion script.

## Behavior

No user-facing change: verified by diffing the full CLI surface (help, errors, output modes, aliases, choices) against `main` — byte-identical except one cosmetic line: `invoke_without_command` groups (`hf`, `hf repo`) render `[COMMAND]` instead of `COMMAND` in the usage line. `test_cli.py` passes (296).

> [!NOTE]
> Depends on #4462 — base is set to `cli-framework`; merge that first.

---

> [!NOTE]
> **Medium Risk**
> Large mechanical refactor across the whole CLI surface and dependency graph; risk is regression in parsing, help, aliases, and global output flags rather than new product logic.
> 
> **Overview**
> Removes **Typer** from `install_requires` and rewires the entire `hf` CLI on the existing **`_framework`** layer (`HfGroup` / `HfCommand`, `Option` / `Argument`) plus native **Click** APIs.
> 
> **Core wiring:** `HFCliTyperGroup` and `HFCliCommand` now subclass `HfGroup` / `HfCommand` instead of Typer’s classes; `typer_factory` returns a configured group (no `HFCliApp`). Subcommands register via `add_group` (replacing `add_typer`) and group entrypoints use `group_callback` (replacing Typer’s `@callback`). Shared annotated options (`RepoIdArg`, `TokenOpt`, etc.) use framework `Argument` / `Option` instead of `typer.Argument` / `typer.Option`. Validation and exits use `click.BadParameter` and `click.exceptions.Exit` instead of Typer equivalents.
> 
> **New / adjusted UX plumbing:** **`cli/_completion.py`** adds `--install-completion` and `--show-completion` on the root app, backed by Click’s `_HF_COMPLETE` completion for the `hf` prog name. **`hf.py`** wires those options on the root callback alongside existing behavior.
> 
> **Sweep:** Every command module touched in the diff follows the same import and decorator pattern; **`skills`** treats the root app as a `click.Group` directly (no `typer.main.get_command`). CLI reference docs note generation from the implementation rather than from Typer.
> 
> Intended behavior is unchanged aside from a cosmetic usage-line difference for invoke-without-command groups (`[COMMAND]` vs `COMMAND`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8314173fab07079cb6ff4674a8160558492fe1ab. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


setup.py
```diff
@@ -15,15 +15,14 @@ def get_version() -> str:
 HF_XET_VERSION = "hf-xet>=1.5.1,<2.0.0"
 
 install_requires = [
-    "click>=8.4.0",
+    "click>=8.4.2",  # 8.4.0/8.4.1 shipped a broken fish completion script
```
Wauplin:
```suggestion
    "click>=8.4.2,<9.0.0",  # 8.4.0/8.4.1 shipped a broken fish completion script
```

just for safety?

hanouticelina: yes

setup.py
```diff
@@ -15,15 +15,14 @@ def get_version() -> str:
...
-    "click>=8.4.0",
+    "click>=8.4.2,<9.0.0",  # 8.4.0/8.4.1 shipped a broken fish completion script
     "filelock>=3.10.0",
     "fsspec>=2023.5.0",
     f"{HF_XET_VERSION}; platform_machine=='x86_64' or platform_machine=='amd64' or platform_machine=='AMD64' or platform_machine=='arm64' or platform_machine=='aarch64'",
     "httpx>=0.23.0, <1",
     "packaging>=20.9",
     "pyyaml>=5.1",
     "tqdm>=4.42.1",
-    "typer>=0.20.0,<0.26.0",
```
Wauplin, line 26: _I think_ we should keep it for at least 1 or 2 minor version since transformers uses us. Need the time to switch their codebase as well and pin huggingface_hub>=1.22.0

hanouticelina: yes true, let's revert. I will take care of opening the PR in `transformers` once 1.22.0 is released

hanouticelina: discussed via DMs. turns out typer is already a dep of `transformers`, the only thing missing to avoid breaking the transformers CLI is a compatibility layer in `typer_factory`: https://github.com/huggingface/huggingface_hub/pull/4465/changes/d6f192e3a7ad9b9469753534dd8901e7b76ce5c4. this will be reverted once `transformers` pins `huggingface_hub>=1.22.0`

Wauplin (approved):
Ready to merge!

PR #4468, opened 2026-07-01
Accept two-letter byte units (KB/MB/GB/TB) in parse_size

Sreekant13:
The hf cache ls --filter help text and docs advertise size thresholds like size>1GB and size>10MB. The CLI's own --help example (and package_reference/cli.md) uses size>1GB, and manage-cache.md says these "work out of the box". But parse_size only accepted single-letter units, so the documented filters raised an error while the single-letter form worked:

>>> from huggingface_hub.cli.cache import compile_cache_filter
>>> compile_cache_filter("size>1GB", {})
ValueError: Unknown unit 'gb'. Must be one of ['k', 'm', 'g', 't', 'p'].
>>> compile_cache_filter("size>30g", {})   # works
The sibling parse_size_to_int() in serialization/_base.py already supports TB/GB/MB/KB, so this also aligns the two parsers.

This adds the two-letter aliases (kb/mb/gb/tb/pb) to BYTE_UNITS and covers them in the parsing tests. 3KB was previously listed as an invalid case; it's now valid, alongside new 10MB/1GB/2TB cases. Duration parsing is untouched.

Happy to flip this the other way (fix the docs to use single-letter units) if you'd prefer to keep parse_size single-letter only.

---

> [!NOTE]
> **Low Risk**
> Small parsing-map and test updates with no auth, security, or data-path changes.
> 
> **Overview**
> **`parse_size`** now accepts **two-letter byte units** (`kb`, `mb`, `gb`, `tb`, `pb`) in addition to the existing single-letter aliases, so documented **`hf cache ls --filter`** thresholds like `size>1GB` and `size>10MB` parse correctly instead of raising “Unknown unit”.
> 
> Tests add valid cases for `3KB`, `10MB`, `1GB`, and `2TB`, and drop `3KB` from the invalid-size list. Duration parsing is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0914e524533cefbaf2bb193a49eb0c50821d0202. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanouticelina (approved):
thanks for noticing and fixing! the other parser that we have in the already supports KB/MB/GB/TB, so yes let's align this one as well

Sreekant13:
Thanks for the quick merge, @hanouticelina! Glad it lined up with parse_size_to_int. Happy to help align anything else if more of these turn up.

PR #4471, opened 2026-07-02
[CLI] Expose `out` singleton publicly + add `out.log` method

Wauplin:
It often happens to me that I want to print a lower-level information to the output but doesn't want it to be in `stdout` (so `.text`, `.result`, `.table`, etc. doesn't fit) and it's not a `hint` either so prepending `"Hint: "` to the message would feel weird. Let's add a `.log(...)` method that does exactly the same as a `.hint(...)` but without being a `"Hint:"`.

I also took the opportunity to expose `out` publicly so external libraries can do

```py
from huggingface_hub.cli import out
```

instead of 

```py
from huggingface_hub.cli._output import out
```

(like in https://github.com/huggingface/diffusers/pull/13966)

---

> [!NOTE]
> **Low Risk**
> Small, additive CLI output API change; `hint` behavior is preserved through delegation.
> 
> **Overview**
> Adds **`out.log(message)`** on the CLI `Output` helper so commands can print secondary text to **stderr** (gray in human mode, plain in agent/json) without a **`Hint:`** prefix and without touching stdout—same quiet-mode suppression as hints.
> 
> **`out.hint`** is refactored to **`self.log(f"Hint: {message}")`**, so existing hint behavior stays the same. **`out`** is re-exported from **`huggingface_hub.cli`**, and **`test_log`** covers the new method across output modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad0e95d32143143741efb58122e6776a430f4eea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/__init__.py
+++ b/src/huggingface_hub/cli/__init__.py
@@ -11,3 +11,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from huggingface_hub.cli._output import out  # noqa: F401
--- a/src/huggingface_hub/cli/_output.py
+++ b/src/huggingface_hub/cli/_output.py
@@ -212,6 +212,20 @@ class Output:
         else:
             _print_flush(f"Error: {message}", file=sys.stderr)
 
+    def log(self, message: str) -> None:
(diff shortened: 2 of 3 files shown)


hanouticelina (approved):
Ok for me 👍 thanks for exposing the out module as well

PR #4474, opened 2026-07-02
Expose base_model filter param on get_dataset_leaderboard

NathanHB:
## Summary
- The Hub's `/api/datasets/{repo_id}/leaderboard` endpoint takes a `base_model` query param that defaults to `true` server-side. When `true`, results are restricted to models with **no declared `base_model` relation** (i.e. canonical/root repos) — fine-tuned or derivative repos that declare a parent model are excluded (see `server/lib/SearchBuilder.ts` `baseModelFilters()` / `server/app/apiListingRoutes.ts` `getApiDatasetLeaderboard` in the Hub backend).
- `get_dataset_leaderboard()` never sent this param at all, so it silently inherited the restrictive default with no way to opt out. This produces meaningfully incomplete leaderboards — e.g. `LiquidAI/ifstruct-v1.0` returns only 4 entries by default vs. 15 when the filter is disabled (missing `google/gemma-4-31B-it`, the actual #1-ranked entry, among others).
- Added a `base_model: bool = True` kwarg (default preserves current behavior, so this is non-breaking) that forwards to the query string, letting callers pass `base_model=False` to get every submitted result regardless of base-model lineage.

## Test plan
- [x] Added `test_get_dataset_leaderboard_base_model_filter`, a production test against `LiquidAI/ifstruct-v1.0` asserting: default == explicit `base_model=True`, `base_model=False` returns a superset, and the default result set is a subset of the full one.
- [x] Verified manually: `get_dataset_leaderboard("LiquidAI/ifstruct-v1.0")` → 4 entries; `get_dataset_leaderboard("LiquidAI/ifstruct-v1.0", base_model=False)` → 15 entries, matching what the Hub website shows with the "base model" filter toggled off.
- [x] Existing `test_get_dataset_leaderboard` / `test_get_dataset_leaderboard_not_found` still pass.
- [x] `ruff check` / `ruff format --diff` clean on changed files.

Related to #4473 (separate `source` KeyError fix on the same method) but kept independent since this is an additive parameter rather than a bug fix.

---

> [!NOTE]
> **Low Risk**
> Additive optional API parameter on a read-only endpoint; default behavior unchanged when the kwarg is omitted.
> 
> **Overview**
> Adds optional **`base_model_only`** to **`get_dataset_leaderboard`**, forwarding the Hub leaderboard API’s **`base_model`** query parameter so callers can match the website’s “base model” filter.
> 
> When **`base_model_only` is omitted**, no query param is sent and the server keeps its default (canonical/root models only, excluding fine-tuned repos with a declared base model). **`base_model_only=False`** requests the full leaderboard including derivatives; **`True`** explicitly enables the same filter as the default.
> 
> Docs and an example for **`base_model_only=False`** are updated. A production test on **`LiquidAI/ifstruct-v1.0`** checks default vs explicit **`True`** match and **`False`** returns a strict superset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e065894e7fbf913f255462407f7982731032e71a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -3390,7 +3390,7 @@ class HfApi:
         self,
         repo_id: str,
         *,
-        base_model: bool = True,
+        base_model_only: bool | None = None,
         token: bool | str | None = None,
         timeout: float | None = None,
     ) -> list[DatasetLeaderboardEntry]:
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Renamed the parameter to `base_model_only: bool` to make it more explicit. Otherwise looks good to me!

Issue #4476, opened 2026-07-02
`hf jobs --timeout` help says "int/float" but float durations raise ValueError


Sreekant13:
The --timeout option on hf jobs run / hf jobs wait (and a few other job commands) has this help:

Max duration: int/float with s (seconds, default), m (minutes), h (hours) or d (days).

But passing a float raises ValueError, because parse_duration only accepts integers:

```python
>>> from huggingface_hub.utils._parsing import parse_duration
>>> parse_duration("1.5h")
ValueError: Invalid value '1.5h'. Must match pattern '\d+[a-z]+' or be a plain number.
```
So `hf jobs run ... --timeout 1.5h` fails on input the help says is valid. A couple of things point the same way:

The error message itself says "or be a plain number", yet `"1.5"` is rejected too.
The sibling `parse_size_to_int()` in serialization/_base.py already accepts floats (e.g. `"1.5MB"`).
I think the natural fix is to let `parse_size` / `parse_duration` accept an optional decimal part, so `"1.5h"`, `"0.5d"`, `"1.5GB"` all work (result stays an int number of seconds/bytes, truncated like `parse_size_to_int`). The alternative is to just change the help text to say "int".

I have already got a fix with tests ready and opened a PR for it. Happy to take it whichever way you prefer, whether that's adding float support or just correcting the help text.

PR: https://github.com/huggingface/huggingface_hub/pull/4477

Sreekant13:
Resolved by #4477. Tthe `--timeout` help now correctly advertises `int` (use e.g. `90m` instead of `1.5h`). Closing.

PR #4477, opened 2026-07-02
Correct hf jobs/spaces --timeout help: durations are int, not int/float

Sreekant13:
Following the review, instead of adding float support this just corrects the misleading help.

The `--timeout` option on `hf jobs run` / `hf jobs wait` / `hf spaces wait` (and the matching `HfApi.run_job` timeout docstrings) advertised `int/float with s/m/h/d`, but the string-with-unit form only accepts integers, so `--timeout 1.5h` raised `ValueError`. This drops `float` from the help so the advertised format matches what's accepted. Users can write `90m` instead of `1.5h`. Regenerated the CLI reference doc (package_reference/cli.md) to match.

Addresses #4476.


Wauplin:
Thanks for noticing! Preferable to just update the help text yes. User can do "90m" instead of "1.5h" to avoid decimal values.

Wauplin (approved):
thank you!

Wauplin:
@Sreekant13 can you update the PR title and description accordingly now? other than that we should be good to merge once CI is green :hugs:

Sreekant13:
Thank you so much @Wauplin! Switched to just fixing the help text as you suggested and dropped "float" from the `--timeout` help on jobs/spaces, updated the matching HfApi.run_job docstrings, and regenerated the CLI reference doc. "90m" it is 🙂

Will update the Title and Description now as well.

Wauplin:
Great, merged it!

Sreekant13:
Thanks a lot, @Wauplin!

PR #4498, opened 2026-07-07
[Cache] Add cross-repo shared blob store

hanouticelina:
## Summary

Adds a cache-wide shared blob store that deduplicates Xet files across cached repos. Identical files are stored once on disk and reused on download without re-fetching, across repos and even after the repo that first downloaded them was deleted. Enabled by default, opt out with `HF_HUB_DISABLE_SHARED_BLOBS=1`.

## How it works

- Payloads are stored once at `<HF_HUB_CACHE>/blobs/<2-hex-prefix>/<xet_hash>`, read-only on POSIX, keyed by the Xet file hash (known from the HEAD response and verified by `hf_xet`). Each per-repo `blobs/<etag>` is a relative symlink to the store entry; the snapshot layout is unchanged.
- On download, a store hit is symlinked into the repo with no transfer. Otherwise the file is downloaded through the verified Xet path, moved into the store and symlinked back. The plain HTTP fallback is never published. A store entry is only used if it is a regular file of the expected size that the current user can read; anything else is skipped with a warning and replaced by the next verified download.
- Each entry has a `<xet_hash>.refs` manifest listing the repo blobs referencing it, appended and fsynced before the symlink becomes visible. The manifest is a hint: lines are validated against the filesystem, and an entry is removed only when no valid reference remains. Missing or unreadable metadata means the payload is kept.
- `hf cache rm` checks only the manifests of the blobs it touches. `hf cache prune` additionally sweeps entries with no valid reference left, which reclaims payloads after `rm -rf <repo folder>` or an old-client deletion.
- A per-hash lock serializes linking, publication and garbage collection. A version marker identifies the store, so an existing unmarked `blobs/` directory is never adopted.
- Payloads are published with modes mirroring the cache root so multi-user caches work, and any failure to share silently falls back to repo-local storage. Per-repo sizes stay logical, the cache-wide `size_on_disk` is physical.
- Implementation: `huggingface_hub/utils/_shared_blobs.py`, used by `file_download.py`, `utils/_cache_manager.py` and `cli/cache.py`.

## Backward compatibility

- Reading and downloading are unchanged for old `huggingface_hub`, `huggingface.js`, `hf-hub`, `llama.cpp` and any tool that follows symlinks. Old clients keep downloading into the same repo layout.
- Cache deletion requires an up-to-date client: old deletion tools resolve the full symlink chain and can delete a canonical payload still referenced by other repos. Affected files are simply re-downloaded, and payloads left behind by manual or old-client deletion of whole repos are reclaimed by `hf cache prune`. Documented in the caching guide.
- Old scanners report the top-level `blobs/` directory as an unknown cache entry (warning only).
- Shared payloads are read-only once published, so editing cached files in place now fails with a permission error instead of silently affecting every repo sharing them.
- Without symlink support (Windows without Developer Mode, `HF_HUB_DISABLE_SYMLINKS=1`) the store is inactive and the existing degraded layout is used. The symlink-dependent tests are skipped on Windows CI for that reason.

## Test the PR

<details>
<summary>Transcript (made by Claude) </summary>

```console
$ export HF_HUB_CACHE=$(mktemp -d)/hub HF_HUB_DISABLE_PROGRESS_BARS=1

### 1. First download populates the store
$ hf download NousResearch/Llama-2-7b-hf tokenizer.model
path=$HF_HUB_CACHE/models--NousResearch--Llama-2-7b-hf/snapshots/8efe6c9b93655b934e27bd9981e3ec13e55aee9d/tokenizer.model

$ ls -l $HF_HUB_CACHE/models--NousResearch--Llama-2-7b-hf/blobs/ | tail -n +2 | awk '{print $1, $9, $10, $11}'
lrwxr-xr-x@ 9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347 -> ../../blobs/91/91bf184ab12793d0754344f9095332759432e666320cc6c07f637af50e36db6f

$ cat $HF_HUB_CACHE/blobs/91/*.refs
models--NousResearch--Llama-2-7b-hf/blobs/9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347

### 2. Same file from another repo: store hit, no transfer
$ HF_DEBUG=1 hf download NousResearch/Llama-2-7b-chat-hf tokenizer.model 2>&1 | grep -i 'shared blob'
Blob '$HF_HUB_CACHE/models--NousResearch--Llama-2-7b-chat-hf/blobs/9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347' reused from shared blob store (no download needed).

$ cat $HF_HUB_CACHE/blobs/91/*.refs
models--NousResearch--Llama-2-7b-hf/blobs/9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347
models--NousResearch--Llama-2-7b-chat-hf/blobs/9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347

$ hf cache ls
id	size	last_accessed	last_modified	refs
model/NousResearch/Llama-2-7b-chat-hf	499.7K	a few seconds ago	a few seconds ago	['main']
model/NousResearch/Llama-2-7b-hf	499.7K	a few seconds ago	a few seconds ago	['main']

Found 2 repo(s) for a total of 2 revision(s) and 499.7K on disk.

### 3. Deletion only frees the payload with its last reference
$ hf cache rm model/NousResearch/Llama-2-7b-hf -y
About to delete 1 repo(s) totalling 0.0.
  - model/NousResearch/Llama-2-7b-hf (entire repo)
Cache deletion done. Saved 0.0.
repos_deleted=1 revisions_deleted=1 freed=0.0

$ hf cache rm model/NousResearch/Llama-2-7b-chat-hf -y
About to delete 1 repo(s) totalling 499.7K.
  - model/NousResearch/Llama-2-7b-chat-hf (entire repo)
Cache deletion done. Saved 499.7K.
repos_deleted=1 revisions_deleted=1 freed=499.7K

$ find $HF_HUB_CACHE/blobs -type f -name '9*' | wc -l | tr -d ' '
0

### 4. Manual deletion of a repo folder leaves an orphan; prune reclaims it
$ hf download NousResearch/Llama-2-7b-hf tokenizer.model

$ rm -rf $HF_HUB_CACHE/models--NousResearch--Llama-2-7b-hf

$ hf cache ls
No results found.

$ hf cache prune --dry-run
About to delete 1 unreferenced shared blob(s) (499.7K total).
dry_run=True revisions=0 incomplete=0 shared_blobs=1 size=499.7K

$ hf cache prune -y
About to delete 1 unreferenced shared blob(s) (499.7K total).
Cache deletion done. Saved 0.0.
revisions_deleted=0 incomplete_deleted=0 shared_blobs_deleted=1 freed=499.7K

### 5. An unreadable store entry is not used, and a fresh download replaces it
$ hf download NousResearch/Llama-2-7b-hf tokenizer.model

$ chmod 000 $HF_HUB_CACHE/blobs/91/91*[0-9a-f]

$ hf download NousResearch/Llama-2-7b-chat-hf tokenizer.model
Shared blob '$HF_HUB_CACHE/blobs/91/91bf184ab12793d0754344f9095332759432e666320cc6c07f637af50e36db6f' is not readable. Not using it.
Shared blob '$HF_HUB_CACHE/blobs/91/91bf184ab12793d0754344f9095332759432e666320cc6c07f637af50e36db6f' is not readable. Not using it.

$ ls -l $HF_HUB_CACHE/blobs/91/ | tail -n +2 | awk '{print $1, $9}'
-r--r--r--@ 91bf184ab12793d0754344f9095332759432e666320cc6c07f637af50e36db6f
-rw-rw-rw-@ 91bf184ab12793d0754344f9095332759432e666320cc6c07f637af50e36db6f.refs

### 6. Opt-outs fall back to the plain repo layout
$ HF_HUB_DISABLE_SHARED_BLOBS=1 HF_HUB_CACHE=/tmp/plain-cache hf download NousResearch/Llama-2-7b-hf tokenizer.model >/dev/null; ls /tmp/plain-cache
CACHEDIR.TAG
models--NousResearch--Llama-2-7b-hf

$ HF_HUB_DISABLE_XET=1 HF_HUB_CACHE=/tmp/noxet-cache hf download NousResearch/Llama-2-7b-hf tokenizer.model >/dev/null; ls /tmp/noxet-cache
CACHEDIR.TAG
models--NousResearch--Llama-2-7b-hf

### 7. Old client reads the new layout (huggingface_hub 1.10.0)
$ uv venv /tmp/hfh-old -q --python 3.11 && uv pip install -q --index-url https://pypi.org/simple -p /tmp/hfh-old/bin/python 'huggingface_hub==1.10.0' typer click && /tmp/hfh-old/bin/hf version
1.10.0

$ /tmp/hfh-old/bin/python -c "from huggingface_hub import hf_hub_download; p = hf_hub_download('NousResearch/Llama-2-7b-chat-hf', 'tokenizer.model', cache_dir='$HF_HUB_CACHE'); print(len(open(p, 'rb').read()), 'bytes read through the symlink')"
499723 bytes read through the symlink

$ /tmp/hfh-old/bin/hf cache ls --cache-dir $HF_HUB_CACHE
ID                                    SIZE     LAST_ACCESSED     LAST_MODIFIED     REFS
------------------------------------- -------- ----------------- ----------------- ----
model/NousResearch/Llama-2-7b-chat-hf   499.7K a few seconds ago a few seconds ago main
model/NousResearch/Llama-2-7b-hf        499.7K a few seconds ago a few seconds ago main

Found 2 repo(s) for a total of 2 revision(s) and 999.4K on disk.

Error: No such option: --show-warnings

$ /tmp/hfh-old/bin/python -c "from huggingface_hub import scan_cache_dir; r = scan_cache_dir('$HF_HUB_CACHE'); print(sorted(x.repo_id for x in r.repos)); print([str(w) for w in r.warnings])"
['NousResearch/Llama-2-7b-chat-hf', 'NousResearch/Llama-2-7b-hf']
['Repo path is not a valid HuggingFace cache directory: $HF_HUB_CACHE/blobs']
```

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes core download and cache deletion semantics on disk; mistakes could corrupt caches or free space incorrectly, though manifests and conservative GC aim to avoid deleting still-referenced blobs.
> 
> **Overview**
> Introduces a **cache-wide shared blob store** so identical **Xet** files are stored once under `<cache>/blobs/<prefix>/<xet_hash>` and reused across repos via repo-local symlinks, with opt-out via **`HF_HUB_DISABLE_SHARED_BLOBS=1`**.
> 
> **Download path:** `hf_hub_download` tries to link from the store before downloading, publishes verified Xet payloads after download (not HTTP fallback), and adjusts symlink/`force_download` handling so replacements do not write through shared symlinks.
> 
> **Cache management:** `scan_cache_dir` reports a **physical** total size (deduplicated blobs, including orphans) while per-repo sizes stay logical; `hf cache rm` / deletion strategies sweep shared payloads using `.refs` manifests; **`hf cache prune`** also removes unreferenced shared blobs and incomplete downloads.
> 
> Documentation covers the layout, compatibility caveats for older clients, and the new env var.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4ce9bda1d9f92da849bcab5b9ce38dc637389af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanouticelina:
bugbot run

docs/source/en/guides/cli.md
```diff
@@ -1781,6 +1781,7 @@ Copy-and-paste the text below in your GitHub issue.
 - HF_HUB_DISABLE_EXPERIMENTAL_WARNING: False
 - HF_HUB_DISABLE_IMPLICIT_TOKEN: False
 - HF_HUB_DISABLE_XET: False
+- HF_HUB_ENABLE_SHARED_BLOBS: False
```
hanouticelina: not sure if we should have an env for it or just enable it by default?

src/huggingface_hub/utils/_cache_manager.py
```diff
@@ -22,6 +22,7 @@
 
 from huggingface_hub.errors import CacheNotFound, CorruptedCacheException
 
+from .. import _shared_blobs
```
hanouticelina: imported as a module to tolerate the circular import

src/huggingface_hub/utils/_cache_manager.py
```diff
@@ -22,6 +22,11 @@
 
 from huggingface_hub.errors import CacheNotFound, CorruptedCacheException
 
+# Imported as a module (not names): `_shared_blobs` is only partially initialized when
+# this import runs from a direct `import huggingface_hub._shared_blobs` (its own imports
+# trigger the `utils` package init, which loads this module). Attribute access at call
+# time is safe; importing names here would crash on the partial module.
+from .. import _shared_blobs
```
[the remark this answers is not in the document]
hanouticelina: (from claude) since Python 3.7, `from package import submodule` falls back to `sys.modules` when the parent attribute isn't set yet, which is exactly the circular-import case here

hanouticelina:
bugbot run

Wauplin:
Opened https://github.com/huggingface/huggingface_hub/pull/4600 on top with some heavy modifications, especially switching from hardlinks to soft symlinks. See rational (and my prompt) in the PR description. Let's discuss this further once we all back from summer break.

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,607 @@
...
+    except OSError as e:
+        logger.debug(f"Could not publish '{blob_path}' to shared blob store: {e}")
+        if blob_moved and not os.path.lexists(blob_path_obj):
+            try:
+                shutil.copyfile(store_path, blob_path_obj)
+            except OSError as restore_error:
+                raise OSError(
+                    f"Could not restore repo blob '{blob_path}' after shared-store failure"
+                ) from restore_error
+        return False
```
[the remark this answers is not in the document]
hanouticelina: this is very unlikely but since it's only a 5 line fix then let's fix it

hanouticelina:
Everything on the branch since the initial version, in order:

1. #4600: merged, already reviewed

2. `hf cache ls` footer reports the physical total when unfiltered: https://github.com/huggingface/huggingface_hub/pull/4498/commits/1c7de69d4370a3155ee784d7dae60cb53fdb733e

3. #4797: merged, already reviewed

4. Fixes from @hanouticelina's review: readable store entries, one marker read per scan, test cleanup, module moved to utils/, shorter comments, tighter guide section
   https://github.com/huggingface/huggingface_hub/compare/e68149e1...6bcd3506

5. Bugbot follow-ups: restore the repo blob under the lock, always keep the first symlink hop when scanning, dry run respects HF_HUB_DISABLE_SYMLINKS: https://github.com/huggingface/huggingface_hub/compare/0ba8f38a...d6439b54

The PR description reflects the current state.

docs/source/en/guides/manage-cache.md
```diff
@@ -204,6 +204,50 @@ When symlinks are not supported, a warning message is displayed to the user to a
...
+  files are re-downloaded on next use. Use an up-to-date `huggingface_hub` for
+  `hf cache rm`, `hf cache prune`, and programmatic deletion.
+- Their cache scanners may report the top-level `blobs` directory as an unknown entry.
+
+The store requires the symlink-based cache layout and is disabled by
+`HF_HUB_DISABLE_XET=1`. Any failure to share a file, such as an unsupported filesystem,
+a permission error, or a pre-existing unmarked `blobs` directory, silently falls back to
+regular repo-local storage. Set
+[`HF_HUB_DISABLE_SHARED_BLOBS=1`](../package_reference/environment_variables#hfhubdisablesharedblobs)
+to opt out entirely.
```
```suggestion
### Shared blobs across repos

By default, Xet files are also deduplicated across repos. A Xet file downloaded through `hf_xet` is stored once at `<CACHE_DIR>/blobs/<prefix>/<xet_hash>`, and the repo's `blobs/<etag>` entry is a relative symlink to it. When another repo needs the same file, it gets a symlink instead of a download: no bytes are transferred and no extra space is used. The snapshot layout does not change. A marker file identifies the store, so a `blobs` directory that was not created by `huggingface_hub` is never touched.

Each shared file has a `<xet_hash>.refs` manifest listing the repo blobs that use it. `hf cache rm` reads it to check only the shared files affected by a deletion, and `hf cache prune` removes shared files that no cached repo uses anymore, for example after a repo folder was deleted by hand or with an older client. The manifest is only a hint: every entry is checked against the filesystem, and a file with missing or unreadable metadata is kept. Cleanup prefers leaving reclaimable data behind over breaking a valid cache entry.

Per-repo sizes stay logical, so a shared file is counted for every repo that uses it. The cache-wide `size_on_disk` is physical: each shared file is counted once, including files no repo uses anymore.

On multi-user clusters, symlinks work across users, unlike hardlinks, which require permission to link another user's file. On POSIX systems, shared files are made read only and readable by every user who can traverse the cache root; manifests and locks stay writable. Set the intended group and the setgid bit on the cache root. Anyone who can write to a shared cache can affect all of its users, so do not put private repos in a cache exposed to untrusted users.

Older clients (`huggingface_hub`, `huggingface.js`, `hf-hub`, `llama.cpp`, and anything that follows symlinks) keep reading and downloading normally, in the same repo folders.

Two limitations:
- Their cache deletion tools may delete a shared file still used by other repos. Affected files are re-downloaded on next use. Use an up-to-date `huggingface_hub` for `hf cache rm`, `hf cache prune`, and programmatic deletion.
- Their cache scanners may report the top-level `blobs` directory as an unknown entry.

The store requires the symlink-based cache layout and is disabled by `HF_HUB_DISABLE_XET=1`. Any failure to share a file, such as an unsupported filesystem, a permission error, or a pre-existing unmarked `blobs` directory, silently falls back to regular repo-local storage. Set [`HF_HUB_DISABLE_SHARED_BLOBS=1`](../package_reference/environment_variables#hfhubdisablesharedblobs) to opt out entirely.
```

docs/source/en/guides/manage-cache.md
```diff
@@ -667,9 +711,9 @@ Dry run: no files were deleted.
 When working outside the default cache location, pair the command with
 `--cache-dir PATH`.
 
-To clean up cache garbage in bulk, run `hf cache prune`. It automatically deletes both
-revisions that are no longer referenced by a branch or tag and any leftover `.incomplete`
-files from interrupted downloads:
+To clean up cache garbage in bulk, run `hf cache prune`. It automatically deletes
+revisions that are no longer referenced by a branch or tag, leftover `.incomplete` files
+from interrupted downloads, and shared blobs that no cached repo references anymore:
```
```suggestion
To clean up cache garbage in bulk, run `hf cache prune`. It automatically deletes revisions that are no longer referenced by a branch or tag, leftover `.incomplete` files from interrupted downloads, and shared blobs that no cached repo references anymore:
```

docs/source/en/package_reference/environment_variables.md
```diff
@@ -200,6 +200,16 @@ Set to disable using `hf-xet`, even if it is available in your Python environmen
...
 
+### HF_HUB_DISABLE_SHARED_BLOBS
+
+Set to `1` to disable the cache-wide shared blob store. By default, successfully
+downloaded Xet files are stored in `<CACHE_DIR>/blobs/<prefix>/<xet_hash>` and symlinked
+into repo caches, so identical files are stored on disk only once across repos and can
+be reused without re-fetching. The optimization requires Xet
+(`HF_HUB_DISABLE_XET` must not be set) and the symlink-based cache layout; it falls back
+silently to regular repo-local storage otherwise. See the
+[caching guide](../guides/manage-cache#shared-blobs-across-repos) for details.
```
Wauplin:
```suggestion
Set to `1` to disable the cache-wide shared blob store. By default, successfully downloaded Xet files are stored in `<CACHE_DIR>/blobs/<prefix>/<xet_hash>` and symlinked into repo caches, so identical files are stored on disk only once across repos and can be reused without re-fetching. The optimization requires Xet (`HF_HUB_DISABLE_XET` must not be set) and the symlink-based cache layout; it falls back silently to regular repo-local storage otherwise. See the [caching guide](../guides/manage-cache#shared-blobs-across-repos) for details.
```

docs/source/en/guides/manage-cache.md
```diff
@@ -204,6 +204,24 @@ When symlinks are not supported, a warning message is displayed to the user to a
...
 by setting the `HF_HUB_DISABLE_SYMLINKS_WARNING` environment variable to true.
 
+### Shared blobs across repos
+
+By default, Xet files are also deduplicated across repos. A Xet file downloaded through `hf_xet` is stored once at `<CACHE_DIR>/blobs/<prefix>/<xet_hash>`, and the repo's `blobs/<etag>` entry is a relative symlink to it. When another repo needs the same file, it gets a symlink instead of a download: no bytes are transferred and no extra space is used. The snapshot layout does not change. A marker file identifies the store, so a `blobs` directory that was not created by `huggingface_hub` is never touched.
+
+Each shared file has a `<xet_hash>.refs` manifest listing the repo blobs that use it. `hf cache rm` reads it to check only the shared files affected by a deletion, and `hf cache prune` removes shared files that no cached repo uses anymore, for example after a repo folder was deleted by hand or with an older client. The manifest is only a hint: every entry is checked against the filesystem, and a file with missing or unreadable metadata is kept. Cleanup prefers leaving reclaimable data behind over breaking a valid cache entry.
```
```suggestion
Each shared file has a `<xet_hash>.refs` manifest listing the repo blobs that use it. `hf cache rm` reads it to check only the shared files affected by a deletion, and `hf cache prune` removes shared files that no cached repo uses anymore. The manifest is only a hint: every entry is checked against the filesystem, and a file with missing or unreadable metadata is kept. Cleanup prefers leaving reclaimable data behind over breaking a valid cache entry.
```

docs/source/en/guides/manage-cache.md
```diff
@@ -204,6 +204,24 @@ When symlinks are not supported, a warning message is displayed to the user to a
...
+
+Per-repo sizes stay logical, so a shared file is counted for every repo that uses it. The cache-wide `size_on_disk` is physical: each shared file is counted once, including files no repo uses anymore.
+
+On multi-user clusters, symlinks work across users, unlike hardlinks, which require permission to link another user's file. On POSIX systems, shared files are made read only and readable by every user who can traverse the cache root; manifests and locks stay writable. Set the intended group and the setgid bit on the cache root. Anyone who can write to a shared cache can affect all of its users, so do not put private repos in a cache exposed to untrusted users.
+
```
```suggestion
```

I feel this paragraph is not needed

docs/source/en/guides/manage-cache.md
```diff
@@ -204,6 +204,24 @@ When symlinks are not supported, a warning message is displayed to the user to a
...
+Older clients (`huggingface_hub`, `huggingface.js`, `hf-hub`, `llama.cpp`, and anything that follows symlinks) keep reading and downloading normally, in the same repo folders.
+
+Two limitations:
+- Their cache deletion tools may delete a shared file still used by other repos. Affected files are re-downloaded on next use. Use an up-to-date `huggingface_hub` for `hf cache rm`, `hf cache prune`, and programmatic deletion.
+- Their cache scanners may report the top-level `blobs` directory as an unknown entry.
```
```suggestion
Older clients (`huggingface_hub`, `huggingface.js`, `hf-hub`, `llama.cpp`, and anything that follows symlinks) keep reading and downloading normally, in the same repo folders. Two limitations:
- Their cache deletion tools may delete a shared file still used by other repos. Affected files are re-downloaded on next use. Use an up-to-date `huggingface_hub` for `hf cache rm`, `hf cache prune`, and programmatic deletion.
- Their cache scanners may report the top-level `blobs` directory as an unknown entry.
```

docs/source/en/package_reference/environment_variables.md
```diff
@@ -200,6 +200,10 @@ Set to disable using `hf-xet`, even if it is available in your Python environmen
 > Now that the Hugging Face Hub is fully powered by the Xet storage backend, all file transfers go through the `hf-xet` binary package. It provides efficient transfers using a chunk-based deduplication strategy and integrates seamlessly with `huggingface_hub`.
 > This means `hf_transfer` can't be used anymore. If you are interested in higher performance, check out the [`HF_XET_HIGH_PERFORMANCE` section](#hf_xet_high_performance)
 
+### HF_HUB_DISABLE_SHARED_BLOBS
+
+Set to `1` to disable the cache-wide shared blob store. By default, successfully downloaded Xet files are stored in `<CACHE_DIR>/blobs/<prefix>/<xet_hash>` and symlinked into repo caches, so identical files are stored on disk only once across repos and can be reused without re-fetching. The optimization requires Xet (`HF_HUB_DISABLE_XET` must not be set) and the symlink-based cache layout; it falls back silently to regular repo-local storage otherwise. See the [caching guide](../guides/manage-cache#shared-blobs-across-repos) for details.
```
Wauplin:
```suggestion
Set to `1` to disable the cache-wide shared blob store. By default, successfully downloaded Xet files are stored in `<CACHE_DIR>/blobs/<prefix>/<xet_hash>` and symlinked into repo caches, so identical files are stored on disk only once across repos and can be reused without re-fetching. This optimization is enabled by default, but is disabled if Xet is unavailable (e.g. if `HF_HUB_DISABLE_XET=1` is set) or if symlinks are not supported locally. In those cases it silently falls back to regular repo-local storage. See the [caching guide](../guides/manage-cache#shared-blobs-across-repos) for details.
```

src/huggingface_hub/utils/_cache_manager.py
```diff
@@ -947,7 +1034,7 @@ def _try_delete_path(path: Path, path_type: str) -> None:
     """
     logger.debug(f"Delete {path_type}: {path}")
     try:
-        if path.is_file():
+        if path.is_file() or path.is_symlink():
```
Wauplin: AFAIU, the `path.is_symlink` is useful when a path is a symlink but not a symlink to a file, i.e. when it's a symlink to a directory => we never have symlinks to directories, right? so this is not necessary ?

Wauplin: or is it for the case where the symlink is broken?

tests/test_utils_cache.py
```diff
@@ -558,7 +558,8 @@ def test_delete_all_revisions(self) -> None:
             self.cache_info, "repo_A_rev_detached", "repo_A_rev_pr_1", "repo_A_rev_main"
         )
         expected = DeleteCacheStrategy(
-            expected_freed_size=4444,
+            # Sum of the unique blob sizes: the freed size is computed per blob.
+            expected_freed_size=11111,
```
Wauplin, line 562: but then it means this value is overstated ?

src/huggingface_hub/file_download.py
```diff
@@ -1240,21 +1256,48 @@ def _hf_hub_download_to_cache_dir(
 
     # Local file doesn't exist or etag isn't a match => retrieve file from remote (or cache)
 
+    # Cross-repo dedup needs the symlink layout: `xet_hash` stays set only when the shared store
+    # can be used for this download (see `utils/_shared_blobs.py`).
+    xet_hash = xet_file_data.file_hash if xet_file_data is not None else None
+    if xet_hash is not None and not (shared_blobs_enabled() and are_symlinks_supported(cache_dir)):
+        xet_hash = None
```
Wauplin:
```suggestion
    xet_hash = xet_file_data.file_hash if xet_file_data is not None else None
    use_shared_blobs = xet_hash is not None and shared_blobs_enabled() and are_symlinks_supported(cache_dir)
```

I feel that using the `xet_hash` variable for the xet hash value and as a "are we using the shared blob" is misleading when reasoning about the code (the comment explains it here but below it's easy to forget about it). More explicit to have 2 different variables IMO

src/huggingface_hub/file_download.py
```diff
@@ -1240,21 +1256,48 @@ def _hf_hub_download_to_cache_dir(
...
-            url_to_download=url_to_download,
-            headers=headers,
-            expected_size=expected_size,
-            filename=filename,
-            force_download=force_download,
-            etag=etag,
-            xet_file_data=xet_file_data,
-            tqdm_class=tqdm_class,
+        blob_reused_from_store = (
+            xet_hash is not None
```
Wauplin:
```suggestion
            use_shared_blobs
```

(see above)

src/huggingface_hub/file_download.py
```diff
@@ -1240,21 +1256,48 @@ def _hf_hub_download_to_cache_dir(
...
+                url_to_download=url_to_download,
+                headers=headers,
+                expected_size=expected_size,
+                filename=filename,
+                force_download=force_download,
+                etag=etag,
+                xet_file_data=xet_file_data,
+                tqdm_class=tqdm_class,
+            )
+            if xet_hash is not None and will_download and is_xet_available():
```
Wauplin:
```suggestion
            if use_shared_blobs and will_download:
```

src/huggingface_hub/file_download.py
Wauplin: (nit) or maybe `should_use_shared_blobs` is even more explicit?

src/huggingface_hub/file_download.py
```diff
@@ -2034,7 +2077,43 @@ def _chmod_and_move(src: Path, dst: Path) -> None:
...
+                    raise move_error
+                try:
+                    backup_dst.unlink()
+                    backup_holds_previous_entry = False
+                except OSError as cleanup_error:
+                    logger.warning(f"Could not remove previous destination backup '{backup_dst}': {cleanup_error}")
+            finally:
+                staged_dst.unlink(missing_ok=True)
+                if not backup_holds_previous_entry:
+                    backup_dst.unlink(missing_ok=True)
```
Wauplin: I feel this logic could be moved to a `_replace_no_matter_what` helper method (same as current `_copy_no_matter_what`)

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+    return shared_blobs_dir(cache_dir) / xet_hash[:2] / xet_hash
+
+
+def shared_blob_manifest_path(cache_dir: str | Path, xet_hash: str) -> Path:
+    """Return the reference manifest path for a given Xet file hash."""
+    return shared_blob_path(cache_dir, xet_hash).with_name(f"{xet_hash}{_MANIFEST_SUFFIX}")
+
+
+def _is_regular_file(path: Path) -> bool:
+    try:
```
Wauplin, line 77: adding a comment for future us with my understanding of why this alias is needed

```suggestion
    # checks if path is a regular file without following symlinks
    try:
```

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+        return stat.S_ISREG(path.lstat().st_mode)
+    except OSError:
+        return False
+
+
+def _is_directory(path: Path) -> bool:
+    try:
```
Wauplin, line 85:
```suggestion
    # checks if path is a regular directory without following symlinks
    try:
```

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+    if not _is_directory(store_dir) or not _is_regular_file(marker_path):
+        return False
+    try:
+        return marker_path.read_text() == f"{SHARED_BLOBS_LAYOUT_VERSION}\n"
+    except OSError:
+        return False
+
+
+def _cleanup_abandoned_marker_temps(store_dir: Path) -> bool:
+    """Remove owned marker temporaries, refusing unexpected directory content."""
```
Wauplin: let's document what the return value refers to (is it just whether the directory is empty?)

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+    try:
+        store_dir.mkdir(exist_ok=True)
+        if is_shared_blobs_dir(store_dir):
+            return True
+        if not _is_directory(store_dir) or not _cleanup_abandoned_marker_temps(store_dir):
+            logger.debug(f"Refusing to use unmarked shared blob directory '{store_dir}'.")
+            return False
+        _repair_shared_directory_mode(store_dir, cache_dir)
+
+        tmp_marker = marker_path.with_name(f"{marker_path.name}.{uuid.uuid4().hex[:8]}.tmp")
```
Wauplin:
```suggestion
        tmp_marker = marker_path.with_name(f"{marker_path.name}.{secrets.token_hex(4)}.tmp")
```

(nit) feels more natural to general a random 8-characters hexa string

requires a `import secrets` above

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+    return store_path.with_name(f"{store_path.name}{_MANIFEST_SUFFIX}")
+
+
+def _lock_path_for_store_path(store_path: Path) -> Path:
+    return store_path.with_name(f"{store_path.name}{_LOCK_SUFFIX}")
+
+
+@contextmanager
+def _shared_blob_lock(store_path: Path) -> Generator[None, None, None]:
+    """Lock publication, reference creation, and GC for one content hash."""
```
Wauplin, line 318: would be great to have a 1-line explainer on why not using filelock/softfilelock dircetly

src/huggingface_hub/utils/_shared_blobs.py
```diff
@@ -0,0 +1,615 @@
...
+        line = f"{relative_blob_path}\n".encode()
+        if os.write(fd, line) != len(line):
+            raise OSError(f"Could not append a complete reference to '{manifest_path}'.")
+        os.fsync(fd)
+    finally:
+        os.close(fd)
+
+
+def _make_temporary_symlink(blob_path: Path, store_path: Path) -> Path:
+    tmp_link = blob_path.with_name(f".{blob_path.name}.{uuid.uuid4().hex[:8]}.shared")
```
Wauplin:
```suggestion
    tmp_link = blob_path.with_name(f".{blob_path.name}.{secrets.token_hex(4)}.shared")
```

Wauplin (commented):
Looks good! Lots of comments but mostly cosmetic. Otherwise works fine in my local experiments 🙌 

(note: I haven't properly reviewed the tests but they seem to unit test most methods so I assumed they are good)

src/huggingface_hub/utils/_cache_manager.py
hanouticelina: yes, broken symlinks. `is_file()` follows the link and returns `False` when the target is gone, so a repo blob whose shared payload was deleted (e.g. by an old client) would fall into `rmtree` and fail. i added a comment in e4ce9bda.

tests/test_utils_cache.py
hanouticelina: not overstated, - 11111 is the correct sum of the five unique blobs in that mock (1 + 10 + 100 + 1000 + 10000). the previous 4444 was an arbitrary `repo_A.size_on_disk` mock value

src/huggingface_hub/file_download.py
hanouticelina: went with a single `shared_blob_hash` (the hash, or `None` when the store must not be used) instead of a separate boolean: mypy does not narrow `xet_hash` through an aliased flag, so `use_shared_blobs` would have needed a redundant `is not None` at each call site. done in e4ce9bda

hanouticelina:
thanks @Wauplin for the review! every comment is addressed in  869da12 and e4ce9bd

Wauplin (approved):
Let's go! ✔️

Wauplin:
:fire: :fire: :fire:

PR #4502, opened 2026-07-08
Add sort option to hf jobs ls

### LINKED ISSUE #4374: Add --sort and --limit flags to hf jobs ps

`hf jobs ps -a` returns the full list of jobs in API/creation order with no way to reorder or trim the output, so once you have more than ~20 finished jobs the most recent ones (or the longest-running ones) get buried in the middle of the table and the command becomes hard to scan during cleanup or post-mortem review.

Add two flags to `hf jobs ps` mirroring what `hf models ls` and `hf datasets ls` already expose:

- `--sort <field>` to sort by `created`, `status`, `runtime`, or `id` (default `created`, with `--reverse` / `-r` to flip the order).
- `--limit <N>` to cap the number of rows shown (default `50`, with `--limit 0` to list all, matching the convention used by `hf repos ls`).

Sorting can be done client-side over the existing `JobInfo` payload (it already carries `created_at`, `status.stage`, and `durations.running_secs`), so no API changes are required.

- Pipe through `jq` / `sort` manually — works but breaks `--format json` and is annoying for one-off inspections.
- Add a server-side `?sort=` query param to the Jobs API — strictly better but a much larger change for what is currently a CLI ergonomics gap.
- Just always default-sort by `created desc` — would help most cases but still leaves no way to sort by runtime or limit output.

**Additional context**

Repro:

```
hf jobs ps -a
# jobs returned in unsorted order, no flag to change it
```

Currently the only flags are `-a/--all`, `-f/--filter`, and the global `--format` / `--quiet`. The function is `jobs_ps` in `src/huggingface_hub/cli/jobs.py`, which already builds `items` with `created`, `status`, and `runtime` columns — the data is there, just no knob to sort/limit it.

lxingy3:
Summary
- add `--sort` to `hf jobs ls/ps` for created time, status, runtime, and id
- add `--reverse` to flip the selected sort order
- keep the existing fast limit path when no sort is requested
- document the new sorting options in the CLI guide

Tests
- `python -m ruff format --check src\huggingface_hub\cli\jobs.py tests\test_cli.py`
- `python -m ruff check src\huggingface_hub\cli\jobs.py tests\test_cli.py`
- `python -m pytest tests\test_cli.py::TestJobsCommand -q`

Closes #4374

---

> [!NOTE]
> **Low Risk**
> Localized CLI listing behavior with a preserved fast path when sort is omitted; covered by new unit tests.
> 
> **Overview**
> **`hf jobs ls` / `ps`** can now order results with **`--sort`** (`created`, `status`, `runtime`, or `id`) and flip direction with **`--reverse`**. For `created` and `runtime`, the default order is newest / longest first unless reversed.
> 
> When **`--sort`** is set, the command loads the matching jobs, sorts client-side, then applies **`--limit`** so the top N rows are the right slice (e.g. longest-running completed jobs). Without sort, behavior is unchanged: streaming **`--limit`** without pulling the full list. **`--reverse`** alone errors with a clear **`CLIError`**.
> 
> The CLI guide documents the new flags; tests cover sort-before-limit, reverse, and the reverse-without-sort guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a98e2bde8af42e8cbfb8e06255654f0ef2b2a7ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -2045,6 +2045,16 @@ By default `hf jobs ps` displays at most 100 Jobs to avoid bloating the terminal
 >>> hf jobs ps -a --limit 0
 ```
 
+Use `--sort` to order Jobs by `created`, `status`, `runtime`, or `id`. Created time and runtime sort newest or longest first by default; pass `--reverse` to flip the order:
+
+```bash
+# Show the longest-running completed Jobs
+>>> hf jobs ps -a --status completed --sort runtime --limit 10
+
+# Show Jobs by ID in descending order
+>>> hf jobs ps -a --sort id --reverse
+```
+
 > [!WARNING]
 > `-f`/`--filter` is deprecated in favor of `--status` and `--label`. Matching is exact: glob patterns (`data-*`) and negation (`key!=value`) are not supported, and filtering by `id`, `image` or `command` is not available.
 
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -19,6 +19,7 @@ import multiprocessing.pool
 import shutil
 import time
 from collections.abc import Callable, Iterable
+from enum import Enum
 from fnmatch import fnmatch
 from pathlib import Path
 from queue import Empty, Queue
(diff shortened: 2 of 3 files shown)


Wauplin:
Hi @lxingy3 thanks for your PR. However I will not merge it as the sort is only happening client-side. Since server doesn't return all Jobs (it has a limit), the sorting doesn't make sense. For instance if someone does 

```
hf jobs ls --sort runtime --reverse --limit 10
```

they will expect to have the top 10 Job by runtime. But they will only get the top 10 Jobs, then sorted by runtime. This cannot be solved client-side. As long as we don't have a hard requirement on that, we wont implement sorting.

PR #4526, opened 2026-07-14
[Repocard] Fix catastrophic backtracking (ReDoS) in REGEX_YAML_BLOCK

sohumt123:
## Summary

`REGEX_YAML_BLOCK` in `repocard.py` is vulnerable to catastrophic backtracking (ReDoS). A card whose content is a `---` opener followed by a long run of newlines and **no** closing fence makes `re.search()` run for seconds-to-minutes on a few KB of input, so parsing untrusted card content can pin a CPU core.

I hit this while feeding community-authored README frontmatter through `ModelCard`, and traced it to the regex.

### Root cause

```python
REGEX_YAML_BLOCK = re.compile(r"^(\s*---[\r\n]+)([\S\s]*?)([\r\n]+---(\r\n|\n|$))")
```

Both delimiters use unbounded `[\r\n]+` newline runs, and the body `[\S\s]*?` between them is lazy. When the input is a valid `---` opener followed by many newlines with no valid closing fence, the engine can split any long newline run three ways — between the opener tail (group 1), the lazy body (group 2), and the closer head (group 3). Every failed attempt at a closing `---` forces it to redistribute the split and retry, giving quadratic/cubic blowup. It fails *slowly* instead of failing fast.

Measured on the current pattern (LF opener + N newlines, no closing fence):

| N newlines | `re.search` time |
| ---------: | ---------------: |
| 500        | ~48 ms           |
| 1000       | ~338 ms          |
| 1500       | ~1.15 s          |
| 2000       | ~2.77 s          |
| 2500       | ~5.5 s           |

Doubling the input roughly 8x's the time. A ~6 KB CRLF payload runs for well over a minute.

Minimal repro (< 30s):

```python
import re, time
rx = re.compile(r"^(\s*---[\r\n]+)([\S\s]*?)([\r\n]+---(\r\n|\n|$))")
payload = "---\n" + "\n" * 3000  # ~3 KB, valid opener, no closing fence
t = time.perf_counter()
rx.search(payload)               # returns None ... eventually
print(round(time.perf_counter() - t, 1), "s")   # ~13 s on my machine
```

This constant is reached from public entry points that parse untrusted card content:

- `RepoCard(content)` / `ModelCard(content)` / `DatasetCard(content)` (content setter)
- `RepoCard.load(repo_id)` (downloads a README then parses it)
- `metadata_load(...)`
- `metadata_update(...)` → `metadata_save(...)`

`ignore_metadata_errors=True` does not help, because the time is spent inside `re.search` before any YAML parsing runs.

## Fix

Reduce both delimiter newline runs to a single token, removing the three-way ambiguity that drives the backtracking:

```python
REGEX_YAML_BLOCK = re.compile(r"^(\s*---[\r\n])([\S\s]*?)([\r\n]---[ \t]*(\r\n|\n|$))")
```

- Opener `[\r\n]+` → `[\r\n]` (single newline).
- Closer `[\r\n]+---` → `[\r\n]---` (single newline).
- `[ \t]*` after the closing `---` tolerates trailing spaces/tabs after the fence.

I verified this is behavior-preserving for everything observable through the public API. Across LF, CRLF, blank-line-padded, leading-whitespace and model-index cards:

- the overall match boolean is unchanged,
- `match.end()` is byte-identical (so `self.text = content[match.end():]` and all linebreak-preservation behavior are unchanged), and
- `yaml.safe_load(match.group(2))` yields identical dicts.

The only difference is internal: a newline that used to land in group 1 / group 3 now lands in the body group 2, and no caller inspects those raw groups. A possessive-only tweak (`[\r\n]++`) is *not* enough here — the interior lazy body still backtracks quadratically — so both delimiter runs have to be reduced to single tokens.

### Note on server-side sync

The comment above this constant says it must stay identical to the Hub server (`moon-landing` `ViewMarkdown.ts#L18`). I updated the comment to explain the ReDoS rationale and flag that the server-side pattern should be mirrored, otherwise the server stays vulnerable and the two diverge. The change only *tightens* delimiter matching (it accepts the same frontmatter), so mirroring it server-side is safe.

## Testing

Added `TestRegexYamlBlock.test_no_catastrophic_backtracking` (parametrized over LF, LF-with-title-and-body, and CRLF payloads of 200,000 newlines with no closing fence). Each asserts `REGEX_YAML_BLOCK.search(payload) is None` **and** that the search completes in under 1 second.

- On the old regex the test fails hard (even a bounded n=2500 payload takes ~5.5s, far over the 1s bound; the 200k payloads run for minutes).
- On the fixed regex all three run in ~3-6 ms.

```
$ pytest tests/test_repocard.py::TestRegexYamlBlock -q
6 passed in 0.05s
```

The three existing `TestRegexYamlBlock` correctness tests are unchanged and still pass, and I confirmed the rest of `tests/test_repocard.py` is green (57 passed), including CRLF/linebreak preservation, model-index parsing, and the `metadata_load` / `metadata_update` round-trips.

`make style` and `make quality` both pass.

---

> [!NOTE]
> **Medium Risk**
> Touches shared frontmatter parsing used whenever cards or metadata are loaded; the regex is intentionally tightened but behavior on edge-case line endings was wrong before and is now explicitly tested.
> 
> **Overview**
> Fixes **catastrophic backtracking (ReDoS)** in `REGEX_YAML_BLOCK`, which parses YAML frontmatter for repo cards. Untrusted README content with a `---` opener and a long newline run but **no** closing fence could make `re.search()` run for seconds or minutes on paths like `RepoCard` / `ModelCard` / `DatasetCard`, `metadata_load`, and `metadata_save`.
> 
> The pattern now uses a **single** newline after the opening `---` and before the closing `---` (with optional `[ \t]*` after the closing fence), instead of unbounded `[\r\n]+` on both sides. That removes the three-way split that caused exponential backtracking while keeping observable parsing behavior for normal LF/CRLF cards.
> 
> Tests add CRLF correctness cases (including a regression for mis-parsed double-fence cards) and **`test_no_catastrophic_backtracking`**, which asserts large no-closing-fence payloads finish in under one second.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 305a81734dc7dee9c95424de1da593034cefbaf4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/repocard.py
```diff
@@ -29,9 +29,16 @@
...
+# Based on the same regex as in the Hub server (moon-landing ViewMarkdown.ts#L18), with a
+# fix for catastrophic backtracking (ReDoS). The original `[\r\n]+` runs at both delimiters
+# let a long newline run be ambiguously redistributed between the opener tail, the lazy body
+# and the closer head, causing quadratic/cubic backtracking on inputs like a `---` opener
+# followed by many newlines and no closing fence. Each delimiter now consumes exactly ONE
+# complete line ending via a bounded non-capturing alternation ordered `\r\n|\r|\n` (the
+# two-char CRLF sequence is tried first so a CRLF opener cannot strand a `\n` that fakes an
+# early closer). No unbounded repetition remains at either delimiter, so the ReDoS fix is
+# preserved. The server-side regex should be mirrored.
+REGEX_YAML_BLOCK = re.compile(r"^(\s*---(?:\r\n|\r|\n))([\S\s]*?)((?:\r\n|\r|\n)---[ \t]*(\r\n|\n|$))")
```
Wauplin, line 34:
```suggestion
# exact same regex as in the Hub server. Please keep in sync.
# See https://github.com/huggingface/moon-landing/blob/main/server/lib/ViewMarkdown.ts#L18
REGEX_YAML_BLOCK = re.compile(r"^(\s*---(?:\r\n|\r|\n))([\S\s]*?)((?:\r\n|\r|\n)---[ \t]*(\r\n|\n|$))")
```

Wauplin (approved):
Thanks for the report and fix @sohumt123 ! Tested that myself and confirmed the issue. We've also fixed the regex server-side to avoid ReDos. I'll merge as soon as the CI is complete :)

src/huggingface_hub/repocard.py
```diff
@@ -29,9 +29,18 @@
...
 
+# Based on the same regex as in the Hub server (moon-landing ViewMarkdown.ts#L18), with a
+# fix for catastrophic backtracking (ReDoS). The original `[\r\n]+` runs at both delimiters
+# let a long newline run be ambiguously redistributed between the opener tail, the lazy body
+# and the closer head, causing quadratic/cubic backtracking on inputs like a `---` opener
+# followed by many newlines and no closing fence. Each delimiter now consumes exactly ONE
+# complete line ending via a bounded non-capturing alternation ordered `\r\n|\r|\n` (the
+# two-char CRLF sequence is tried first so a CRLF opener cannot strand a `\n` that fakes an
+# early closer). No unbounded repetition remains at either delimiter, so the ReDoS fix is
+# preserved. The server-side regex should be mirrored.
```
Wauplin:
```suggestion
```

Wauplin:

PR #4540, opened 2026-07-17
[Download] Reject absolute/UNC/traversal filenames on local_dir & cache paths

Wauplin:
## Summary

Client-side hardening of the filename → local path mapping used when downloading/uploading to a `local_dir` or to the cache.

- The `--local-dir` paths in `_local_folder.py` only had a Windows-only `..\` string check and **no containment**. Absolute (`C:\...`), drive-relative (`D:foo`), root-relative (`\foo`) and UNC (`\\host\share\...`) filenames are not `..\`, so they passed the check — and `local_dir / filename` discards `local_dir` once the right-hand side is anchored → arbitrary file write on Windows clients pulling a malicious repo (no MITM needed). The UNC form additionally makes the Windows client authenticate to the attacker's SMB server during path resolution, leaking a NetNTLMv2 hash.
- The **cache** path was already protected by the `_get_pointer_path` containment check; the `_local_folder.py` refactor never got an equivalent guard. So `--local-dir` was strictly weaker than the cache path.

Fix: one shared `_validate_relative_filename` helper, applied to **both** the local_dir (download + upload) and the cache download paths. It rejects `..` traversal and anchored paths (absolute / drive / root / UNC), interpreting the name under **both** POSIX and Windows rules on **all** platforms — so a file materialized on Linux can't escape when later consumed on Windows. The existing `_get_pointer_path` containment stays as defense-in-depth.

Refs CVE-2026-15717.

## Behavior change

The old check was Windows-only and only caught `..\`. The new check runs everywhere and is stricter. In particular, a filename like `folder/..\..\..\file` — previously downloadable on Linux (backslashes are literal chars there) — is now rejected on all platforms. Legitimate repo filenames never contain a `..` path segment or a drive/UNC anchor, so real downloads are unaffected. The two Linux-only regression tests that asserted this weird name downloads fine were updated to assert rejection.

## Example

```python
>>> from huggingface_hub._local_folder import get_local_download_paths
>>> from pathlib import Path
>>> get_local_download_paths(Path("/home/me/models"), "C:\\Windows\\System32\\evil")
ValueError: Invalid filename 'C:\Windows\System32\evil': cannot be an absolute, drive-relative or UNC path. ...
>>> get_local_download_paths(Path("/home/me/models"), "\\\\attacker\\share\\evil")
ValueError: Invalid filename '\\attacker\share\evil': cannot be an absolute, drive-relative or UNC path. ...
>>> get_local_download_paths(Path("/home/me/models"), "path/in/repo.txt").file_path
PosixPath('/home/me/models/path/in/repo.txt')   # normal files unaffected
```

---

> [!NOTE]
> **High Risk**
> Security fix for arbitrary file write and UNC/SMB credential leak on malicious repo filenames; behavior is stricter on all platforms with minimal impact on legitimate repo paths.
> 
> **Overview**
> **Hardens Hub download/upload path mapping** against malicious repo filenames that could write outside the intended directory (CVE-2026-15717 follow-up).
> 
> Adds shared **`_validate_relative_filename`**, applied before any filesystem join in **`get_local_download_paths`**, **`get_local_upload_paths`**, and **cache downloads** via `_hf_hub_download_to_cache_dir`. It rejects `..` segments and anchored paths (POSIX absolute, Windows drive/root/UNC, including nested segments), evaluated under **both POSIX and Windows rules on all platforms**. The old **Windows-only** `..\` string checks are removed in favor of this unified guard; existing **`_get_pointer_path`** containment remains for the cache.
> 
> **Behavior change:** names like `folder/..\..\..\file` that could previously download on Linux are now rejected everywhere. Regression tests cover unsafe/safe filename lists and expect **`ValueError`** from `hf_hub_download` for traversal attempts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 044f6672d83dbbc658f17b52a486678df52dcfc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Wauplin:
(unrelated CI failures)

PR #4546, opened 2026-07-18
Fix Windows crash when downloading into a deep local_dir

askalf:
On Windows the default is LongPathsEnabled=0, downloading into a deep local directory crash while creating the. cache/huggingface directory. That guard prefixes paths above 255 chars, but directory creation already fails at ≥248
(MAX_PATH minus 12 for legacy 8.3 names), so the 248-255 window slips through and throws.

The repro is local_dir whose absolute path lands in that 248-255 window + an ordinary file name -> the crash. The fix
is to lower the guard threshold to >247 so \\?\ extended-length prefix applies across the whole failing range.

Testing is a parametrized regression test at [252, 300] chars, guarded to windows, then executing in your windows-latest CI matrix. AI assisted me with this PR, I've reviewed the change and understand every line.

---

> [!NOTE]
> **Low Risk**
> Scoped to Windows filesystem path handling in local-dir cache setup; behavior elsewhere is unchanged and covered by a targeted regression test.
> 
> **Overview**
> Fixes **WinError 206** when downloading into a very deep `local_dir` on Windows without long-path support enabled.
> 
> **`_huggingface_dir`** now applies the `\\?\` extended-length prefix for `mkdir`, `CACHEDIR.TAG`, and `.gitignore` when the absolute `.cache/huggingface` path exceeds **247** characters (directory limit before the existing **255**-char guard on download/upload paths). It still returns the normal unprefixed path so downstream path logic is unchanged.
> 
> Adds a **Windows-only** parametrized test (`252` and `300` char cache paths) that exercises `get_local_download_paths` with a long `local_dir`, not just a long filename.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02efc9720a8fcf67b4414bc6f1161291ffe581cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_local_folder.py
+++ b/src/huggingface_hub/_local_folder.py
@@ -431,15 +431,25 @@ def _huggingface_dir(local_dir: Path) -> Path:
     """Return the path to the `.cache/huggingface` directory in a local directory."""
     # Wrap in lru_cache to avoid overwriting the .gitignore file if called multiple times
     path = local_dir / ".cache" / "huggingface"
-    path.mkdir(exist_ok=True, parents=True)
+    # Some Windows versions do not allow for paths longer than 255 characters, so creating the
+    # `.cache/huggingface` directory (and the bookkeeping files below) fails for a deep `local_dir`.
+    # Use the extended-length "\\?\" prefix for the filesystem operations, matching what
+    # `get_local_download_paths`/`get_local_upload_paths` already do for the download/upload paths.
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Looks good! Thanks for reporting and fixing it :)

Wauplin:

PR #4547, opened 2026-07-18
fix: bound safetensors metadata Range requests with a timeout

### LINKED ISSUE #4377: get_safetensors_metadata can hang indefinitely: no timeout on shard Range requests

sergioperezcheco:
Fixes #4377.

HfApi.get_safetensors_metadata and HfApi.parse_safetensors_file_metadata fetched each shard header via a Range GET with no timeout, and fanned the requests out via thread_map with no overall deadline. A stalled or half-open connection to the CDN (one that accepts the request then never delivers the body and never closes) would block the worker thread indefinitely, hanging the whole call and any caller for the life of the process.

Both methods now accept a timeout parameter (default None, which falls back to constants.HF_HUB_DOWNLOAD_TIMEOUT, consistent with the *_info methods, get_hf_file_metadata, and file_download). The value is forwarded to every Range GET in parse_safetensors_file_metadata, and get_safetensors_metadata threads it through to the per-file parse call.

Added TestSafetensorsMetadataTimeout with three offline, mocked tests covering the default-timeout fallback, an explicit override on both Range requests (the small-header and the full-metadata paths), and the forwarding through get_safetensors_metadata. All three pass; ruff check + format clean.

---

> [!NOTE]
> **Low Risk**
> Targeted networking change with backward-compatible defaults; behavior matches existing download timeout patterns elsewhere in the library.
> 
> **Overview**
> Adds a **`timeout`** parameter to **`get_safetensors_metadata`** and **`parse_safetensors_file_metadata`**, forwarding it to every Range GET when reading safetensors headers (fixes indefinite hangs on stalled CDN connections, #4377).
> 
> **`None`** uses **`HF_HUB_DOWNLOAD_TIMEOUT`** (same as other Hub downloads); **`0`** disables the bound by mapping to httpx **`None`** (not a zero-second timeout). Sharded repos pass the same timeout through each per-shard parse.
> 
> **`TestSafetensorsMetadataTimeout`** adds mocked regression tests for default timeout, explicit override on one- and two-request paths, forwarding from **`get_safetensors_metadata`**, and **`timeout=0`** behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9abf5dc7cd60d6eec5f5ffe0b6d4526273b807c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -6889,8 +6889,15 @@ class HfApi:
             [`SafetensorsParsingError`]:
                 If a safetensors file header couldn't be parsed correctly.
         """
+        # Normalize the timeout before handing it to httpx. `None` resolves to the
+        # configured default (`HF_HUB_DOWNLOAD_TIMEOUT`), while `0` is the documented
+        # way to disable the bound entirely. Note that for httpx `0` means a
+        # zero-second timeout (i.e. immediate failure), so we translate it to `None`,
+        # which is the only value that actually disables timeouts.
         if timeout is None:
             timeout = constants.HF_HUB_DOWNLOAD_TIMEOUT
+        elif timeout == 0:
+            timeout = None
         url = hf_hub_url(
             repo_id=repo_id, filename=filename, repo_type=repo_type, revision=revision, endpoint=self.endpoint
         )
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -6694,6 +6694,7 @@ class HfApi:
         repo_type: str | None = None,
         revision: str | None = None,
         token: bool | str | None = None,
+        timeout: float | None = None,
     ) -> SafetensorsRepoMetadata:
         """
         Parse metadata for a safetensors repo on the Hub.
@@ -6720,6 +6721,12 @@ class HfApi:
                 token, which is the recommended method for authentication (see
                 https://huggingface.co/docs/huggingface_hub/quick-start#authentication).
                 To disable authentication, pass `False`.
+            timeout (`float` or `None`, *optional*):
+                Timeout for each Range request, in seconds. If `None` (default), falls back to
+                [`~constants.HF_HUB_DOWNLOAD_TIMEOUT`] (defaults to 10s). Pass `0` to disable. Without a
+                bound, a stalled or half-open connection to a shard can block the worker indefinitely
+                and hang the whole call (and any caller). See
+                https://github.com/huggingface/huggingface_hub/issues/4377.
 
         Returns:
             [`SafetensorsRepoMetadata`]: information related to safetensors repo.
@@ -6771,6 +6778,7 @@ class HfApi:
                 repo_type=repo_type,
                 revision=revision,
                 token=token,
+                timeout=timeout,
             )
             return SafetensorsRepoMetadata(
                 metadata=None,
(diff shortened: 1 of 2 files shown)


sergioperezcheco:
Thanks for the review. The bot finding is stale against the current diff: the timeout normalization block already maps `0` → `None` before the value is forwarded to httpx, so `timeout=0` does disable the bound as documented (httpx treats `None` as "no timeout", while a literal `0` would be a zero-second timeout). The existing regression test `test_parse_safetensors_file_metadata_timeout_zero_disables` in `TestSafetensorsMetadataTimeout` asserts exactly this (`kwargs["timeout"] is None`). Happy to adjust if I've misread the comment.

sergioperezcheco:
Resolved in a9abf5d (the follow-up commit on this branch). The earlier commit d4c6d6d advertised `timeout=0` as disabling the bound but forwarded 0 to httpx unchanged, which is exactly the zero-second/immediate-failure behavior Cursor flagged. The follow-up now explicitly translates `0` to `None` (the only value that disables timeouts in httpx), so the documented disable path works:

```python
if timeout is None:
    timeout = constants.HF_HUB_DOWNLOAD_TIMEOUT
elif timeout == 0:
    timeout = None
```

`test_parse_safetensors_file_metadata_zero_disables_timeout` covers it and asserts the forwarded `timeout` is `None`.

Wauplin:
closed by https://github.com/huggingface/huggingface_hub/pull/4378

PR #4549, opened 2026-07-19
Support scheme-less custom endpoints with ports

PEliet:
# Support scheme-less custom endpoints with ports in HF URI parsing

## Summary

`parse_hf_uri` accepts scheme-less Hugging Face web URLs, including custom endpoints passed through `endpoint`. However, a custom endpoint with a port, such as `http://localhost:8080/hf`, is rejected when the input omits the scheme:

```python
parse_hf_uri(
    "localhost:8080/hf/my-org/my-model",
    endpoint="http://localhost:8080/hf",
)
```

This change makes scheme-less host detection parse the hostname with `urlsplit()` instead of relying on string-prefix comparisons. The port is therefore handled separately from the hostname, while the existing recognized-host allowlist remains in place.

## Changes

- Parse scheme-less URL candidates with a temporary `//` prefix.
- Compare the parsed hostname against `_recognized_hosts(endpoint)`.
- Treat malformed scheme-less netlocs as ordinary invalid HF URIs instead of leaking `ValueError`.
- Add a regression test for a scheme-less `localhost:8080` endpoint.
- Add regression coverage for malformed IPv6-like scheme-less inputs.

## Validation

```text
python -m pytest tests/test_utils_hf_uris.py -q
142 passed
```

```text
git diff --check
```

The existing behavior for URLs with a scheme and for the public Hugging Face hosts is covered by the existing test cases.

### DIFF
--- a/src/huggingface_hub/utils/_hf_uris.py
+++ b/src/huggingface_hub/utils/_hf_uris.py
@@ -344,8 +344,16 @@ def _looks_like_hf_url(uri: str, endpoint: str | None = None) -> bool:
     lowered = uri.lower()
     if lowered.startswith(("http://", "https://")):
         return True
-    # Scheme-less host (e.g. 'huggingface.co/org/model').
-    return any(lowered == host or lowered.startswith(host + "/") for host in _recognized_hosts(endpoint))
+    # Scheme-less host (e.g. 'huggingface.co/org/model'). Parse it instead of
+    # comparing string prefixes so custom endpoints with a port are supported.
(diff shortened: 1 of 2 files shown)


Wauplin:
Not worth the change in my opinion. Not a problem to accept only URLs with schemas when passing a local endpoint.

PEliet:
> Not worth the change in my opinion. Not a problem to accept only URLs with schemas when passing a local endpoint.

Okay.

PR #4552, opened 2026-07-20
fix(cli): allow '=' in hf jobs ls --label values

sergioperezcheco:
The label-filter parser in jobs_ps used item.split("=") with no maxsplit, so a label value that itself contains an '=' (common for base64 tokens, URLs with query strings, or JSON fragments like config=key=value) crashed with 'ValueError: too many values to unpack' instead of being sent to the API. Only the simple key=value case worked.

This switches to split("=", 1) so everything after the first '=' is treated as the value. That matches how the same module already parses env/secret entries (jobs.py:1306) and how the legacy --filter path works (jobs.py:1051), so the behavior is now consistent across the jobs CLI.

Added a regression test that passes --label with values containing multiple '=' signs and asserts they round-trip unchanged into the list_jobs call.

---

> [!NOTE]
> **Low Risk**
> Small CLI parsing fix with a targeted test; no auth, API contract, or server behavior changes.
> 
> **Overview**
> **`hf jobs ls --label`** no longer crashes when a label value contains `=` (e.g. base64, URLs with query strings).
> 
> The filter parser in **`jobs_ps`** now uses **`split("=", 1)`** so only the first `=` separates key from value, matching **`_parse_labels_map`** and the scheduled jobs **`--filter`** path. A CLI regression test asserts multi-`=` values are passed through to **`list_jobs`** unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88394a8869d2f8cd04007a9d5775cfb301a64ca4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -637,7 +637,7 @@ def jobs_ps(
     for item in label or []:
         if "=" not in item:
             raise CLIError(f"Invalid label filter '{item}': must be in the form 'key=value'")
-        key, value = item.split("=")
+        key, value = item.split("=", 1)
         labels[key] = value
 
     jobs_iter = api.list_jobs(namespace=namespace, status=server_statuses, labels=labels or None)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -3443,6 +3443,23 @@ class TestJobsCommand:
         assert kwargs["status"] == ["completed", "scheduling"]
         assert kwargs["labels"] == {"model": "Qwen3-06B"}
 
+    def test_ls_label_filter_accepts_value_containing_equals_sign(self, runner: CliRunner) -> None:
+        """Regression: a label value with an embedded '=' (e.g. a base64 blob, URL, or JSON
+        fragment) must not crash the CLI. Before the fix, `item.split("=")` raised
+        `ValueError: too many values to unpack` because it yielded more than 2 parts."""
+        with patch("huggingface_hub.cli.jobs.get_hf_api") as api_cls:
+            api = api_cls.return_value
+            api.list_jobs.return_value = self._make_mock_jobs()
+            result = runner.invoke(
+                app, ["jobs", "ls", "--label", "token=abc=def==ghi", "--label", "url=https://example.co/x?a=1"]
+            )
+        assert result.exit_code == 0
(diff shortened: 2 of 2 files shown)


Wauplin:
Hi @sergioperezcheco , do you have a real world example for this use case? I don't see why not just doing `key=value` . As-is I'm a bit reluctant to merge it

sergioperezcheco:
Hey @Wauplin, the bug isn't about the key=value format itself — it's that the parser used `item.split("=")` with no maxsplit, so a value that contains another `=` crashes with `ValueError: too many values to unpack`. Reproducer:

```
$ hf jobs ls --label config=a==b
ValueError: too many values to unpack (expected 2)
```

With `split("=", 1)` the value keeps everything after the first `=`: `key="config"`, `value="a==b"`, and that gets forwarded to `list_jobs` unchanged.

Real-world examples where this bites: base64-encoded tokens used as label values (`auth=Bearer dGVzdA==`), URLs with query strings (`redirect=https://x?a=b`), and label values that carry small JSON fragments (`config={"k":"v"}` — a lot of people store `key={...}` with a stray `=` after a serializer quirk). All of those are valid from the API side but the CLI rejects them today.

The same module already does `split("=", 1)` for env/secret parsing (`jobs.py:1306`) and for the legacy `--filter` path (`jobs.py:1051`), so this just makes `--label` consistent with those two rather than being the one codepath that breaks on embedded `=`. Happy to drop it if you'd rather just document the limitation, but it felt like an accidental inconsistency more than a design choice.

Wauplin:
what I meant is: did you have a real use case yourself and couldn't get around it? or is it just a theoretical one "just in case someone wants to use this or that format" ? If not a real request from a real user for a real use case, then I'll just close this one.

sergioperezcheco:
Honest answer @Wauplin: no real-world report, I hit it reading the parser while looking at something else and noticed `--label` was the one path that did a bare split. I would not push it if you would rather keep the surface minimal — happy to close. The only thing I would note is that `--env`/`--filter` already use split with maxsplit=1 in the same file, so if you do want to keep `--label` strict, the inconsistency is the one downside. Either way your call, no hard feelings if close.

Wauplin:
Thanks for your answer, let's drop this PR for now and reassess in case of community interest

PR #4555, opened 2026-07-20
feat(cli): add --sort/--reverse to hf jobs ps

### LINKED ISSUE #4374: Add --sort and --limit flags to hf jobs ps

Solaris-star:
## Summary
`hf jobs ps -a` had no way to reorder Jobs before display. This adds:

- `--sort {created,status,runtime,id}` (default `created`)
- `-r` / `--reverse`

`--limit` was already present; sorting now runs **before** the limit so `--sort created --reverse --limit 10` returns the 10 newest Jobs.

## Testing
- `python -m py_compile src/huggingface_hub/cli/jobs.py`
- Full CLI suite not run here (no live Jobs token); CI should cover `jobs` CLI tests if present

Fixes #4374

---

> [!NOTE]
> **Low Risk**
> CLI display-only change; loading all filtered jobs before limit may use more memory for very large job lists.
> 
> **Overview**
> **`hf jobs ps`** can now reorder the table with **`--sort`** (`created`, `status`, `runtime`, `id`; default `created`) and **`-r` / `--reverse`**.
> 
> Listing no longer applies **`--limit`** on the raw API iterator first. It loads all jobs that match the existing filters, builds rows, sorts in the CLI (the API has no sort parameter), then trims with **`--limit`**. That way combinations like **`--sort created --reverse --limit 10`** show the ten newest jobs, not an arbitrary first page from the server.
> 
> Runtime sorting uses raw seconds internally; missing runtime sorts as **0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43c73c60ce06066fc72d052180c51ab7562c23a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -592,6 +592,22 @@ def jobs_ps(
             help="Maximum number of Jobs to display. Set to 0 to show all (no limit).",
         ),
     ] = 100,
+    sort: Annotated[
+        str,
+        Option(
+            "--sort",
+            click_type=SoftChoice(["created", "status", "runtime", "id"]),
+            help="Sort Jobs by field: created, status, runtime, or id. Default: created.",
+        ),
+    ] = "created",
+    reverse: Annotated[
+        bool,
+        Option(
+            "-r",
+            "--reverse",
+            help="Reverse the sort order.",
+        ),
+    ] = False,
     namespace: NamespaceOpt = None,
     token: TokenOpt = None,
     filter: Annotated[
@@ -606,7 +622,8 @@ def jobs_ps(
     """List Jobs.
 
     Use `--status` to filter by status (see [`JobStage`] for possible values) and `--label` to filter by `key=value`
-    labels. A Job must match every filter to be listed.
+    labels. A Job must match every filter to be listed. Use `--sort` / `--reverse` to reorder and `--limit` to trim
+    the table.
     """
     api = get_hf_api(token=token)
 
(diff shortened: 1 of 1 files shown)


Wauplin:
same answer as in https://github.com/huggingface/huggingface_hub/pull/4502#issuecomment-4914680966

we don't want to implement that client side as long as the server API doesn't allow for it

PR #4556, opened 2026-07-21
docs: clarify upload_folder has no Xet server-side fallback

### LINKED ISSUE #4507: Clarification on `upload_folder` fallback behavior when the server doesn't support XET
### Describe the bug

Hi team, first, thanks for all the recent work on `upload_folder` and the XET pipeline. The multi-commit and resumable behavior in 1.20.0 is really nice to have.

I ran into a case I'd like your guidance on, and I want to make sure I'm reading the docs correctly before assuming anything is off.

The [upload guide](https://github.com/huggingface/huggingface_hub/blob/main/docs/source/en/guides/upload.md) and the `upload_folder` docstring both describe the fallback as:

> If `hf_xet` is not installed, `upload_folder` falls back to the legacy behavior

That covers the client-side case very clearly. The situation I'm not sure about is when the client does have `hf_xet` installed but the target Hub doesn't support XET server-side (e.g. `/api/{type}/{repo}/xet-write-token/{revision}` returns 404, or `xetEnabled` is `false`). In that case `upload_folder` raises a `ConnectionError` from inside the XET pipeline and there doesn't seem to be a fallback path.

Setting `HF_HUB_DISABLE_XET=1` works around it cleanly, so I have an unblock. I just wanted to check whether this case is intended to be user-handled via that env var, or whether the docs' "falls back to the legacy behavior" was intended to cover it too. I noticed the pre-1.20.0 client used to negotiate transports server-side via the LFS batch endpoint (removed in [`97ea04b`](https://github.com/huggingface/huggingface_hub/commit/97ea04be543c4808aa4050fcd3f7ead9b81d3f0e)), which is what got me wondering.

If the intent is XET-first everywhere going forward, a short note in the upload guide saying "non-XET Hub deployments should set `HF_HUB_DISABLE_XET=1`" would help folks in similar setups discover the workaround faster. And if a server-side fallback ever fits the roadmap, that would be great too, but no pressure either way.

Thanks again for maintaining this so actively. Happy to test any changes or provide more info if useful.

### Reproduction

Reproducible against any Hub endpoint that does not implement `/api/{repo_type}s/{repo_id}/xet-write-token/{revision}` (returns 404), or where the repo has `xetEnabled: false`.

```python

# Point at any Hub deployment that does not support XET server-side
api = HfApi(endpoint="https://<hub-without-xet-support>")
api.upload_folder(
    folder_path="./folder",
    repo_id="user/repo",
    repo_type="model",
)
```

Setting `HF_HUB_DISABLE_XET=1` before importing `huggingface_hub` routes through the legacy `create_commit` LFS path and works correctly.

### Logs

```shell
Shape of the failure (endpoint host redacted):


  File ".../huggingface_hub/hf_api.py", in upload_folder
    return pipelined_upload(...)
  File ".../huggingface_hub/_upload_pipeline.py", in run
    self._coordinator_loop()
  File ".../huggingface_hub/_upload_pipeline.py", in _coordinator_loop
    batch.xet_commit = self.xet_session.new_upload_commit(...)
ConnectionError: Network error: Request error: HTTP status client error (404 Not Found),
domain: https://<hub-without-xet-support>/api/models/user/repo/xet-write-token/main
```

### System info

```shell
- huggingface_hub version: 1.20.0 (also observed on later releases through 1.22.0)
- hf_xet: installed (default since 1.20.0)
- ENDPOINT: a Hub deployment that does not support XET server-side (returns 404 on `/xet-write-token/{revision}`)
- HF_HUB_DISABLE_XET: False (setting to True is the working workaround)
```

Solaris-star:
## Summary

Clarify `upload_folder` Xet behavior when the **client** has `hf_xet` but the **target Hub** does not support Xet server-side (#4507).

Previously the upload guide only described the client-side case (`hf_xet` not installed → legacy path). When `hf_xet` is installed and the Hub returns 404 on the Xet write-token endpoint (or `xetEnabled: false`), there is **no automatic transport fallback** — users should set `HF_HUB_DISABLE_XET=1` (or uninstall `hf_xet`).

Also notes the same in the `upload_folder` docstring.

## Testing

Docs/docstring only; no runtime change.

Fixes #4507

---

> [!NOTE]
> **Low Risk**
> Documentation and docstring updates only; no code path or upload logic changes.
> 
> **Overview**
> Documents a gap that was only partially covered before: when **`hf_xet` is installed** but the **Hub does not support Xet server-side** (404 on the xet write-token route or `xetEnabled: false`), **`upload_folder` does not automatically fall back** to the legacy hash-then-HTTP path.
> 
> The upload guide now states that self-hosted or non-Xet deployments should set **`HF_HUB_DISABLE_XET=1`** (or uninstall `hf_xet`) instead of expecting silent client-side fallback. A missing period on the existing “`hf_xet` not installed” sentence is fixed.
> 
> The same guidance is added to the **`upload_folder`** docstring in `hf_api.py`. **No runtime behavior changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11e5ed71850105fe659a5a53747d40f8c41d3635. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/upload.md
+++ b/docs/source/en/guides/upload.md
@@ -93,7 +93,9 @@ Found 5,000 files to upload
   Committing  ██████████████████░░  4,580 / 5,000  6 commits
 ```
 
-If `hf_xet` is not installed, [`upload_folder`] falls back to the legacy behavior: hash everything first, upload over HTTP, then create a single commit. We always recommend to keep `hf_xet` installed for better robustness
+If `hf_xet` is not installed, [`upload_folder`] falls back to the legacy behavior: hash everything first, upload over HTTP, then create a single commit. We always recommend to keep `hf_xet` installed for better robustness.
+
+If `hf_xet` **is** installed but the target Hub does not support Xet server-side (for example `/api/{type}/{repo}/xet-write-token/{revision}` returns 404, or the repo reports `xetEnabled: false`), there is **no automatic transport fallback**. Set [`HF_HUB_DISABLE_XET=1`](../package_reference/environment_variables.md#hfhubdisablexet) to force the legacy upload path. Self-hosted or non-Xet Hub deployments should use that env var (or uninstall `hf_xet`) rather than expecting a silent client-side fallback.
 
 ## Upload from the CLI
 
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -5751,7 +5751,7 @@ class HfApi:
         read pass), and large folders are automatically committed in several batches to stay below server limits
         (follow-up commits get a ` (part N)` suffix on the commit message). If the upload is interrupted, re-running
         the same call resumes it: already-committed files are skipped and already-uploaded data is deduplicated. When
-        `hf_xet` is not installed, falls back to a single commit created with [`create_commit`].
+        `hf_xet` is not installed, falls back to a single commit created with [`create_commit`]. If `hf_xet` is installed but the Hub does not support Xet server-side, set `HF_HUB_DISABLE_XET=1` to force the legacy path (there is no automatic transport fallback).
 
         Args:
             repo_id (`str`):


Wauplin:
closing this one without merging (as per https://github.com/huggingface/huggingface_hub/issues/4507#issuecomment-4922843419)

PR #4557, opened 2026-07-21
fix(cli/jobs): correct cell truncation when terminal-fit loop compresses columns below width 3

sergioperezcheco:
The private _tabulate helper used by `hf jobs ps` / `hf jobs ls` truncates overflowing cells with `value[: col_width - 3] + "..."`. The terminal-fit loop above it (`while len(headers) + sum(col_widths) > terminal_width: col_widths[...] //= 2`) can compress a column down to width 0, 1, or 2 when the terminal is narrow relative to the content. In that range the slice `col_width - 3` goes negative, so instead of truncating it drops trailing characters and then the appended "..." makes the cell longer than both the original value and the column. The visible effect is that `hf jobs ps` rows become ragged/overflowing in narrow terminals instead of staying aligned.

Fix: extracted the truncation into _truncate_to_width, which hard-truncates to the exact column width when there is no room for an ellipsis (col_width <= 3) and otherwise keeps the existing `value[: col_width - 3] + "..."` behavior. Added tests/test_cli_jobs_tabulate.py covering the narrow-width cases and a full narrow-terminal table render asserting no row overflows.

Ran ruff check/format on the two changed files (clean) and the new tests pass. Note: I could not run the full `make quality` locally because generate_cli_reference.py fails in my environment on a click/Typer version mismatch unrelated to this change.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 1e2dfc5241d4b3d33161a95fa4cd351093964571. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/jobs.py
+++ b/src/huggingface_hub/cli/jobs.py
@@ -1312,6 +1312,22 @@ def _parse_labels_map(labels: list[str] | None, *, name: str | None = None) -> d
     return labels_map
 
 
+def _truncate_to_width(value: str, col_width: int) -> str:
+    """Truncate ``value`` to ``col_width`` columns, appending ``...`` when truncated.
+
+    Handles ``col_width < 3`` (which the terminal-fit loop in :func:`_tabulate` can
+    produce): the previous ``value[: col_width - 3] + "..."`` form turned into a
+    negative slice that dropped trailing chars and then overflowed the column
+    with the ellipsis.
+    """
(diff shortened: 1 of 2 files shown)


Wauplin:
thanks for the PR but it's already fine like this. We don't want to optimize for columns less than 3 characters. If you encounter such issues, I recommend you passing  `--json` + using jq to do whatever you prefer.

PR #4564, opened 2026-07-22
[Inference] Send max_tokens for Together and Novita text-generation

rahulrshetty45:
## Summary

- `InferenceClient.text_generation(...)` always sends the TGI-style `max_new_tokens` parameter.
- The Together and Novita **text-generation** tasks route to OpenAI-compatible completion endpoints (`/v1/completions` and `/v3/openai/completions`) and parse the OpenAI response shape (`choices[0]["text"]`), but they inherit the base payload builder, so they forwarded `max_new_tokens` unchanged. OpenAI-style endpoints expect `max_tokens` and silently ignore the unknown field — so the user's length limit had no effect.
- Fix: rename `max_new_tokens` → `max_tokens` in both `_prepare_payload_as_dict`, exactly as the sibling `DeepInfraTextGenerationTask` and `FeatherlessTextGenerationTask` already do for the same kind of endpoint. HF-inference (TGI) keeps `max_new_tokens`, so the base class is intentionally left unchanged.

### Before (payload sent to Together)

```python
{"prompt": "hi", "max_new_tokens": 50, "temperature": 0.7, "model": "..."}   # max_new_tokens ignored by the endpoint
```

### After

```python
{"prompt": "hi", "temperature": 0.7, "max_tokens": 50, "model": "..."}
```

Added payload tests for both providers.

---

> [!NOTE]
> **Low Risk**
> Targeted inference payload mapping and CLI formatting with regression tests; no auth or broad API surface changes.
> 
> **Overview**
> Fixes **Together** and **Novita** text-generation so `InferenceClient.text_generation` length limits actually apply: both providers hit OpenAI-style completion APIs that expect **`max_tokens`**, but payloads still forwarded TGI-style **`max_new_tokens`**, which those endpoints ignore. Each provider now overrides **`_prepare_payload_as_dict`** to rename the field (same pattern as DeepInfra/Featherless); HF Inference / base text-generation behavior is unchanged.
> 
> Separately, **agent (TSV) CLI tables** now render missing/`None` cells as an **empty field** instead of the literal string `"None"`, matching human/json handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca659376303168e579647fd8549da8f45b447ff7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/_output.py
+++ b/src/huggingface_hub/cli/_output.py
@@ -352,6 +352,8 @@ def _truncate_columns(
 
 def _format_table_cell_agent(value: Any) -> str:
     """Format a cell value for agent TSV output (ISO timestamps, tabs escaped)."""
+    if value is None:
+        return ""
     if isinstance(value, datetime.datetime):
         return value.isoformat()
     return _single_line(str(value))
--- a/src/huggingface_hub/inference/_providers/novita.py
+++ b/src/huggingface_hub/inference/_providers/novita.py
@@ -23,6 +23,15 @@ class NovitaTextGenerationTask(BaseTextGenerationTask):
         # there is no v1/ route for novita
         return "/v3/openai/completions"
 
+    def _prepare_payload_as_dict(
+        self, inputs: Any, parameters: dict, provider_mapping_info: InferenceProviderMapping
+    ) -> dict | None:
+        params = filter_none(parameters.copy())
+        # Novita's completion endpoint is OpenAI-compatible and expects `max_tokens`.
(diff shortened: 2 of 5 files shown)


Wauplin:
Thanks for the report @rahulrshetty45 . Could you please provide a real example (i.e. a standalone snippet of code that one can run) that fails on Together and Novita and that is fixed by this PR. Thanks in advance!

PR #4566, opened 2026-07-22
[Jobs] Auto-name Jobs on creation

Wauplin:
## Summary

Follow-up to #4532 (named Jobs). Jobs now get an automatic `name` when the user doesn't provide one, following: **auto-name == the script (uv) or image (docker) you run**, with a **short hash of the command line appended** so reruns of the same command group together while different runs stay distinct.

- `run_job` / `create_scheduled_job` → base from `image`: keep the final `name:tag` component (registry/namespace dropped), or the `namespace/repo` id for Spaces.
- `run_uv_job` / `create_scheduled_uv_job` → base from `script` (filename without `.py`).
- Suffix `-<hash>` where `<hash>` is the first 8 hex chars of `sha256(image/script + command line)`.
- Explicit `--name` / `name=` / `name` label still take precedence.

Job names are validated server-side against `^[a-zA-Z0-9._-]*$` **and** are also stored as tags (which additionally forbid `.`). So `:`, `/` and `.` can't be kept verbatim — every disallowed char in the base is replaced with `-`:

| input | auto-name |
|---|---|
| `python:3.12 foo --truc` | `python-3-12-7c6db949` |
| `python:3.12 foo --bar` | `python-3-12-2c9bffe9` |
| `pytorch/pytorch:2.6.0-cuda12.4-cudnn9-devel ...` | `pytorch-2-6-0-cuda12-4-cudnn9-devel-<hash>` |
| `hf.co/spaces/lhoestq/duckdb ...` | `lhoestq-duckdb-<hash>` |
| `train.py --epochs 3` (uv) | `train-2f161011` |
| `lighteval` (uv) | `lighteval-<hash>` |

Same image, same command → same name (groups reruns); same image, different command → different name.

Also updated CLI hints, `--name` help and the jobs/cli guides.

## Tests

Parametrized tests for the two derivation helpers (no mock / no HTTP), plus a test asserting `python:3.12 foo --truc` and `python:3.12 foo --bar` get different names while the same command is stable.

## Manual testing (real `hf` CLI)

```
$ hf jobs run --detach python:3.12 foo --truc
  id: 6a60b85c13e6ef894d54b949
Hint: Job auto-named 'python-3-12-7c6db949'. Pass `--name` or run `hf jobs labels Wauplin/... --name NAME` to rename it.

$ hf jobs uv run --detach train.py          # uv job -> named from script, not the uv image
Hint: Job auto-named 'train-<hash>'. ...

$ hf jobs run --detach --name my-custom python:3.12 python -c 'print("hi")'
✓ Job started                                # explicit --name: no auto-name hint

$ hf jobs run --detach --label name=explicit-lbl python:3.12 python -c 'print("hi")'
✓ Job started                                # name via --label: also treated as explicit, no auto hint
```

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -2028,7 +2028,7 @@ Add labels to a Job using `-l` or `--label`. Labels are a key=value pairs that a
 
 The my-label key doesn't specify a value so its value defaults to an empty string ("").
 
-Use `--name` to add the `name` label when creating a Job. Names make Jobs easier to find and identify in the UI; they are optional and do not have to be unique. If you don't pass `--name`, a name is derived automatically from the Docker image or the script, e.g. `python:3.12` → `python-3-12` You can also rename an existing Job:
+Use `--name` to add the `name` label when creating a Job. Names make Jobs easier to find and identify in the UI; they are optional and do not have to be unique. If you don't pass `--name`, a name is derived automatically from the Docker image or the script, plus a short hash of the command so reruns of the same command share a name (e.g. `python:3.12 foo --truc` → `python-3-12-1a2b3c4d`). You can also rename an existing Job:
 
 ```bash
 >>> hf jobs run --name training-v2 python:3.12 python train.py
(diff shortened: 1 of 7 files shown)


Wauplin (commented):
✔️

Wauplin:

PR #4595, opened 2026-07-27
[Download] Reject redacted Xet hashes from tree cache

### LINKED ISSUE #895: Download fails with "Task error: Unable to parse string as hex hash value" (hf-xet 1.5.1)
## Environment
- hf-xet 1.5.1 (latest), huggingface_hub 1.23.0, Python 3.13, macOS 15.6 (arm64)

## What happened
`snapshot_download` on a gated private dataset repo (14 files; one 11.9 GB xet-backed blob) fails shortly after starting:

```
File ".../huggingface_hub/file_download.py", line 1920, in _download_to_tmp_and_move
    xet_get(
File ".../huggingface_hub/file_download.py", line 563, in xet_get
    with session.new_file_download_group(
        token_refresh_url=xet_file_data.refresh_route,
        ...
    ) as group:
RuntimeError: Task error: Unable to parse string as hex hash value
```

Small files in the same snapshot download fine; the failure happens on the large blob. Auth token is valid (the same call succeeds with xet disabled).

## Workaround
`HF_HUB_DISABLE_XET=1` downloads the same snapshot successfully over plain HTTP.

Happy to provide more details or run a patched build — the repo is private so I can't share a public repro, but the failure is reproducible on this snapshot.

seanses:
## Summary

Fixes [xet-core#895](https://github.com/huggingface/xet-core/issues/895): xet downloads failing with `Unable to parse string as hex hash value (got '****************************************************************')`.

- Root cause: [#4394](https://github.com/huggingface/huggingface_hub/pull/4394) added a tree-listing cache and skips per-file HEAD calls for xet files when metadata can be rebuilt from `/tree`. On gated repos, `/tree` can return a redacted `xetHash` (`64 × '*'`) when the caller lacks content access — the real hash only comes from HEAD (`X-Xet-Hash` on `/resolve`).
- Add `is_valid_xet_hash()` and reject invalid/masked hashes in `_xet_file_metadata_from_tree_cache` → fall back to HEAD instead of passing `'****...'` to hf-xet.
- Add `is_valid_tree_entries()` and enforce it directly in the tree-cache layer: `read_tree_cache` treats invalid/redacted trees as cache misses, while `write_tree_cache` skips them. This keeps `snapshot_download` unchanged and lets newly granted gate access refresh the metadata.

Repro the redacted tree hash:

```bash
curl -s 'https://huggingface.co/api/models/google/translategemma-4b-it/tree/main?recursive=true' \
  | jq '[.[] | select(.xetHash != null) | {path, xetHash}] | .[0]'
# → "xetHash": "****************************************************************"
```

---

> [!NOTE]
> **Low Risk**
> Targeted validation on download/cache paths with tests; behavior change is to reject bad cached metadata and retry via HEAD, which is the safe fallback.
> 
> **Overview**
> Fixes Xet downloads failing when the Hub returns a **redacted** `xetHash` (`64 × '*'`) from `/tree` on gated repos without content access—the tree cache was treating that value as real and skipping per-file HEAD, so `hf-xet` tried to parse it as hex.
> 
> Adds **`is_valid_xet_hash()`** (64 hex characters) and uses it in **`_xet_file_metadata_from_tree_cache`** so invalid hashes fall back to HEAD (where `X-Xet-Hash` is correct). **`read_tree_cache`** / **`write_tree_cache`** enforce **`is_valid_tree_entries()`** so redacted listings are never stored or served from cache; after access is granted, a fresh tree can be cached normally.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c6b4477284503018916543775dbbbe21dbeaacc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -20,6 +20,7 @@ from .errors import (
 from .file_download import REGEX_COMMIT_HASH, DryRunFileInfo, hf_hub_download, repo_folder_name
 from .hf_api import DatasetInfo, HfApi, KernelInfo, ModelInfo, RepoFile, SpaceInfo
 from .utils import OfflineModeIsEnabled, filter_repo_objects, logging, validate_hf_hub_args
+from .utils._xet import is_valid_xet_hash
 from .utils._xet_progress_reporting import (
     XET_BYTES_BAR_FORMAT,
     XET_TRANSFER_BAR_FORMAT,
@@ -381,7 +382,7 @@ def snapshot_download(
                 blob_id=f.blob_id,
                 lfs_sha256=f.lfs.sha256 if f.lfs is not None else None,
                 lfs_size=f.lfs.size if f.lfs is not None else None,
-                xet_hash=f.xet_hash,
+                xet_hash=f.xet_hash if f.xet_hash is not None and is_valid_xet_hash(f.xet_hash) else None,
             )
             for f in api.list_repo_tree(repo_id=repo_id, recursive=True, revision=commit_hash, repo_type=repo_type)
             if isinstance(f, RepoFile)
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -380,10 +380,6 @@ def snapshot_download(
 
     # Retrieve /tree listing from cache or fetch it
     tree_entries = read_tree_cache(tree_cache_folder, commit_hash)
-    if tree_entries is not None and not is_valid_tree_entries(tree_entries):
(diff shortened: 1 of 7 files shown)


Wauplin:
Thanks @seanses for the PR! Makes perfect sense. I took the liberty to update the PR description and content to invalidate any cached tree that contain a redacted xet hash. This way any legacy cached tree file will be overwritten with correct values as soon as the user passes the model's gate. I'll wait for the CI to go green and merge it. Should be released later this week.

Wauplin (approved):

PR #4608, opened 2026-07-29
[CLI] Install & refresh the hf-cli skill (installer, update, hints)

Wauplin:
## Summary

Implements the solutions discussed in [this Slack thread](https://huggingface.slack.com/archives/C0ACCEFPU9W/p1785184000472449):

1. **Installer (opt-out)**: the standalone installer runs `hf skills add hf-cli --global --claude` by default and prints a disclaimer saying the skill was installed automatically. Pass `--exclude-skill` (bash) / `-ExcludeSkill` (powershell) to skip it. The CI installer check asserts the installed skill matches `hf version`.
2. **`hf update`**: refreshes the globally installed `hf-cli` skill. Runs in a subprocess since the CLI code was just replaced on disk while the current process still runs the old version. Only global installs (`~/.agents/skills`, `~/.claude/skills`) - we don't touch project directories. Hints at `hf skills add` if not installed. For standalone installs, `hf update` re-runs the installer, so it checks whether the skill is installed *before* updating and passes `--exclude-skill` / `-ExcludeSkill` when it isn't: opting out once (or removing the skill later) is never silently undone.
3. **Any `hf` command**: at most once a day, hints at `hf skills add` (not installed) or `hf skills update` (SKILL.md generated by another `hf` version). Never auto-installs or auto-updates. Skipped on `hf skills ...` (the user is already managing skills), on `hf update` (it handles the skill itself, so the hint would be redundant or contradictory), and when `HF_HUB_DISABLE_UPDATE_CHECK=1`.

Only `hf-cli` is covered, other marketplace skills are left alone. Skills directories (`.agents/skills`, `.claude/skills`, local + global) moved to `constants.py`.

Note: the installer's uninstall instructions deliberately don't list the skill directories - keeping that block short matters more than being exhaustive there.

Drive-by: `iwr <url> | iex -SomeFlag` doesn't work (`Invoke-Expression` has no such parameter), so the PowerShell installer flags were never actually reaching the script. Switched to `& ([scriptblock]::Create((iwr -useb <url>))) -SomeFlag`, which also fixes the pre-existing `-WithTransformers` command surfaced in the update hint.

## Examples

```
# 1. installer (default)
$ curl -LsSf https://hf.co/cli/install.sh | bash -s
[INFO] Installing the hf-cli skill for AI agents...
Installed 'hf-cli' to central location: ~/.agents/skills/hf-cli
Created symlink: ~/.claude/skills/hf-cli
[INFO] The hf-cli skill was installed automatically so AI agents know how to use the hf CLI.
[INFO] Pass --exclude-skill to skip it.
...
[INFO] To uninstall the Hugging Face CLI, run:
[INFO]   rm -rf ~/.hf-cli
[INFO]   rm -f ~/.local/bin/hf

# 1b. opt-out
$ curl -LsSf https://hf.co/cli/install.sh | bash -s -- --exclude-skill
[INFO] Skipping the hf-cli skill (--exclude-skill)

# 2. hf update (after a successful update)
Updating the `hf-cli` skill...
hf-cli: up_to_date

# 3. daily hint
$ hf version
Hint: The `hf-cli` skill is not installed. Run `hf skills add -g --claude` to teach your AI agents how to use the `hf` CLI.
version=1.26.0.dev0

$ hf version
Hint: The installed `hf-cli` skill was generated by another `hf` version. Run `hf skills update hf-cli -g --claude` to refresh it.
version=1.26.0.dev0
```

Note: the `windows-installer` check is red for a pre-existing reason, tracked in #4609 (`hf version` crashes on Windows when stdout isn't UTF-8). Same failure on `main` since 2026-06-10; the workflow only runs when `utils/installers/**` changes, which is why it surfaces here.

---

> [!NOTE]
> **Medium Risk**
> Touches install/update paths and user home directories (~/.agents/skills), but changes are opt-out aware and hints-only at CLI startup; subprocess refresh after update is the main behavioral edge case.
> 
> **Overview**
> **End-to-end `hf-cli` skill lifecycle** for AI agents: default install via standalone installers, refresh on `hf update`, and optional daily CLI hints—without forcing installs on users who opted out.
> 
> The **bash and PowerShell installers** now run `hf skills add hf-cli --global --claude --force` after the CLI is installed, with **`--exclude-skill` / `-ExcludeSkill`** to skip. CI installer jobs set `HOME` / `USERPROFILE` hermetically and assert `SKILL.md` exists and matches `hf version`.
> 
> **`hf update`** detects whether the global `hf-cli` skill is present; standalone re-installs pass **`exclude_skill`** when it is not, so opt-out is preserved. When the skill is installed, update runs **`hf skills update hf-cli -g --claude`** in a subprocess (fresh binary after upgrade). PowerShell self-update commands are fixed to use a **scriptblock** so installer flags (including pre-existing `-WithTransformers`) actually reach the script.
> 
> A **daily local check** (`check_skill_update`, gated by `HF_HUB_DISABLE_UPDATE_CHECK`) hints at `hf skills add` or `hf skills update` when missing or version-stale; it is skipped for `hf skills` and `hf update`. Skill directory paths move to **`constants.py`**; Claude installs **copy instead of symlink** on Windows when symlinks fail.
> 
> Docs and env-var docs describe installer behavior, update refresh rules, and the new hint.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96b92d0172fe65d04c5e96a740a619ac607e18b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


utils/installers/install.sh
```diff
@@ -378,6 +383,18 @@ expose_cli_command() {
...
+install_skills() {
+    if [ "$INCLUDE_SKILLS" != "true" ]; then
+        return
+    fi
+
+    log_info "Installing the hf-cli skill for AI agents..."
+    if ! "$BIN_DIR/hf" skills add hf-cli --global --claude --force; then
+        log_warning "Failed to install the hf-cli skill. Install it later with: hf skills add -g --claude"
+    fi
+}
```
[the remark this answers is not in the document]
Wauplin: Not a bug: the disclaimer was intentionally dropped in ebbcf51. Nothing is auto-installed here - the skill is only installed when the user explicitly passes `--include-skills` / `-IncludeSkills`, so there's no need to tell them it happened or how to undo it. The "promised disclaimer" came from an earlier revision of the PR description, now updated (along with a couple of doc sentences that could read as if the skill were installed automatically).

docs/source/en/guides/cli.md
```diff
@@ -18,6 +18,8 @@ The `huggingface_hub` Python package comes with a built-in CLI called `hf`. This
 > # includes the above + Claude Code
 > hf skills add --claude
 > ```
+> The standalone installer can install it for you with `--include-skills` (see below), and `hf update` refreshes it.
+> Otherwise, the CLI merely suggests `hf skills add` / `hf skills update` at most once a day.
```
```suggestion
```

Wauplin (commented):
✔️

utils/installers/install.sh
```diff
@@ -378,6 +383,22 @@ expose_cli_command() {
...
+    fi
+
+    log_info "Installing the hf-cli skill for AI agents..."
+    if ! "$BIN_DIR/hf" skills add hf-cli --global --claude --force; then
+        log_warning "Failed to install the hf-cli skill. Install it later with: hf skills add -g --claude"
+        return
+    fi
+    log_info "The hf-cli skill was installed automatically so AI agents know how to use the hf CLI."
+    log_info "Pass --exclude-skill to skip it."
+}
```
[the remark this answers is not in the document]
Wauplin: Good catch, fixed in 27dfa706b.

`hf update` now checks whether the `hf-cli` skill is installed globally *before* re-running the installer, and passes `--exclude-skill` / `-ExcludeSkill` when it isn't. So opting out at install time (or removing the skill later) is never silently undone.

While at it: `iwr <url> | iex -ExcludeSkill` wouldn't have worked anyway (`Invoke-Expression` has no such parameter), so the PowerShell paths now use `& ([scriptblock]::Create((iwr -useb <url>))) -ExcludeSkill`. That also fixes the pre-existing `-WithTransformers` variant.

src/huggingface_hub/cli/hf.py
```diff
@@ -123,6 +124,8 @@ def main():
         if not constants.HF_DEBUG:
             logging.set_verbosity_info()
         check_cli_update("huggingface_hub")
+        if sys.argv[1:2] != ["skills"]:  # don't nag while the user is already managing skills
+            check_skill_update()
```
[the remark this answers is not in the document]
Wauplin: Right, fixed in 51f04b45c — the daily hint is now skipped on `hf update` too, since `update()` handles the skill itself.

```python
# Don't nag while the user is already managing skills, nor on `hf update` which handles the
# skill itself (it would print a redundant or contradictory hint before doing so).
if sys.argv[1:2] not in (["skills"], ["update"]):
    check_skill_update()
```

Added a parametrized test covering the three cases (`hf version` → check runs, `hf skills add` / `hf update` → skipped).

src/huggingface_hub/cli/skills.py
hanouticelina: now that `install.ps1` runs `hf skills add hf-cli --global --claude --force`, this line is run on windows and creating a symlink on windows requires Dev Mode or admin rights (double-checked with Opus). what we already do in the library is probe, fallback, and never fail on a symlink
```suggestion
    try:
        link_path.symlink_to(os.path.relpath(central_skill_path, agent_skills_dir))
    except OSError:
        # Windows needs Developer Mode or admin rights for symlinks (see `are_symlinks_supported`).
        # Fall back to a copy: `hf skills update` walks both roots, so the copy stays in sync.
        shutil.copytree(central_skill_path, link_path)
    return link_path
```

src/huggingface_hub/cli/_skills.py
```diff
@@ -309,3 +314,73 @@ def _apply_single_update(
...
+    if stamp_path.exists() and (time.time() - stamp_path.stat().st_mtime) < 24 * 3600:
+        return
+    stamp_path.parent.mkdir(parents=True, exist_ok=True)
+    stamp_path.touch()
+
+    global_dirs = _installed_hf_cli_dirs(constants.AGENTS_SKILLS_GLOBAL_PATH, constants.CLAUDE_SKILLS_GLOBAL_PATH)
+    local_dirs = _installed_hf_cli_dirs(constants.AGENTS_SKILLS_LOCAL_PATH, constants.CLAUDE_SKILLS_LOCAL_PATH)
+
+    if not global_dirs and not local_dirs:
+        out.hint(
```
hanouticelina, line 358: let's skip dev and rc versions
```suggestion
    if not global_dirs and not local_dirs:
        if any(tag in __version__ for tag in ["rc", "dev"]):
            return
        out.hint(
```

hanouticelina (approved):
Looks good to me! left two minor comments, I will address them and ship this PR!

src/huggingface_hub/cli/skills.py
```diff
@@ -334,7 +331,12 @@ def _create_symlink(agent_skills_dir: Path, skill_name: str, central_skill_path:
     link_path = agent_skills_dir / skill_name
 
     _remove_existing(link_path, force)
-    link_path.symlink_to(os.path.relpath(central_skill_path, agent_skills_dir))
+    try:
+        link_path.symlink_to(os.path.relpath(central_skill_path, agent_skills_dir))
+    except OSError:
+        # Windows needs Developer Mode or admin rights for symlinks (see `are_symlinks_supported`).
+        # Fall back to a copy: `hf skills update` walks both roots, so the copy stays in sync.
```
hanouticelina:
```suggestion
        # Windows needs Developer Mode or admin rights for symlinks.
        # Fall back to a copy: `hf skills update` walks both roots, so the copy stays in sync.
```

PR #4619, opened 2026-07-31
[CLI] Fix scheduled upload of a single file in a subfolder

dfedoryshchev:
## Summary

`hf upload <repo> <dir>/<file> --every N` never commits anything. It prints `Scheduling commits every N minutes to <repo>` and then runs forever without uploading the file.

Two problems in the single-file branch of the `--every` path, both from the same mismatch: `CommitScheduler` matches `allow_patterns` against paths relative to `folder_path` and re-uploads them under `path_in_repo`, while the command was passing the full local path.

**1. The allow pattern can never match.**

```python
# _commit_scheduler.py, push_to_hub()
relpath_to_abspath = {
    path.relative_to(self.folder_path).as_posix(): path
    for path in sorted(self.folder_path.glob("**/*")) if path.is_file()
}
for relpath in filter_repo_objects(relpath_to_abspath.keys(), allow_patterns=self.allow_patterns, ...):
```

For `hf upload my-model checkpoints/model.safetensors --every 5` the scheduler gets `folder_path="checkpoints"` and `allow_patterns=["checkpoints/model.safetensors"]`, while the only relpath it sees is `model.safetensors`. Nothing matches, so every run hits `Dropping schedule commit: no changed file to upload.` It bails at `_commit_scheduler.py:249` before any request is sent, so there is no server-side trace either.

**2. Only the folder part of `path_in_repo` can be honoured.**

The scheduler globs a folder and re-uploads whatever it finds underneath `path_in_repo`, so the destination file name is always the local one and a rename is not expressible. Stripping the file name with `endswith` mis-sliced any destination whose last segment merely ends with the local name: for a local `model.safetensors`, a destination of `adapter_model.safetensors` produced the prefix `adapter_` and would have landed the file at `adapter_/model.safetensors`. The prefix is now taken from the directory part only, and a destination file name that differs from the local one warns rather than being silently dropped.

## Which invocations are affected

This is the documented single-file form. The CLI guide's own example is a file in a subfolder:

```bash
# docs/source/en/guides/cli.md
>>> hf upload Wauplin/my-cool-model ./models/model.safetensors
https://huggingface.co/Wauplin/my-cool-model/blob/main/model.safetensors
```

Add `--every` to that documented command and it silently uploads nothing.

| command | before | after |
| --- | --- | --- |
| `hf upload repo model.safetensors --every 5` | uploads to `model.safetensors` | unchanged |
| `hf upload repo models/model.safetensors --every 5` | **nothing uploaded** | uploads to `model.safetensors` |
| `hf upload repo /abs/path/model.safetensors --every 5` | **nothing uploaded** | uploads to `model.safetensors` |
| `hf upload repo models/model.safetensors weights/model.safetensors --every 5` | **nothing uploaded** | uploads to `weights/model.safetensors` |
| `hf upload repo model.safetensors adapter_model.safetensors --every 5` | uploads to `adapter_/model.safetensors` | warns, uploads to `model.safetensors` |

Only a relative path with no directory component and no destination override worked. An absolute path failed too, even without a subfolder, since `os.path.dirname` is non-empty there as well.

## Reproduce (before)

```console
$ mkdir -p /tmp/hf-repro/models && echo weights > /tmp/hf-repro/models/model.safetensors
$ cd /tmp/hf-repro
$ HF_DEBUG=1 hf upload <user>/my-cool-model ./models/model.safetensors --every 1 --private
Scheduling commits every 1 minutes to <user>/my-cool-model.
# ...then, on every tick:
#     Dropping schedule commit: no changed file to upload.
$ hf models ls <user>/my-cool-model   # empty
```

The same command without `--every` uploads correctly, which is what makes the scheduled path a bug rather than a missing feature.

## Notes on the implementation

- `PATH_IN_REPO` is a positional argument, not a `--path-in-repo` flag.
- The rename notice uses `out.warning()` rather than `warnings.warn()`, matching the CLI convention: it goes to stderr in every output mode, including `--format quiet` and `--format json`, and is not subject to Python's warning filters or once-per-call-site dedup. The four pre-existing `warnings.warn("Ignoring --x ...")` calls in this file predate the `out` layer and are left alone to keep this diff on topic.
- The notice only fires on a real rename. An empty or `.` destination name means a directory, so `weights/`, `data/` and `.` are honoured silently rather than warned about.

## Tests

Two tests in `tests/test_cli.py`, following the harness used by the existing `--every` tests. The single-file branch had no coverage at all before.

- `test_upload_single_file_with_every` is parametrized over the default destination, an explicit destination folder, and a destination folder with a trailing slash. It asserts the full `CommitScheduler` call and that no rename warning is emitted.
- `test_upload_single_file_with_every_warns_on_rename` pins that a differing destination file name warns and that `path_in_repo` keeps only the directory part, never the destination file name or a slice of it.


hanouticelina (approved):
Thanks for flagging and fixing! I pushed a couple of updates to rewrite the tests and add a safeguard: one parametrized test instead of three, `out.warning` instead of `warnings.warn`, and the notice now only fires on a real rename (it also matched `weights/` before).
Also fixed the description: `PATH_IN_REPO` is positional, and absolute paths fail too

PR #4628, opened 2026-08-04
[Inference Endpoints] Allow container command/args without custom image + support them in update

gary149:
## Summary

Fast-follow to #4329, closing the two gaps that PR listed as out of scope:

1. **`--container-command` / `--container-args` no longer require `--custom-image`.** That gate was conservative CLI scoping, not an API constraint: `model.command` and `model.args` are top-level fields of the endpoint model payload and apply to managed images too (the Inference Endpoints UI exposes "Container Arguments" on managed engine deploys, and the [vLLM engine docs](https://huggingface.co/docs/inference-endpoints/en/engines/vllm) recommend passing engine flags that way). `--health-route` and `--port` still require `--custom-image`, since they only exist on the custom image payload.
2. **Container command/args can now be changed after deploy.** `hf endpoints update` had no flags for them and `update_inference_endpoint` had no parameters for them, even though the API's model-update schema accepts `command` / `args`. Changing engine flags on a running endpoint (e.g. enabling tool calling) previously meant using the UI or hand-writing the PUT payload.

## Changes

**CLI**
- `hf endpoints deploy`: validation now only gates `--health-route` / `--port` behind `--custom-image`.
- `hf endpoints update`: new `--container-command` / `--container-args` options (quoted string, `shlex`-split into tokens, same as deploy). Passing an empty string sends `[]`, which resets to the image default; omitting the flag leaves the current value untouched.

**SDK**
- `HfApi.update_inference_endpoint`: new `container_command` / `container_args` (`list[str]`) parameters mapping to `model.command` / `model.args`, mirroring `create_inference_endpoint`.
- `InferenceEndpoint.update`: same two parameters, passed through.

**Docs**: guides updated accordingly, CLI reference regenerated with `utils/generate_cli_reference.py --update`.

## Example

```bash
# Engine flags at deploy time, no custom image required anymore
hf endpoints deploy my-endpoint --repo gpt2 --framework pytorch \
  --accelerator cpu --instance-size x2 --instance-type intel-icl \
  --region us-east-1 --vendor aws \
  --container-args "--max-model-len 8192"

# Change engine flags on an existing endpoint (previously UI / raw API only)
hf endpoints update my-endpoint --container-args "--enable-auto-tool-choice --tool-call-parser lfm2"

# Reset to the image defaults
hf endpoints update my-endpoint --container-args ""
```

## Out of scope

- Selecting managed engine images from the CLI: the API models engines as `model.image` variants (`vLLM`, `sGLang`, `tgi`, ...) and its `framework` enum is `custom | pytorch | llamacpp`, so something like `--framework vllm` is not a valid deploy today (a vLLM deploy is `framework: "pytorch"` + `image: {"vLLM": {...}}`). A dedicated engine selector for `hf endpoints deploy` would be a separate feature. The related SDK gap (the `custom_image` image-dict passthrough does not recognize the newer `vLLM` / `vLLMNeuron` / `sGLang` / `hfServe` variants) is now tracked in #4629.
- `--custom-image` on `hf endpoints update` (the SDK already supports `custom_image` there).

## Tests

Mocked unit tests, no network:
- CLI: deploy forwards `container_args` without `custom_image`; `--port` / `--health-route` still error without `--custom-image`; `update` forwards split tokens; empty string resets to `[]`.
- SDK: `test_update_inference_endpoint_container_command_and_args_payload` asserts the PUT payload contains `model.command` / `model.args` and no `image` key, mirroring the existing create-side test.

---

> [!NOTE]
> **Medium Risk**
> Changes live endpoint configuration (engine flags trigger redeploys) and replace semantics can wipe catalog-tuned args if users overwrite without checking `describe` first; no auth or data-path changes.
> 
> **Overview**
> **Inference Endpoints** can now set and change container entrypoint flags on **managed** engine images (vLLM, SGLang, etc.), not only custom Docker deploys.
> 
> `hf endpoints deploy` no longer requires `--custom-image` for `--container-command` / `--container-args`; `--health-route` and `--port` still do. `hf endpoints update` gains the same two flags (quoted strings, `shlex`-split). Updates **replace** the full `model.command` / `model.args` payload—empty string clears to `[]`.
> 
> The Python API mirrors this: `HfApi.update_inference_endpoint` and `InferenceEndpoint.update` accept `container_command` and `container_args`, wired into the PUT body like create already did. Docs and CLI reference describe managed-image use and the replace semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de0fa434637c5e65b32efdc5c117b02eda2e516b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -392,6 +393,27 @@ def update(
...
+        ),
+    ] = None,
+    container_args: Annotated[
+        str | None,
+        Option(
+            "--container-args",
+            help=(
+                "Arguments appended to the container entrypoint, as a quoted string split into tokens "
+                '(e.g. "--enable-auto-tool-choice --tool-call-parser lfm2"). '
+                "Pass an empty string to reset to the image default."
```
hanouticelina:
```suggestion
                '(e.g. "--enable-auto-tool-choice --tool-call-parser lfm2"). Replaces the arguments currently '
                "set on the endpoint rather than adding to them, so include the ones you want to keep, run "
                "'hf endpoints describe NAME' first to see them. Pass an empty string to clear them."
```

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -201,7 +201,7 @@ def deploy(
             "--container-args",
             help=(
                 "Arguments appended to the container entrypoint, as a quoted string split into tokens "
-                '(e.g. "--tp 8 --reasoning-parser qwen3"). Requires --custom-image.'
+                '(e.g. "--tp 8 --reasoning-parser qwen3"). Not limited to --custom-image deployments.'
```
hanouticelina: (nit) removing "Requires --custom-image." already carries the message
```suggestion
                '(e.g. "--tp 8 --reasoning-parser qwen3").'
```

src/huggingface_hub/cli/inference_endpoints.py
```diff
@@ -392,6 +393,27 @@ def update(
             help="The task on which to deploy the model (e.g. 'text-classification').",
         ),
     ] = None,
+    container_command: Annotated[
+        str | None,
+        Option(
+            "--container-command",
+            help=(
+                "Override the container entrypoint, as a quoted string split into tokens "
+                '(e.g. "python -m sglang.launch_server"). Pass an empty string to reset to the image default.'
```
hanouticelina: `--container-command` / `--container-args` replace the whole list rather than merging into it, so someone adding one flag silently drops the others. Verified against the API with this branch, on a paused endpoint running a managed `{"huggingface": {}}` image:

```
update(container_args=["--foo", "bar"])   ->  args = ['--foo', 'bar']
update(container_args=["--baz"])          ->  args = ['--baz']          # replaced, not appended
update(container_args=[])                 ->  args unset
```

So "reset to the image default" is not wrong mechanically, the field really is cleared. The reason to reword is that it sounds restorative: endpoints deployed from the catalog already have tuned flags in `model.args`, and those came from the catalog, not from the image:

```bash
curl -s -H "Authorization: Bearer $HF_TOKEN" \
  https://endpoints.huggingface.co/api/catalog/recipe/MiniMaxAI/MiniMax-M2.1 \
  | jq '.endpointConfig.model.args'
```

```json
["--tool-call-parser", "minimax_m2",
 "--reasoning-parser", "minimax_m2_append_think",
 "--enable-auto-tool-choice", "--trust-remote-code"]
```

That is the payload `hf endpoints catalog deploy` POSTs. So running `hf endpoints update my-endpoint --container-args "--max-model-len 8192"` on such an endpoint leaves only `["--max-model-len", "8192"]`: `--trust-remote-code` is gone and the model no longer loads. Same for `zai-org/GLM-5-FP8` and `dots-studio/dots.ocr`, other catalog entries return `null`.

```suggestion
                '(e.g. "python -m sglang.launch_server"). Replaces the current value; '
                "pass an empty string to clear it."
```

src/huggingface_hub/cli/inference_endpoints.py
hanouticelina: (nit) worth surfacing the new flag in `--help` and the generated CLI reference:

```suggestion
@ie_cli.command(
    examples=[
        "hf endpoints update my-endpoint --min-replica 2",
        'hf endpoints update my-endpoint --container-args "--enable-auto-tool-choice --tool-call-parser lfm2"',
    ]
)
```

src/huggingface_hub/hf_api.py
```diff
@@ -9806,6 +9808,12 @@ def update_inference_endpoint(
             custom_image (`dict`, *optional*):
                 A custom Docker image to use for the Inference Endpoint. This is useful if you want to deploy an
                 Inference Endpoint running on the `text-generation-inference` (TGI) framework (see examples).
+            container_command (`list[str]`, *optional*):
+                Override the container entrypoint command (maps to `model.command` in the API payload). Typically
+                used together with `custom_image`.
```
```suggestion
                Override the container entrypoint command (maps to `model.command` in the API payload). Works with
                both managed engine images (e.g. vLLM, SGLang) and custom images.
```

src/huggingface_hub/hf_api.py
```diff
@@ -9414,8 +9414,8 @@ def create_inference_endpoint(
                 Override the container entrypoint command (maps to `model.command` in the API payload). Typically
                 used together with `custom_image`.
```
```suggestion
                Override the container entrypoint command (maps to `model.command` in the API payload). Works with
                both managed engine images (e.g. vLLM, SGLang) and custom images.
```

src/huggingface_hub/_inference_endpoints.py
```diff
@@ -302,6 +304,12 @@ def update(
             custom_image (`dict`, *optional*):
                 A custom Docker image to use for the Inference Endpoint. This is useful if you want to deploy an
                 Inference Endpoint running on the `text-generation-inference` (TGI) framework (see examples).
+            container_command (`list[str]`, *optional*):
+                Override the container entrypoint command (maps to `model.command` in the API payload). Typically
+                used together with `custom_image`.
```
hanouticelina:
```suggestion
                Override the container entrypoint command (maps to `model.command` in the API payload). Works with
                both managed engine images (e.g. vLLM, SGLang) and custom images.
```

docs/source/en/guides/cli.md
```diff
@@ -2324,6 +2324,13 @@ To deploy your own Docker image instead of a Hugging Face managed one, pass `--f
       --type authenticated
 ```
 
+`--container-args` and `--container-command` are not limited to custom images: they map to `model.args` and `model.command` in the API payload, which apply to managed images as well. You can also change them on an existing endpoint (pass an empty string to reset to the image default):
```
hanouticelina: double checked the claim with a live test 
```suggestion
`--container-args` and `--container-command` are not limited to custom images: they map to `model.args` and `model.command` in the API payload, which apply to managed images as well. On an existing endpoint both flags replace the current value rather than adding to it, so pass the full list you want and use an empty string to clear it. Endpoints deployed from the catalog already come with tuned engine flags, so check `hf endpoints describe` before overwriting them:
```

hanouticelina (approved):
Thanks!

PR #4630, opened 2026-08-04
[HfFileSystem] Fix bucket prefix collisions

lewtun:
## Issue

The Buckets API applies `prefix` lexically, but `HfFileSystem` needs path-component semantics. If a bucket contains `logs_existing/result.txt` but no `logs/`, listing `logs` returns the sibling. `HfFileSystem` then caches it under its real parent, and `exists(".../logs/new-result.txt")` raises `KeyError` instead of returning `False`; `ls(".../logs")` can also return the unrelated sibling.

## Reproducer

```python
from uuid import uuid4

import fsspec
from huggingface_hub import batch_bucket_files, create_bucket, delete_bucket, whoami


namespace = whoami()["name"]
bucket_id = f"{namespace}/hf-fs-prefix-repro-{uuid4().hex[:10]}"
create_bucket(bucket_id, private=True)

try:
    batch_bucket_files(
        bucket_id,
        add=[(b"test", "logs_existing/result.txt")],
    )

    url = f"hf://buckets/{bucket_id}/logs/new-result.txt"
    fs, path = fsspec.core.url_to_fs(url, skip_instance_cache=True)
    print(fs.exists(path))  # Expected: False; before this fix: KeyError
finally:
    delete_bucket(bucket_id, missing_ok=True)
```

## Fix

Filter bucket tree results in `_list_bucket_tree_with_folders()` so entries are accepted only when the requested prefix is empty, matches exactly, or is followed by `/`. This keeps `HfApi.list_bucket_tree()`'s documented lexical behavior unchanged while restoring normal filesystem behavior:

- `exists()` returns `False`
- `glob()` returns no matches
- `ls()` raises `FileNotFoundError`

Includes regression coverage using the existing `data/` fixture and colliding missing prefix `dat`.

## Validation

- Prefix-collision regression: `1 passed`
- Full bucket filesystem suite: `38 passed, 2 skipped` (2 transient staging 504 reruns)
- `make style`: passed
- `make quality`: passed

---

> [!NOTE]
> **Low Risk**
> Small, localized filter in bucket listing with a regression test; no auth or API contract changes.
> 
> **Overview**
> Fixes **bucket path prefix collisions** in `HfFileSystem` when `list_bucket_tree` returns lexically similar paths (e.g. listing `dat` also returns `data/...`).
> 
> In `_list_bucket_tree_with_folders`, results are now filtered so only entries that equal the prefix or sit under `prefix/` are kept—restoring normal filesystem behavior for **`exists`**, **`glob`**, and **`ls`** without changing the API’s lexical prefix behavior.
> 
> Adds **`test_prefix_collision`** against the shared `data/` fixture and the missing prefix `dat`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd1699683e760af4d208bd80a43671c7dad1f424. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lewtun:
The failing tests seem unrelated to this PR AFAICT

src/huggingface_hub/hf_file_system.py
```diff
@@ -663,6 +663,8 @@ def _list_bucket_tree_with_folders(
         out: list[BucketFile | BucketFolder] = []
 
         for bucket_entry in bucket_files:
+            if prefix and bucket_entry.path != prefix and not bucket_entry.path.startswith(f"{prefix}/"):
+                continue
```
hanouticelina, line 670: adding a small comment to document this line
```suggestion
            # The server matches `prefix` lexically, so listing "logs" also returns "logs_existing/...".
            # Filesystem semantics require path components => drop entries that aren't `prefix` itself
            # or below it.
            if prefix and bucket_entry.path != prefix and not bucket_entry.path.startswith(f"{prefix}/"):
                continue
```

hanouticelina (approved):
The fix looks good to me, thanks!

PR #4631, opened 2026-08-05
Add require_complete to snapshot_download

### LINKED ISSUE #4624: snapshot_download: no way to require that the cached snapshot was actually checked

Offline, `snapshot_download` returns the same thing whether it checked the snapshot for completeness or checked nothing at all, and I can't tell which happened.

Here are two ways to fetch one file of a two-file repo, then ask for the snapshot offline. Save as `repro.py` and `uv run repro.py`:

```python
# /// script
# requires-python = ">=3.9"
# dependencies = ["huggingface_hub==1.26.0"]
# ///
import os
import tempfile

from huggingface_hub import hf_hub_download, snapshot_download

REPO = "hf-internal-testing/dataset_with_data_files"  # data/train.txt, data/test.txt


def offline_snapshot(cache):
    try:
        path = snapshot_download(REPO, repo_type="dataset", cache_dir=cache, local_files_only=True)
        return sorted(os.listdir(os.path.join(path, "data")))
    except Exception as e:
        return f"{type(e).__name__}: {e}"


cache = tempfile.mkdtemp()
hf_hub_download(REPO, "data/train.txt", repo_type="dataset", cache_dir=cache)
print(f"{'hf_hub_download:':<17}{offline_snapshot(cache)}")

cache = tempfile.mkdtemp()
snapshot_download(REPO, repo_type="dataset", cache_dir=cache, allow_patterns="data/train.txt")
print(f"{'allow_patterns:':<17}{offline_snapshot(cache)}")
```

```
hf_hub_download: ['train.txt']
allow_patterns:  IncompleteSnapshotError: The cached snapshot for
'hf-internal-testing/dataset_with_data_files' (revision 'main', commit c19550d3...) is incomplete:
2 file(s) are missing (.gitattributes, data/test.txt). Outgoing traffic is disabled
('local_files_only=True'). Re-run the download with network access to complete the snapshot.
```

The same single file is on disk either way, and I get opposite answers. The second one is what I want. The first hands me a partial snapshot with no indication that anything is missing.

The difference is that a filtered `snapshot_download` caches the repo's full file listing and `hf_hub_download` doesn't, and [`_raise_if_incomplete_snapshot`](https://github.com/huggingface/huggingface_hub/blob/7d62d95fa01c7e1db9f8680cdf6f96b681f031d7/src/huggingface_hub/_snapshot_download.py#L543) says as much:

> If the tree listing is not cached we cannot tell, so we do nothing and the caller keeps returning the folder as-is.

That's a sensible default, but as a caller I can't find out which case I'm in, so I can't safely build on the result. A cache filled by `hf_hub_download` has no listing, and neither does one written before the tree cache existed.

A `require_complete` flag on `snapshot_download`, defaulting to `False`. With `require_complete=True`, `IncompleteSnapshotError` is raised when no tree listing is cached for the resolved commit, on top of the existing case where a cached listing shows files missing. Nothing changes when a listing is present, and nothing changes online, where the listing is fetched anyway.

I have this working, with a test, and would be glad to open a PR if the shape looks right to you.

I can work around it from outside: call `get_cached_repo_tree` to establish that a listing exists for the resolved commit, then call `snapshot_download` again so its own check runs against a listing now known to be there. That works, but only in that order. An online download writes the tree cache before it fetches the files the cache names ([`_snapshot_download.py:398-522`](https://github.com/huggingface/huggingface_hub/blob/7d62d95fa01c7e1db9f8680cdf6f96b681f031d7/src/huggingface_hub/_snapshot_download.py#L398-L522)), so checking first and reading the listing afterwards can see a listing published by a concurrent download moments after an unchecked snapshot was handed back.

Making the strict behaviour the default rather than putting it behind a flag. That would start failing for caches written before the tree cache existed, and for any cache filled file by file by `hf_hub_download`, so a flag seemed like the safer shape.

**Additional context**

I hit this trying to add offline dataset loading to `datasets`. It isn't a blocker there, but building from a partial snapshot means a training run on silently truncated data, and being able to rule that out would make the work much easier.

Same shape as #4500, which put `snapshot_path` on this error so `diffusers` could find the cached files.

`huggingface_hub` 1.26.0, Python 3.12, macOS.

eaubin:
## Summary

- add an opt-in `require_complete` check to `snapshot_download` and `HfApi.snapshot_download`
- raise `IncompleteSnapshotError` during local cache fallback when the requested snapshot cannot be shown to contain every requested file
- keep the current fallback by default so caches created by older clients or by `hf_hub_download` continue to work
- document the strict mode in the cache guide

Closes #4624.

Offline completeness checks depend on the cached tree listing for the resolved commit. A cache populated file by file with `hf_hub_download`, or written before tree listings were cached, may contain only part of a repository without a listing that lets `snapshot_download` detect the missing files.

`require_complete=True` makes that uncertainty an error:

```python
snapshot_download(
    "hf-internal-testing/dataset_with_data_files",
    repo_type="dataset",
    local_files_only=True,
    require_complete=True,
)
```

Manual check after caching only `data/train.txt` with `hf_hub_download`:

```python
import tempfile

from huggingface_hub import hf_hub_download, snapshot_download
from huggingface_hub.errors import IncompleteSnapshotError


cache_dir = tempfile.mkdtemp()
hf_hub_download(
    "hf-internal-testing/dataset_with_data_files",
    "data/train.txt",
    repo_type="dataset",
    cache_dir=cache_dir,
)
print("cached: data/train.txt")
try:
    snapshot_download(
        "hf-internal-testing/dataset_with_data_files",
        repo_type="dataset",
        cache_dir=cache_dir,
        local_files_only=True,
        require_complete=True,
    )
except IncompleteSnapshotError as error:
    print(f"{type(error).__name__}: {error}")
```

```text
cached: data/train.txt
IncompleteSnapshotError: The cached snapshot for 'hf-internal-testing/dataset_with_data_files' (revision 'main', commit c19550d35263090b1ec2bfefdbd737431fafec40) cannot be shown to be complete: no tree listing is cached for this commit, so which files the repository holds is unknown. Outgoing traffic is disabled ('local_files_only=True'). Re-run the download with network access to cache the listing.
```

The same call without `require_complete=True` keeps returning the cached folder, as before. 

“Complete” keeps the existing `snapshot_download` meaning: every file selected after `allow_patterns` and `ignore_patterns` exists locally. It does not verify file contents or checksums. Strict mode also rejects an existing non-empty `local_dir` when its revision cannot be resolved locally; otherwise that path could still return a snapshot that was never checked.

Doing the check inside `snapshot_download` avoids splitting the guarantee across `get_cached_repo_tree` and a later download call. A concurrent download can publish the tree cache before it finishes writing every file, so those two calls do not prove that the returned snapshot was checked against the listing they observed.

---

> [!NOTE]
> **Medium Risk**
> Changes error behavior on an opt-in flag only; default paths stay backward compatible, but callers enabling strict mode depend on tree-cache semantics for local/offline downloads.
> 
> **Overview**
> Adds opt-in **`require_complete=True`** to `snapshot_download` and `HfApi.snapshot_download` for stricter offline/local fallback behavior.
> 
> By default, when the Hub is unreachable and completeness cannot be verified (no cached `trees/` listing for the commit, or a non-empty `local_dir` whose revision cannot be resolved locally), the existing local snapshot is still returned. With **`require_complete=True`**, those cases raise **`IncompleteSnapshotError`** instead. When a tree listing exists, behavior is unchanged: missing files still raise **`IncompleteSnapshotError`**; the new flag only treats “cannot prove complete” as an error.
> 
> **`_raise_if_incomplete_snapshot`** centralizes the “no tree cache” path; **`IncompleteSnapshotError`** docs and the cache guide are updated. Tests cover unresolved `local_dir` and `hf_hub_download`-only caches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e55eaee5a78691aae81241086eb6120cb0a51772. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/manage-cache.md
+++ b/docs/source/en/guides/manage-cache.md
@@ -107,6 +107,8 @@ Both [`snapshot_download`] and [`hf_hub_download`] read this cache to avoid netw
 
 Because the cached file list describes exactly what a commit should contain, [`snapshot_download`] can also tell whether a local snapshot is complete. If the Hub cannot be reached (you are offline, the connection fails, or you passed `local_files_only=True`) and some expected files are missing from the local snapshot, [`snapshot_download`] raises [`~errors.IncompleteSnapshotError`] instead of returning a partial folder. Before this, an incomplete snapshot was returned silently, which could leave you working with missing files without knowing it. Files excluded by `allow_patterns` or `ignore_patterns` are not counted as missing. The exception exposes the path to the incomplete snapshot via its `snapshot_path` attribute, so you can still locate the partially cached files if needed.
 
+When [`snapshot_download`] falls back to an existing local snapshot but no file list is cached for its commit, completeness cannot be checked and the snapshot is returned as-is. An existing non-empty `local_dir` is also returned as-is if its revision cannot be resolved locally. Pass `require_complete=True` to raise [`~errors.IncompleteSnapshotError`] in either case.
+
 ### .no_exist (advanced)
 
 In addition to the `blobs`, `refs` and `snapshots` folders, you might also find a `.no_exist` folder
--- a/src/huggingface_hub/_snapshot_download.py
+++ b/src/huggingface_hub/_snapshot_download.py
@@ -50,6 +50,7 @@ def snapshot_download(
     force_download: bool = False,
     token: bool | str | None = None,
     local_files_only: bool = False,
+    require_complete: bool = False,
     allow_patterns: list[str] | str | None = None,
     ignore_patterns: list[str] | str | None = None,
     max_workers: int = 8,
@@ -75,6 +76,7 @@ def snapshot_download(
     force_download: bool = False,
     token: bool | str | None = None,
     local_files_only: bool = False,
+    require_complete: bool = False,
     allow_patterns: list[str] | str | None = None,
     ignore_patterns: list[str] | str | None = None,
     max_workers: int = 8,
@@ -100,6 +102,7 @@ def snapshot_download(
     force_download: bool = False,
     token: bool | str | None = None,
     local_files_only: bool = False,
+    require_complete: bool = False,
     allow_patterns: list[str] | str | None = None,
     ignore_patterns: list[str] | str | None = None,
     max_workers: int = 8,
@@ -125,6 +128,7 @@ def snapshot_download(
     force_download: bool = False,
     token: bool | str | None = None,
     local_files_only: bool = False,
+    require_complete: bool = False,
     allow_patterns: list[str] | str | None = None,
     ignore_patterns: list[str] | str | None = None,
     max_workers: int = 8,
@@ -183,6 +187,10 @@ def snapshot_download(
             The Hub endpoint to send the request to. Defaults to the value of `HF_ENDPOINT`.
         local_files_only (`bool`, *optional*, defaults to `False`):
             If `True`, do not download any files even if they are not in `cache_dir` or `local_dir`.
+        require_complete (`bool`, *optional*, defaults to `False`):
+            If `True`, when falling back to an existing local snapshot, raise
+            [`~errors.IncompleteSnapshotError`] instead of returning it if completeness cannot be checked
+            because no tree listing is cached for the commit or the revision cannot be resolved locally.
         allow_patterns (`list[str]` or `str`, *optional*):
             If provided, only files matching at least one pattern are downloaded.
         ignore_patterns (`list[str]` or `str`, *optional*):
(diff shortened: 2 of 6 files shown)


hanouticelina:
Thanks @eaubin for the repro! I'd rather not add a `require_complete` flag, since you can already
get this from outside with public API:

```python
    def verified_snapshot(repo_id, *, repo_type=None, revision=None, cache_dir=None, **kwargs):
        locate = dict(repo_type=repo_type, revision=revision, cache_dir=cache_dir)
        get_cached_repo_tree(repo_id, **locate)   # CachedRepoTreeNotFoundError if unverifiable
        return snapshot_download(repo_id, **locate, local_files_only=True, **kwargs)
```
which gives exactly the three outcomes you want: unverifiable, verified-incomplete, verified-complete. I checked, including with `allow_patterns`.

on the race: it fails safe in that order. If a concurrent download publishes the tree between the two calls, `snapshot_download` reads the new listing, sees missing files and raises. Only the reverse order is unsafe.A new public kwarg on `snapshot_download` is a permanent commitment, so for four lines of caller code I'd suggest handling this user-side for now. Closing this one

PR #4639, opened 2026-08-05
[Cache] Stop deleting snapshot files twice when deleting a revision

hanouticelina:
## Summary

Reported in https://huggingface.slack.com/archives/C02V5EA0A95/p1785884638328099 (private link)

- `scan_cache_dir` resolves each snapshot file to its blob with `file_path.resolve()`. For a file that isn't a symlink this returns the file itself, so it gets recorded as a blob living inside the snapshot dir. That happens for files created by the user inside a snapshot dir (in the report: `build/lib/...` and `*.egg-info/...` left behind by a `pip install` in a Space snapshot) and for copies made when symlinks aren't supported (Windows).
- `DeleteCacheStrategy.execute()` deletes snapshot dirs first and blobs last, so those files were already gone when their turn came, and each one logged a `FileNotFoundError` with `exc_info=True`.
- Fix: skip files whose blob path is inside the snapshot dir being deleted. Their size is still counted, so the reported freed size doesn't change.
- Also toned down the logs: the expected `FileNotFoundError` no longer logs a traceback, and the per-path `Delete <type>: <path>` line moved to `debug` (the CLI sets verbosity to `info`, so pruning a large cache printed one line per file). Both are still visible with `HF_DEBUG=1`.

Reproduced with a small handmade cache (a Space revision with `build/lib/...` files inside the snapshot dir): before the fix the output matched the report line for line, after it the output is a handful of lines and the deletion is still complete (detached snapshot, its stray files, its blob and the `.incomplete` file gone; `main` revision and shared blob untouched), with the same total freed size as before.

No new test: `test_scan_and_delete_cache_no_symlinks` already covers a cache whose snapshot files aren't symlinks, so I updated its expectations (it used to assert the in-snapshot copy was scheduled as a blob, and only checked that `execute()` didn't raise).

---

> [!NOTE]
> **Low Risk**
> Localized cache deletion and logging behavior; freed-size accounting is preserved and covered by an existing test.
> 
> **Overview**
> Fixes cache revision deletion when snapshot files are **not** symlinks into `blobs/` (no-symlink caches, user artifacts under a snapshot, etc.). `delete_revisions` no longer queues those paths for a separate blob delete—they’re removed with the snapshot tree—but their size still counts toward `expected_freed_size`.
> 
> **Logging:** per-path delete lines are `debug` instead of `info`, and missing-path `FileNotFoundError` is logged without a traceback so large prunes don’t flood the terminal.
> 
> `test_scan_and_delete_cache_no_symlinks` expectations were updated to match (blob set excludes in-snapshot copies; asserts freed size and snapshot removal after `execute()`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 118a9487af2e58e8a221161f19d868e7f339a0d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanouticelina:
reproduced the noisy logs and since it's a low impact addition, bypassing the review for this one

PR #4648, opened 2026-08-09
Fix metadata fetch on endpoints that return absolute redirects

### LINKED ISSUE #4637: snapshot_download fails with misleading LocalEntryNotFoundError when HF_ENDPOINT mirror 308-redirects to the Hub (x-repo-commit missing); relative-redirect follower drops query string

vollegrewar:
Fixes #4637 (defect 2)

## Problem

When `HF_ENDPOINT` is set to a mirror that answers `/resolve/...` requests with **absolute** 308 redirects (`Location` pointing to another host, e.g. the CDN or `huggingface.co`), the metadata fetch loop only follows redirects with an empty netloc (relative redirects). The absolute redirect response is returned as-is:

- `get_hf_file_metadata` looks for the `X-Repo-Commit` header on the 308 response → it's missing → `FileMetadataError("Distant resource does not seem to be on huggingface.co...")` is raised.
- `_raise_on_head_call_error` misclassifies that `FileMetadataError` as a connection problem → the user sees a misleading `LocalEntryNotFoundError("check your connection")`.

## Fix

1. **`src/huggingface_hub/utils/_http.py`** — `_httpx_follow_relative_redirects_with_backoff` now follows absolute redirects too, but **only when the target netloc belongs to a trusted set**:
   - `huggingface.co`
   - the host of the configured `HF_ENDPOINT`
   - `hub-staging.huggingface.co`
   
   This is a security constraint: `build_hf_headers()` unconditionally attaches the `Authorization` header, so following absolute redirects to arbitrary CDN hosts would leak the token. Redirect chains are capped with `_MAX_REDIRECTS = 10` to avoid loops.

2. **`src/huggingface_hub/file_download.py`** — `_raise_on_head_call_error` now re-raises `FileMetadataError` as-is instead of wrapping it into the "check your connection" `LocalEntryNotFoundError`, preserving the accurate diagnostic.

## Tests

- `tests/test_utils_http.py::TestFollowRedirects` — 4 cases: relative redirect regression, absolute redirect to a trusted hub host is followed, absolute redirect to a non-trusted host is refused, redirect loop is capped.
- `tests/test_file_download.py::TestRaiseOnHeadCallError` — 2 cases: `FileMetadataError` propagates unchanged; other network errors still get wrapped.

All new tests pass; full `tests/test_utils_http.py` suite passes (65 passed, 2 pre-existing environment-related failures unrelated to this change). `ruff check` and `ruff format` clean.

## Related

- #4644 fixes the relative-redirect branch (query-string loss); this PR complements it on the absolute-redirect branch and the error classification. No overlap.

---

> [!NOTE]
> **Medium Risk**
> Touches redirect-following and when auth headers are sent or stripped during metadata/download; behavior is constrained to an allowlist and HTTPS, but mistakes could break mirrors or leak tokens.
> 
> **Overview**
> Fixes metadata fetch failures when `HF_ENDPOINT` mirrors answer `/resolve` with **absolute** 308 redirects (e.g. back to `huggingface.co`), which previously left a bare 3xx response and triggered a misleading “check your connection” error (#4637).
> 
> **Redirect handling** in `_httpx_follow_relative_redirects_with_backoff` still follows relative redirects, and now also follows **HTTPS absolute** redirects whose host is in `_trusted_hub_netlocs()` (`ENDPOINT`, default prod/staging). CDN and other untrusted hosts, plus **plaintext `http`**, are never followed so the Hub `Authorization` header is not leaked. Chains are capped at 10 hops with a warning.
> 
> **Download path** aligns auth stripping with the same trust model: tokens are dropped only for cross-host targets that are not trusted Hub HTTPS endpoints.
> 
> **Errors:** `_raise_on_head_call_error` re-raises `FileMetadataError` instead of wrapping it in `LocalEntryNotFoundError`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e73e1f6418049793322196b240edaacd9cc22a76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vollegrewar:
Verified this PR against the live mirror repro from #4637 (hf-mirror.com still 308-redirects to huggingface.co as of today), clean HF_HOME, `HF_ENDPOINT=https://hf-mirror.com`:

- `get_hf_file_metadata("https://hf-mirror.com/Systran/faster-whisper-base/resolve/main/config.json")`: returns `commit_hash=ebe41f70d5b6dfa9166e2c581c45c9c0cfc57b66`, `etag=867cf1a0…` — on `main` this returns `commit_hash=None` because the absolute 308 is left unfollowed.
- `snapshot_download("Systran/faster-whisper-base", allow_patterns=["config.json"])`: succeeds, file downloaded — on `main` this raises `LocalEntryNotFoundError("…check your connection…")`.

The end-to-end mirror flow is fixed: metadata is fetched from the followed redirect target, the download completes, and the re-raised `FileMetadataError` path is only reached when metadata is genuinely absent from a successful response.

Puneet15556 (commented):
Thanks for putting this together — and for the direct answer on the error message, that's exactly the confirmation I was hoping for.

Agreed that a separate D2 PR would just duplicate what you have here, so I won't open one. On A vs B: your argument is fair — since A only surfaces when metadata is genuinely missing from a *successful* response, the precise type is more useful there than the generic one. I'd only note for the maintainers that it is technically a breaking change for anyone catching `LocalEntryNotFoundError` around downloads today, so it's their call whether that warrants a changelog note.

Two things on the redirect-following design, since you asked:

**1. The allowlist constrains the host but not the scheme.** `urlparse("http://huggingface.co/x").netloc` is `huggingface.co`, which is in `_TRUSTED_HUB_NETLOCS`, so a `Location: http://huggingface.co/...` would be followed and the `Authorization` header would go out over plaintext. Since the allowlist exists specifically to keep that header off untrusted hops, it seems worth also requiring `https` (or requiring the target scheme to match the original request's).

**2. `_TRUSTED_HUB_NETLOCS` is evaluated at import time.** `constants.ENDPOINT` is read once when `_http.py` is first imported, so a later change to it isn't reflected — including `tests/conftest.py`, which patches `constants` per test. Setting `HF_ENDPOINT` in the environment before import works fine, but computing the set inside the function (or via a small cached helper keyed on the current endpoint) would track the configured endpoint more reliably.

**3. Minor / coordination:** this branch is off `main`, so the relative-redirect line here is still `_replace(path=parsed_target.path)` — it will conflict with #4644. Whichever lands first, I'm happy to rebase mine on top of yours; no preference on ordering.

One more thought: when `_MAX_REDIRECTS` is hit the loop breaks and the 3xx is returned as the final response, which lands back in the same "no `x-repo-commit`" state this PR is fixing. A log line (or an explicit error) there might save someone the same debugging session.

Puneet15556 (commented):
I looked into the Bugbot finding above and I think it's a real interaction rather than a false positive — though it's narrower than the summary suggests, so worth pinning down before deciding.

The relevant heuristic in `file_download.py` is:

```python
# If url domain is different => we are downloading from a CDN => url is signed => don't send auth
if xet_file_data is None and url != metadata.location:
    url_to_download = metadata.location
    if urlparse(url).netloc != urlparse(metadata.location).netloc:
        headers.pop("authorization", None)
```

With a mirror endpoint, `url` stays `hf-mirror.com/...` while `metadata.location` now resolves to `huggingface.co/...` (via `location=response.headers.get("Location") or str(response.request.url)`), so the netlocs differ and the header is dropped — even though the target is a trusted hub host rather than a signed CDN URL. The two places now disagree on what "a different host" means: `_http.py` has learned about trusted hub netlocs, `file_download.py` still assumes different netloc ⇒ CDN.

Two things that make it narrower than it first reads:

- It's **not a regression this PR introduced in isolation** — before it, the mirror flow died earlier at `commit_hash=None`, so this branch was simply unreachable. The PR makes metadata resolution succeed, which is what exposes the next step.
- It only bites when `xet_file_data is None` **and** the repo actually needs auth. Your repro (`Systran/faster-whisper-base`, public) wouldn't show it; a gated or private repo over a mirror endpoint would 401.

A minimal fix that keeps the two in agreement would be to reuse the same notion of trust, e.g. keep the header when the target netloc is a trusted hub netloc:

```python
target_netloc = urlparse(metadata.location).netloc
if target_netloc != urlparse(url).netloc and target_netloc not in _trusted_hub_netlocs():
    headers.pop("authorization", None)
```

That preserves the CDN protection (signed URLs are on other hosts) while not stripping auth on a hub→hub hop. Worth a test with a gated repo id if one is available in the test fixtures, since that's the only case that distinguishes the two behaviours.

Happy to be wrong on the reachability if `xet_file_data` is set for most repos these days — I haven't verified how often that branch is taken in practice.

vollegrewar:
@Puneet15556 Addressed in 7797e1fb.

The download-side redirect policy now uses the same trusted-Hub policy as metadata resolution:

- https://hf-mirror.com/... -> https://huggingface.co/... keeps the Authorization header, so a private or gated repo does not lose authentication on a trusted Hub-to-Hub hop.
- Redirects to an external CDN still remove Authorization.
- Redirects to a trusted hostname over plaintext HTTP also remove Authorization.

Added unit regression coverage by mocking the metadata response and asserting header mutation for all three cases above. This covers the auth-header decision point directly; it is not a live gated-mirror integration test.

Validation:
- pytest tests/test_file_download.py -q: 68 passed, 4 skipped
- pytest tests/test_utils_http.py -k FollowRedirects -q: 6 passed
- ruff check, ruff format --check, and ty check src: passed

Could you please re-review?

Puneet15556 (approved):
Re-reviewed 7797e1f — the fix addresses the finding correctly, and I ran it locally rather than reading it only:

- `tests/test_file_download.py::TestMetadataDownloadHeaders` and `::TestRaiseOnHeadCallError` — 5 passed
- redirect-focused tests in `tests/test_utils_http.py` — 6 passed

I also exercised the three cases directly against `_get_metadata_or_catch_error` to confirm the behaviour matches the intent:

| `metadata.location` | Authorization |
| --- | --- |
| `https://huggingface.co/...` (trusted, https) | kept |
| `https://cdn.example.com/...` (untrusted) | removed |
| `http://huggingface.co/...` (trusted, plaintext) | removed |

Sharing the same `_trusted_hub_netlocs()` between `_http.py` and `file_download.py` is the right call — the two can't drift out of sync now.

**One remaining gap, non-blocking and pre-existing:** the condition short-circuits on the netloc comparison, so the scheme is never checked when the host is unchanged:

```python
if target.netloc != urlparse(url).netloc and (
    target.scheme != "https" or target.netloc not in _trusted_hub_netlocs()
):
```

With `url = https://hf-mirror.com/...` and `metadata.location = http://hf-mirror.com/...`, the netlocs are equal, the `and` short-circuits, and the token is sent over plaintext — I confirmed this with the same harness as the table above. This is the behaviour on `main` too, so it isn't something this PR introduced; I only mention it because "plaintext HTTP targets still remove it" isn't quite true for the same-host case, and the parametrised test only covers the cross-host variant.

If you want to close it, requiring https unconditionally would do it:

```python
if target.scheme != "https" or (
    target.netloc != urlparse(url).netloc and target.netloc not in _trusted_hub_netlocs()
):
```

Entirely your call whether that belongs here or in a follow-up — the PR as it stands is a clear improvement over the current behaviour and I'd be happy to see it merged either way.

vollegrewar:
Gentle nudge from the author — this PR has been ready for review for about three weeks (opened Aug 9), covering the absolute-redirect leg of #4637 (mirror 308 → Hub) with a trusted-host allowlist so the `Authorization` header never leaves hub hosts, plus the `FileMetadataError` re-raise (option A). @Puneet15556 reviewed and approved it after the Bugbot finding was addressed in `7797e1f`, and the tests pass locally. #4644 is the separate, smaller relative-redirect fix; both are independent and both are waiting on maintainer review. No rush at all — if you'd prefer option B for the exception type or any other adjustment before merge, I'm happy to change it.

vollegrewar:
Closing as superseded: #4739 (merged Aug 28) implements this fix upstream with the same approach — following redirects between Hub hosts via an allowlist, keeping the `Authorization` header on Hub-to-Hub hops while stripping it for CDN/plaintext redirects, and bounding the redirect loop. That resolves the absolute-redirect leg of #4637, which is now closed.

Thank you @Puneet15556 for the careful reviews — the https enforcement, the import-time evaluation catch, and confirming the Bugbot finding all made this substantially better, and the trust-boundary shape #4739 landed is essentially the one this review process converged on.

PR #4666, opened 2026-08-13
[CLI] Fix silent drop of --include when filenames are set

### LINKED ISSUE #4665: `hf download REPO --include A B C` silently drops A and exits 0
### Describe the bug

`hf download` accepts both a positional `filenames` argument and an `--include` option. When both are non-empty, `filenames` silently wins: the CLI emits a `UserWarning` and **exits 0**, so a partial download is reported as success.

This is easy to hit by accident, because `--include` consumes **one value per occurrence**. Extra shell words after it fall through to the positional argument, so

```
hf download REPO --include "A.gguf" "B.gguf" "C.gguf"
```

parses as `include=["A.gguf"]`, `filenames=["B.gguf", "C.gguf"]` — and `A.gguf` is then dropped. In a script the warning goes to stderr and the exit code says success, so the missing file surfaces much later as a confusing downstream failure.

### Reproduction

Tiny public repo, three files, ~1.1 KB total, runs in a couple of seconds:

```bash
hf download hf-internal-testing/tiny-random-gpt2 \
  --include config.json special_tokens_map.json tokenizer_config.json \
  --local-dir ./repro
```

```
.../cli/download.py:158: UserWarning: Ignoring `--include` since filenames have been explicitly set.
path=./repro

$ echo $?
0
$ ls repro
special_tokens_map.json
tokenizer_config.json
```

`config.json` was requested and never fetched, and the exit code is 0. There is no traceback to attach — nothing is raised, which is exactly the problem.

### Expected behavior

Either all requested files are downloaded, or the command fails. Silently narrowing the selection while reporting success is the part that bites.

Failing looks right to me, because the same class of conflict is **already** fatal a few lines earlier in the same function. In `src/huggingface_hub/cli/download.py` (`run_download`):

- **lines 143-153** — a subfolder argument (`art/`) combined with `--include`/`--exclude` raises `CLIError`, exit 1.
- **lines 155-160** — regular filenames combined with `--include`/`--exclude` calls `warnings.warn` and continues, exit 0.

Two mutually-exclusive-selection conflicts, ten lines apart, resolved opposite ways.

The `warnings.warn` branch dates to #1617, the original CLI download implementation. The `CLIError` branch arrived later in #3822, which added subfolder downloads and deliberately rejected mixing them with `--include`/`--exclude`. So the hard failure reads as the newer, considered convention and the warning as the branch nobody revisited.

Since the root cause is the one-value-per-occurrence parsing, an error here could also point at the working form (`--include "A.gguf" --include "B.gguf"`), which would have saved me the debugging.

One thing worth flagging: changing this is a behavior break, exit 0 becomes exit 1. What makes it feel safe is that `--include` cannot widen a set already pinned by explicit filenames, so in that combination the option has never actually done anything — there's no behavior to depend on, only a warning that the selection was narrowed. If you'd rather not take the break, keeping the warning but exiting non-zero would still remove the false success signal.

I have a patch and tests ready if that's useful, but I'll leave the call to you given you generally prefer to implement these yourselves.

### Environment

```
- huggingface_hub version: 1.28.0.dev0
- Platform: macOS-26.5-arm64-arm-64bit
- Python version: 3.12.13
- Running in iPython ?: No
- Running in notebook ?: No
- Has saved token ?: True
- Configured git credential helpers: osxkeychain
- Run by AI agent ?: Yes
- Installation method: unknown
- httpx: 0.28.1
- hf_xet: 1.6.0
- gradio: N/A
- tensorboard: N/A
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
```

Reproduced against `main` (installed editable as `1.28.0.dev0`); originally noticed on 1.27.0.

<details>
<summary>Related prior work</summary>

#4337 touched these exact lines to fix the warning's grammar and tighten `test_download_with_ignored_patterns` to assert the exact warning strings. It was explicitly copy-only with no behavior change, so any fix here would need to rewrite that test.

</details>

<sub>Refiling this cleanly — I closed #4663 while sorting out the report format. Same bug, no duplicate intended.</sub>

sohampanda000:
Fixes #4665. Would you mind assigning that issue to me? I don't have permission to self-assign.

## Summary

- `hf download REPO --include A B C` silently dropped `A` and exited 0. `--include` takes one value per occurrence, so `B` and `C` fell through to the positional `filenames` argument, which then took precedence.
- Raise `CLIError` instead of warning, matching what the subfolder branch ten lines above already does.
- The message points at the repeated-flag form, since that's the actual mistake being made.
- Added a note to the CLI guide — it presents filenames and `--include` as "two ways" without saying they're exclusive.

## Before

```
$ hf download hf-internal-testing/tiny-random-gpt2 \
    --include config.json special_tokens_map.json tokenizer_config.json \
    --local-dir ./repro
.../cli/download.py:158: UserWarning: Ignoring `--include` since filenames have been explicitly set.
path=./repro

$ echo $?
0
$ ls repro
special_tokens_map.json
tokenizer_config.json
```

`config.json` was asked for and never fetched, exit 0.

## After

```
$ hf download hf-internal-testing/tiny-random-gpt2 \
    --include config.json special_tokens_map.json tokenizer_config.json \
    --local-dir ./repro
Error: Cannot combine filenames ('special_tokens_map.json') with `--include`.
`--include` takes one pattern per occurrence: repeat it
(e.g. `--include "config.json" --include "special_tokens_map.json"`) or pass filenames only.
Hint: set HF_DEBUG=1 as environment variable for full traceback.

$ echo $?
1
$ ls repro
ls: repro: No such file or directory
```

Nothing is written — it fails before touching the filesystem.

And the form it suggests works:

```
$ hf download hf-internal-testing/tiny-random-gpt2 \
    --include config.json --include special_tokens_map.json --include tokenizer_config.json \
    --local-dir ./repro
$ ls repro
config.json
special_tokens_map.json
tokenizer_config.json
```

## Breaking change

Yes — an invocation that exited 0 now exits 1.

What makes it feel safe: `--include` can't widen a set already pinned by explicit filenames, so in that combination the option has never had any effect. There's no behavior to depend on, only a warning saying the selection got narrowed. The realistic way to reach this branch is the accident above, where the current behavior loses files.

If you'd rather not take the break, the smaller version is to keep the warning and just exit non-zero. Leaves two conflict styles in one function, but kills the false success. Happy to switch.

## Decisions

- **Error rather than merging filenames into `allow_patterns`.** Merging would guess at intent, and #3822 already set the precedent of erroring instead of guessing for this exact conflict.
- **`test_download_with_ignored_patterns` is replaced, not extended.** It asserted the old warning strings (tightened in #4337), so it can't survive. Replaced by `test_download_filenames_with_include_raises_error` and `..._exclude_raises_error`, named to match the adjacent subfolder tests. No coverage lost — `test_download_multiple_files` already covers filenames-only → `snapshot_download(allow_patterns=...)`. Kept to two tests per the AGENTS.md note about not testing every parameter permutation.
- **`import warnings` dropped from both files** because this removes its last use in each.
- Both new tests fail before the `download.py` change (`DID NOT RAISE CLIError`) and pass after.

`make style` and `make quality` are clean, `pytest tests/test_cli.py` is 314 passed. `uvx mypy src` reports 3 errors, all pre-existing in `_framework.py` and `_http.py`, none from this change.

## Disclosure

I used an AI agent for the investigation and the patch, run from the repo root so it picked up `AGENTS.md`. The terminal output above is real, captured from actual runs rather than written by hand. Happy to walk through any part of the diff or the reasoning behind it if anything looks off.

---

> [!NOTE]
> **Low Risk**
> CLI-only validation change: previously ignored flag combinations now exit with an error, which is intentional and limits silent partial downloads.
> 
> **Overview**
> **`hf download` no longer warns and continues when positional filenames are used together with `--include` or `--exclude`.** It raises `CLIError` instead, aligned with the existing subfolder + pattern checks, so mistaken invocations (e.g. extra tokens after a single `--include`) fail before any download instead of exiting 0 while dropping patterns.
> 
> Error text explains that patterns are **one per flag** and suggests repeating `--include` / `--exclude` or using filenames only; the exclude path avoids steering users toward `--include` when they meant to skip files.
> 
> The CLI guide now states that filenames and `--include`/`--exclude` are **mutually exclusive**, and that multiple patterns require repeated flags. Tests replace the old “ignored patterns” warning case with include/exclude error coverage; unused `warnings` imports are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49a44cc2854d1f7e35bbe866686c3dadc0cce77f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -303,6 +303,8 @@ Fetching 8 files: 100%|███████████████████
 /home/wauplin/.cache/huggingface/hub/models--stabilityai--stable-diffusion-xl-base-1.0/snapshots/462165984030d82259a11f4367a4eed129e94a7b
 ```
 
+The two approaches cannot be combined: passing filenames together with `--include` or `--exclude` raises an error. Note that `--include` takes a single pattern per occurrence, so repeat the option to pass several of them (`--include "*.json" --include "*.txt"`).
+
 ### Download a dataset or a Space
 
 The examples above show how to download from a model repository. To download a dataset or a Space, use the `--repo-type` option:
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -303,7 +303,7 @@ Fetching 8 files: 100%|███████████████████
 /home/wauplin/.cache/huggingface/hub/models--stabilityai--stable-diffusion-xl-base-1.0/snapshots/462165984030d82259a11f4367a4eed129e94a7b
 ```
 
-The two approaches cannot be combined: passing filenames together with `--include` or `--exclude` raises an error. Note that `--include` takes a single pattern per occurrence, so repeat the option to pass several of them (`--include "*.json" --include "*.txt"`).
+The two approaches cannot be combined: passing filenames together with `--include` or `--exclude` raises an error. Note that both options take a single pattern per occurrence, so repeat them to pass several (`--include "*.json" --include "*.txt"`).
 
 ### Download a dataset or a Space
 
--- a/src/huggingface_hub/cli/download.py
+++ b/src/huggingface_hub/cli/download.py
@@ -163,7 +163,9 @@ def download(
             if exclude is not None and len(exclude) > 0:
                 raise CLIError(
                     f"Cannot combine filenames ('{regular_filenames[0]}') with `--exclude`. "
-                    f'Please use `--include "{regular_filenames[0]}"` with `--exclude` instead.'
+                    "`--exclude` takes one pattern per occurrence: repeat it "
+                    f'(e.g. `--exclude "{exclude[0]}" --exclude "{regular_filenames[0]}"`), '
+                    f'or use `--include "{regular_filenames[0]}"` with `--exclude` to select files explicitly.'
                 )
 
         # Single file to download (not a subfolder): use `hf_hub_download`
--- a/src/huggingface_hub/cli/download.py
+++ b/src/huggingface_hub/cli/download.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Contains command to download files from the Hub with the CLI."""
 
-import warnings
 from typing import Annotated
 
 from huggingface_hub import constants
@@ -152,12 +151,20 @@ def download(
                     f'Please use `--include "{subfolders[0]}*"` with `--exclude` instead.'
                 )
 
-        # Warn user if patterns are ignored (only if regular filenames are provided)
+        # Error if regular filenames are combined with --include/--exclude
+        # Both select what to download, so silently letting filenames win would drop files the user asked for
         if len(regular_filenames) > 0:
             if include is not None and len(include) > 0:
-                warnings.warn("Ignoring `--include` since filenames have been explicitly set.")
+                raise CLIError(
+                    f"Cannot combine filenames ('{regular_filenames[0]}') with `--include`. "
+                    "`--include` takes one pattern per occurrence: repeat it "
+                    f'(e.g. `--include "{include[0]}" --include "{regular_filenames[0]}"`) or pass filenames only.'
+                )
             if exclude is not None and len(exclude) > 0:
-                warnings.warn("Ignoring `--exclude` since filenames have been explicitly set.")
+                raise CLIError(
+                    f"Cannot combine filenames ('{regular_filenames[0]}') with `--exclude`. "
+                    f'Please use `--include "{regular_filenames[0]}"` with `--exclude` instead.'
+                )
 
         # Single file to download (not a subfolder): use `hf_hub_download`
         if len(regular_filenames) == 1 and len(subfolder_patterns) == 0:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1205,7 +1205,7 @@ class TestDownloadImpl:
     @patch("huggingface_hub.cli.download.hf_hub_download")
     def test_download_filenames_with_include_raises_error(self, mock_download: Mock, mock_snapshot: Mock) -> None:
         """Test that combining filenames with --include raises an error instead of silently dropping the patterns."""
-        with pytest.raises(CLIError, match="Cannot combine filenames"):
+        with pytest.raises(CLIError, match=r"Cannot combine filenames.*`--include` takes one pattern per occurrence"):
             download(
                 repo_id="author/model",
                 filenames=["README.md", "config.json"],
@@ -1219,7 +1219,9 @@ class TestDownloadImpl:
     @patch("huggingface_hub.cli.download.hf_hub_download")
     def test_download_filenames_with_exclude_raises_error(self, mock_download: Mock, mock_snapshot: Mock) -> None:
         """Test that combining filenames with --exclude raises an error instead of silently dropping the patterns."""
-        with pytest.raises(CLIError, match="Cannot combine filenames"):
+        # Guidance must be about `--exclude` itself: an earlier version suggested converting the filename into
+        # `--include`, which inverts intent when the user meant to skip that file.
+        with pytest.raises(CLIError, match=r"Cannot combine filenames.*`--exclude` takes one pattern per occurrence"):
             download(
                 repo_id="author/model",
                 filenames=["README.md", "config.json"],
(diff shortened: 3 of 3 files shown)


sohampanda000:
Good catch from both bots — this was a real inversion, not just awkward wording.

The `--exclude` message reused the subfolder template, which only makes sense when the positional argument really was meant as a file. For the accident this PR is about, it gives backwards advice:

```
$ hf download REPO --exclude "*.bin" "*.h5"
Error: Cannot combine filenames ('*.h5') with `--exclude`.
Please use `--include "*.h5"` with `--exclude` instead.
```

`*.h5` fell through to `filenames`, and the suggestion is to *include* the file the user was trying to skip. Following it downloads exactly the wrong thing.

Fixed in e597cd0d — the `--exclude` branch now leads with the same one-value-per-occurrence explanation as `--include`, and keeps the `--include` fallback as a second option instead of the only advice:

```
$ hf download REPO --exclude "*.bin" "*.h5"
Error: Cannot combine filenames ('*.h5') with `--exclude`. `--exclude` takes one pattern
per occurrence: repeat it (e.g. `--exclude "*.bin" --exclude "*.h5"`), or use
`--include "*.h5"` with `--exclude` to select files explicitly.
```

Verified the suggested form does what the user meant:

```
$ hf download hf-internal-testing/tiny-random-gpt2 --exclude "*.bin" --exclude "*.h5" --local-dir ./out
$ ls out
config.json  merges.txt  model.safetensors  special_tokens_map.json
tokenizer.json  tokenizer_config.json  vocab.json
```

`pytorch_model.bin` and `tf_model.h5` are correctly absent.

I also fixed the same asymmetry in the CLI guide sentence, which said only `--include` takes one pattern per occurrence when it's true of both.

`make quality` clean, `pytest tests/test_cli.py` still 314 passed.

sohampanda000:
@Wauplin @hanouticelina — this is ready for review whenever you have a moment. Tagging you two since you own the CLI; happy to be redirected if someone else is a better fit.

Both bot findings are addressed and the branch is up to date with `main`. Current state:

- `make style` and `make quality` clean, `pytest tests/test_cli.py` 314 passed
- The two new tests are unmarked, and pass under both `-m "not xet"` and `-m "not no_xet"` — they mock `hf_hub_download`/`snapshot_download`, so nothing touches Xet
- Tests fail before the `download.py` change and pass after

One thing I'd genuinely like your call on, since I can argue it either way: this turns an exit-0 invocation into exit 1, which is a breaking CLI change. I went with the break because `--include` can't widen a set already pinned by explicit filenames, so in that combination the option has never had any effect — there's no behavior to depend on, only a warning that the selection got narrowed. But if you'd rather not take it, keeping the warning and just exiting non-zero removes the false-success signal with no break. Say the word and I'll switch.

If you'd prefer to take this one in-house per CONTRIBUTING, that's completely fine — close it and the issue stands on its own.

Also, could one of you assign #4665 to me? I don't have permission to self-assign.

hanouticelina:
Hi @sohampanda000, it isn't actually silent though: `--include` already warns when filenames are set ("Ignoring `--include` since filenames have been explicitly set"), and that warning is in your own before/after output.

So this is turning an existing warning into a hard error for a usage mistake, which is a breaking change for anyone scripting it. Did you hit this yourself, or find it reading the code? Without a real report I'd rather keep the warning.

Closing this one and #4665, thanks anyway!

sohampanda000:
Fair on both counts.

To answer the question directly: I hit it, before reading any of the code. Pulling three GGUF shards from one repo on 1.27.0/macOS with `--include "A.gguf" "B.gguf" "C.gguf"` — the first file was the one that went missing, and I only noticed later when the thing consuming it failed. Opening `download.py` came after, to work out why.

You're right that "silently" was the wrong word, though. The warning is there, it's in my own output, and I shouldn't have led with a title my own repro contradicts. What actually got me was the exit code — the warning goes to stderr and my script only checked status. That's a narrower complaint than the one I filed.

Understood on keeping the warning. Thanks for the explanation rather than just a close.

PR #4668, opened 2026-08-14
[CLI] Accept a Hugging Face URL in `hf download`

### LINKED ISSUE #4099: hf download with URL
It would be useful if `uv run hf download` would support passing direct URL to a specific file in private/public huggingface repository rather than specifying repo type / path

NoiceHax:
`parse_hf_uri` already handles Hugging Face web URLs (blob, resolve, raw and tree routes) and turns them into the canonical `hf://` form. But `hf download` only ran the repo id through it when the string started with `hf://`, so pasting a file URL from the browser fell through to the plain repo id path and failed with an opaque `HFValidationError`.

This branches on the URL shape as well, so a pasted URL resolves to a repo type, revision and file path just like an `hf://` URI. The `--repo-type` / `--revision` conflict errors now say "Hub URI or URL" since they cover both. Docs and the CLI examples get a URL example.

Tests are in tests/test_cli.py::TestDownloadImpl: a blob URL and a resolve URL with `?download=true`, a repo page URL that pulls the whole repo, and a non-Hub host that still gets rejected. They fail on main and pass with this change.

Closes #4099

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -326,7 +326,7 @@ The examples above show how to download from the latest commit on the main branc
 ...
 ```
 
-### Use an hf:// URI
+### Use an hf:// URI or a URL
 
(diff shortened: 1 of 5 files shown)


hanouticelina:
Hi, thanks for the PR, but i'm going to close it. This is a new feature rather than a fix and we'd like to handle and discuss  the design internally, keeping the issue open for now

PR #4671, opened 2026-08-14
[Inference Endpoints] Support managed engine images in `custom_image`

### LINKED ISSUE #4629: INFERENCE_ENDPOINT_IMAGE_KEYS is missing newer engine image variants (vLLM, vLLMNeuron, sGLang, hfServe)
### Context

Agent-driven deploys mostly use custom images anyway: an agent can research the best engine version and flags for a given model (day-0 releases, forks) and iterate on the config itself, which managed images can't match. That is probably why nobody hit this bug before, the custom path works fine and it's what programmatic users reach for. Still, the managed variants are valid API payloads and the SDK silently corrupts them, worth a one-line fix (I just hope it won't steer agents towards outdated engine versions instead of picking the best one themselves :)).

### Describe the bug

`create_inference_endpoint` accepts a full image dict via `custom_image` and forwards it as-is when its first key is a known image type:

https://github.com/huggingface/huggingface_hub/blob/7d62d95f/src/huggingface_hub/hf_api.py#L9534-L9539

The known-keys list is:

```python
INFERENCE_ENDPOINT_IMAGE_KEYS = [
    "custom",
    "huggingface",
    "huggingfaceNeuron",
    "llamacpp",
    "tei",
    "tgi",
    "tgiNeuron",
]
```

The Inference Endpoints API ([OpenAPI spec](https://api.endpoints.huggingface.cloud/)) has since gained more `model.image` variants: `vLLM`, `vLLMNeuron`, `sGLang` and `hfServe`. Because those keys are missing from the list, a valid image dict for any of them is silently mis-wrapped into a `custom` image and the request fails (or creates the wrong thing):

```python
api.create_inference_endpoint(..., custom_image={"vLLM": {"tensorParallelSize": 8}})
# sends: "image": {"custom": {"vLLM": {"tensorParallelSize": 8}}}   <- invalid
# expected: "image": {"vLLM": {"tensorParallelSize": 8}}
```

This means the SDK currently has no way to create an endpoint running the managed vLLM / SGLang engine images, even though the API supports them (they are what the web UI deploys for those engines).

Related: `update_inference_endpoint` wraps unconditionally (`payload["model"]["image"] = {"custom": custom_image}`), so on update even the listed variants (`tgi`, `tei`, ...) cannot be set, only `custom`.

### Suggested fix

- Add `"vLLM"`, `"vLLMNeuron"`, `"sGLang"`, `"hfServe"` to `INFERENCE_ENDPOINT_IMAGE_KEYS` (the casing is the literal JSON key in the API schema).
- Optionally reuse the same first-key check in `update_inference_endpoint` instead of always wrapping in `custom`.

Happy to open a PR for this. Found while working on #4628.

### Reproduction

```python

api = HfApi()
api.create_inference_endpoint(
    "vllm-managed",
    repository="meta-llama/Llama-3.1-8B-Instruct",
    framework="pytorch",
    task="text-generation",
    accelerator="gpu",
    vendor="aws",
    region="us-east-1",
    instance_size="x1",
    instance_type="nvidia-a10g",
    custom_image={"vLLM": {}},
)
```

Inspect the request payload: `model.image` is `{"custom": {"vLLM": {}}}` instead of `{"vLLM": {}}`.

### System info

```
huggingface_hub main (1.27.0.dev0), also affects latest release
```

hanouticelina:
## Summary

Fixes #4629.

`INFERENCE_ENDPOINT_IMAGE_KEYS` was a stale allowlist: the Inference Endpoints API has since gained `vLLM`, `vLLMNeuron`, `sGLang` and `hfServe` `model.image` variants. Since anything whose first key isn't in the list gets wrapped as `{"custom": ...}`, passing a valid engine image dict silently produced an invalid payload.

Refreshing the list would just push the same bug to the next engine the API ships, so the allowlist is gone and `url` discriminates instead:

- every `model.image` variant except `huggingface` inherits `BaseContainer`, which requires `url`, and no variant is named `url`. A top-level `url` therefore means "flat custom container" and nothing else. `_build_endpoint_image_payload` wraps those in `{"custom": ...}` and forwards everything else untouched. Engines added to the API later work without a `huggingface_hub` release, and a typo like `vllm` instead of `vLLM` gets reported by the API rather than silently turned into a malformed custom container.
- the same helper is now used by `update_inference_endpoint`, which used to wrap unconditionally in `{"custom": ...}`. Even the already-listed variants (`tgi`, `tei`, ...) could not be set on update, and passing `{"custom": {...}}` produced a double-wrapped `{"custom": {"custom": {...}}}`. `EndpointModelUpdate.image` is the same `EndpointModelImage` union, so the same rule applies.

Flat dicts (`{"url": ..., "port": ...}`) keep being wrapped as `custom`, so no behavior change there.

## Breaking change

`constants.INFERENCE_ENDPOINT_IMAGE_KEYS` is removed. It is what made this bug possible and there is nothing left to allow-list. It was never exported at the package root, never in `__all__`, and never documented in the API reference, but it has been reachable as `huggingface_hub.constants.INFERENCE_ENDPOINT_IMAGE_KEYS` since #3076, so anyone reading it directly will get an `AttributeError`.

## Payload compatibility

No working call changes: every input that produced a payload the API accepted before produces a byte-identical payload now. Checked against the live API. `PUT` deserializes the request body before it looks up the endpoint, so the rejection cases below were probed against a non-existent endpoint name, with no side effects.

| `custom_image` | before | after |
| --- | --- | --- |
| flat, with `url` | accepted | accepted, identical |
| `{"tgi": ...}`, `{"custom": ...}`, engine variants | create: accepted / update: **broken** | accepted on both |
| flat, no `url` | `400 model.image.custom: missing field 'url'` | `400 model.image: unknown variant 'port', expected one of ...` |
| `{}` | create: `StopIteration` / update: `400 missing field 'url'` | `400 model.image: expected value` |
| mistyped `{"vllm": ...}` | `400 model.image.custom: missing field 'url'` | `400 unknown variant 'vllm', expected one of ... 'vLLM' ...` |

That last row is the practical upside of dropping the allowlist: the API answers with every valid variant, correct casing included, instead of a `url` error that says nothing about the actual mistake.

Also confirmed while checking: `{"vLLM": {}}` returns `400 model.image.vLLM: missing field 'url'`. Engine variants do not get a server-side default image, so the docs are right to make callers pass `url`.

### One regression worth knowing about

Passing a variant key *and* a top-level `url` in the same dict, i.e. mixing the keyed form with the flat form:

```python
custom_image={"tgi": {...}, "url": "ghcr.io/huggingface/text-generation-inference:latest"}
```

On create this used to go out unwrapped and get a `400`. It now gets wrapped as `{"custom": {"tgi": {...}, "url": ...}}`, which the API **accepts**: `CustomContainer` only requires `url`, and unknown fields inside a container are ignored, whereas they are rejected one level up at `image`. So the endpoint deploys as a plain custom container and the `tgi` block is silently dropped. On update this is not new, since update already wrapped everything unconditionally.

Nothing that worked stops working, because the old payload was rejected too. What is lost is the error. I don't think it's worth guarding client-side: `CustomContainer.credentials` is itself a dict, so `{"url": ..., "credentials": {...}}` is structurally identical to the bad input, and telling them apart means hardcoding container field names, which is the same staleness this PR removes.

### Before

```python
>>> api.create_inference_endpoint(..., custom_image={"vLLM": {"url": "vllm/vllm-openai:v0.23.0", "tensorParallelSize": 8}})
# model.image == {"custom": {"vLLM": {"url": "vllm/vllm-openai:v0.23.0", "tensorParallelSize": 8}}}   <- invalid
```

### After

```python
>>> api.create_inference_endpoint(..., custom_image={"vLLM": {"url": "vllm/vllm-openai:v0.23.0", "tensorParallelSize": 8}})
# model.image == {"vLLM": {"url": "vllm/vllm-openai:v0.23.0", "tensorParallelSize": 8}}

>>> api.update_inference_endpoint("my-endpoint", custom_image={"sGLang": {"url": "lmsysorg/sglang:v0.5.2"}})
# model.image == {"sGLang": {"url": "lmsysorg/sglang:v0.5.2"}}

# an engine this version has never heard of is forwarded too
>>> api.create_inference_endpoint(..., custom_image={"futureEngine": {"url": "some.registry/future:v1"}})
# model.image == {"futureEngine": {"url": "some.registry/future:v1"}}
```

## Tests

The normalization rule is a pure function now, so it is unit-tested directly. The mocked tests only cover what the helper can't: that the image reaches `model.image`, that `custom_image=None` still defaults to `{"huggingface": {}}`, and the #4629 update regression.

```
tests/test_hf_api.py::test_build_endpoint_image_payload[flat_dict] PASSED
tests/test_hf_api.py::test_build_endpoint_image_payload[keyed_custom] PASSED
tests/test_hf_api.py::test_build_endpoint_image_payload[keyed_engine] PASSED
tests/test_hf_api.py::test_build_endpoint_image_payload[unknown_variant] PASSED
tests/test_hf_api.py::test_create_inference_endpoint_custom_image_payload[no_custom_image] PASSED
tests/test_hf_api.py::test_create_inference_endpoint_custom_image_payload[custom_image] PASSED
tests/test_hf_api.py::test_update_inference_endpoint_custom_image_payload PASSED

7 passed, 342 deselected in 0.08s
```

Checked against `main`: the create case and the update regression both fail there. Checked against the intermediate allowlist commit: `unknown_variant` fails there, which is what pins the removal.

### End to end against production

Created a real endpoint with an engine-keyed image, switched it to another engine, switched it to a flat custom container, then deleted it. The interesting part is what the API echoes back for `model.image`:

```
1) CREATE with engine-keyed vLLM image
   server echoed model.image = {"vLLM": {"healthRoute": "/health", "port": 8000, "url": "vllm/vllm-openai:v0.11.0", "tensorParallelSize": 1}}
   -> kept as vLLM (not wrapped in custom): True

2) UPDATE to engine-keyed sGLang image
   server echoed model.image = {"sGLang": {"healthRoute": "/health", "port": 30000, "url": "lmsysorg/sglang:v0.5.2rc1-cu126"}}
   -> switched to sGLang, not double-wrapped: True

3) UPDATE to a flat custom container (must still wrap in `custom`)
   server echoed model.image = {"custom": {"healthRoute": "/health", "port": 8000, "url": "vllm/vllm-openai:v0.11.0"}}
   -> wrapped as custom: True

4) DELETED zz-pr4671-check
```

`tensorParallelSize` surviving the round trip is the actual #4629 fix: on `main` that dict lands inside `{"custom": {...}}` and the engine option never reaches vLLM.

## Note

Left the CLI alone: `hf endpoints deploy --custom-image` only takes an image URL and builds a `custom` container, so it's unaffected. Exposing engine images from the CLI would be a separate (bigger) design question.

---

> [!NOTE]
> **Medium Risk**
> Changes how inference endpoint create/update API payloads are built; wrong wrapping could break deployments, but behavior is narrowed with clear rules and targeted tests.
> 
> **Overview**
> Fixes incorrect `model.image` payloads when deploying Inference Endpoints with **managed engine images** (`vLLM`, `sGLang`, `tgi`, etc.) via `custom_image`.
> 
> **Create** no longer relies on a hardcoded `INFERENCE_ENDPOINT_IMAGE_KEYS` allowlist (removed from `constants.py`). Shared helper `_build_endpoint_image_payload` wraps only **flat** container dicts that include a top-level `url` as `{"custom": ...}`; engine-keyed dicts and unknown future variants are sent to the API unchanged.
> 
> **Update** (`update_inference_endpoint`) now uses the same helper instead of always setting `{"custom": custom_image}`, which had broken engine images and could double-wrap `{"custom": {...}}`.
> 
> Docs and `custom_image` docstrings describe engine-keyed usage (e.g. vLLM with `tensorParallelSize`). Tests cover the helper, create/update payloads, and the #4629 update regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a45636dff4fabec83a795384574296726fa4dda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -102,22 +102,6 @@ INFERENCE_ENDPOINT = os.environ.get("HF_INFERENCE_ENDPOINT", "https://api-infere
 INFERENCE_ENDPOINTS_ENDPOINT = "https://api.endpoints.huggingface.cloud/v2"
 INFERENCE_CATALOG_ENDPOINT = "https://endpoints.huggingface.co/api/catalog"
 
-# Variants of the `model.image` union, i.e. the managed engine images plus `custom`.
-# See https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-
-INFERENCE_ENDPOINT_IMAGE_KEYS = [
-    "custom",
-    "hfServe",
-    "huggingface",
-    "huggingfaceNeuron",
-    "llamacpp",
-    "sGLang",
-    "tei",
-    "tgi",
-    "tgiNeuron",
-    "vLLM",
-    "vLLMNeuron",
-]
-
 # Proxy for third-party providers
 INFERENCE_PROXY_TEMPLATE = "https://router.huggingface.co/{provider}"
 
--- a/src/huggingface_hub/constants.py
+++ b/src/huggingface_hub/constants.py
@@ -102,15 +102,20 @@ INFERENCE_ENDPOINT = os.environ.get("HF_INFERENCE_ENDPOINT", "https://api-infere
 INFERENCE_ENDPOINTS_ENDPOINT = "https://api.endpoints.huggingface.cloud/v2"
 INFERENCE_CATALOG_ENDPOINT = "https://endpoints.huggingface.co/api/catalog"
 
+# Variants of the `model.image` union, i.e. the managed engine images plus `custom`.
 # See https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-
 INFERENCE_ENDPOINT_IMAGE_KEYS = [
     "custom",
+    "hfServe",
     "huggingface",
     "huggingfaceNeuron",
     "llamacpp",
+    "sGLang",
     "tei",
     "tgi",
     "tgiNeuron",
+    "vLLM",
+    "vLLMNeuron",
 ]
 
 # Proxy for third-party providers
--- a/docs/source/en/guides/inference_endpoints.md
+++ b/docs/source/en/guides/inference_endpoints.md
@@ -100,6 +100,24 @@ By default the Inference Endpoint is built from a docker image provided by Huggi
 
 The value to pass as `custom_image` is a dictionary containing a url to the docker container and configuration to run it. For more details about it, checkout the [Swagger documentation](https://api.endpoints.huggingface.cloud/#/v2%3A%3Aendpoint/create_endpoint).
 
+`custom_image` also accepts the managed engine images maintained by Hugging Face, by keying the dictionary with the engine name instead of describing a container. The supported keys are listed in `constants.INFERENCE_ENDPOINT_IMAGE_KEYS` (`vLLM`, `sGLang`, `tgi`, `tei`, `llamacpp`, ...):
+
+```python
+# Start an Inference Endpoint running the managed vLLM image
+>>> endpoint = create_inference_endpoint(
+...     "llama-3-1-8b-instruct-vllm",
+...     repository="meta-llama/Llama-3.1-8B-Instruct",
+...     framework="pytorch",
+...     task="text-generation",
+...     accelerator="gpu",
+...     vendor="aws",
+...     region="us-east-1",
+...     instance_size="x1",
+...     instance_type="nvidia-a10g",
+...     custom_image={"vLLM": {}},
+... )
+```
+
 For containers that need a custom entrypoint or runtime flags, pass `container_command` and/or `container_args` (each a list of tokens). They map to `model.command` and `model.args` in the API payload. They are not tied to custom images: managed engine images (e.g. vLLM, SGLang) accept engine flags through `container_args` as well. The same is available from the CLI via `hf endpoints deploy ... --container-command "..." --container-args "..."`.
 
 ### Get or list existing Inference Endpoints
--- a/docs/source/en/guides/inference_endpoints.md
+++ b/docs/source/en/guides/inference_endpoints.md
@@ -100,7 +100,7 @@ By default the Inference Endpoint is built from a docker image provided by Huggi
 
 The value to pass as `custom_image` is a dictionary containing a url to the docker container and configuration to run it. For more details about it, checkout the [Swagger documentation](https://api.endpoints.huggingface.cloud/#/v2%3A%3Aendpoint/create_endpoint).
 
-`custom_image` also accepts the engine images managed by Hugging Face, by keying the dictionary with the engine name instead of leaving it flat. The supported keys are listed in `constants.INFERENCE_ENDPOINT_IMAGE_KEYS` (`vLLM`, `sGLang`, `tgi`, `tei`, `llamacpp`, ...) and each engine takes its own tuning options on top of the container ones:
+`custom_image` also accepts the engine-specific container types supported by the API, by keying the dictionary with the engine name (`vLLM`, `vLLMNeuron`, `sGLang`, `tgi`, `tgiNeuron`, `tei`, `llamacpp`, `hfServe`, ...) instead of leaving it flat. Each engine takes the usual container fields (`url`, `port`, `healthRoute`) plus its own tuning options. Any dict without a top-level `url` is forwarded to the API untouched, so engines added to the API later work without upgrading `huggingface_hub`:
 
 ```python
 # Start an Inference Endpoint running the vLLM engine image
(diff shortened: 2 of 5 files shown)


src/huggingface_hub/constants.py
```diff
@@ -102,17 +102,6 @@ def _as_int(value: str | None) -> int | None:
...
-INFERENCE_ENDPOINT_IMAGE_KEYS = [
-    "custom",
-    "huggingface",
-    "huggingfaceNeuron",
-    "llamacpp",
-    "tei",
-    "tgi",
-    "tgiNeuron",
-]
-
```
hanouticelina, line 115: I asked Opus to remove this list because #4629 exists because `INFERENCE_ENDPOINT_IMAGE_KEYS` was a hardcoded mirror of a server side union. the next time Endpoints ships an engine, the exact same silent bug reappears for everyone who hasn't upgraded the lib and it's silent, because an unknown key gets transformed into` {"custom": {...}}` instead of passed through and rejected with a clear message. yes, it's a breaking change since it's exposed in the `huggingface_hub.constants module` but I think it's very low impact an,d it's not in the docs or exported at package root.

hanouticelina (commented):
✅ all good!

PR #4672, opened 2026-08-14
[Inference Endpoints] Add `hf endpoints hardware` to list available instances

### LINKED ISSUE #4638: [Inference Endpoints] Add hf endpoints hardware to list available instances
## Problem

`hf endpoints deploy` requires five hardware flags — `--accelerator`, `--instance-type`, `--instance-size`, `--vendor`, `--region` — all free text, and there is no way to learn valid values from the CLI: `--help` only shows example literals (`'cpu'`, `'x4'`, `'intel-icl'`, `'aws'`), and `hf endpoints catalog` lists models, not hardware. The only sources today are the web UI picker or calling the platform API by hand. This is item 1 ("discovery wall") of #4314: neither a human nor an agent can build a valid `deploy` command from the CLI alone.

## The server API already exists and is public

`GET https://api.endpoints.huggingface.cloud/v2/provider` is documented in the [Endpoints API reference](https://api.endpoints.huggingface.cloud/) and marked **[PUBLIC]** in the OpenAPI spec. It returns the full vendor → region → compute matrix; `GET /v2/provider/{namespace}` returns the same with the caller's actual quota numbers.

```bash
curl -s https://api.endpoints.huggingface.cloud/v2/provider | jq '.vendors[0].regions[0].computes[0]'
```

```json
{
  "id": "aws-us-east-1-nvidia-l4-x1",
  "accelerator": "gpu",
  "instanceType": "nvidia-l4",
  "instanceSize": "x1",
  "architecture": "Nvidia L4",
  "gpuMemoryGb": 24,
  "pricePerHour": 0.8,
  "status": "available",
  "quota": { "maxAccelerators": 0, "usedAccelerators": 0 }
}
```

So this is client-side wiring of an existing public endpoint.

## Proposal

**SDK:** `HfApi.list_endpoint_hardware(namespace=None, vendor=None, region=None, accelerator=None, instance_type=None, include_unavailable=False)` returning a list of `EndpointHardware` dataclasses (the API compute object plus `vendor`/`region`). With `namespace` + auth it uses `/v2/provider/{namespace}` so quota numbers are real.

Naming note: deliberately *not* "provider" in the public name despite the API path — "Inference Providers" is a different product. "Hardware" matches the existing `deploy --help` wording ("The hardware accelerator to be used…").

**CLI:** `hf endpoints hardware`

```
$ hf endpoints hardware
VENDOR  REGION     ACCEL  INSTANCE-TYPE  SIZES        GPU-MEM(x1)  $/HR(x1)
aws     us-east-1  gpu    nvidia-t4      x1,x4        16GB         0.50
gcp     us-east4   gpu    nvidia-l4      x1           24GB         0.70
aws     us-east-1  gpu    nvidia-l4      x1,x4        24GB         0.80
aws     eu-west-1  gpu    nvidia-a10g    x1           24GB         1.00
aws     us-east-1  gpu    nvidia-a100    x1,x2,x4,x8  80GB         2.50
aws     eu-west-1  cpu    intel-spr      x1,x2,x4,x8  —            0.033
...

Deploy with e.g.:
  hf endpoints deploy my-ep --repo <repo> --framework vllm --accelerator gpu \
    --vendor aws --region us-east-1 --instance-type nvidia-l4 --instance-size x1
```

- Defaults to `status == "available"`; `--all` includes deprecated/unavailable.
- Filter flags mirror the deploy flags (`--vendor`, `--region`, `--accelerator`, `--instance-type`), so the filter vocabulary *is* the deploy vocabulary.
- Human format collapses sizes into one row per (vendor, region, type); `--format json|agent` emit full per-size rows (standard formatting options apply).
- When authenticated, show a `QUOTA used/max` column — answers "what can I actually deploy right now".
- Output ends with a copy-pasteable deploy hint.

hanouticelina:
## Summary

Closes #4638 (item 1 of #4314, the "discovery wall").

`hf endpoints deploy` needs five free-text hardware flags (`--vendor --region --accelerator --instance-type --instance-size`) and there was no way to learn valid values from the CLI: `--help` only shows example literals and `hf endpoints catalog` lists models, not hardware. The vendor → region → compute matrix is already exposed by `GET /v2/provider/{namespace}`, so this is client-side wiring.

- **SDK**: `HfApi.list_inference_endpoints_hardware(namespace=..., token=...)` flattens the nested response into a list of `InferenceEndpointHardware` dataclasses (the API compute object + `vendor`/`region`, quota flattened into `max_accelerators`/`used_accelerators`).
- **CLI**: `hf endpoints hardware`, with `--vendor / --region / --accelerator / --instance-type` filters and `-a`/`--all`. By default it only lists hardware the namespace can deploy on right now.
- `hf endpoints deploy` and `create_inference_endpoint` now point at it, since that is where you stand when you hit the wall.

## Example

```bash
$ hf endpoints hardware --vendor aws --accelerator gpu
VENDOR REGION    ACCELERATOR INSTANCE_TYPE       INSTANCE_SIZE MEMORY_GB GPU_MEMORY_GB PRICE_PER_HOUR QUOTA STATUS
------ --------- ----------- ------------------- ------------- --------- ------------- -------------- ----- ---------
aws    eu-west-1 gpu         nvidia-a10g         x1                 30.0            24            1.0 0/16  available
aws    eu-west-1 gpu         nvidia-t4           x1                 15.0            16            0.5 1/30  available
aws    us-east-1 gpu         nvidia-a100         x1                145.0            80            2.5 0/4   available
aws    us-east-1 gpu         nvidia-a100         x2                290.0           160            5.0 0/4   available
aws    us-east-1 gpu         nvidia-a100         x4                580.0           320           10.0 0/4   available
aws    us-east-1 gpu         nvidia-a10g         x1                 30.0            24            1.0 0/16  available
aws    us-east-1 gpu         nvidia-a10g         x4                186.0            96            5.0 0/16  available
aws    us-east-1 gpu         nvidia-l4           x1                 30.0            24            0.8 1/16  available
aws    us-east-1 gpu         nvidia-l4           x4                185.0            96            3.8 1/16  available
aws    us-east-1 gpu         nvidia-l40s         x1                 30.0            48            1.8 1/16  available
aws    us-east-1 gpu         nvidia-l40s         x4                380.0           192            8.3 1/16  available
aws    us-east-1 gpu         nvidia-l40s         x8               1532.0           384           23.5 1/16  available
aws    us-east-1 gpu         nvidia-t4           x1                 15.0            16            0.5 1/30  available
aws    us-east-1 gpu         nvidia-t4           x4                192.0            64            3.0 1/30  available
aws    us-east-2 gpu         nvidia-rtx-pro-6000 x1                256.0            96           2.75 0/4   available
aws    us-east-2 gpu         nvidia-rtx-pro-6000 x2                512.0           192            5.5 0/4   available
aws    us-east-2 gpu         nvidia-rtx-pro-6000 x4               1024.0           384           11.0 0/4   available
aws    us-west-2 gpu         nvidia-h200         x1                256.0           141            5.0 1/2   available
Hint: Deploy on one of these, e.g.: hf endpoints deploy my-endpoint --repo <repo> --framework <framework> --vendor aws --region eu-west-1 --accelerator gpu --instance-type nvidia-a10g --instance-size x1
```

Only what the namespace can deploy on right now is listed. `--all` adds the rest, and the `QUOTA` and `STATUS` columns say why each extra row was hidden (real output, `--vendor gcp --accelerator gpu`):

```bash
$ hf endpoints hardware --vendor gcp --accelerator gpu
VENDOR REGION   ACCELERATOR INSTANCE_TYPE INSTANCE_SIZE MEMORY_GB GPU_MEMORY_GB PRICE_PER_HOUR QUOTA STATUS
------ -------- ----------- ------------- ------------- --------- ------------- -------------- ----- ---------
gcp    us-east4 gpu         nvidia-a100   x1                174.0            80            3.6 0/2   available
gcp    us-east4 gpu         nvidia-a100   x2                348.0           160            7.2 0/2   available
gcp    us-east4 gpu         nvidia-l4     x1                 12.0            24            0.7 0/8   available
gcp    us-east4 gpu         nvidia-l4     x4                186.0            96            3.8 0/8   available
gcp    us-east4 gpu         nvidia-t4     x1                 11.0            16            0.5 0/5   available

$ hf endpoints hardware --all --vendor gcp --accelerator gpu
VENDOR REGION   ACCELERATOR INSTANCE_TYPE INSTANCE_SIZE MEMORY_GB GPU_MEMORY_GB PRICE_PER_HOUR QUOTA STATUS
------ -------- ----------- ------------- ------------- --------- ------------- -------------- ----- ---------
gcp    us-east4 gpu         nvidia-a100   x1                174.0            80            3.6 0/2   available
gcp    us-east4 gpu         nvidia-a100   x2                348.0           160            7.2 0/2   available
gcp    us-east4 gpu         nvidia-a100   x4                696.0           320           14.4 0/2   available
gcp    us-east4 gpu         nvidia-a100   x8               1392.0           640           28.8 0/2   reserved
gcp    us-east4 gpu         nvidia-h100   x1                240.0            80           10.0 0/0   available
gcp    us-east4 gpu         nvidia-h100   x2                480.0           160           20.0 0/0   available
gcp    us-east4 gpu         nvidia-h100   x4                960.0           320           40.0 0/0   available
gcp    us-east4 gpu         nvidia-l4     x1                 12.0            24            0.7 0/8   available
gcp    us-east4 gpu         nvidia-l4     x4                186.0            96            3.8 0/8   available
gcp    us-east4 gpu         nvidia-t4     x1                 11.0            16            0.5 0/5   available
```

The five extra rows are all quota-bound: `a100 x4` and `x8` need 4 and 8 accelerators against a quota of 2 (and `x8` is `reserved` on top of that), and the namespace has no `h100` quota at all. Note that all three `h100` rows report `available` despite a `0/0` quota — status alone is not enough to tell whether a deploy will succeed.

`--format json` keeps the whole record, including the per-replica specs the table drops:

```bash
$ hf endpoints hardware --vendor aws --instance-type nvidia-l4 --format json | jq '.[0]'
{
  "id": "aws-us-east-1-nvidia-l4-x1",
  "vendor": "aws",
  "region": "us-east-1",
  "accelerator": "gpu",
  "instance_type": "nvidia-l4",
  "instance_size": "x1",
  "architecture": "Nvidia L4",
  "num_accelerators": 1,
  "num_cpus": 7,
  "memory_gb": 30.0,
  "gpu_memory_gb": 24,
  "price_per_hour": 0.8,
  "status": "available",
  "max_accelerators": 16,
  "used_accelerators": 1,
  "quota": "1/16"
}
```

SDK side:

```py
>>> from huggingface_hub import list_inference_endpoints_hardware
>>> [hw.id for hw in list_inference_endpoints_hardware() if hw.accelerator == "neuron"]
['aws-us-east-1-inf2-x1', 'aws-us-east-1-inf2-x12']
```

## Decisions taken (and why they deviate from the issue)

- **Always hits `/v2/provider/{namespace}`**, defaulting to the current user via `_get_namespace()` like every other `*_inference_endpoint*` method — rather than the issue's "public `/v2/provider` by default, namespaced when asked". One code path instead of two, and the quota numbers are always real: the unauthenticated route returns `{"maxAccelerators": 0, "usedAccelerators": 0}` for everything, which reads as "you can't deploy anything". The namespaced route also surfaces namespace-specific hardware the public one hides (a `reserved` `gcp-us-east4-nvidia-a100-x8` shows up for my namespace but not on the public route). Cost: no anonymous browsing, which is consistent with the rest of `hf endpoints` (all of it needs a token) and gives the standard `Not logged in. Run 'hf auth login' first.` error.
- **One row per (vendor, region, type, size) in every format**, not "human collapses sizes / json expands them" as the issue proposed. Mode-dependent *rows* would mean `if out.mode == ...` inside the command, and per-size rows are what you need anyway since `--instance-size` is a deploy flag. `--vendor`/`--accelerator` keep the list short.
- **Filtering and the deployability default live in the CLI, not the SDK.** The API has no server-side filtering, so SDK-side filters would just be a list comprehension with extra parameters. The SDK mirrors the API (everything, all statuses), the CLI owns the UX default.
- **"Deployable" is a status denylist plus a quota check, not `status == "available"`.** Two reasons an allowlist of one is wrong. (1) `ComputeStatus` has five values and the deployable set is the complement of two: the Endpoints UI's own `Cloud.getComputeCompatibility` computes `statusOk = status !== 'not_available' && status !== 'deprecated'`, so it treats `reserved` and `low_availability` as selectable (its instance picker renders `reserved` computes with "Reserved" in place of the price). `low_availability` is in the spec but not yet returned in prod, so an allowlist would have silently dropped it from the default view the day the server starts sending it — exactly what keeping `status` a plain `str` (next bullet) is meant to avoid. (2) Status alone does not decide whether a deploy succeeds: 3 of the 52 `available` computes in my namespace have `maxAccelerators: 0`, and `a100 x4` has a quota of 2 against 4 needed. The UI treats that as a separate `notEnoughQuota` state (`quotaOk = maxAccelerators - usedAccelerators >= numAccelerators * replicas`). So the CLI hides `{deprecated, not_available}` and anything without enough remaining quota for one replica, which is what `--min-replica` defaults to. For my namespace that is 43 rows by default out of 67.
- **`status` is a plain `str`, not an enum.** Same for `accelerator`. New server-side values shouldn't need a client release — the spec already has an unused `low_availability` status and an unused `zero_gpu` accelerator.
- **`architecture`, `num_cpus` and `num_accelerators` are dropped from the human/agent table** but kept in the items, so `--format json` still carries them (same pattern as `hf jobs ps`). `architecture` restates `instance_type` (`"Nvidia L4"` vs `nvidia-l4`), `num_accelerators` restates `instance_size` (always the multiplier, CPU rows included), and `num_cpus` is `null` on every CPU row. `memory_gb` is kept alongside `gpu_memory_gb`: the latter is `null` on CPU hardware, so without it CPU rows carry no memory information at all, and it is the field the UI's `isComputeMemOk` checks for `cpu`.

## Notes

- The OpenAPI spec describes `gpuMemoryGb` as "GPU memory **per accelerator**", but the data says otherwise (`nvidia-l4 x1` = 24, `x4` = 96, `nvidia-h200 x8` = 1128 = 141×8), so it is documented as the per-replica total. Worth a spec fix on the server side.
- Neither region nor vendor status is used for filtering, only compute status. The only `not_available` region today (`aws ap-northeast-2`) has all of its computes marked `deprecated`, so the per-compute check already covers it, and all three vendors are currently `available`.
- The name is `list_inference_endpoints_hardware`, plural, to match the two existing siblings `list_jobs_hardware` and `list_spaces_hardware` (and to sit next to `list_inference_endpoints`). The dataclass stays singular, `InferenceEndpointHardware`, like `JobHardwareInfo`.

## Tests

- `test_list_hardware` (`tests/test_inference_endpoints.py`): mocks the HTTP layer with a trimmed real `/v2/provider` payload and checks the flattening — vendor/region pulled from the enclosing objects, `numCpus` missing and `gpuMemoryGb` null on CPU hardware, non-deployable hardware returned with its status.
- `TestInferenceEndpointsHardwareCommand` (`tests/test_cli.py`): `test_default_lists_only_deployable_hardware` is parametrized over all five statuses and four quota cases (plenty left / exactly enough / none at all / not enough for an `x8`), `--all` includes both a deprecated and an out-of-quota entry, filters are case-insensitive and combine, both memory columns are shown, quota renders as `used/max`, and the empty result hints at `--all` only when something was actually filtered out.

No production test: unlike `hf spaces hardware`, this route requires auth, and the `production` marker doesn't provide a token.

Verified manually against the live API (all output above is real).

---

> [!NOTE]
> **Low Risk**
> Additive API/CLI and documentation; no changes to deploy or auth flows beyond a new read-only provider endpoint call.
> 
> **Overview**
> Adds **hardware discovery** for Inference Endpoints so users can see valid `--vendor`, `--region`, `--accelerator`, `--instance-type`, and `--instance-size` values before deploying.
> 
> The SDK exposes `list_inference_endpoints_hardware` / `InferenceEndpointHardware`, flattening the authenticated `GET /v2/provider/{namespace}` response (pricing, quota, specs). The new **`hf endpoints hardware`** command filters that list, shows quota and hourly price, and by default only rows the namespace can deploy on now (status + accelerator quota for one replica); `--all` shows everything. **`hf endpoints deploy`** and **`create_inference_endpoint`** docs now point to this command.
> 
> CLI and API tests cover flattening, deployability rules, filters, and empty-result hints.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ecefcb6ecf612e5517ded6da037201a606cf82f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/source/en/guides/cli.md
```diff
@@ -2307,6 +2310,23 @@ Use `hf endpoints` to list, deploy, describe, and manage Inference Endpoints dir
...
+
+```bash
+>>> hf endpoints hardware --vendor aws --accelerator gpu
+VENDOR REGION    ACCELERATOR INSTANCE_TYPE INSTANCE_SIZE MEMORY_GB GPU_MEMORY_GB PRICE_PER_HOUR QUOTA STATUS
+------ --------- ----------- ------------- ------------- --------- ------------- -------------- ----- ---------
+aws    eu-west-1 gpu         nvidia-t4     x1                 15.0            16            0.5 1/30  available
+aws    us-east-1 gpu         nvidia-l4     x1                 30.0            24            0.8 1/16  available
+aws    us-east-1 gpu         nvidia-l4     x4                185.0            96            3.8 1/16  available
+aws    us-west-2 gpu         nvidia-h200   x1                256.0           141            5.0 1/2   available
+...
```
```suggestion
...
Hint: Deploy on one of these, e.g.: hf endpoints deploy my-endpoint --repo <repo> --framework <framework> --vendor aws --region eu-west-1 --accelerator gpu --instance-type nvidia-a10g --instance-size x1
```

davanstrien:
Tried this out locally against my own namespace and works well. 

Some small DX things

- **Price formatting**: `hf jobs hardware` and `hf spaces hardware` both show `$`-formatted costs (`$0.40`), while this prints a bare float (`0.5`). Not a big deal, but maybe nice to align? The bare floats are a bit more agent friendly but maybe worth having the currency somewhere to make it clear?
- **Case-insensitive filters are one-sided**: `hw.vendor == vendor.lower()` lowers only the user input, so the filter silently breaks if a server value is ever not lowercase. I guess it shouldn't be the case server-side, so probably fine. 
- **Typo'd filter gives a bare "No results found."**: the `--all` hint only fires when matching rows were filtered out, so `--vendor asw` gets no help at all. Maybe worth a hint here too?
- **Docs**: quota is per-namespace, so someone deploying into an org needs `hf endpoints hardware --namespace <org>` to see the quota that actually applies; maybe worth adding a sentence on this.

hanouticelina:
thanks @davanstrien for the feedback! all of them are addressed, good catch on the price formatting and the Case insensitive filters!

PR #4681, opened 2026-08-16
Add cancel_event to hf_hub_download for per-download cancellation

### LINKED ISSUE #4632: [BUG] Cancelling one Xet download aborts every other download, and the only way to do it is private API

### The situation

I have an application that downloads several models, showing each as a row the user can cancel individually. Cancelling one must leave the others running.

### HTTP handles this fine

`http_get` calls the progress callback inline in its read loop, so raising from the callback stops that one download and nothing else.

### Xet does not

Progress comes from a thread `hf_xet` owns. An exception raised in the callback is swallowed there. The download runs to completion behind a row that says "cancelled", and keeps consuming bandwidth other downloads are waiting for.

### The only thing that works is private, and too broad

`abort_xet_session()` lives in `huggingface_hub.utils._xet`. It stops the transfer, but it aborts the whole process-wide session, so every concurrent download dies with it.

I had to drop my download concurrency to 1 to avoid that.

```python
import threading, time
from huggingface_hub.utils._xet import abort_xet_session  # private

BIG = ("unsloth/Qwen3-4B-Instruct-2507-GGUF", "Qwen3-4B-Instruct-2507-Q4_K_M.gguf")
SMALL = ("unsloth/Qwen3-1.7B-GGUF", "Qwen3-1.7B-Q4_K_M.gguf")

done = {}

def fetch(name, repo, filename):
    try:
        hf_hub_download(repo_id=repo, filename=filename)
        done[name] = "completed"
    except Exception as exc:
        done[name] = f"{type(exc).__name__}: {exc}"

threading.Thread(target=fetch, args=("A", *BIG)).start()
threading.Thread(target=fetch, args=("B", *SMALL)).start()
time.sleep(10)

abort_xet_session()   # intended: cancel A

# B is cancelled too, and B was never the one I wanted to stop.
```

Cancellation of a single download, in the public API, behaving the same on Xet and HTTP. A few shapes would each solve it, and they are not mutually exclusive.

### A scoped session

Good when downloads are grouped and the group has a natural lifetime. Leaving the block stops whatever is still running.

```python
from huggingface_hub import download_session

with download_session() as session:
    session.download(repo_id=..., filename=...)
    session.download(repo_id=..., filename=...)
    # session.cancel() stops only this session's transfers
```

### A handle returned per download

Good when each download is tracked as its own object, which is what a task list wants.

```python
handle = hf_hub_download(repo_id=..., filename=..., return_handle=True)
handle.cancel()      # stops this download only
handle.result()      # raises DownloadCancelled
```

### A cancel object accepted by `hf_hub_download`

Good when the caller already has its own concurrency model and does not want a new one. Accepting anything with `is_set()` covers `threading.Event`, `asyncio.Event` and custom flags without committing users to threads.

```python
from huggingface_hub import CancelToken

token = CancelToken()
hf_hub_download(repo_id=..., filename=..., cancel=token)

token.cancel()   # from wherever the caller's UI lives
```

Whichever shape, the important part is that cancelling one download does not touch the others, and that it raises something specific rather than surfacing as a bare `RuntimeError`.

### Raising from the progress callback

Works on HTTP, silently swallowed on Xet.

### Passing my own session to `xet_get`

`xet_get` takes no session argument and calls `get_xet_session()` internally, so there is no injection point.

### Running each download in its own process

Gives real isolation because the session holder is per-process, and makes cancellation a kill. It costs IPC for progress reporting, which is a lot of machinery to buy one feature.

**Additional context**

### The isolation may already exist

In xet-core at `v1.5.2`, `XetSessionBuilder::build()` only reuses a runtime from thread-local storage, so a session built on a fresh worker thread gets its own `XetRuntime`, and `sigint_abort` tears down only `runtime_cell_if_owned()`.

If that reading is correct, this is about exposing a capability rather than building one. I have not confirmed it experimentally.

### Cancelled transfers are not resumable

Aborting a 2.5GB download after 15s left an empty `blobs/` directory and no new entries in the Xet chunk cache, so retrying starts from zero. The HTTP path resumes from its `.incomplete` blob. There is a pre-existing issue for this filed. 

### Environment

- `huggingface_hub` 1.26.0, `hf_xet` 1.5.2
- macOS arm64 and Linux x86_64

chintan-diwakar:
Fixes #4632.

## What

Adds a `cancel_event` argument to `hf_hub_download` (threaded through to `http_get` and `xet_get`): any object with an `is_set()` method (`threading.Event`, `asyncio.Event`, a custom flag). Once set, that download raises a new `DownloadCancelledError`; concurrent downloads are unaffected.

```python
import threading
from huggingface_hub.errors import DownloadCancelledError

cancel = threading.Event()
# from the UI thread: cancel.set()
try:
    hf_hub_download(repo_id=..., filename=..., cancel_event=cancel)
except DownloadCancelledError:
    ...
```

## How

- **HTTP path**: the event is checked in the existing chunk loop of `http_get` (and forwarded through the retry recursion).
- **Xet path**: the key realization is that no Rust changes are needed for isolation. `hf_xet` already exposes `XetFileDownloadGroup.abort()`, which cancels a single group while leaving the process-wide session usable (its docs contrast it with `sigint_abort` explicitly), and `xet_get` already creates one group per file. A small watcher thread polls the event and calls `group.abort()` when set. The abort deliberately does not happen inside the progress callback: `abort()` joins the progress thread, so aborting from the callback would deadlock (and callbacks raised from `hf_xet`'s thread are swallowed anyway, per the issue).
- `DownloadCancelledError(OSError)` added to `errors.py` and to the errors reference docs.

## Tests

`TestDownloadCancellation` in `tests/test_file_download.py`, no network needed:

- pre-set event: `http_get` raises without issuing the request; `xet_get` raises without creating a session
- mid-stream: `http_get` stops at the cancellation point (mocked stream, same pattern as `TestHttpGet`)
- xet: watcher aborts the download group (fake group that blocks until aborted, mirroring `wait_to_finish` semantics)
- plumbing: `hf_hub_download(..., cancel_event=...)` forwards the event to the transport

`ruff check`, `ruff format --check`, and `ty check src` are clean on the changed files (the two pre-existing `ty` diagnostics in `cli/_output.py` are untouched).

## Verified end to end against the Hub

Two concurrent `hf_hub_download` calls (xet-backed), cancelling one 3s in: the cancelled one raised `DownloadCancelledError` with ~0.1s latency and the other completed normally. This is the exact scenario from #4632 that previously required `abort_xet_session()` and killed both.

## Known limitation (upstream)

While verifying with `/proc/net/dev`, I found that after `group.abort()` the aborted group's transfer keeps downloading in the background at full bandwidth until the file would have completed, then discards the bytes (chunk cache does not grow). So on the xet path this PR gives prompt API-level cancellation and correct isolation, but the bandwidth of the cancelled download is only truly released once the underlying transfer finishes. That is an `hf_xet` issue rather than something fixable from Python; filed with measurements as huggingface/xet-core#942. The HTTP path releases bandwidth immediately.

---

> [!NOTE]
> **Medium Risk**
> Touches core download paths (`hf_hub_download`, HTTP streaming, Xet) with a new public API and threading on Xet; behavior is scoped to opt-in cancellation and is well tested, but download failures/cleanup on cancel merit reviewer attention.
> 
> **Overview**
> Adds optional **`cancel_event`** to **`hf_hub_download`** (and internal **`http_get`** / **`xet_get`** paths) so callers can stop a single in-flight download without affecting others. Setting the event raises a new **`DownloadCancelledError`** (`OSError` subclass), documented in the errors reference.
> 
> On **HTTP**, cancellation is checked before the request and on each chunk; **`cancel_event`** is preserved across download retries. On **Xet**, a daemon watcher polls the event and calls **`group.abort()`** on that file’s download group only (not the shared session), avoiding abort from inside the progress callback to prevent deadlock.
> 
> **`TestDownloadCancellation`** covers pre-set cancel, mid-stream HTTP stop, Xet group abort, and plumbing from **`hf_hub_download`** to **`http_get`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec559da95071caec05fc65e407addbdacb57e07a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/package_reference/utilities.md
+++ b/docs/source/en/package_reference/utilities.md
@@ -245,6 +245,10 @@ user as possible.
 
 [[autodoc]] huggingface_hub.errors.OfflineModeIsEnabled
 
+#### DownloadCancelledError
+
+[[autodoc]] huggingface_hub.errors.DownloadCancelledError
+
 ## Telemetry
 
 `huggingface_hub` includes a helper to send telemetry data. This information helps us debug issues and prioritize new features.
--- a/src/huggingface_hub/errors.py
+++ b/src/huggingface_hub/errors.py
@@ -252,6 +252,13 @@ class FileMetadataError(OSError):
     """
 
 
+# DOWNLOAD ERRORS
+
+
+class DownloadCancelledError(OSError):
+    """Raised when a download is cancelled via the `cancel_event` argument of [`hf_hub_download`]."""
+
+
 # BUCKET ERRORS
 
 
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -4,12 +4,13 @@ import os
 import re
 import shutil
 import stat
+import threading
 import time
 import uuid
 import warnings
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, BinaryIO, Literal, NoReturn, overload
+from typing import Any, BinaryIO, Literal, NoReturn, Protocol, overload
 from urllib.parse import quote, urlparse
 
 import httpx
@@ -26,6 +27,7 @@ from ._local_folder import (
 from ._revision import ResolvedRevision
 from ._tree_cache import read_tree_cache, tree_cache_folder_for_local_dir
 from .errors import (
+    DownloadCancelledError,
     FileMetadataError,
     GatedRepoError,
     HfHubHTTPError,
@@ -67,6 +69,16 @@ _CACHED_NO_EXIST_T = Any
 
 # Regex to get filename from a "Content-Disposition" header for CDN-served files
 HEADER_FILENAME_PATTERN = re.compile(r'filename="(?P<filename>.*?)";')
+
+
+class CancelEvent(Protocol):
+    """Anything with an `is_set()` method, e.g. `threading.Event` or `asyncio.Event`.
+
+    Once set, the download it was passed to is cancelled and raises [`~errors.DownloadCancelledError`].
+    """
+
+    def is_set(self) -> bool: ...
+
 
 # Regex to check if the revision IS directly a commit_hash
 REGEX_COMMIT_HASH = re.compile(r"^[0-9a-f]{40}$")
@@ -331,6 +343,7 @@ def http_get(
     expected_size: int | None = None,
     displayed_filename: str | None = None,
     tqdm_class: type[base_tqdm] | None = None,
+    cancel_event: CancelEvent | None = None,
     _nb_retries: int = 5,
     _tqdm_bar: tqdm | None = None,
 ) -> None:
@@ -357,7 +370,13 @@ def http_get(
         displayed_filename (`str`, *optional*):
             The filename of the file that is being downloaded. Value is used only to display a nice progress bar. If
             not set, the filename is guessed from the URL or the `Content-Disposition` header.
+        cancel_event (`CancelEvent`, *optional*):
+            An event-like object with an `is_set()` method (e.g. `threading.Event`). Once set, the download stops
+            and [`~errors.DownloadCancelledError`] is raised. Only this download is affected.
     """
+    if cancel_event is not None and cancel_event.is_set():
+        raise DownloadCancelledError(f"Download cancelled: {displayed_filename or url}")
+
     if expected_size is not None and resume_size == expected_size:
         # If the file is already fully downloaded, we don't need to download it again.
         return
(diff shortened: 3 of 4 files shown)


hanouticelina:
Hi there, thanks, but I'm going to close this one: we don't want to add a new parameter to `hf_hub_download`. out of curiosity, are you running into this yourself? this also doesn't quite solve #4632 I think. that issue is about abort_xet_session() being process wide, and per your own known limitation the xet transfer keeps running after cancellation

PR #4683, opened 2026-08-16
Fix double percent-decode of URL revisions in parse_hf_uri

icn5381:
## Bug

A branch or tag name containing a literal percent sequence is decoded twice when parsed from a web URL, silently changing the revision.

```py
>>> parse_hf_uri("https://huggingface.co/org/repo/resolve/a%2520b/file.txt").revision
'a b'        # expected 'a%20b' — the branch is literally named "a%20b"
>>> parse_hf_uri("hf://org/repo@a%2520b/file.txt").revision
'a%20b'      # the equivalent hf:// form is correct
```

Git allows `%` in ref names, so such branches exist. A user copy-pasting the URL gets a silently wrong revision (`EntryNotFoundError` on download) while the equivalent `hf://` URI works — the same resource parses two different ways depending on input form.

Root cause: `_url_to_uri_body` percent-decodes each URL segment once, then hands the revision to the canonical parser, which applies a second `unquote` (`_parse_repo_body`). The path-in-repo is unaffected (it is not unquoted downstream); only the revision double-decodes.

## Fix

- The revision segment of a URL is now decoded with a dedicated helper that re-escapes `%` → `%25` (and keeps a decoded `/` as `%2F`, as before), so the canonical parser's single `unquote` restores the exact branch name.
- `to_uri` now escapes `%` in revisions the same way it already escaped `/`, so `parse → to_uri → parse` stays idempotent for such revisions.

## Verification

- New regression cases in all three parametrized suites: URL input, `hf://` input with `to_uri` round-trip, and the `to_url` round-trip (which already asserted that the revision must reparse exactly — it fails on `main` for these cases).
- Confirmed the wrong output on `main` directly (`'a b'`) vs the fixed output (`'a%20b'`).
- Full `tests/test_utils_hf_uris.py`: 143 passed. `ruff check` / `ruff format --check` clean.

---

> [!NOTE]
> **Low Risk**
> Localized change to HF URI/URL encoding in `utils/_hf_uris.py` with targeted tests; no auth, network, or data-path changes beyond fixing revision strings.
> 
> **Overview**
> Fixes **incorrect revision parsing** when a branch or tag name contains a literal `%` (e.g. `a%20b`). Web URLs were decoded twice—once when normalizing the URL and again in the canonical `hf://` parser—so the revision became `a b` instead of `a%20b`, while the same resource via `hf://` parsed correctly.
> 
> URL normalization now uses **`_decode_url_revision_segment`** for the revision path segment, re-escaping `%` → `%25` (and `/` → `%2F` as before) so the downstream single `unquote` restores the exact ref name. **`HfUri.to_uri`** applies the same `%` escaping for canonical round-trips. Regression tests cover `hf://`, Hub URLs, and `to_url` re-parse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20b2c962fa317377d5baf061c7df68a6c4041bc9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_hf_uris.py
+++ b/src/huggingface_hub/utils/_hf_uris.py
@@ -142,11 +142,12 @@ class HfUri:
         parts: list[str] = [constants.HF_PROTOCOL, _TYPE_TO_PREFIX[self.type], "/", self.id]
         if self.revision is not None:
             # Encode '/' as '%2F' for revisions that would otherwise be split as '<revision>/<path>'
-            # at parse time. Special refs ('refs/pr/N', 'refs/convert/<name>') are kept verbatim
+            # at parse time, and '%' as '%25' because the parser unquotes the revision exactly once.
+            # Special refs ('refs/pr/N', 'refs/convert/<name>') are kept verbatim
             # because the parser matches them eagerly.
             revision = self.revision
-            if "/" in revision and _SPECIAL_REFS_REVISION_REGEX.fullmatch(revision) is None:
-                revision = revision.replace("/", "%2F")
+            if ("/" in revision or "%" in revision) and _SPECIAL_REFS_REVISION_REGEX.fullmatch(revision) is None:
+                revision = revision.replace("%", "%25").replace("/", "%2F")
             parts.append(f"@{revision}")
         if self.path_in_repo:
             parts.append(f"/{self.path_in_repo}")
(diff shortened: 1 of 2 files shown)


hanouticelina:
Hi, is this a real bug you faced in a real use case? if yes, please provide a real use case scenario, otherwise, I'd rather close this PR

PR #4689, opened 2026-08-16
Fix ResolvedRevision silently becoming "main" after pickling

### LINKED ISSUE #4688: ResolvedRevision silently becomes "main" after pickling
## Problem

`ResolvedRevision` is a `str` subclass, but its string value does not survive pickling. A round-trip returns an object whose `repr` is unchanged but whose actual string value has become `"main"`. That combination makes it hard to spot when it causes issues: the object looks correct under `repr` and in a debugger, and only misbehaves when actually used.

## Example real-world impact

vLLM [#49990](https://github.com/vllm-project/vllm/pull/49990) stores the result of `resolve_revision()` in its model config, and its V1 engine pickles that config into a sub-process, so `--revision <branch>` is honored in the parent and silently ignored in the child. Every repository that serves weights from a non-default branch fails to load, with a confusing error: the child fetches from `main`, finds no `config.json`, and reports `Unrecognized model ... Should have a 'model_type' key in its 'config.json'`, followed by a tokenizer instantiation failure.

Repos that keep everything on `main` are unaffected, which is why this can sit unnoticed for a while: it looks like a per-repository problem rather than a serialization one.

## Believed cause

`ResolvedRevision.__new__` takes the resolution arguments, but pickling a `str` subclass reconstructs it through the `str` protocol, so `initial` is restored as an attribute while the string value is rebuilt from the default (`initial=None` becomes `"main"`, per the documented behavior).

### Reproduction

```python
import pickle

rr = HfApi().resolve_revision("turboderp/Qwen3-0.6B-exl3", revision="3.0bpw")
rt = pickle.loads(pickle.dumps(rr))

print("before:", repr(str(rr)), repr(rr))
print("after :", repr(str(rt)), repr(rt))
```

```
before: '3.0bpw' ResolvedRevision(initial='3.0bpw', resolved='d46a12a64d5a0a9a546d862855c8f390b16b7436')
after : 'main'   ResolvedRevision(initial='3.0bpw', resolved='d46a12a64d5a0a9a546d862855c8f390b16b7436')
```

### Logs

```shell

```

### System info

```shell
- huggingface_hub version: 1.27.0
- Platform: Linux-6.12.0-254.el10.x86_64-x86_64-with-glibc2.39
- Python version: 3.12.13
- Running in iPython ?: No
- Running in notebook ?: No
- Token path ?: /home/ypell/.cache/huggingface/token
- Has saved token ?: True
- Who am I ?: yeasah
- Configured git credential helpers: 
- Run by AI agent ?: No
- Installation method: pip
- httpx: 0.28.1
- hf_xet: 1.6.0
- gradio: N/A
- tensorboard: N/A
- HF_HUB_CACHE: /home/ypell/.cache/huggingface/hub
- HF_ASSETS_CACHE: /home/ypell/.cache/huggingface/assets
- HF_TOKEN_PATH: /home/ypell/.cache/huggingface/token
- HF_STORED_TOKENS_PATH: /home/ypell/.cache/huggingface/stored_tokens
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
```

ManthanKeim:
## Summary
Fixes #4688.

`ResolvedRevision` is a `str` subclass whose string value is the user-requested revision (`initial`, or `"main"` when `initial` is `None`) while `.resolved` holds the commit hash. Pickling a `str` subclass reconstructs it with `__new__(cls, *args)` from `__getnewargs__`. The inherited `str.__getnewargs__` only returns the string value, so unpickling called `__new__(cls, initial)`:

- `resolved` was set to the original string value
- `initial` defaulted to `None`
- the immutable string value became `"main"`
- `__dict__` restore then put the original attributes back, so `repr()` still looked correct

That combination is what made the bug hard to spot. Callers such as vLLM that pickle a resolved revision into a subprocess then fetch from `main` instead of the requested branch.

This adds `__getnewargs__` so unpickling calls `__new__(cls, resolved, initial)` and preserves both the string value and the commit hash. Tests cover every pickle protocol.

## Test plan
- [x] `PYTHONPATH=src python3` pickle round-trip of `ResolvedRevision(..., initial="3.0bpw")` keeps `str(...) == "3.0bpw"` on protocols 0–5
- [x] `pytest tests/test_snapshot_download.py -k "test_revision_str or test_revision_survives_pickle"` — 7 passed
- [x] `ruff check` / `ruff format --check` on the two changed files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/huggingface_hub/_revision.py
+++ b/src/huggingface_hub/_revision.py
@@ -39,5 +39,12 @@ class ResolvedRevision(str):
         revision.resolved = resolved
         return revision
 
+    def __getnewargs__(self) -> tuple[str, str | None]:
+        # `str` subclasses pickle through `__new__(cls, *args)`. The inherited
+        # `str.__getnewargs__` only returns the string value, so unpickling would call
+        # `__new__(cls, initial)` and treat `initial` as `resolved`. `initial` then
+        # defaults to `None` and the string value silently becomes `"main"`.
+        return (self.resolved, self.initial)
+
     def __repr__(self) -> str:
         return f"ResolvedRevision(initial={self.initial!r}, resolved={self.resolved!r})"
--- a/tests/test_snapshot_download.py
+++ b/tests/test_snapshot_download.py
@@ -1,4 +1,5 @@
 import os
+import pickle
 from pathlib import Path
 from unittest.mock import MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


hanouticelina:
closing in favor of https://github.com/huggingface/huggingface_hub/pull/4692

PR #4691, opened 2026-08-17
[Tests] Don't require real symlink support in test_cache_no_symlinks

### LINKED ISSUE #4684: Windows: test_cache_no_symlinks tests require real symlink support and fail on a machine that actually cannot symlink
### Describe the bug

Three tests in `tests/test_cache_no_symlinks.py::TestCacheLayoutIfSymlinksNotSupported` — the class specifically about the no-symlinks environment — fail on a Windows machine that genuinely cannot create symlinks (no Developer Mode, non-elevated, i.e. the default consumer setup). They implicitly require real symlink support.

CI can't see this: `windows-latest` runners are elevated, so `os.symlink` actually works there, and the `DISABLE_SYMLINKS_IN_WINDOWS_TESTS` + `FakeSymlinkDict` fixture from `tests/conftest.py` protects none of the three (details below). So the combination "Windows + symlinks truly unavailable" — the situation this test class is named after — is exercised neither in CI nor by these tests, and a Windows contributor without Developer Mode gets 3 baffling failures out of the box.

For context, this is the complete cache/symlink failure set from a full run of the Windows CI selection (`-m "not no_xet and not inference"`, 2,422 items) on this machine — everything else around the cache passes.

**Mechanisms, per test:**

**1. `test_are_symlinks_supported_default`** patches the memo dict with the imported constant as key:

```python
mocker.patch("huggingface_hub.file_download._are_symlinks_supported_in_dir", {HF_HUB_CACHE: True})
assert are_symlinks_supported()
```

but the autouse `patch_constants` fixture (`tests/conftest.py`) rebinds `constants.HF_HUB_CACHE` to a fresh temp dir for every test, while the test module's `HF_HUB_CACHE` was captured at import time. The keys never match, so `are_symlinks_supported()` misses the dict and runs the real probe. On a symlink-capable machine the probe succeeds and the assert passes *by accident*; here it honestly returns `False` and the test fails. (I verified the key miss by printing the resolved runtime key vs. the patched key inside the test — they're two different temp/real paths.)

**2. `test_download_no_symlink_existing_file`** and **3. `test_scan_and_delete_cache_no_symlinks`** both mock `are_symlinks_supported` to `True` at some point and then run a real `hf_hub_download`, which reaches the real `os.symlink`:

```
  File "...\src\huggingface_hub\file_download.py", line 675, in _create_symlink
    os.symlink(src_rel_or_abs, abs_dst)
OSError: [WinError 1314] A required privilege is not held by the client: '..\\..\\blobs\\d9e9f15b...' -> '...\\snapshots\\60b8d3fe...\\config.json'
```

WinError 1314 maps to plain `OSError` with `errno 22` (not `PermissionError`), so `_create_symlink`'s `except PermissionError` fallback doesn't catch it either — which is correct behavior for the library, the tests just assume the call can succeed.

Why the CI fixture doesn't help: test 1 *replaces* the `FakeSymlinkDict` with its own dict, and tests 2–3 bypass the dict entirely by mocking `are_symlinks_supported` itself. Meanwhile `test_are_symlinks_supported_windows_specific_dir` in the same class mocks `os.symlink` directly — and that one passes everywhere, including here.

Happy to send a PR per CONTRIBUTING's issue-first flow — my suggestion would be: mock `os.symlink` in 2–3 the way `test_are_symlinks_supported_windows_specific_dir` already does (or skip them when the machine really can't symlink), and have 1 assert against a directly probed temp dir instead of the patched-dict shortcut that never fires.

### Reproduction

On any Windows machine where symlink creation is unavailable (check: the snippet below raises `[WinError 1314]`):

```python
import os, tempfile
d = tempfile.mkdtemp()
open(os.path.join(d, "s.txt"), "w").write("x")
os.symlink(os.path.join(d, "s.txt"), os.path.join(d, "l.txt"))  # OSError: [WinError 1314]
```

then:

```
python -m pytest tests/test_cache_no_symlinks.py -v
```

### Logs

```
FAILED tests/test_cache_no_symlinks.py::TestCacheLayoutIfSymlinksNotSupported::test_are_symlinks_supported_default - assert False
 +  where False = are_symlinks_supported()
FAILED tests/test_cache_no_symlinks.py::TestCacheLayoutIfSymlinksNotSupported::test_download_no_symlink_existing_file - OSError: [WinError 1314] A required privilege is not held by the client: ...
FAILED tests/test_cache_no_symlinks.py::TestCacheLayoutIfSymlinksNotSupported::test_scan_and_delete_cache_no_symlinks - OSError: [WinError 1314] A required privilege is not held by the client: ...
3 failed, 3 passed, 1 warning in 3.37s
```

### System info

```
- huggingface_hub version: 1.28.0.dev0 (main @ 72b5d66c)
- Python version: 3.13.13
- Run by AI agent ?: Yes
- Installation method: pip (editable, [dev] extras)
- httpx: 0.28.1
- hf_xet: 1.6.0
```

tonycoder-hub:
Fixes #4684

## Summary

Test-only change. No product code touched.

Three tests in `TestCacheLayoutIfSymlinksNotSupported` implicitly required a machine that can create symlinks, so they failed on a stock Windows setup (no Developer Mode, non-elevated). CI never noticed because `windows-latest` runners are elevated.

- `test_are_symlinks_supported_default` patched the memo dict with `HF_HUB_CACHE` captured at import time, while the autouse fixture repoints `constants.HF_HUB_CACHE` to a temp dir. The key never matched, so `are_symlinks_supported()` ran the real filesystem probe. It now mocks `os.symlink` the same way `test_are_symlinks_supported_windows_specific_dir` already does, and asserts the memo entry.
- `test_are_symlinks_supported_disabled_by_env` had the same dead key. Fixed the same way; it now asserts the env var short-circuits before probing.
- `test_download_no_symlink_existing_file` and `test_scan_and_delete_cache_no_symlinks` mock `are_symlinks_supported` to `True` then run a real `hf_hub_download`, which hits real `os.symlink` and raises `[WinError 1314]`. They now skip when the machine genuinely cannot symlink, via a module-level OS probe (not `are_symlinks_supported()`, so Windows CI still runs them).

`WinError 1314` is `OSError` with `errno 22`, not `PermissionError`, so `_create_symlink`'s fallback is correct library behavior. The tests were the ones assuming the call can succeed.

## Testing

Reproduced the three failures on Linux by making `os.symlink` raise `OSError(errno=22)` at import time. After the fix: 4 passed, 2 skipped on a no-symlink host; 14 passed (`test_cache_no_symlinks.py` + `test_cache_layout.py`) on a symlink-capable host. `make style` and `make quality` are clean.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to tests; no library behavior is modified.
> 
> **Overview**
> **Test-only** fix so `TestCacheLayoutIfSymlinksNotSupported` no longer assumes the host can create real symlinks (e.g. stock Windows without Developer Mode).
> 
> `test_are_symlinks_supported_default` and `test_are_symlinks_supported_disabled_by_env` no longer seed the memo dict with import-time `HF_HUB_CACHE`, which did not match the autouse `patch_constants` temp cache. They now mock `os.symlink`, inject an empty memo, and assert the expected memo key (`constants.HF_HUB_CACHE`) or that probing is skipped when `HF_HUB_DISABLE_SYMLINKS` is set.
> 
> Two integration tests that call real `os.symlink` via `hf_hub_download` are gated with a module-level `_can_create_symlink()` probe and `@requires_symlinks`, so they skip on hosts that cannot symlink instead of failing with WinError 1314.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit faa615c3dd75ac30954d7cdcbdb839479ef2c821. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_cache_no_symlinks.py
+++ b/tests/test_cache_no_symlinks.py
@@ -1,30 +1,62 @@
+import os
 import warnings
 from pathlib import Path
 
 import pytest
 from pytest_mock import MockerFixture
 
-from huggingface_hub import hf_hub_download, scan_cache_dir
-from huggingface_hub.constants import CONFIG_NAME, HF_HUB_CACHE
+from huggingface_hub import constants, hf_hub_download, scan_cache_dir
+from huggingface_hub.constants import CONFIG_NAME
 from huggingface_hub.file_download import are_symlinks_supported
+from huggingface_hub.utils import SoftTemporaryDirectory
 
 from .testing_constants import DUMMY_MODEL_ID
+
+
+def _can_create_symlink() -> bool:
+    """Whether this machine is actually allowed to create symlinks.
+
+    On Windows this requires Developer Mode or an elevated process. Without either, `os.symlink`
+    raises `OSError` ([WinError 1314] "A required privilege is not held by the client").
+    """
+    with SoftTemporaryDirectory() as tmpdir:
+        src_path = Path(tmpdir) / "dummy_file_src"
+        src_path.touch()
+        try:
+            os.symlink(src_path, Path(tmpdir) / "dummy_file_dst")
+        except OSError:
+            return False
+        return True
+
+
+requires_symlinks = pytest.mark.skipif(
+    not _can_create_symlink(),
+    reason="requires a machine that can create symlinks (on Windows: Developer Mode or an elevated process)",
+)
 
 
 @pytest.mark.production
 class TestCacheLayoutIfSymlinksNotSupported:
     def test_are_symlinks_supported_default(self, mocker: MockerFixture) -> None:
-        mocker.patch(
-            "huggingface_hub.file_download._are_symlinks_supported_in_dir",
-            {HF_HUB_CACHE: True},
-        )
+        """Called without argument, the default cache directory is the one probed and memoized."""
+        mocker.patch("huggingface_hub.file_download.os.symlink")  # symlink creation succeeds
+        memo: dict[str, bool] = {}
+        mocker.patch("huggingface_hub.file_download._are_symlinks_supported_in_dir", memo)
+
         assert are_symlinks_supported()
+        # `HF_HUB_CACHE` is read from the module and not imported: the `patch_constants` fixture
+        # (tests/conftest.py) repoints it to a fresh temporary directory for every test.
+        assert memo == {str(Path(constants.HF_HUB_CACHE).expanduser().resolve()): True}
 
     def test_are_symlinks_supported_disabled_by_env(self, mocker: MockerFixture) -> None:
         """Setting HF_HUB_DISABLE_SYMLINKS=1 forces are_symlinks_supported() to return False."""
         mocker.patch("huggingface_hub.file_download.constants.HF_HUB_DISABLE_SYMLINKS", True)
-        mocker.patch("huggingface_hub.file_download._are_symlinks_supported_in_dir", {HF_HUB_CACHE: True})
+        mocker.patch("huggingface_hub.file_download.os.symlink")  # symlink creation would succeed
+        memo: dict[str, bool] = {}
+        mocker.patch("huggingface_hub.file_download._are_symlinks_supported_in_dir", memo)
+
         assert not are_symlinks_supported()
+        assert memo == {}  # returned early: the directory has not even been probed
 
     def test_are_symlinks_supported_windows_specific_dir(self, mocker: MockerFixture) -> None:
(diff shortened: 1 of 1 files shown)


hanouticelina:
this is purely a test limitation on Windows without Developer Mode (the library itself handles missing symlink support correctly), so it has no impact on users

PR #4700, opened 2026-08-18
[Tree cache] Rebuild HEAD metadata from tree listing without Xet

p0we7:
## Summary

The tree-cache optimization (`_xet_file_metadata_from_tree_cache`) only kicked in when Xet was available: it returned `None` as soon as `is_xet_available()` was `False`, forcing a **per-file HEAD call** for every downloaded file. In setups where Xet must be disabled — e.g. `HF_HUB_DISABLE_XET=1` together with `HF_ENDPOINT` pointed at a self-hosted mirror/reverse proxy to download models in regions where `huggingface.co` is not directly reachable — this HEAD call cannot be relied on, and downloads of any repo with non-LFS metadata files fail hard.

This PR makes the tree cache usable **without Xet**: metadata (etag/size) is rebuilt from the cached `/tree` listing for both LFS and regular files, saving one round-trip per file and making proxy/mirror setups work.

## Why this is needed (real-world background)

For a China-based server we built a Hugging Face reverse proxy (nginx → `huggingface.co`) behind Cloudflare, and download models with:

```
HF_ENDPOINT=https://hf-mirror.example.com HF_HUB_DISABLE_XET=1 hf download <repo> --local-dir ./model
```

- `HF_HUB_DISABLE_XET=1` is mandatory here: with Xet enabled the download bytes go to `cas-server.xethub.hf.co` over the Xet protocol, bypassing the proxy entirely.
- With Xet disabled, every file needs a metadata HEAD call. Modern hub servers answer `HEAD /resolve/...` with a `307` to `/api/resolve-cache/...`, which the client follows.

The failure we hit: **Cloudflare (and similar edge proxies) strips the `Content-Length` header from HEAD responses** — in our testing, for any response > ~1KB whose content looks compressible, regardless of `cache-control: no-transform`, compression rules, or disabling every dynamic transformation setting (Automatic HTTPS Rewrites, Email Obfuscation, Rocket Loader, etc.). The result:

```
huggingface_hub.errors.FileMetadataError: Distant resource does not have a Content-Length.
```

every single non-LFS metadata file (README.md, config.json, tokenizer.json, manifests, ...) failed the download. LFS weight files (`*.safetensors`) were unaffected because their HEAD returns a `302` carrying `x-linked-size`/`x-linked-etag`, so the client never needs to follow up.

We also verified the workaround of `hf download --include "*.safetensors"` (weights only), and confirmed the tree-cache path is the right fix: the `/tree` listing already contains everything needed (`size`, `blob_id`, `lfs_sha256`, `lfs_size`, `xet_hash`) — it was just gated behind `is_xet_available()`.

## Changes

- Rename `_xet_file_metadata_from_tree_cache` → `_file_metadata_from_tree_cache` (it now serves all files, not just Xet ones).
- Remove the `is_xet_available()` early-return.
- LFS-tracked files: etag/size from `lfs_sha256`/`lfs_size` (mirroring the server HEAD response); `XetFileData` is attached only when Xet is available **and** the entry has a valid Xet hash — otherwise `xet_file_data=None` and the download falls back to the regular HTTP `/resolve` flow, which is exactly what a mirror/proxy setup wants.
- Regular (non-LFS) files: etag/size from the git blob (`blob_id`/`size`), matching what a HEAD call would return.
- Still returns `None` (→ caller does the HEAD call as before) when the tree listing or the entry metadata is not cached.

## Tests

- `tests/test_tree_cache.py`: 19 passed. Updated the three tests that asserted the old behavior (`test_non_xet_file_returns_none`, `test_xet_file_returns_none_when_xet_unavailable`, `test_get_metadata_heads_for_non_xet_file`) to assert the new one (metadata rebuilt without Xet, `xet_file_data=None`, no HEAD call).
- Manually verified end-to-end against the proxy described above: `hf download` of a full repo (metadata + weights) now succeeds, with bytes flowing through the reverse proxy.
- `ruff check` passes on both changed files.

---

> [!NOTE]
> **Medium Risk**
> Changes core download metadata resolution for all cached-tree downloads; incorrect etag/size from tree entries could affect cache keys or integrity checks, though behavior is limited to pinned commit hashes with existing tree cache.
> 
> **Overview**
> **Tree-cache metadata** no longer depends on Xet being enabled. `_xet_file_metadata_from_tree_cache` is renamed to `_file_metadata_from_tree_cache` and can rebuild per-file HEAD metadata (etag, size, URL, optional `XetFileData`) for **any** file in a cached `/tree` listing when the revision is a commit hash.
> 
> **LFS files** use `lfs_sha256` / `lfs_size` from the cache; `XetFileData` is attached only when Xet is available and the entry has a valid Xet hash—otherwise downloads use normal HTTP via `/resolve` (mirror/proxy friendly). **Non-LFS files** use `blob_id` / `size`. The old `is_xet_available()` gate and requirement that Xet/LFS fields all be present for every hit are removed.
> 
> Tests now expect regular files and Xet-disabled LFS files to skip the per-file HEAD call when the tree is cached; branch/tag revisions still HEAD as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86c6824a672267f28d98095bb5a2549c31448abf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -1686,7 +1686,7 @@ def _get_metadata_or_catch_error(
 
     # Skip the per-file HEAD call when the file metadata can be rebuilt from a tree listing cached on disk.
     if tree_cache_folder is not None and REGEX_COMMIT_HASH.match(revision):
-        tree_metadata = _xet_file_metadata_from_tree_cache(
+        tree_metadata = _file_metadata_from_tree_cache(
             tree_cache_folder=tree_cache_folder,
             repo_id=repo_id,
             repo_type=repo_type,
@@ -1830,7 +1830,7 @@ def _detach_tracebacks(error: BaseException) -> None:
         pending.extend(exc for exc in (current.__cause__, current.__context__) if exc is not None)
 
 
-def _xet_file_metadata_from_tree_cache(
+def _file_metadata_from_tree_cache(
     *,
     tree_cache_folder: str,
     repo_id: str,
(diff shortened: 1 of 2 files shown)


Wauplin:
Hi @p0we7, the tree listing cache has been added very recently (see https://github.com/huggingface/huggingface_hub/pull/4394). It is meant to be an optimization PR only, not to fix some proxy-related issues. This is why we chose to put it only on the Xet-path, as the pure python implementation is just a slow fallback solution, not something we want to optimize for. I'm therefore a bit reluctant over merging this PR as I would keep things simple in the pure-python part. 

I'll close this PR without merging for now. Reasoning is that fixing a proxy issue with a tree-listing cache feels quite hacky / more of a side effect. If we merge this PR, it means we start to "officially" support this use case which I don't think we want to.

What I'd recommend is 1. test on older versions of `huggingface_hub` if it was working (older than #4394) => I'm expecting not, but at least it confirm it's pre-existing the cache tree solution and 2. please open a separate issue describing the problem and the alternatives you have, and only then we can discuss potential mitigations.

p0we7:
Hi @Wauplin, thanks for the detailed explanation — completely understand keeping the pure-Python path simple and the tree cache as a pure optimization.

Following your suggestions:

1. **Verified on older versions (pre-#4394):** `huggingface_hub` v1.26.0 and v1.23.0 (both before the tree-listing cache, v1.27.0) reproduce the exact same `FileMetadataError: Distant resource does not have a Content-Length.` on metadata files through the proxy. So confirmed: this is a pre-existing limitation of the metadata HEAD flow, not something the tree cache introduced.

2. **Opened an issue** with the full background, our exhaustive testing (no-transform / compression rules / disabling all dynamic transformations are all ineffective against Cloudflare's HEAD Content-Length stripping), and a few alternative mitigations we'd be happy to implement: https://github.com/huggingface/huggingface_hub/issues/4741

Thanks again for the quick review!

PR #4705, opened 2026-08-20
[Download] Add HF_HUB_DOWNLOAD_MODE env var

guenhter:
Introduce a new env var `HF_HUB_DOWNLOAD_MODE` which has the values:
* `auto`: default. Current behaviour like if `HF_HUB_OFFLINE` is not set or `0`
* `prefer_offline`: Only make network communication if files are missing. If everything is available, it doesn't produce network traffic and not even checks for updates.
* `offline`: no network communication. Same as `HF_HUB_OFFLINE=1`

`HF_HUB_OFFLINE` was deprecated (but still functional) as `HF_HUB_DOWNLOAD_MODE` superseds it.

So the main change is `prefer_offline` so cached files skip the Hub freshness check while missing files still download.

## HF_HUB_DOWNLOAD_MODE manual tests

Manual Docker checks used `snapshot_download("hf-internal-testing/tiny-random-bert")` with a shared volume so the cache survived `--network none` runs.

| Step | Network | Mode | Cache | Result |
| --- | --- | --- | --- | --- |
| 1 | on | `offline` | empty | failed (`LocalEntryNotFoundError`) |
| 2 | on | `auto` | empty | downloaded all 10 files |
| 3 | on | `prefer_offline` | empty | downloaded all 10 files |
| 4 | off | `auto` | empty | failed (DNS `ConnectError`) — network really off |
| 5 | off | `offline` | from step 2 | reused cache |
| 6 | off | `prefer_offline` | from step 2 | reused cache |

## References

https://github.com/huggingface/huggingface_hub/issues/3201

---

> [!NOTE]
> **Medium Risk**
> Changes core download and revision-resolution behavior across cache and local_dir paths, though legacy offline semantics are preserved and coverage is extensive.
> 
> **Overview**
> Adds **`HF_HUB_DOWNLOAD_MODE`** (`auto`, `prefer_offline`, `offline`) as the single knob for Hub download behavior, with **`prefer_offline`** as the new middle ground: reuse cached files without freshness HEAD calls while still fetching missing files, pinned to the last known commit so revisions are not mixed.
> 
> **`HF_HUB_OFFLINE`** / **`TRANSFORMERS_OFFLINE`** are deprecated when the new variable is set; otherwise they still map to `offline` vs `auto` at import. **`is_offline_mode()`** remains true only for full `offline`; **`is_prefer_offline_mode()`** gates the download-only optimizations.
> 
> Download paths now honor **`prefer_offline`** in **`hf_hub_download`** (hub cache and **`local_dir`**, including metadata-based commit pinning), **`snapshot_download`** (early return on complete cached snapshots via **`_is_snapshot_complete`**), and **`HfApi.resolve_revision`**. Error messages and telemetry/offline HTTP handling point users at **`HF_HUB_DOWNLOAD_MODE=auto`** instead of unsetting **`HF_HUB_OFFLINE`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6129c5ff145f87cabfc7a6f00e20012182f8067a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## AI Generated

* is most of this AI generated: for sure it is
* did I blindly push this: no. I did a close review on the produced code (even though I'm not an expert in this library) AND checked with the docker tests, that the behaviour is what it should be and not break the old behaviour.

### DIFF
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -1785,6 +1785,7 @@ Copy-and-paste the text below in your GitHub issue.
 - HF_ASSETS_CACHE: /home/wauplin/.cache/huggingface/assets
 - HF_TOKEN_PATH: /home/wauplin/.cache/huggingface/token
 - HF_STORED_TOKENS_PATH: /home/wauplin/.cache/huggingface/stored_tokens
+- HF_HUB_DOWNLOAD_MODE: auto
 - HF_HUB_OFFLINE: False
 - HF_HUB_DISABLE_TELEMETRY: False
 - HF_HUB_DISABLE_PROGRESS_BARS: None
--- a/docs/source/en/package_reference/environment_variables.md
+++ b/docs/source/en/package_reference/environment_variables.md
@@ -76,6 +76,20 @@ Defaults to `"warning"`.
 
 For more details, see [logging reference](../package_reference/utilities#huggingface_hub.utils.logging.get_verbosity).
 
+### HF_HUB_DOWNLOAD_MODE
+
+Controls how `huggingface_hub` talks to the Hub when downloading files. Must be one of
+`{"auto", "prefer_offline", "offline"}` (case-insensitive).
+
+- `auto` (default): A HEAD request checks whether a newer version is available, then the cache is reused or the file is downloaded.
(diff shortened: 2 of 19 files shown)


hanouticelina:
Hi @guenhter, thanks but we'd rather close this PR. the pain point behind #3201 is already handled, `hf_hub_download` automatically falls back to the local cache whenever the freshness check fails, also we definitely don't want to to deprecate `HF_HUB_OFFLINE`

PR #4720, opened 2026-08-21
fix: detect packages in frozen environments (PyInstaller)

### LINKED ISSUE #4717: Xet downloads fall back to HTTP in PyInstaller-packaged applications
### Describe the bug

When a Python application is packaged with PyInstaller, downloads from Xet-backed repositories fall back to regular HTTP even when `hf_xet` is installed in the build environment.

When the same script is run directly with Python, `hf_xet` is detected and the download uses Xet successfully.

### Reproduction

Create a virtual environment and install the packages:

```powershell
python -m venv .venv
.\.venv\Scripts\python.exe -m pip install "huggingface_hub==1.3.1" "hf-xet==1.2.0" "pyinstaller==6.20.0" "pyinstaller-hooks-contrib==2026.5"
```

Create `reproduce.py`:

```python
import sys

from huggingface_hub.utils import logging
from huggingface_hub.utils._runtime import is_xet_available


output_dir = sys.argv[1]
logging.set_verbosity_debug()

print(f"is_xet_available(): {is_xet_available()}", flush=True)

hf_hub_download(
    repo_id="hf-internal-testing/tiny-random-bert",
    filename="pytorch_model.bin",
    local_dir=output_dir,
    force_download=True,
)
```

Run it directly:

```powershell
.\.venv\Scripts\python.exe reproduce.py direct-download
```

Output:

```text
is_xet_available(): True
Xet Storage is enabled for this repo. Downloading file from Xet Storage..
Download complete.
```

Package and run the same script:

```powershell
.\.venv\Scripts\pyinstaller.exe --clean --onedir --name hf-xet-repro reproduce.py
.\dist\hf-xet-repro\hf-xet-repro.exe packaged-download
```

Output:

```text
is_xet_available(): False
Xet Storage is enabled for this repo, but the 'hf_xet' package is not installed. Falling back to regular HTTP download.
Download complete.
```

### Logs

```shell

```

### System info

```shell
- huggingface_hub: 1.3.1
- hf_xet: 1.2.0
- Python: 3.13.1
- Platform: Windows 11
- PyInstaller: 6.20.0
- pyinstaller-hooks-contrib: 2026.5
- Build mode: onedir
```

syahra712:
# What does this PR do?

Fixes #4717

When a Python application is packaged with PyInstaller, `importlib.metadata.version()` cannot find package metadata even though the module itself is bundled. This causes `is_package_available("hf_xet")` to return `False`, so Xet downloads silently fall back to HTTP.

## Root cause

The runtime detection loop in `utils/_runtime.py` only uses `importlib.metadata.version()`. PyInstaller bundles modules but strips their `dist-info` metadata, so `PackageNotFoundError` fires and the package is marked `"N/A"`.

## Fix

Added `importlib.util.find_spec()` as a fallback after the metadata loop. When metadata lookup fails, `find_spec` checks whether the module itself is importable. The version is set to `"unknown"` since metadata isn't available, but `is_package_available()` now correctly returns `True`.

- No behavior change in standard Python environments (metadata lookup succeeds first)
- `get_xet_version()` returns `"unknown"` instead of `"N/A"` in frozen environments
- The fallback also benefits other optional dependencies bundled without metadata

## Who can review?

@Wauplin

### DIFF
--- a/src/huggingface_hub/utils/_runtime.py
+++ b/src/huggingface_hub/utils/_runtime.py
@@ -70,6 +70,15 @@ for candidate_name, package_names in _CANDIDATES.items():
             break
         except importlib.metadata.PackageNotFoundError:
             pass
+    # Fallback: in frozen environments (e.g. PyInstaller) metadata is often
+    # stripped while the module itself is bundled. Use find_spec as a fallback
+    # so that bundled packages are still detected.
+    if _package_versions[candidate_name] == "N/A":
+        try:
+            if importlib.util.find_spec(candidate_name) is not None:
+                _package_versions[candidate_name] = "unknown"
+        except (ModuleNotFoundError, ValueError):
+            pass
 
 
 def _get_version(package_name: str) -> str:


hanouticelina:
Hi, thanks for the PR but I'm closing it as we'd rather fix these things ourselves. by the way, did you run into the same issue mentioned in #4717 ?

syahra712:
> Hi, thanks for the PR but I'm closing it as we'd rather fix these things ourselves. by the way, did you run into the same issue mentioned in #4717 ?

No worries, totally understand! I didn't run into the issue myself — I spotted #4717 while scanning for open issues and the fix looked straightforward. Happy to leave it to you. Thanks for the quick response!

PR #4728, opened 2026-08-24
reject safetensors indexes that point at missing tensors

### LINKED ISSUE #4725: Safetensors metadata does not detect tensors referenced by index but absent from shards
### Describe the bug

`get_safetensors_metadata()` returns both the index `weight_map` and the parsed tensor metadata for every referenced shard, but it does not verify that an indexed tensor exists in the shard to which the index maps it.

This allows an internally inconsistent `SafetensorsRepoMetadata` object to be returned successfully. A related producer-side inconsistency was fixed in #4391; #4033 validates shard filenames and extensions, but neither validates index entries against shard headers on the reader side.

### Reproduction

Using current `main` (`ee0a4510f4bc15e26cb722be2efc2bdb01383ca0`):

```python

repo_id = "nightmedia/Qwen3.8-27B-Brainwaves"
revision = "4cf25125da7eb7bdf672545a027f92cdedab6b6a"

metadata = HfApi().get_safetensors_metadata(repo_id, revision=revision)
for tensor_name, shard_filename in metadata.weight_map.items():
    if tensor_name not in metadata.files_metadata[shard_filename].tensors:
        print(tensor_name, shard_filename)
```

The call currently succeeds and prints:

```text
mtp.fc.weight model-00012-of-00012.safetensors
```

The pinned index contains 1,199 entries, while the 12 parsed shard headers contain 1,198 tensors. No tensor payload, model code, Torch, or Transformers is needed to reproduce this.

### Proposed invariant

For every `(tensor_name, shard_filename)` pair in `weight_map`, require:

```python
tensor_name in files_metadata[shard_filename].tensors
```

I also checked 27 public Safetensors repositories (single-file and 2/3/4/8-shard models) and found no violations of this invariant.

Would you prefer this consistency check to be enforced directly inside `get_safetensors_metadata()`, or exposed in another way?

### System info

```text
- Python version: 3.12.2
- Installation method: pip (editable, dev dependencies)
```

gyanu2507:
Fixes #4725

`get_safetensors_metadata` (and the local equivalent) returns the index `weight_map` plus parsed shard headers, but never checks they agree. A repo can have an index entry whose tensor isn't in the mapped file, and the call still succeeds.

I hit this on `nightmedia/Qwen3.8-27B-Brainwaves` at `4cf25125` — `mtp.fc.weight` is in the index, missing from `model-00012-of-00012.safetensors`.

After parsing shards, require every `(tensor, shard)` in `weight_map` to exist in that shard's header. Raise `SafetensorsParsingError` if not.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -164,6 +164,7 @@ from .utils._auth import _get_token_from_environment, _get_token_from_file, _get
 from .utils._deprecation import _deprecate_arguments, _deprecate_method
 from .utils._http import _httpx_follow_relative_redirects_with_backoff
 from .utils._runtime import is_xet_available
+from .utils._safetensors import _assert_weight_map_matches_shard_headers
 from .utils._typing import CallableT
 from .utils._verification import collect_local_files, resolve_local_root, verify_maps
 from .utils.endpoint_helpers import _is_emission_within_threshold
@@ -6971,6 +6972,8 @@ class HfApi:
                 desc="Parse safetensors files",
                 tqdm_class=hf_tqdm,
             )
+
+            _assert_weight_map_matches_shard_headers(weight_map, files_metadata)
 
             return SafetensorsRepoMetadata(
                 metadata=index.get("metadata", None),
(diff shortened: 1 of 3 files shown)


Wauplin:
Thanks for the report and PR. However I don't think repo metadata validation should be done automatically. `get_safetensors_metadata` is correctly returning the metadata from the remote files. If there is a consistency issue in them, it must be checked downstream. If we merge this PR, we might break some workflows relying on `get_safetensors_metadata` on broken repos.

PR #4482, opened 2026-07-03
[Xet] Remove dead connection info helpers

Wauplin:
I realized that since the `XetSession` integration (#4116 cc @seanses) some legacy helpers are just dead code now.  This PR cleans them.

- Remove unused Python-side Xet connection info fetching/cache helpers.
- Drop stale tests and assertions that only covered that old path.

@seanses would you mind having a look at it to make sure I'm not breaking anything? (doesn't seem to be the case + it's internal stuff only but wanna make sure)

---

> [!NOTE]
> **Medium Risk**
> Large deletion in Xet auth plumbing; behavior should be unchanged if all transfers already use hf_xet token refresh, but any external code importing the removed helpers will break.
> 
> **Overview**
> Removes the legacy Python path for fetching and caching Xet CAS connection tokens (`XetConnectionInfo`, header parsing, `http_backoff` fetches, and the in-memory cache). **Upload/download now rely on `hf_xet`**, which refreshes tokens via `token_refresh_url` / per-file `refresh_route` from `XetFileData`.
> 
> **Public surface:** `huggingface_hub.utils` no longer exports `XetConnectionInfo`, `refresh_xet_connection_info`, or `fetch_xet_connection_info_from_repo_info`. **`xet_connection_info_refresh_url`** and file-metadata parsing stay.
> 
> **API cleanup:** `delete_repo` and `delete_bucket` no longer call `reset_xet_connection_info_cache_for_repo` because that cache is gone.
> 
> Tests drop coverage of the removed helpers (including the old pointer-file backward-compat download path and the post-delete token cache regression) and assert **`refresh_route`** on metadata instead of resolving full connection info.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17850927eb95b63517d57994b23276e8a03a60dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4732, opened 2026-08-24
[Xet] Cache connection info to avoid one token request per file

hanouticelina:
## Summary

- Since the `XetSession` migration (#4116, shipped in v1.19.0), building a xet download group eagerly calls the `xet-read-token` endpoint to discover the CAS endpoint. One group = one file, so `snapshot_download` on a many-files repo makes one Hub API call per file (e.g. 77k files at ~25 files/s ≈ 1,500 calls/min) and eventually gets rate-limited: `hf_xet` retries with backoff for ~6 minutes (looks like a stalled download), then raises `ConnectionError: ... 429 Too Many Requests`.
- This PR restores the Python-side connection info cache (added in #3534 for exactly this problem, removed as dead code in #4482 once #4116 stopped calling it) and seeds `endpoint` / `token` / `token_expiry_unix_secs` on each download group, so `hf_xet` skips the eager per-group token request. The refresh URL is still passed for mid-download token expiry.
- No `hf_xet` change or pin bump needed: the seeding kwargs exist since `hf-xet` 1.5.0 (our floor is 1.5.2).

## Review guide: mostly a partial revert of #4482

The restored helpers come from the pre-#4482 file (`git show 8fa58388~1:src/huggingface_hub/utils/_xet.py`) and the `xet_get()` call site is the exact pre-#4116 line. `XetConnectionInfo`, `refresh_xet_connection_info`, `_is_expired` and the cache constants are verbatim. Deltas:

- `_fetch_xet_connection_info_with_url` / `_cache_key` lose their `params` and `cache_key_prefix` arguments: their only consumer was the upload-path helper `fetch_xet_connection_info_from_repo_info`, which is not restored (uploads request one token per commit, no cache needed). Same reason `reset_xet_connection_info_cache_for_repo` and its `delete_repo` hooks (#3847) don't come back: download cache keys embed the commit SHA via the refresh URL, so a recreated repo can never hit the cache.
- New per-key lock around the fetch: collapses the cold-cache stampede (one fetch per worker, measured 4 with `max_workers=4`) into a single request per key; under a real 429, one thread backs off via `http_backoff` while the others wait instead of all hammering the endpoint. The lock is per cache key so cold fetches for different repos downloaded in parallel don't serialize; a single global lock guards the cache dict and the per-key lock registry.
- `parse_xet_connection_info_from_headers` takes `Mapping` instead of `dict` + `# type: ignore` at the call site (`resp.headers` is `httpx.Headers`; `ty` rejects the old form).
- Helpers stay private (not re-exported in `huggingface_hub.utils`), as preferred in #4116/#4482.

Only `xet_get()` is seeded — the bucket download path already uses a single group per call. Tests intentionally not added yet (will follow before un-drafting); the `tests/test_xet_download.py` change fixes an existing test that now needs the fetch mocked.

## Notes on the hf_xet side

Longer term this cache probably belongs at the `XetSession` level in `hf_xet` (keyed by refresh URL): the eager refresh on group build happens whenever `endpoint` isn't seeded, every binding has to work around it the same way (huggingface.js maintains its own copy of this exact cache), and mid-download token expiry is still refreshed per live group, uncached, on the Rust side. Once a session-level cache exists there, this Python-side one can be dropped again.

## Manual testing

Counted `xet-read-token` calls with a local reverse proxy in front of huggingface.co, downloading 10 files with `max_workers=4` — one call per file on `main`, one call total with this PR:

Before:

```
   10  GET /api/datasets/ArlingtonCL2/DogSpeak_Dataset/xet-read-token/34c9047dcff6744700ce772d6acb313884ba8b14
```

After:

```
    1  GET /api/datasets/ArlingtonCL2/DogSpeak_Dataset/xet-read-token/34c9047dcff6744700ce772d6acb313884ba8b14
```

Simulated a rate-limited token endpoint (proxy returns 429 after the first two calls): on `main` the download stalls for minutes then fails with `ConnectionError: ... 429 Too Many Requests ... /xet-read-token/...`; with this PR the same scenario succeeds with a single token call. Downloaded bytes are identical to the plain-HTTP path (`sha256` match with `HF_HUB_DISABLE_XET=1`), and `pytest tests/test_xet_utils.py tests/test_xet_download.py` passes (33 passed).

---

> [!NOTE]
> **Medium Risk**
> Touches download auth/token handling and shared caching under concurrency; wrong cache keys or expiry logic could cause failed downloads or stale tokens, but scope is limited to the Xet path and mirrors previously shipped behavior.
> 
> **Overview**
> Fixes **rate limiting and stalled multi-file Xet downloads** by restoring a cached Hub `xet-read-token` flow instead of letting each file’s download group trigger its own eager token fetch.
> 
> **`xet_get()`** now calls **`refresh_xet_connection_info()`** once per repo revision (keyed by refresh URL + auth) and passes **`endpoint`**, **`token`**, and **`token_expiry_unix_secs`** into **`new_file_download_group()`**, while still supplying **`token_refresh_url`** for mid-download refresh. **`utils/_xet.py`** adds **`XetConnectionInfo`**, header parsing, an in-memory cache with expiry buffer and size cap, and **per-key locks** so parallel workers don’t stampede the token endpoint on a cold cache.
> 
> Docs in **`file_download.py`** are updated to describe seeded tokens. **`tests/test_xet_download.py`** mocks the new fetch so existing header assertions keep working.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d66643eeba292cdab49bf955c37563720b7b4fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_xet.py
+++ b/src/huggingface_hub/utils/_xet.py
@@ -17,8 +17,8 @@ _REGEX_XET_HASH = re.compile(r"^[0-9a-fA-F]{64}$")
 XET_CONNECTION_INFO_SAFETY_PERIOD = 60  # seconds
 XET_CONNECTION_INFO_CACHE_SIZE = 1_000
 XET_CONNECTION_INFO_CACHE: dict[str, "XetConnectionInfo"] = {}
-_XET_CONNECTION_INFO_LOCKS: dict[str, threading.Lock] = {}
-_XET_CONNECTION_INFO_LOCKS_GUARD = threading.Lock()
+_XET_CONNECTION_INFO_LOCK = threading.Lock()  # guards the cache dict and the per-key lock registry
+_XET_CONNECTION_INFO_KEY_LOCKS: dict[str, threading.Lock] = {}
 
 
 def is_valid_xet_hash(xet_hash: str) -> bool:
@@ -166,14 +166,7 @@ def _fetch_xet_connection_info_with_url(
 
     # A per-key lock collapses concurrent workers into a single token request per expired/missing
     # key, without blocking fetches for other keys (e.g. other repos downloaded in parallel).
-    with _XET_CONNECTION_INFO_LOCKS_GUARD:
-        key_lock = _XET_CONNECTION_INFO_LOCKS.get(cache_key)
-        if key_lock is None:
-            if len(_XET_CONNECTION_INFO_LOCKS) >= XET_CONNECTION_INFO_CACHE_SIZE:
-                _XET_CONNECTION_INFO_LOCKS.pop(next(iter(_XET_CONNECTION_INFO_LOCKS)))
-            key_lock = _XET_CONNECTION_INFO_LOCKS[cache_key] = threading.Lock()
-
-    with key_lock:
+    with _key_lock(cache_key):
         cached_info = XET_CONNECTION_INFO_CACHE.get(cache_key)
         if cached_info is not None and not _is_expired(cached_info):
             return cached_info
@@ -186,17 +179,18 @@ def _fetch_xet_connection_info_with_url(
         if metadata is None:
             raise ValueError("Xet headers have not been correctly set by the server.")
 
-        # Delete expired cache entries
-        for k, v in list(XET_CONNECTION_INFO_CACHE.items()):
-            if _is_expired(v):
-                XET_CONNECTION_INFO_CACHE.pop(k, None)
+        with _XET_CONNECTION_INFO_LOCK:
+            # Delete expired cache entries
+            for k, v in list(XET_CONNECTION_INFO_CACHE.items()):
+                if _is_expired(v):
+                    XET_CONNECTION_INFO_CACHE.pop(k, None)
 
-        # Enforce cache size limit
-        if len(XET_CONNECTION_INFO_CACHE) >= XET_CONNECTION_INFO_CACHE_SIZE:
-            XET_CONNECTION_INFO_CACHE.pop(next(iter(XET_CONNECTION_INFO_CACHE)))
+            # Enforce cache size limit
+            if len(XET_CONNECTION_INFO_CACHE) >= XET_CONNECTION_INFO_CACHE_SIZE:
+                XET_CONNECTION_INFO_CACHE.pop(next(iter(XET_CONNECTION_INFO_CACHE)))
 
-        # Update cache
-        XET_CONNECTION_INFO_CACHE[cache_key] = metadata
+            # Update cache
+            XET_CONNECTION_INFO_CACHE[cache_key] = metadata
 
         return metadata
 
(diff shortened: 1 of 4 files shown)


hanouticelina:
bugbot run

src/huggingface_hub/utils/_xet.py
```diff
@@ -64,6 +78,132 @@ def parse_xet_file_data_from_response(response: httpx.Response, endpoint: str |
...
+    """
+    # Check cache first
+    cache_key = _cache_key(url, headers)
+    cached_info = XET_CONNECTION_INFO_CACHE.get(cache_key)
+    if cached_info is not None:
+        if not _is_expired(cached_info):
+            return cached_info
+
+    # The lock collapses concurrent workers into a single token request per expired/missing key.
+    with _XET_CONNECTION_INFO_LOCK:
```
seanses: Note that this lock remains held throughout `http_backoff`, which may block other independent Xet downloads for other repositories (is that a possible use case?).

seanses (approved):
Thanks for the investigation! The hf-xet related logic looks good!

src/huggingface_hub/utils/_xet.py
hanouticelina: yes true, the lock is global and is hold across the request, so two fetches for different repos would serialize, but the impact is bounded though: the cache is checked before acquiring the lock, so once a repo's token is cached (once per revision per token lifetime) its downloads never touch the lock, only overlapping cache misses wait

Wauplin: can't we make the lock depend on `cache_key`? doesn't feel too much to add and would avoid this corner case

Agree it's a corner case but I can imagine someone fetching e.g. the `default/validation/0000.parquet` file for many datasets repos in parallel (the datasets auto-converted parquet files) and if the Xet files are small the lock on first read can have an impact. Not a big use case but also not too big of a change to handle it.

Wauplin (approved):
Everything looks good to me except for this subtlety https://github.com/huggingface/huggingface_hub/pull/4732/changes#r3864122545

which can be handled like this

```py
_XET_CONNECTION_INFO_LOCK = threading.Lock()  # guards the cache dict and the per-key lock registry
_XET_CONNECTION_INFO_KEY_LOCKS: dict[str, threading.Lock] = {}


def _key_lock(cache_key: str) -> threading.Lock:
    """Return the lock guarding token fetches for `cache_key`, creating it if needed."""
    with _XET_CONNECTION_INFO_LOCK:
        return _XET_CONNECTION_INFO_KEY_LOCKS.setdefault(cache_key, threading.Lock())
```

(and I think that way we are perfectly thread safe)

src/huggingface_hub/utils/_xet.py
hanouticelina: yes, done

hanouticelina:
failing tests are unrelated and should be fixed by https://github.com/huggingface/huggingface_hub/pull/4753

PR #4739, opened 2026-08-25
[Download] Follow redirects between Hub hosts when resolving files

### LINKED ISSUE #4637: snapshot_download fails with misleading LocalEntryNotFoundError when HF_ENDPOINT mirror 308-redirects to the Hub (x-repo-commit missing); relative-redirect follower drops query string

hanouticelina:
Fixes #4637.

## Summary

- When resolving a file, the client only followed redirects with a relative `Location`. An endpoint answering with an absolute redirect to another Hub host (e.g. a mirror set as `HF_ENDPOINT` redirecting to `huggingface.co`) was never followed: the client kept the raw `308`, found no `X-Repo-Commit` header on it, and failed with a misleading `LocalEntryNotFoundError: ... check your connection`.
- New rule: follow a redirect when the target stays on the same host or on a known Hub host (`constants.HF_URL_HOSTS`, i.e. `huggingface.co`, `hf.co`, staging, and the host of the configured `HF_ENDPOINT`). Stop at any other host: for LFS files the metadata lives on the redirect response itself, and the `Authorization` header must not be forwarded to a CDN.
- More accurate error when metadata is genuinely missing: the message now names the URL and the missing header and points at `HF_ENDPOINT`, instead of blaming the connection.

---

> [!NOTE]
> **Medium Risk**
> Changes core HEAD/resolve and auth-forwarding behavior for downloads; logic is tightened around trusted Hub hosts but misconfigured `HF_URL_HOSTS` could affect redirects or token forwarding.
> 
> **Overview**
> Fixes file resolution when `HF_ENDPOINT` (or another Hub mirror) answers with an **absolute redirect** to a known Hub host (e.g. `huggingface.co`). The client previously only chased **relative** `Location` headers, so it could stop on a `308` without `X-Repo-Commit` and surface a misleading “check your connection” cache error.
> 
> **Redirect policy:** `_httpx_follow_relative_redirects_with_backoff` is replaced by `_httpx_follow_hub_redirects_with_backoff`, which follows redirects while the target stays on the **same host** or on `constants.HF_URL_HOSTS` (public Hub, `hf.co`, staging, configured endpoint). Redirects to CDNs/storage are **not** followed so LFS metadata on the redirect response is preserved and **`Authorization` is not sent off-Hub**. `_is_same_or_hub_host` is shared for HEAD chasing and for stripping auth on CDN download URLs.
> 
> **Errors:** Missing `X-Repo-Commit` now names the URL, the header, and `HF_ENDPOINT`. `FileMetadataError` during cache fallback is re-raised as `LocalEntryNotFoundError` with a configuration-focused message instead of a generic connectivity warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 809e4fbbf0bac07911df52b8c0f49e2402fdb110. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -691,6 +691,9 @@ def http_stream_backoff(
     )
 
 
+_MAX_REDIRECTS = 20  # same bound as httpx's default max_redirects
+
+
 def _httpx_follow_hub_redirects_with_backoff(
     method: HTTP_METHOD_T, url: str, *, retry_on_errors: bool = False, **httpx_kwargs
 ) -> httpx.Response:
@@ -722,7 +725,7 @@ def _httpx_follow_hub_redirects_with_backoff(
         {} if retry_on_errors else {"retry_on_exceptions": (), "retry_on_status_codes": ()}
     )
 
-    for _ in range(20):  # same bound as httpx's default max_redirects
+    for _ in range(_MAX_REDIRECTS):
         response = http_backoff(
             method=method,
             url=url,
@@ -743,7 +746,7 @@ def _httpx_follow_hub_redirects_with_backoff(
 
         url = target
 
-    raise httpx.TooManyRedirects(f"Exceeded 20 redirects while resolving '{url}'.")
+    raise httpx.TooManyRedirects(f"Exceeded {_MAX_REDIRECTS} redirects while resolving '{url}'.")
 
 
 def _is_same_or_hub_host(url: str, target: str) -> bool:
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -701,11 +701,8 @@ def _httpx_follow_hub_redirects_with_backoff(
 
     Used to fetch HEAD /resolve on repo or bucket files.
 
-    A redirect is followed when it targets the host of the current request (e.g. a renamed repository, or the
-    `/api/resolve-cache/` route) or another known Hub host (e.g. an `HF_ENDPOINT` mirror redirecting to
-    `huggingface.co`). Redirects to any other host are *not* followed: those point to a CDN or storage bucket, the
-    file metadata is already carried by the redirect response itself, and the authorization header must not be
-    forwarded off the Hub.
+    Redirects to the same host or another Hub host are followed. Redirects to any other host (CDN, storage
+    bucket) are not: the file metadata is on the redirect response itself and the auth header must not leave the Hub.
 
     A backoff mechanism retries the HTTP call on errors (429, 5xx, timeout, network errors).
 
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -691,9 +691,6 @@ def http_stream_backoff(
     )
 
 
-_MAX_HUB_REDIRECTS = 10
-
-
 def _httpx_follow_hub_redirects_with_backoff(
     method: HTTP_METHOD_T, url: str, *, retry_on_errors: bool = False, **httpx_kwargs
 ) -> httpx.Response:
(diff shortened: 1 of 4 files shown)


hanouticelina (commented):
✅

src/huggingface_hub/utils/_http.py
```diff
@@ -728,18 +732,24 @@ def _httpx_follow_relative_redirects_with_backoff(
...
-        if 300 <= response.status_code <= 399:
-            parsed_target = urlparse(response.headers["Location"])
-            if parsed_target.netloc == "":
-                # Relative redirect -> update URL and retry
-                url = urlparse(url)._replace(path=parsed_target.path).geturl()
-                continue
+        if not response.has_redirect_location:
+            return response
+
+        target = urljoin(url, response.headers["Location"])
```
Wauplin, line 738: TIL (again?) `urljoin` behavior is so confusing 😵 

https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin

Wauplin (approved):
LGTM, thank you!

src/huggingface_hub/utils/_http.py
hanouticelina: lol Claude sent me to the same stackoverflow page because i didn't understand much in `urljoin` behavior

Wauplin: well 14 years ago and still top 1 google result so we are not the only ones^^

Puneet15556:
Late follow-up on this after it landed — there's one case the host check doesn't cover.

`_is_same_or_hub_host` compares hostnames only, so a `Location` of `http://huggingface.co/...` is judged to be a Hub host. Both call sites then act on it: `_httpx_follow_hub_redirects_with_backoff` follows the redirect, and `_get_metadata_or_catch_error` keeps the `authorization` header for the download. The token can therefore go out over plaintext http.

Verified against `main`:

| request | target | treated as Hub |
| --- | --- | --- |
| `https://hf-mirror.com/...` | `https://huggingface.co/...` | yes (correct) |
| `https://hf-mirror.com/...` | `http://huggingface.co/...` | **yes** |
| `https://hf-mirror.com/...` | `http://hf-mirror.com/...` | **yes** |
| `https://hf-mirror.com/...` | `https://cdn.example.com/...` | no (correct) |

It needs a `Location` that the endpoint controls, so in practice this is a custom `HF_ENDPOINT` mirror rather than the default endpoint — I'd call it hardening rather than a live vulnerability. But since the point of the host check is to keep the token on the Hub, the downgrade seems worth closing as well. #4648 had ended up with an explicit `https` requirement for the same reason before it was superseded.

The change is small — refuse the target when it downgrades https to http, leaving endpoints deliberately configured over http untouched:

```python
parsed_url, parsed_target = urlparse(url), urlparse(target)
if parsed_url.scheme == "https" and parsed_target.scheme != "https":
    return False
```

I also wrote tests for `_httpx_follow_hub_redirects_with_backoff` and `_is_same_or_hub_host`, which currently have none: target query preserved, base query not inherited, dot-segments resolved, absolute redirect to a Hub host followed, redirect off the Hub returned as-is, downgrade refused, and the `_MAX_REDIRECTS` bound raising `TooManyRedirects`. The three downgrade assertions fail without the change. Locally `tests/test_utils_http.py` is 80 passed and `tests/test_file_download.py` 63 passed / 4 skipped.

It's ready on a branch: https://github.com/Puneet15556/huggingface_hub/tree/fix/hub-redirect-scheme-and-tests — I tried to open it as a PR but PR creation from forks appears to be restricted on this repo at the moment. Happy to open one as soon as that's possible, or a maintainer is welcome to take the branch; equally happy to send just the tests if the scheme change isn't wanted.

PR #4740, opened 2026-08-25
[HfApi] Validate safetensors index entries

### LINKED ISSUE #4725: Safetensors metadata does not detect tensors referenced by index but absent from shards

OllieinCanada:
Fixes #4725

## What does this PR do?

Reject inconsistent sharded safetensors indexes when a tensor is mapped to a shard that does not actually contain it.

Previously, `get_safetensors_metadata()` trusted `weight_map` after parsing every referenced shard. That could return metadata successfully even when a tensor existed only in a different shard than the index claimed, hiding corrupt repository metadata from callers.

The parser now validates each tensor against the exact shard named by `weight_map` and raises `SafetensorsParsingError` with the repository, tensor, and shard context when the mapping is inconsistent.

## Tests

- regression fails before the fix: expected `SafetensorsParsingError`, but none was raised
- regression passes after the fix
- existing safetensors API tests: 7 passed, including single-file, multi-shard, malformed, revision, and large-header cases
- Ruff format and lint checks pass

The regression deliberately puts the misplaced tensor in another parsed shard, proving validation is shard-specific rather than merely checking that the tensor exists somewhere in the repository.

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -6972,6 +6972,13 @@ class HfApi:
                 tqdm_class=hf_tqdm,
             )
 
+            for tensor_name, filename in weight_map.items():
+                if tensor_name not in files_metadata[filename].tensors:
+                    raise SafetensorsParsingError(
+                        f"Safetensors index for '{repo_id}' is inconsistent: tensor '{tensor_name}' is mapped to "
+                        f"'{filename}' but is not present in that file."
+                    )
+
             return SafetensorsRepoMetadata(
                 metadata=index.get("metadata", None),
                 sharded=True,
--- a/tests/test_safetensors_metadata.py
+++ b/tests/test_safetensors_metadata.py
@@ -1,0 +1,74 @@
+# Copyright 2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


Wauplin:
closing with same reason as https://github.com/huggingface/huggingface_hub/pull/4728#issuecomment-5412176619

PR #4743, opened 2026-08-26
fix: remove deprecated labels/tags arguments from zero-shot and list_datasets

### LINKED ISSUE #4742: Removed `labels`/`tags` args still documented, 3 API reference examples raise TypeError
## Describe the bug

Two deprecated arguments were removed in the past. In both cases the cleanup was applied to one method but not to its sibling, so the removed argument is still documented and still used in the examples.

`labels`, renamed to `candidate_labels`, deprecated in #2668 and removed in #2878:

- `zero_shot_classification` lost the parameter and the handling code, but its `Args:` block still documents `labels` (`inference/_client.py:3124`) and its second example still calls with `labels=` (`inference/_client.py:3181`).
- `zero_shot_image_classification` lost the handling code and the `@_deprecate_arguments` decorator, but `labels: list[str] = None` is still in the signature (`inference/_client.py:3218`). Nothing reads it. The `Args:` entry (`:3228`) and the example (`:3253`) were left behind too.

`tags`, deprecated in #3318 and removed in #3464:

- `list_models` was cleaned.
- `list_datasets` still documents `tags` (`hf_api.py:2681`) and its example still calls `api.list_datasets(tags="fiftyone")` (`hf_api.py:2736`).

`AsyncInferenceClient` is generated from `InferenceClient`, so it carries both zero-shot methods unchanged (`inference/_generated/_async_client.py:3175`, `:3271`, `:3281`).

`HfApi`, `InferenceClient` and `AsyncInferenceClient` are all rendered with `[[autodoc]]`, so the phantom arguments and the failing examples are published on the API reference pages.

Two user-visible consequences:

1. Three examples copied from the reference docs raise `TypeError`.
2. `zero_shot_image_classification(..., labels=[...])` is accepted and dropped. Someone migrating off the old name gets no error and no labels sent to the provider.

`utils/check_inference_input_params.py` does not catch this. It runs only for the methods listed in `PARAMETERS_TO_SKIP` (`chat_completion` and `text_generation`), and it checks that signature parameters are documented, never that documented parameters exist.

Present on `main` (1.29.0.dev0) and on v1.28.0.

## Reproduction

```python
import inspect

from huggingface_hub import HfApi, InferenceClient

client = InferenceClient()
api = HfApi()

# 1. Example from the `zero_shot_classification` docstring
try:
    client.zero_shot_classification(
        text="I really like our dinner and I'm very happy. I don't like the weather though.",
        labels=["positive", "negative", "pessimistic", "optimistic"],
        multi_label=True,
        hypothesis_template="This text is {} towards the weather",
    )
except TypeError as e:
    print("1.", e)

# 2. Example from the `zero_shot_image_classification` docstring
try:
    client.zero_shot_image_classification(
        "https://upload.wikimedia.org/wikipedia/commons/thumb/4/43/Cute_dog.jpg/320px-Cute_dog.jpg",
        labels=["dog", "cat", "horse"],
    )
except TypeError as e:
    print("2.", e)

# 3. Example from the `list_datasets` docstring
try:
    api.list_datasets(tags="fiftyone")
except TypeError as e:
    print("3.", e)

# 4. `labels` is still in the signature
print("4.", "labels" in inspect.signature(client.zero_shot_image_classification).parameters)
```

The dead parameter, without any network call. `labels` carries 3 classes and `candidate_labels` carries 1. The guard only looks at `candidate_labels`, which shows `labels` never reaches the request:

```python

InferenceClient().zero_shot_image_classification(b"", ["dog"], labels=["dog", "cat", "horse"])
# ValueError: You must specify at least 2 classes to compare.
```

## Logs

```shell
1. InferenceClient.zero_shot_classification() got an unexpected keyword argument 'labels'
2. InferenceClient.zero_shot_image_classification() missing 1 required positional argument: 'candidate_labels'
3. HfApi.list_datasets() got an unexpected keyword argument 'tags'
4. True
```

## System info

```txt
- Platform: Windows-10-10.0.26200-SP0
- Python version: 3.10.11
- Installation method: pip
- httpx: 0.28.1
- hf_xet: 1.2.0
```

## Suggested fix

Finish both removals the same way they were finished on the sibling methods:

- drop `labels` from the `zero_shot_image_classification` signature and from the three `Args:` blocks,
- use `candidate_labels=` in the two inference examples,
- use `api.list_datasets(filter="fiftyone")` in the `list_datasets` example. That is what the deprecation message pointed at, and it returns the same datasets:

```python
>>> [d.id for d in HfApi().list_datasets(filter="fiftyone", limit=3)]
['Voxel51/APAC-Egocentric-Stereo', 'Voxel51/spatial_lm_dataset', 'Voxel51/STONE']
```

- regenerate `_async_client.py` with `make style`.

Everything above is a docstring correction except one line, the `labels` parameter in the `zero_shot_image_classification` signature. Dropping it turns a silently ignored keyword into a `TypeError`. The deprecation message said it would go in `>=0.30.0` and it already does nothing, so no working call breaks, but tell me if you would rather keep it as a no-op.

I'd like to open a PR for this if you're happy with the approach.

abhi-0203:
Closes #4742

Removed deprecated `labels` argument from `zero_shot_classification` and `zero_shot_image_classification` docstrings and signature. Updated example in `zero_shot_image_classification` to use `candidate_labels`. Updated `list_datasets` example to use `filter="tag:fiftyone"` instead of deprecated `tags`.

This finishes the cleanup that was partially done on sibling methods.

---

> [!NOTE]
> **Low Risk**
> Doc and deprecated-parameter removal only; no runtime behavior changes beyond dropping an unused `labels` kwarg on one method.
> 
> **Overview**
> Finishes deprecation cleanup for Hugging Face Hub client docs and one inference API surface.
> 
> **InferenceClient:** Removes the deprecated `labels` argument from `zero_shot_image_classification` (signature and docs) and strips `labels` from `zero_shot_classification` docs. The image-classification doc example now uses **`candidate_labels`** instead of `labels`.
> 
> **HfApi:** Updates the `list_datasets` doc example from **`tags="fiftyone"`** to **`filter="fiftyone"`**, matching the supported filtering API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea1a7eb5ed6d9fae9eb47e43a92d09a162231e67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -2733,7 +2733,7 @@ class HfApi:
         ... )
 
         # List FiftyOne datasets (identified by the tag "fiftyone" in dataset card)
-        >>> api.list_datasets(tags="fiftyone")
+        >>> api.list_datasets(filter="tag:fiftyone")
         ```
 
         Example usage with the `search` argument:
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -2733,7 +2733,7 @@ class HfApi:
         ... )
 
         # List FiftyOne datasets (identified by the tag "fiftyone" in dataset card)
-        >>> api.list_datasets(filter="tag:fiftyone")
+        >>> api.list_datasets(filter="fiftyone")
         ```
 
         Example usage with the `search` argument:
--- a/src/huggingface_hub/inference/_client.py
+++ b/src/huggingface_hub/inference/_client.py
@@ -3121,8 +3121,6 @@ class InferenceClient:
                 The input text to classify.
             candidate_labels (`list[str]`):
                 The set of possible class labels to classify the text into.
-            labels (`list[str]`, *optional*):
-                (deprecated) List of strings. Each string is the verbalization of a possible label for the input text.
             multi_label (`bool`, *optional*):
                 Whether multiple candidate labels can be true. If false, the scores are normalized such that the sum of
                 the label likelihoods for each sequence is 1. If true, the labels are considered independent and
@@ -3214,8 +3212,6 @@ class InferenceClient:
         *,
         model: str | None = None,
         hypothesis_template: str | None = None,
-        # deprecated argument
-        labels: list[str] = None,  # type: ignore
     ) -> list[ZeroShotImageClassificationOutputElement]:
         """
         Provide input image and text labels to predict text labels for the image.
@@ -3225,8 +3221,6 @@ class InferenceClient:
                 The input image to caption. It can be raw bytes, an image file, a URL to an online image, or a PIL Image.
             candidate_labels (`list[str]`):
                 The candidate labels for this image
-            labels (`list[str]`, *optional*):
-                (deprecated) List of string possible labels. There must be at least 2 labels.
             model (`str`, *optional*):
                 The model to use for inference. Can be a model ID hosted on the Hugging Face Hub or a URL to a deployed
                 Inference Endpoint. This parameter overrides the model defined at the instance level. If not provided, the default recommended zero-shot image classification model will be used.
@@ -3250,7 +3244,7 @@ class InferenceClient:
 
         >>> client.zero_shot_image_classification(
         ...     "https://upload.wikimedia.org/wikipedia/commons/thumb/4/43/Cute_dog.jpg/320px-Cute_dog.jpg",
-        ...     labels=["dog", "cat", "horse"],
+        ...     candidate_labels=["dog", "cat", "horse"],
         ... )
         [ZeroShotImageClassificationOutputElement(label='dog', score=0.956),...]
         ```


src/huggingface_hub/hf_api.py
```diff
@@ -2733,7 +2733,7 @@ def list_datasets(
         ... )
 
         # List FiftyOne datasets (identified by the tag "fiftyone" in dataset card)
-        >>> api.list_datasets(tags="fiftyone")
+        >>> api.list_datasets(filter="tag:fiftyone")
```
```suggestion
        >>> api.list_datasets(filter="fiftyone")
```

Wauplin (approved):
thanks for spotting, and yes makes sense

(note: it's not a breaking change since it was already broken)

Wauplin:
thanks for the report! I just merged https://github.com/huggingface/huggingface_hub/pull/4745 superseeding this PR

PR #4752, opened 2026-08-26
Edit files in-place with open modes "e" and "eb" (and enable "a" and "ab" too)

### LINKED ISSUE #4153: [Buckets] Range-aware file write
Follow-up of https://github.com/huggingface/xet-core/pull/717 and especially @lhoestq 's suggestion in https://github.com/huggingface/xet-core/pull/717#issuecomment-4253676022 for the Python API. I'm starting this separate issue to avoid polluting the Xet implementation PR.

_Original comment from @lhoestq:_ 

> Forgot to share earlier but for those who want to play with it in python I made a simple integration in `huggingface_hub`:
> 
> ```python
> """ 
> This requires 
> 
> - https://github.com/lhoestq/xet-core/tree/lhoestq/py-upload_ranges
> - https://github.com/huggingface/huggingface_hub/tree/upload-ranges
> """ 
> 
> from huggingface_hub.hf_api import HfApi, FileRange, hffs
> 
> hffs.write_text("buckets/username/my-bucket/data.txt", "Hello world")
> 
> ranges = [FileRange(start=6, end=11, content=b"you !")]
> HfApi()._batch_bucket_file_ranges("username/my-bucket", "data.txt", ranges=ranges)
> 
> print(hffs.read_text("buckets/username/my-bucket/data.txt"))
> # Hello you !
> ```
> 
> Also sharing some ideas of what we can have in the future, since this requires the ability for new ranges to be smaller or longer than the overwritten range:
> 
> ```python
> """
> (Brainstorming)
> Example of filesystem API we could have:
> - similar interface as python lists operations
> - open(..., "m") is for for mutate/modify
> - replace/insert/remove bytes in the middle of the file
> """
> from huggingface_hub import hffs
> 
> path = "buckets/username/my-bucket/data.txt"
> hffs.write_text(path, "Hello world")
> 
> with hffs.open(path, "m") as f:
>    # do stuff here
> 
> # List-like API:
> f[6:] = "all"
> f[:0] = "Welcome. ")
> del f[:5]
> 
> # Regular methods
> # f[6:] = "all"
> f.seek(6)
> f.write("all")
> f.truncate()
> # f[:0] = "Welcome. ")
> f.seek(0)
> f.insert("Welcome. ")
> # del f[:5]
> f.seek(0)
> f.delete(5)
> 
> # One-liners for convenience
> # f[6:] = "all"
> f.write("all", index=6)
> f.truncate()
> # f[:0] = "Welcome. ")
> f.insert("Welcome. ", index=0)
> # del f[:5]
> f.delete(5, index=0)
> ```
> 
> cc @Wauplin for viz and feedbacks

lhoestq:
Implements `HfFileSystemEditFile` which is available via `hffs.open("hf://buckets/..", "eb")` and its methods:

- `edit((start, end), data)`
- `insert(loc, data)`
- `append(data)`
- `delete(loc, length)`
- `truncate(size=None)`

This also enables append mode "a" and "ab".

This only uploads the new data (and neighboring chunks), which enables efficient edits even on big files.

Close https://github.com/huggingface/huggingface_hub/issues/4153
Related to https://github.com/huggingface/huggingface.js/pull/2407
Follows https://github.com/huggingface/xet-core/pull/717

Still in draft mode until https://github.com/huggingface/xet-core/pull/951 is merged, but feel free to review already !

PS: Currently editing a big files still requires downloading ~64MB of data from neighboring chunks to generate verification and range hashes. Hopefully this limitation can be removed soon

## Feedback on the mode name

I was thinking of naming it "mutate" for `"m"` and `"mb"` but ended up with "edit" for `"e"` and `"eb"`

## Implementation details

It uses a buffer to keep the edits in memory and send them once the buffer is big enough and enough time has passed. Once ready, the buffer is sent to the Hub in a background thread.

The buffer uses `bytearray` to be able to efficiently support fast small appends.

I had to implement `HfApi.edit_bucket_file()` which calls hf_xet.

## Try it yourself

Requires hf_xet from https://github.com/huggingface/xet-core/pull/951 to get the python bindings for `XetSession.new_range_upload()`

Some examples:

- Write logs progressively using the append mode "a":

```py
from huggingface_hub import hffs

with hffs.open("buckets/username/my-bucket/logs.txt", "a") as f:
    for log in logs:
        f.write(log)
```

- Edit a file header using the edit mode "e":

```py
from huggingface_hub import hffs

header_length = 16
new_header = b"MY_NEW_HEADER_00"
with hffs.open("buckets/username/my-bucket/data.bin", "eb") as f:
    f.edit((0, header_length), new_header)
```

- Remove a certain line using the edit mode "e":

```py
from huggingface_hub import hffs

with hffs.open("buckets/username/my-bucket/doc.txt", "e") as f:
    for line in f:
        if line == "this is a bad line\n":
            break
    f.delete(loc=f.loc - len(line), length=len(line))
```

## TODO

- [ ] tests

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -14701,7 +14701,7 @@ class HfApi:
         if xet_file_data is None:
             raise ValueError(f"Could not parse xet file data for '{remote_path}' in bucket '{bucket_id}'.")
 
-        size = response.headers.get("Content-Length")
+        size = response.headers.get("X-Linked-Size")
         if size is None:
             raise ValueError(f"Could not get size for '{remote_path}' in bucket '{bucket_id}'.")
 
@@ -14957,6 +14957,92 @@ class HfApi:
             quiet=quiet,
             token=token,
         )
+
+    def mutate_bucket_file(
+        self,
+        *,
+        bucket_id: str,
+        remote_path: str,
+        edit: list[tuple[tuple[int, int], bytes]] | None = None,
+        insert: list[tuple[int, bytes]] | None = None,
+        delete: list[tuple[int, int]] | None = None,
+        token: bool | str | None = None,
+    ) -> None:
+        # TODO(QL): TRY IT OUT !!! (+ debug lol)
+        from .utils._xet import (
+            XetTokenType,
+            abort_xet_session,
+            get_xet_session,
+            xet_connection_info_refresh_url,
+            xet_headers_without_auth,
+        )
+        from .utils._xet_progress_reporting import XetUploadProgressReporter
+
+        if edit:
+            edit = [((start, end), data) for (start, end), data in edit if start < end or (start == end and len(data) > 0)]
+        if insert:
+            insert = [(loc, data) for loc, data in insert if len(data) > 0]
+        if delete:
+            delete = [(loc, length) for loc, length in delete if length > 0]
+        if not (edit or insert or delete):
+            return
+
+        headers = self._build_hf_headers(token=token)
+
+        if not are_progress_bars_disabled():
+            _progress = XetUploadProgressReporter(total_files=1)
+        else:
+            _progress = None
+
+        refresh_url = xet_connection_info_refresh_url(
+            token_type=XetTokenType.WRITE,
+            repo_id=bucket_id,
+            repo_type="bucket",
+            endpoint=self.endpoint,
+        )
+        xet_headers = xet_headers_without_auth(headers)
+
+        owns_progress = _progress is None
+        if _progress is not None:
+            progress = _progress
+            progress.reset_for_next_commit()
+            progress_callback = progress.update_progress
+        elif not are_progress_bars_disabled():
+            progress = XetUploadProgressReporter()
+            progress_callback = progress.update_progress
+        else:
+            progress, progress_callback = None, None
+        session = get_xet_session()
+        file_metadata = self.get_bucket_file_metadata(
+            bucket_id=bucket_id,
+            remote_path=remote_path,
+            token=token
+        )
+        try:
+            with session.new_range_upload(
+                file_metadata.xet_file_data.file_hash,
+                file_metadata.size,
+                token_refresh_url=refresh_url,
+                token_refresh_headers=headers,
+                custom_headers=xet_headers,
+                progress_callback=progress_callback,
+            ) as commit:
+                if edit:
+                    for (start, end), data in edit:
+                        commit.insert(start, end).write(data)
+                if insert:
+                    for loc, data in insert:
+                        commit.insert(loc, len(data)).write(data)
+                if delete:
+                    for loc, length in delete:
+                        commit.delete(loc, loc + length)
+        except KeyboardInterrupt:
+            abort_xet_session()
+            raise
+        finally:
+            if owns_progress and progress is not None:
+                progress.close()
+        return
 
 
 def _parse_revision_from_pr_url(pr_url: str) -> str:
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -14968,7 +14968,69 @@ class HfApi:
         delete: list[tuple[int, int]] | None = None,
         token: bool | str | None = None,
     ) -> None:
-        # TODO(QL): TRY IT OUT !!! (+ debug lol)
+        """Mutate an existing file in a bucket in-place, only re-uploading the parts the caller actually rewrites.
+
+        See [`HfFileSystemMutateFile`] to use this feature with a file-like API.
+
+        Args:
+            bucket_id (`str`):
+                The ID of the bucket (e.g. `"username/my-bucket"`).
+            remote_path (`str`):
+                the file to mutate.
+            edit (`list[tuple[tuple[int, int], bytes]]`, *optional*):
+                List edits to apply, in the form `((start, end), data)`.
+                Ranges [`start`, `end`) are replaced with `data`, which can
+                be of any size (not necessarily the size of the replaced range).
+            insert (`list[tuple[int, bytes]]`, *optional*):
+                List of inserts to apply, in the form `(loc, data)`.
+                The content of `data` is inserted at location `loc`, and shifts
+                the rest of the file.
+            delete (`list[tuple[int, bytes]]`, *optional*):
+                List of deletes to apply, in the form `(loc, length)`.
+                The range [`loc`, `loc + length`) is deleted.
+        """
+        from .utils._xet_progress_reporting import XetUploadProgressReporter
+
+
+        if not are_progress_bars_disabled():
+            _progress = XetUploadProgressReporter(total_files=1)
+        else:
+            _progress = None
+
+        file_metadata = self.get_bucket_file_metadata(
+            bucket_id=bucket_id,
+            remote_path=remote_path,
+            token=token
+        )
+        self._mutate_bucket_file(
+            bucket_id=bucket_id,
+            remote_path=remote_path,
+            edit=edit,
+            insert=insert,
+            delete=delete,
+            _progress=_progress,
+            _file_hash=file_metadata.xet_file_data.file_hash,
+            _file_size=file_metadata.size
+        )
+
+
+    def _mutate_bucket_file(
+        self,
+        *,
+        bucket_id: str,
+        remote_path: str,
+        edit: list[tuple[tuple[int, int], bytes]] | None = None,
+        insert: list[tuple[int, bytes]] | None = None,
+        delete: list[tuple[int, int]] | None = None,
+        token: bool | str | None = None,
+        _progress: XetUploadProgressReporter | None = None,
+        _file_hash: str | None = None,
+        _file_size: int | None = None,
+    ) -> str:
+        """
+        Internal method: process a single batch of bucket file mutate operations (upload to XET + call /batch).
+        Returns the new file's xet hash.
+        """
         from .utils._xet import (
             XetTokenType,
             abort_xet_session,
@@ -14987,13 +15049,30 @@ class HfApi:
         if not (edit or insert or delete):
             return
 
-        headers = self._build_hf_headers(token=token)
-
-        if not are_progress_bars_disabled():
-            _progress = XetUploadProgressReporter(total_files=1)
+        owns_progress = _progress is None
+        if _progress is not None:
+            progress = _progress
+            progress.reset_for_next_commit()
+            progress_callback = progress.update_progress
+        elif not are_progress_bars_disabled():
+            progress = XetUploadProgressReporter(total_files=1)
+            progress_callback = progress.update_progress
         else:
-            _progress = None
+            progress, progress_callback = None, None
 
+        if _file_hash is None or _file_size is None:
+            file_metadata = self.get_bucket_file_metadata(
+                bucket_id=bucket_id,
+                remote_path=remote_path,
+                token=token
+            )
+            file_hash = file_metadata.xet_file_data.file_hash
+            file_size = file_metadata.size
+        else:
+            file_hash = _file_hash
+            file_size = _file_size
+
+        headers = self._build_hf_headers(token=token)
         refresh_url = xet_connection_info_refresh_url(
             token_type=XetTokenType.WRITE,
             repo_id=bucket_id,
@@ -15002,30 +15081,18 @@ class HfApi:
         )
         xet_headers = xet_headers_without_auth(headers)
 
-        owns_progress = _progress is None
-        if _progress is not None:
-            progress = _progress
-            progress.reset_for_next_commit()
-            progress_callback = progress.update_progress
-        elif not are_progress_bars_disabled():
-            progress = XetUploadProgressReporter()
-            progress_callback = progress.update_progress
-        else:
-            progress, progress_callback = None, None
-        session = get_xet_session()
-        file_metadata = self.get_bucket_file_metadata(
-            bucket_id=bucket_id,
-            remote_path=remote_path,
-            token=token
-        )
         try:
-            commit = session.new_range_upload(
-                file_metadata.xet_file_data.file_hash,
-                file_metadata.size,
+            commit = get_xet_session().new_range_upload(
+                file_hash,
+                file_size,
                 token_refresh_url=refresh_url,
                 token_refresh_headers=headers,
                 custom_headers=xet_headers,
                 progress_callback=progress_callback,
+            )
+            logger.debug(
+                f"About to commit to a file on the hub: {len(edit or [])} edit(s), {len(insert or [])} insert(s) and"
+                f" {len(delete or [])} deletion(s)."
             )
             try:
                 if edit:
@@ -15072,7 +15139,7 @@ class HfApi:
                 ],
                 token=token,
             )
-        return
+        return report.file_info.hash
 
 
 def _parse_revision_from_pr_url(pr_url: str) -> str:
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -14989,19 +14989,24 @@ class HfApi:
                 List of deletes to apply, in the form `(loc, length)`.
                 The range [`loc`, `loc + length`) is deleted.
         """
+        if edit:
+            edit = [
+                ((start, end), data) for (start, end), data in edit if start < end or (start == end and len(data) > 0)
+            ]
+        if insert:
+            insert = [(loc, data) for loc, data in insert if len(data) > 0]
+        if delete:
+            delete = [(loc, length) for loc, length in delete if length > 0]
+        if not (edit or insert or delete):
+            return
         from .utils._xet_progress_reporting import XetUploadProgressReporter
-
 
         if not are_progress_bars_disabled():
             _progress = XetUploadProgressReporter(total_files=1)
         else:
             _progress = None
 
-        file_metadata = self.get_bucket_file_metadata(
-            bucket_id=bucket_id,
-            remote_path=remote_path,
-            token=token
-        )
+        file_metadata = self.get_bucket_file_metadata(bucket_id=bucket_id, remote_path=remote_path, token=token)
         self._mutate_bucket_file(
             bucket_id=bucket_id,
             remote_path=remote_path,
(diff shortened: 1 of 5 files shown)


Wauplin:
Will have a look at it this week! Sorry for the delay

tests/test_buckets_hf_file_system.py
```diff
@@ -83,6 +83,22 @@ def _bucket(self):
...
+
+    def test_edit_file(self):
+        with self.hffs.open(self.text_file, "e") as f:
+            f.insert(0, "this is ")
+            f.edit((8, 13), "a fantastic")
+            f.delete(24, 5)
+            f.append("!")
+            import warnings
+
+            warnings.warn(str(f.buffer.ranges))
```
Wauplin:
```suggestion
```

(probably an oversight)

src/huggingface_hub/hf_api.py
```diff
@@ -15001,6 +14959,180 @@ def sync_bucket(
...
         )
 
+    def edit_bucket_file(
+        self,
+        *,
+        bucket_id: str,
+        remote_path: str,
+        edit: list[tuple[tuple[int, int], bytes]] | None = None,
+        insert: list[tuple[int, bytes]] | None = None,
+        delete: list[tuple[int, int]] | None = None,
```
Wauplin, line 14969: I feel that having 3 lists of operations to perform on a file is error-prone, especially to determine which operation takes precedence over the other.  What do you think of considering that:
- an edit is defined by some bytes and a start+end position
- an insert is defined by some bytes and a start position => similar to an edit but with `end == start`
- a delete is defined by a start+end position => similar to an edit but with  `data == b""`

The API could be simplified to something like this:

```py
api.edit_bucket_file(
    bucket_id="user/bucket",
    remote_path="file.bin",
    edits=[
        (0, 16, b"updated data"),   # edit
        (32, 32, b"inserted data"), # insert
        (64, 80, b""),              # delete
    ],
)
```

WDYT ?

src/huggingface_hub/hf_api.py
```diff
@@ -15001,6 +14959,180 @@ def sync_bucket(
...
+            try:
+                if edit:
+                    for (start, end), data in edit:
+                        commit.edit((start, end), len(data)).write(data)
+                if insert:
+                    for loc, data in insert:
+                        commit.insert(loc, len(data)).write(data)
+                if delete:
+                    for loc, length in delete:
+                        commit.delete(loc, loc + length)
```
Wauplin, line 15098: here the order of operations is arbitrary and can be misleading to user

src/huggingface_hub/hf_file_system.py
```diff
@@ -814,7 +919,34 @@ def cp_file(self, path1: str, path2: str, revision: str | None = None, **kwargs)
         if isinstance(resolved_path1, HfFileSystemResolvedBucketPath) or isinstance(
             resolved_path2, HfFileSystemResolvedBucketPath
         ):
-            raise NotImplementedError("Copy from/to buckets is not available yet")
+            if not isinstance(resolved_path1, HfFileSystemResolvedBucketPath) or not isinstance(
+                resolved_path2, HfFileSystemResolvedBucketPath
+            ):
+                raise NotImplementedError("Copy between repos and buckets is not available yet")
+            if resolved_path1.bucket_id != resolved_path2.bucket_id:
+                raise NotImplementedError("Copy between different buckets is not available yet")
```
Wauplin, line 927: note for a separate PR: copy between buckets is actually possible. And copy from a repo to a bucket is also possible

src/huggingface_hub/hf_file_system.py
```diff
@@ -385,19 +389,120 @@ def invalidate_cache(self, path: str | None = None) -> None:
...
+                text_kwargs = {k: kwargs.pop(k) for k in ["encoding", "errors", "newline"] if k in kwargs}
+                buffer = super().open(
+                    path,
+                    mode=mode,
+                    block_size=block_size,
+                    cache_options=cache_options,
+                    compression=compression,
+                    **kwargs,
+                )
+                return EditTextIOWrapper(buffer, **text_kwargs, write_through=True)
```
Wauplin, line 483: as a start could we support only `ab` and `eb` ? I feel this PR is already quite complex and adding non-binary mode as well is potentially error-prone typically with index, data length, etc. which depends on encoding. 

Even if we ship "only" binary mode, it's quite straightforward for users to edit/append utf-8 text to a fie


(it would also reduce PR size)

src/huggingface_hub/hf_api.py
```diff
@@ -15001,6 +14959,180 @@ def sync_bucket(
...
+                copy=[
+                    _BucketCopyFile(
+                        destination=remote_path,
+                        xet_hash=report.file_info.hash,
+                        source_repo_type="bucket",
+                        source_repo_id=bucket_id,
+                    )
+                ],
+                token=token,
+            )
```
Wauplin, line 15133: wondering if we should add a `parent_xet_hash` similar to `parent_commit` in the `create_commit` API. This is not implemented yet server-side (AFAIK) but would prevent concurrent edits on a file to succeed and silently erase some edits

(taken from granted that moon-landing is able to atomically upsert a file xet_hash while checking its previous value)

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
         self._stream_iterator = self.response.iter_bytes()
 
 
+class HfFileSystemEditFile(fsspec.spec.AbstractBufferedFile):
```
Wauplin, line 1526: haven't reviewed this class. Is it mostly taken from fsspec with a few tweaks? are 100% new code? (asking whether it's worth investing time on it or not)

src/huggingface_hub/hf_api.py
```diff
@@ -15001,6 +14959,180 @@ def sync_bucket(
...
+            return
+        from .utils._xet_progress_reporting import XetUploadProgressReporter
+
+        if not are_progress_bars_disabled():
+            _progress = XetUploadProgressReporter(total_files=1)
+        else:
+            _progress = None
+
+        file_metadata = self.get_bucket_file_metadata(bucket_id=bucket_id, remote_path=remote_path, token=token)
+        self._edit_bucket_file(
```
Wauplin, line 15011: (nit) should forward `token=token` to internal method

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
...
+        if whence == 0:
+            nloc = loc
+        elif whence == 1:
+            nloc = self.loc + loc
+        elif whence == 2:
+            nloc = self.size + loc
+        else:
+            raise ValueError(f"invalid whence ({whence}, should be 0, 1 or 2)")
+        if nloc < 0:
+            raise ValueError("Seek before start of file")
```
Wauplin, line 1759: shouldn't we also have a `if nloc > self.size: raise ValueError`?

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
...
+            loc (`int`):
+                Where to insert the data.
+            length (`int`):
+                Number of bytes to delete.
+        """
+        return self.edit((loc, loc + length), b"")
+
+    def truncate(self, size: int | None = None):
+        """
+        Write data to buffer.
```
Wauplin, line 1935:
```suggestion
        Truncates file.
```

(+ document explicity what it means)

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
...
+        Args:
+            size (`int`, *optional*):
+                Resize the file to the given size in bytes.
+        """
+        if size is None:
+            size = self.loc
+        if size < self.size:
+            self.delete(size, self.size - size)
+        elif self.size < size:
+            self.append(b"\0" * (size - self.size))
```
Wauplin, line 1950: I feel that a "truncate" bigger than the file size should be forbidden rather than implicitly filled with "0"s

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
...
+            defer (`bool`):
+                Send the buffer in the background, non-blocking.
+        """
+
+        if self.closed:
+            raise ValueError("Flush on closed file")
+        if force or (
+            self.buffer_size >= self.blocksize
+            and self.last_update_time
+            and (time.time() - self.last_update_time) > self.send_interval
```
Wauplin, line 1980: instead of

_"if the last non-flushed update was more than 10 seconds ago, then flush"_

I think we should do a

_"if the oldest non-flushed update was more than 10 seconds ago, then flush"_

Otherwise if a script writes a log every 5s, we will never write to Xet until the end of the process

src/huggingface_hub/hf_file_system.py
```diff
@@ -1391,6 +1523,593 @@ def _open_connection(self):
...
+            and (self.task is None or self.task.done())
+        ):
+            self._upload_ranges(defer=defer)
+            return True
+        else:
+            # Defer write on small block or quick update
+            return False
```
Wauplin, line 1987: is this the default for `fsspec`? I found it weird that an explicit `file.flush()` call might not flush (i.e. save the file). In my understanding "flushing" means "persisting changes right now"

Wauplin (commented):
Hey @lhoestq , sorry for the long delay! Finally got some time to get my hands on this PR. I've reviewed the API and high level behavior as much as I could and made some comments. Let me know what you think


(and deferring real review for later once the PR is ready :) )

src/huggingface_hub/hf_api.py
```diff
@@ -14678,6 +14635,7 @@ def _payload_as_ndjson() -> Iterable[bytes]:
                         "xetHash": op.xet_hash,
                         "sourceRepoType": op.source_repo_type,
                         "sourceRepoId": op.source_repo_id,
+                        "mtime": op.mtime,
```
Wauplin, line 14638: thanks for noticing about that one. Definitely an omission. Let's fix it in its own PR: https://github.com/huggingface/huggingface_hub/pull/4920

src/huggingface_hub/hf_file_system.py
lhoestq: yes fsspec does this by default 

https://github.com/fsspec/filesystem_spec/blob/2ff04f9fb24e6ab36d15ecb4da245839a77b1637/fsspec/spec.py#L2134-L2136

src/huggingface_hub/hf_file_system.py
lhoestq: it's the same behavior as builtin f.truncate()

```python
>>> f = open("tmp.bin", "wb")
>>> f.truncate(10)
10
>>> f = open("tmp.bin", "rb")
>>> f.read()
b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

src/huggingface_hub/hf_file_system.py
lhoestq: 100% new code except flush() which is the fsspec one with a few tweaks

src/huggingface_hub/hf_api.py
lhoestq: Very good point. Not just for this use case but also for conditional overwrites which are needed in important data frameworks iirc

src/huggingface_hub/hf_api.py
lhoestq: yes sounds good !

src/huggingface_hub/hf_file_system.py
Wauplin: ok, didn't know about that :+1:

tests/test_buckets_hf_file_system.py
lhoestq: Empty suggestion block 🙂 (copy/paste oversight I guess). The test file has been reworked in 0436b678 anyway.

src/huggingface_hub/hf_api.py
lhoestq: Done in 0436b678: the API is now a single `edits: list[tuple[int, int, bytes]]` where insert is `(loc, loc, data)` and delete is `(start, end, <empty bytes>)`. Much less error-prone on our side too: the buffered ranges now translate to a single flat list of edits.

src/huggingface_hub/hf_api.py
lhoestq: Good point: addressed in 0436b678. All the locations are now documented as referring to the file *as it currently is on the Hub* (i.e. all the edits apply at the same time), so they must not overlap and their order in the list is irrelevant (xet sorts them internally). Same wording in the `HfApi.edit_bucket_file` docstring + example.

src/huggingface_hub/hf_file_system.py
lhoestq: Thanks! I've enabled same-bucket copies in `HfFileSystem.cp_file` in this PR (server-side copy by xet hash). Cross-bucket and repo -> bucket copies are kept for a follow-up PR: added a `TODO` in `cp_file` + a note in the PR description.

src/huggingface_hub/hf_file_system.py
lhoestq: Agreed, done in 0436b678: only `ab` and `eb` are supported, the `EditTextIOWrapper` class is gone (-100 lines) and text modes now raise a `NotImplementedError` with an explicit message. Docs/tests/examples updated accordingly.

src/huggingface_hub/hf_api.py
lhoestq: Added a `TODO` in `_edit_bucket_file` (+ note in the PR description) to pass the previous xet hash as a precondition once the Hub supports it. Would be great to have it for conditional overwrites as well as you mentioned.

src/huggingface_hub/hf_api.py
lhoestq: Fixed in 0436b678 (`token=token` is forwarded). Also fixed the same kind of issue in `_fetch_range`, which now uses `self.fs.token`.

src/huggingface_hub/hf_file_system.py
lhoestq: Done: `seek()` now raises `ValueError(Seek

src/huggingface_hub/hf_file_system.py
lhoestq: Done, docstring now reads `Truncate file to `size` bytes, and return the new size of the file.` and explains the 3 cases explicitly (cut short / extended with zero bytes, like the built-in `f.truncate()` / no-op).

src/huggingface_hub/hf_file_system.py
lhoestq: Makes sense, done: the buffer now tracks `oldest_update_time` (the oldest non-flushed edit) instead of the most recent one, so a stream of small appends is not starved forever. `flush()` docstring updated.

src/huggingface_hub/hf_file_system.py
lhoestq: For the record the docstring now states it explicitly: a regular `flush()` might not send anything and `f.flush(force=True)` is needed to persist the buffered edits right away (same semantics as fsspec).

lhoestq:
Pushed `0436b678` with all the feedback addressed (replies in each thread). Quick summary:

**API**

- `HfApi.edit_bucket_file()` now takes a single `edits: list[tuple[int, int, bytes]]` instead of 3 separate lists. Insert is `(loc, loc, data)`, delete is `(start, end, b"")`.
- Locations all refer to the file *as it currently is on the Hub*, so the order in the list doesn't matter (documented in the docstring).

**Scope**

- Binary modes only (`"ab"` / `"eb"`): `EditTextIOWrapper` removed, text modes raise `NotImplementedError`.
- Same-bucket `cp_file` enabled, cross-bucket / repo → bucket left as a `TODO` for a follow-up PR.
- `parent_xet_hash` precondition: `TODO` added, waiting on Hub support.

**Bugs found while reviewing/tweaking**

- The buffered `ranges` → xet edits translation was wrong for inserts: new bytes were sent as an edit consuming original bytes, which both shifted/lost original bytes and produced *overlapping* edits (rejected by xet). New bytes are now pure inserts and gaps between kept ranges are deletes.
- An edit landing at EOF (or in a file emptied by `truncate(0)`) was silently dropped.
- `flush()` policy used the *most recent* update, which could starve small appends forever → now uses the oldest non-flushed update.
- `_upload_ranges_inner` read `self.original_size` from the main thread while running in the background thread, and the deferred send shared mutable `bytearray`s with the main thread. Both fixed.
- Missing `token` forwarding, extra `info()` call in `__init__`, seek/edit past end of file.

**Validation**

`XetSession.new_range_upload` is not released yet in `hf_xet` (1.6.0), so the integration tests are gated behind a `skipif` and I could not run them here. I validated the edit translation offline against a fake xet server that enforces the documented `upload_ranges` invariants (sorted, non-overlapping, `end <= original_size`) and compares the resulting file content: random sequences of `append`/`insert`/`edit`/`delete`/`truncate`/`seek+write` (with and without intermediate flushes) + read-after-write through `_fetch_range`, all matching. Worth re-running the real tests on staging once the binding is out.

`make quality` is clean except for two pre-existing files (`serialization/_torch.py`, `tests/test_repocard_data.py`) that my local ruff disagrees with — untouched here.

lhoestq:
Follow-up now that I have a dev build of `hf_xet` with `XetSession.new_range_upload()` locally — the real tests do pass on staging:

```
tests/test_buckets_hf_file_system.py tests/test_hf_file_system.py  -> 167 passed, 1 skipped
tests/test_buckets.py                -> 43 passed
tests/test_buckets_cli.py            -> 94 passed
```

So the edit/insert/delete translation is confirmed end-to-end (`test_edit_file` buffers `insert(0) + edit((8,13)) + delete(24,5) + append` and gets `this is a fantastic text!` back when re-reading the file).

A few extra manual checks against the Hub, not covered by the test suite:

- Many small appends with `block_size=4, send_interval=0` => several *deferred* (background-thread) sends, final content correct.
- `f.flush()` returns `False` (nothing sent) while `f.flush(force=True)` returns `True`; reads in between are served from the local buffer.
- `edit_bucket_file()` with an edit + an insert + a delete in a single call, and with the edits listed out of order.
- `"ab"` on a path that does not exist yet (creates it), then `seek(0)` + `read()`.

Still need `hf_xet` to ship `new_range_upload` before the append/edit tests stop being skipped in CI.

Issue #4757, opened 2026-08-27
feat: --watch flag for `hf buckets sync` (keep a bucket mirrored to a local directory)


morphym:
### Summary

`hf buckets sync` is a one-shot operation: compute a plan, execute it, exit. For the
"keep a bucket up to date with a local directory" use case you currently have to wrap
it in a shell loop, which re-pays process startup and gives no control over how
in-progress files are handled.

I'd like a `--continuous` (or `--watch`) flag that re-runs the sync on an interval,
uploading only what changed since the last pass.

```bash
hf buckets sync ./checkpoints hf://buckets/user/my-bucket --continuous \
  --include "*.safetensors" --exclude "*.tmp"
```

All the existing filtering flags (`--include`, `--exclude`, `--filter-from`,
`--existing`, `--ignore-existing`) would keep working unchanged.

### Why

The motivating case is a long-running job — training, data generation, log
accumulation — whose output directory should be mirrored to a bucket as it's produced,
so it survives the machine and can be watched from elsewhere.

Today the options are:

- **`while true; do hf buckets sync ...; sleep 30; done`.** Works, but re-imports the
  library and re-authenticates every pass, and has no notion of "don't upload a file
  that's still being written."
- **`hf-mount` with a bucket mount.** Solves it for files written *through the mount*,
  but requires FUSE/NFS setup and doesn't help when a job writes to an ordinary local
  directory you don't control. (I opened
  huggingface/hf-mount#222 about the related gap there — files held open never flush.)

A flag on `sync` is the lighter-weight answer for the common case, and it composes with
the include/exclude filtering that already exists here.

### Why it looks tractable

`_compute_sync_plan()` and `_execute_plan()` are stateless functions of
`(source, dest, filters)` — nothing carries across calls. So the core of `--continuous`
is a loop around the existing pair plus a sleep, with no new dependency: change
detection is already size + mtime via `os.walk` in `_list_local_files()`.

That said, three things are real design decisions rather than plumbing, and I'd rather
agree on them before writing code:

1. **Files still being written.** A polling sync will happily upload a file mid-write —
   size+mtime look "changed", and what lands in the bucket is truncated. Some kind of
   settle window (skip files modified in the last N seconds) seems necessary. Should
   that be on by default, and what's a sane N?

2. **Cost of re-listing the remote.** `_list_remote_files()` walks the whole bucket
   every pass. At a short interval on a large bucket that's significant API traffic. In
[...]

PR #4760, opened 2026-08-27
[experimental] feat(buckets): continuous/watch sync with volatility-aware uploads

morphym:
> **Status: draft / experimental — not asking for review.** A maintainer has already indicated in
> #4757 that this is not something the project wants to take on short term, and that's a reasonable
> call. This branch was written before that reply landed, so it is published as a reference for the
> design and the stability math rather than as a proposal. Happy for it to be closed.

## What this adds

`hf buckets sync` (and its `hf sync` alias) currently runs one pass and exits. This adds
`--continuous` / `--watch`, which keeps it running and re-syncs on an interval until interrupted —
for mirroring a job's output directory to a bucket while the job is still writing to it.

```bash
hf sync ./checkpoints hf://buckets/user/my-bucket --watch
```

The hard part is not the loop. It is **not uploading a file that is still being written**. An
unattended sync that ships a half-written checkpoint has done something worse than nothing, so most
of this PR is about deciding when a file has gone quiet enough to be worth uploading.

### 1. Calm period, scaled by size

Every file must sit **unchanged for a calm period** before it is eligible. That period scales with
size, because the cost of guessing wrong scales with size: re-uploading a 4 KB log you caught
mid-write is free, re-uploading a 20 GB checkpoint is not.

Two anchors, both flags:

| size | calm period |
| --- | --- |
| ≤ 1 GiB | 60s (`--calm-min`) |
| ≥ 10 GiB | 600s (`--calm-max`) |

Between them the period is interpolated **in log space**, since file sizes span decades and linear
interpolation across a 10× range would spend almost all its resolution on the top of the range:

```
ratio = log10(size / CALM_SMALL_SIZE) / log10(CALM_LARGE_SIZE / CALM_SMALL_SIZE)
calm  = calm_min + ratio * (calm_max - calm_min)
```

The anchors happen to span exactly one decade, so this reduces to
`calm = 60 + 540 * log10(size / 1GiB)` at the defaults — e.g. a 3.16 GiB file (10^0.5) gets 330s.
The ramp is flat outside the anchors, so the function is continuous and monotonic across the whole
size domain.

### 2. Calmness is measured from mtime, not from first sight

```
quiet_for = now - mtime
calm      = quiet_for >= calm_required
```

Using the file's own mtime rather than "when this process first noticed it" matters: it means a
directory of **static files syncs immediately on the first pass**. Only files actually being
written wait. A first-sight-based timer would have imposed a 60s delay on every ordinary sync,
which would be an unacceptable regression in the common case.

### 3. Churn rate

Each pass records `(size, mtime)` per file and appends a timestamp whenever either changed. Over
the last ≤ 8 change observations:

```
mean_interval = (1 / (n-1)) * Σ (t[i+1] - t[i])        , n >= 2
```

This feeds the prediction below and the on-screen meter. Two caveats, both real:

- It needs two observed changes, so it is `None` on first contact.
- It is **quantized to `--interval`** — a file written every 2s under a 5s poll reads as ~5s. The
  estimate is therefore conservative (it under-reports churn), which is the safe direction for the
  prediction that follows.

### 4. Refusing uploads that would be overtaken

A large file can pass the calm check and still be a bad upload: if it changes every few minutes and
the transfer takes longer than that, the copy is stale before it lands and the bandwidth is wasted.

Throughput is **measured, not assumed** — an EWMA over completed transfers, seeded at 50 MiB/s only
until the first real measurement replaces it outright:

```
observed = bytes / seconds
B <- observed                      (first measurement)
B <- 0.7 * B + 0.3 * observed      (thereafter)
eta = size / B
predicted_changes = eta / mean_interval
```

The upload is refused when `eta >= 60s AND predicted_changes >= 2` — i.e. only when the transfer is
expensive enough to care about *and* the file would be overtaken at least twice mid-flight. Both
conditions are required, so a fast upload of a churning file still proceeds.

**This is a pre-flight refusal, not a mid-flight abort.** Once bytes are handed to the Hub client
there is no interrupt point without reaching into the upload path, so the check decides whether to
*start*. Worth flagging explicitly since "abort the upload" is the more obvious reading.

### 5. Why the churn rate is not itself a gate

The first version also blocked any file whose `mean_interval` was under 10s. **That was a bug**: the
samples never age out, so a file that churned once and then went permanently quiet would have been
blocked *forever*. Reaching the eligibility check already proves the file has been quiet for the
full calm period, so the observed rate is now used only to predict the future and to render the
meter — never as an independent gate. There is a named regression test
(`test_file_that_churned_then_settled_is_eventually_uploaded`).

## Terminal output

Watch mode draws a small panel that repaints in place rather than scrolling:

```text
watching:: ./data → hf://buckets/user/my-bucket
calm 60s (≤1GiB) → 10m00s (≥10GiB)   lazy mode active

2 files changed
  train.log               ▅▅  ██████░░   45s/1m00s  ready in 15s
  model-00001.safetensors ▁   ████████      stable  eta 4m10s
uploaded 1 file in 2s
```

Per row: a **churn meter** (`▇▇▇` for a write every second or two, down to `▁` once settled; blank
until two samples exist), a **stability meter** showing progress toward the calm period with
elapsed/required beside it, and a tail saying what happens next — `ready in 15s` while settling, or
the predicted `eta` once stable. Files that have not changed are not listed at all.

**Non-TTY is not a second-class path.** When stdout is not a terminal the panel degrades to plain
appended lines, reporting a file only when its status actually changes. Watch output is routinely
redirected to a log file, where cursor movement would be garbage. (An earlier iteration of this
branch had a real bug here — Python block-buffers a non-TTY stdout, so a redirected watch produced a
*completely empty* logfile until exit. Everything the panel writes is now flushed.)

The scanning `StatusLine` and the transfer progress bars are suppressed while the panel is active;
they would scribble over it, and the `eta` column covers the same ground.

## Safety rules

- **`--delete` is refused with `--continuous`.** An unattended loop that propagates deletions turns
  one transient local failure — a directory briefly unreadable, a disk hiccup — into unrecoverable
  remote data loss with nobody watching. Run a one-shot `sync --delete` to prune.
- **`--continuous` is refused with `--plan` / `--apply` / `--dry-run`**, which each describe a
  single pass.
- A failed pass is logged and retried with exponential backoff to a 5-minute ceiling rather than
  ending the loop. `Ctrl-C` exits cleanly and returns the last plan, so the Python API contract is
  unchanged.

## Implementation notes

`_compute_sync_plan` and `_execute_plan` were already pure functions of `(source, dest, filters)`,
so each pass recomputes from scratch and no state carries between iterations — a crash mid-loop
leaves nothing to reconcile. The only things carried across passes are the `StabilityTracker` and
the display.

`_compute_sync_plan` gained a `local_filter` predicate (replacing a narrower `min_age_seconds`),
which keeps volatility policy out of plan computation. It refuses to combine with `delete`: an
excluded file is invisible to the comparison, so it would otherwise look like a remote deletion
candidate — the same failure mode as the `--delete` rule above, one layer down.

## Known limitations

- **Every pass re-lists the whole remote bucket**, which dominates cost on large buckets at short
  intervals. In the upload direction this process is the only writer, so remote state could be
  cached across passes; deliberately left out of a first version.
- **A file appended to forever never uploads while its writer runs.** That follows directly from
  "wait until it is calm", but it is a real consequence — documented with a warning box, and
  `--calm-time` lowers the bar for anyone who would rather accept partial copies.
- Downloads (bucket → local) go through the same machinery, but the stability logic only applies to
  the local side, so a remote file being rewritten mid-download is not protected.

## Testing

53 offline tests in `tests/test_buckets_sync_continuous.py` — the existing bucket tests all require
staging plus a token, so these are self-contained: the calm ramp and its endpoints, mtime-based
calmness, the overtake prediction in both directions, EWMA throughput, loop behaviour under failure
and `KeyboardInterrupt`, every refusal path, and the display (in-place rewind, leftover-row
clearing, plain-mode degradation, meter rendering).

Also verified end-to-end against a real bucket, driving the actual CLI under a pty and replaying the
ANSI stream: a file appended every 2s was held for its entire write phase and uploaded once writing
stopped, while a file written once uploaded as soon as its calm period elapsed. Nine passes produced
2720 bytes of terminal output total.

## Open questions

1. **Are the defaults right?** 60s/1 GiB → 600s/10 GiB is a guess informed by nothing but
   plausibility. Real checkpoint-writing patterns should set these.
2. **Should `--delete` + `--continuous` be refused outright, or gated behind a confirmation or a
   "refuse to delete more than X% of remote files" bound?** I picked the most conservative option;
   it is the most likely thing here to be wrong.
3. **Is the overtake heuristic worth its complexity**, or should an expensive churning file simply
   be left to the calm period?
4. **Remote-listing caching** — worth doing now, or premature?

### DIFF
--- a/docs/source/en/guides/buckets.md
+++ b/docs/source/en/guides/buckets.md
@@ -709,21 +709,21 @@ A continuous sync runs unattended, so it must not upload a file that is still be
 
 This is on by default and needs no flags. Files that are simply sitting on disk are **not** delayed — calmness is measured from the file's own mtime, so a directory of static files syncs immediately on the first pass. Only files actually being written wait.
 
-While a file is waiting, the sync says so, along with how long it has been quiet and how long it needs:
+On a terminal, watch mode draws a small panel that repaints in place rather than scrolling — it shows what is being watched, the calm period in force, and a row per file that has changed:
 
 ```text
-Continuous sync: ./data -> hf://buckets/user/my-bucket (every 5s, calm period 60s (<=1GiB) to 10m00s (>=10GiB), Ctrl-C to stop)
-  waiting: train.log (changed 3s ago, needs 60s quiet)
-  waiting: model.safetensors (changed 12s ago, needs 2m10s quiet)
-  upload: config.json (new file)
-Pass 4: 1 file(s) in 2s.
+watching:: ./data → hf://buckets/username/my-bucket
+calm 60s (≤1GiB) → 10m00s (≥10GiB)   lazy mode active
+
+2 files changed
+  train.log               ▅▅  ██████░░   45s/1m00s  ready in 15s
+  model-00001.safetensors ▁   ████████      stable  eta 4m10s
+uploaded 1 file in 2s
 ```
 
-Once a file has been quiet long enough it is uploaded, and the observed rate of change is reported alongside it:
+Each row carries two meters. The first is the **churn meter**, showing how rapidly the file is changing (`▇▇▇` for a write every second or two, down to `▁` for a file that has settled; blank until there are two samples to compare). The second is the **stability meter**, showing progress towards the calm period, with the elapsed/required times beside it. The tail says what has to happen next: `ready in 15s` while a file is still settling, or `eta 4m10s` — the predicted upload time — once it is stable.
 
-```text
-  upload: train.log (size differs, changes ~5s apart)
-```
+When output is not a terminal (redirected to a log file, or running under a supervisor), the panel degrades to plain appended lines, and a file is reported only when its status actually changes rather than on every pass.
 
 To override the ramp, use `--calm-time` for a fixed period regardless of size, or tune the two anchors:
 
@@ -749,7 +749,7 @@ A file that is appended to forever — an active log with a write every few seco
 A large file can pass the calm check and *still* be a bad upload: if it changes every few minutes and the upload itself would take longer than that, the copy is stale before it lands and the bandwidth is wasted. Continuous sync predicts this from the file's observed rate of change and the measured upload throughput, and refuses to start an upload that would be overtaken at least twice while in flight:
 
 ```text
-  waiting: checkpoint.safetensors (~4m10s upload would be overtaken ~3x)
+  checkpoint.safetensors  ▃   ████████      stable  eta 4m10s, overtaken ~3x
 ```
 
 Throughput is learned from the uploads it actually performs, so the estimate improves as the sync runs. The file is uploaded on a later pass, once it has calmed down enough for the upload to finish before the next write.
--- a/docs/source/en/guides/buckets.md
+++ b/docs/source/en/guides/buckets.md
@@ -677,6 +677,33 @@ The filter file uses `+` (include) and `-` (exclude) prefixes. Lines starting wi
 + *.json
 ```
 
+### Continuous sync
+
+By default `sync` runs a single pass and exits. Pass `--continuous` to keep it running, re-syncing on an interval until you interrupt it with `Ctrl-C`. This is useful for mirroring a job's output directory to a bucket while the job is still writing to it:
+
+```bash
+# Mirror ./checkpoints to the bucket, re-checking every 30s (default)
+>>> hf buckets sync ./checkpoints hf://buckets/username/my-bucket --continuous
+
+# Check more often, and only ship model weights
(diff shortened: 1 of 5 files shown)


Wauplin:
This is a 1000+ loc changes for an undiscussed feature. Thanks for the effort but I prefer to close it for now until details have been fully discussed in https://github.com/huggingface/huggingface_hub/issues/4757

morphym:
> This is a 1000+ loc changes for an undiscussed feature. Thanks for the effort but I prefer to close it for now until details have been fully discussed in #4757

alright

PR #4767, opened 2026-08-27
[Download] A `ResolvedRevision` only pins the repo it was resolved for

Wauplin:
Follow-up on #4604 and #4692.

### TODO once that's merged/shipped, we will revert https://github.com/vllm-project/vllm/pull/51260 which added a workaround to disable revision resolution due to this bug

---

---

Claude description:

## Problem

Calling `resolve_revision(revision)` twice for the same revision (e.g. `"main"`) but on two different repos returned the same commit oid, because the repo id / repo type were not checked. A commit hash means nothing outside of the repo it comes from, so a `ResolvedRevision` was silently reused for another repo — and nothing on the object said which repo it belonged to.

It already bit downstream integrations twice:

- vllm-project/vllm#51260 — a GGUF model reads its config from the base repo and its weights from the GGUF repo; the base repo's commit was handed to the weights repo => `RevisionNotFoundError`.
- huggingface/transformers#47611 — `from_pretrained("<adapter repo>")` resolves the adapter repo, then loads the base model it points to.

## Fix

`ResolvedRevision` now remembers the repo it was resolved against (`repo_id` + `repo_type`, kept private), and `HfApi.resolve_revision` only returns it as is when it matches the repo being asked about. Otherwise the revision initially requested is resolved again, so the result is what a plain `revision="main"` would have given:

```py
>>> revision = resolve_revision("openai-community/gpt2")
>>> resolve_revision("openai-community/gpt2", revision=revision) is revision  # same repo => as is
True
>>> resolve_revision("openai-community/gpt2-medium", revision=revision).resolved  # another repo => resolved again
'6dcaa7a952f72f9298047fd5137cd6e4f05f41da'
>>> resolve_revision("openai-community/gpt2", repo_type="dataset", revision=revision)  # another repo type too
RepositoryNotFoundError: ...
```

`__reduce__` carries the two new attributes, so this survives pickling and `copy.deepcopy` (#4692).

**Scope**: this is the only behavior change. Download helpers (`hf_hub_download`, `snapshot_download`, `get_cached_repo_tree`) are untouched: a `ResolvedRevision` passed to them is still trusted as is (`revision.resolved`), whatever repo they are called with. We always consider that the caller passes a revision that is correct for the repo it is downloading from — libraries that download from several repos should call `resolve_revision` once per repo (which is now safe to do by passing the `ResolvedRevision` back, see the guide). An earlier iteration of this PR also made the download helpers fall back to the requested ref for a foreign `ResolvedRevision`, via a `ResolvedRevision._commit_hash_for(repo_id, repo_type)` helper — dropped, since it added a private API and made the download path guess what the caller meant.

## Testing

`tests/test_snapshot_download.py` and `tests/test_file_download.py` pass (89 tests, serially — the class-scoped repo fixtures race under `-n 4`). New test covers a cross-repo resolution against the CI Hub (fails without the fix), and the pickle round-trip now checks the repo is preserved.

---

> [!NOTE]
> **Medium Risk**
> Changes core revision-resolution behavior used by multi-repo loaders (adapters, base models); the fix corrects a serious bug but any code that relied on incorrect cross-repo reuse will see different commit hashes.
> 
> **Overview**
> **`ResolvedRevision` is now tied to the repo (`repo_id` + `repo_type`) it was resolved against**, so reusing the same object across repos no longer reuses another repo’s commit hash.
> 
> **`HfApi.resolve_revision`** still returns the same instance when the target repo matches. If the repo differs, it resolves again from the original ref (`revision.initial`, e.g. `"main"`) instead of returning the foreign commit. Instances built without repo metadata (`_repo_id is None`) keep the old “trust anywhere” behavior. New `ResolvedRevision` values from resolution (including raw commit hashes) are stamped with the repo they belong to; **pickle/deepcopy** preserve that via an updated `__reduce__`.
> 
> **Docs** in `manage-cache.md` explain multi-repo flows: pass a `ResolvedRevision` back into `resolve_revision` per additional repo. **Tests** add cross-repo resolution, repo-type mismatch, and pickle round-trips for repo fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e10b1b8545a846b2d8a295cf7e3907884253e627. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/close-unscoped-community-prs.yml
+++ b/.github/workflows/close-unscoped-community-prs.yml
@@ -6,10 +6,8 @@
 # Uses `pull_request_target` because `GITHUB_TOKEN` is read-only for PRs opened from forks.
 # The PR branch is never checked out, so no untrusted code runs here.
 #
-# The automatic trigger is dry-run only for now: it logs its decision without commenting or
-# closing, so we can watch it against real PRs. To enable enforcement, drop
-# `github.event_name == 'pull_request_target'` from `DRY_RUN` below. Until then, act on a
-# single PR with `workflow_dispatch` and `dry_run` unchecked.
+# A single PR can also be evaluated manually with `workflow_dispatch`, with `dry_run` checked
+# to only log the decision without commenting or closing.
 
 name: Close unscoped community PRs
 
@@ -48,7 +46,7 @@ jobs:
         env:
           MAINTAINERS: Wauplin,hanouticelina
           PR_NUMBER: ${{ github.event.pull_request.number || inputs.pr_number }}
-          DRY_RUN: ${{ inputs.dry_run || github.event_name == 'pull_request_target' }}
+          DRY_RUN: ${{ inputs.dry_run || false }}
         with:
           script: |
             const MARKER = "<!-- close-unscoped-community-pr -->";
--- a/docs/source/en/guides/cli.md
+++ b/docs/source/en/guides/cli.md
@@ -2171,6 +2171,11 @@ Manage scheduled jobs using
 # List your active scheduled jobs
 >>> hf jobs scheduled ls
 
+# Same filters as `hf jobs ls`: --status, --label and --name
+>>> hf jobs scheduled ls --all
+>>> hf jobs scheduled ls --status suspended
+>>> hf jobs scheduled ls --name hourly-task --label env=prod
+
 # Inspect the status of a job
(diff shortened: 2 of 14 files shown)


Wauplin (commented):
✔️

hanouticelina (approved):

PR #4008, opened 2026-03-30
[Internal] Bump minimum Python version from 3.9 to 3.10

hanouticelina:
This PR: 
- Raises the minimum supported Python version from 3.9 to 3.10.
- Migrates type annotations: replace `Union[X, Y]` with `X | Y` and `Optional[X]` with `X | None` across the entire codebase (~3,100 occurrences in 130+ files).
- Uses `match`/`case` statements where they improve readability.

entirely done by Claude Code (this PR costs a lot of tokens 😄) and some tiny help from my side.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes packaging/CI to drop Python 3.9 support, which can break downstream users still pinned to 3.9. Most other edits are mechanical typing/import updates with low behavioral impact.
> 
> **Overview**
> Raises the project’s minimum supported Python version to **3.10** by updating `setup.py` (`python_requires`/classifiers), CI/release GitHub Actions to run on 3.10+, and installation docs across locales.
> 
> Refactors the codebase to use Python 3.10+ syntax and stdlib typing locations (e.g., `X | Y`/`X | None`, `collections.abc` imports), and removes now-unnecessary version-gated extras (e.g., `gradio` extra always available). Also replaces a few deprecated `EnvironmentError` raises with `OSError`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0a509f8ff2c6c0a4d1f8960d56a090ffe999c6a3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4776, opened 2026-08-29
[CI] Fix torch_1.11 case-arm label to match renamed matrix entry

### LINKED ISSUE #4775: CI: torch_1.11 lane has not installed torch since the matrix rename in #4008
The `torch_1.11` CI lane has not installed torch since March, so the tests it exists to run have been skipping and the job has been green anyway.

`.github/workflows/python-tests.yml:95` still matches on `"Python 3.9, torch_1.11")`, but commit `4c7b420f` (#4008, 2026-03-31) renamed the matrix entry at line 50 to `"Python 3.10, torch_1.11"` and updated only the run-step arm at line 140. The install `case` has no `*)` default, so no arm matches and the torch install is skipped silently. The run step still selects `test_hub_mixin*` and `test_serialization.py`, those tests skip with "Test requires torch", and the job passes.

Your own run 31806513458 shows the difference between the two lanes:

```
build-ubuntu (3.12, torch_latest)             torch==2.13.0   46 passed / 47 passed
build-ubuntu (3.10, Python 3.10, torch_1.11)  no torch        25 passed + 21 skipped / 18 passed + 29 skipped
```

That is about 50 tests not running, and torch 1.11 back-compatibility unverified for 138 days. It matters for anyone pinned to `torch~=1.11` using `PyTorchModelHubMixin` or `save_torch_state_dict`, which is the case the lane was added for.

Locally the skip is easy to see:

```
cd src && python -m pytest ../tests/test_hub_mixin_pytorch.py -q -rs
21 skipped  --  Test requires torch
```

The fix is one token in the case label. I would rather flag it than open the PR blind, because I cannot run Python 3.10 here, and once torch 1.11 actually installs those 50 tests may not all pass. If they fail, that is 138 days of drift surfacing rather than anything the one-token change introduced, and you would probably want to decide how to handle it before a patch lands.

MohammedAlkindi:
Closes https://github.com/huggingface/huggingface_hub/issues/4775.

The `torch~=1.11` matrix entry's `test_name` was renamed to `"Python 3.10, torch_1.11"` in #4008 (`4c7b420f`, 2026-03-31), but the install step's `case` arm still matches the old `"Python 3.9, torch_1.11"`. There is no `*)` default arm, so the case falls through, torch is never installed on that lane, and its tests skip with "Test requires torch" while the job still reports green.

This changes that one label so the arm matches the entry it belongs to. Nothing else.

`git grep "Python 3.9, torch_1.11"` returns nothing after the change, so no other reference to the old label remains.

What I have not verified: whether the lane passes once torch actually installs. These tests have not run for 138 days, so some may fail on drift that accumulated while the lane was silently skipping. That would be a separate finding rather than a reason to leave the lane disabled, but it is worth expecting on the first green run.

---

> [!NOTE]
> **Low Risk**
> Single string fix in CI install logic; restores intended torch 1.11 coverage with no runtime or security impact.
> 
> **Overview**
> Fixes a **silent CI gap** on the `torch~=1.11` lane: the matrix `test_name` was already `"Python 3.10, torch_1.11"`, but the dependency install `case` arm still matched the old `"Python 3.9, torch_1.11"` string.
> 
> Because that workflow has **no default `*)` branch**, the mismatch meant **torch was never installed** on that job. Hub-mixin/serialization tests could skip for missing torch while the job still looked successful.
> 
> The change updates that one install `case` label so it matches the matrix entry and the existing test-run arm; **no application code** is touched.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf7b8a1dfdb84e10c7b63de8826016600bcc2de9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -92,7 +92,7 @@ jobs:
               uv pip install --upgrade torch
               ;;
 
-            "Python 3.9, torch_1.11")
+            "Python 3.10, torch_1.11")
               uv pip install "huggingface_hub[torch] @ ."
               uv pip install torch~=1.11
               ;;


Wauplin:
Looks good, thanks for fixing the typo!

Wauplin (approved):

MohammedAlkindi:
Post-merge result on the one thing I flagged as unverified: the lane passes and nothing broke.

Both suites collect the same 101 tests before and after. On `4cc16bc2` they were `25 passed, 21 skipped` and `26 passed, 29 skipped` with no torch installed; on `94cb25da` they are `46 passed` and `50 passed, 5 skipped` with `torch==1.13.1` (what `torch~=1.11` resolves to). So 45 tests went from skipped to passing, with zero failures on either side.

Numbers from the job logs of runs 33176176467 and 33391842386.

PR #4801, opened 2026-09-03
fix(auth): clearer whoami network/DNS error without silent retry

### LINKED ISSUE #4796: hf auth login aborts on transient DNS failure: whoami has no connection retry
### Describe the bug

`hf auth login` intermittently fails during token validation with a transient DNS error and aborts the whole login, even though retrying the exact command moments later succeeds.

Root cause: login validates the token via `whoami` **before** saving it (`_validate_and_save_token` → `whoami`), and `HfApi._inner_whoami` calls `get_session().get(...)` **directly**, with no retry:

```python
# src/huggingface_hub/hf_api.py
def _inner_whoami(self, token: str) -> dict:
    r = get_session().get(
        f"{self.endpoint}/api/whoami-v2",
        headers=self._build_hf_headers(token=token),
    )
```

The default client (`default_client_factory`) uses an httpx transport with `retries=0`, so a single transient DNS failure (Windows `WSATRY_AGAIN` → `[Errno 11002] getaddrinfo failed`) surfaces immediately as `httpx.ConnectError` and login fails. Because validation happens before the token is saved, nothing is written.

Interestingly, `httpx.ConnectError` is a subclass of `httpx.NetworkError`, which is already in `http_backoff`'s `_DEFAULT_RETRY_ON_EXCEPTIONS` — so `http_backoff` *would* retry this error, but the whoami/login path doesn't use `http_backoff`.

This also explains a confusing secondary symptom: after a failed attempt, a **new terminal** may report *"already logged in"*. That check (`interpreter_login`) is a purely local token-file read (no network), so once any single attempt briefly succeeds and writes the token, later invocations report success without touching DNS.

I intend to submit a PR for this (I have a branch ready): route `_inner_whoami` through the existing `http_backoff`, scoped to connection-level exceptions via `retry_on_status_codes=()` so 401/429 handling stays unchanged (in particular `/whoami-v2` remains intentionally un-retried on 429). Happy to open it if you'd like, or leave it to you.

### Reproduction

Simulate a transient `getaddrinfo` failure (fail the first 2 lookups, then let DNS recover) and issue the same GET that login performs:

```python
import socket, httpx

HOST, URL = "huggingface.co", "https://huggingface.co/api/whoami-v2"
_real = socket.getaddrinfo

def flaky(fail_first):
    st = {"n": 0}
    def f(host, *a, **k):
        if host == HOST:
            st["n"] += 1
            if st["n"] <= fail_first:
                raise socket.gaierror(11002, "getaddrinfo failed")  # WSATRY_AGAIN
        return _real(host, *a, **k)
    return f

def go(retries):
    socket.getaddrinfo = flaky(2)
    try:
        with httpx.Client(transport=httpx.HTTPTransport(retries=retries), timeout=10) as c:
            return f"OK HTTP {c.get(URL).status_code}"
    except httpx.ConnectError as e:
        return f"FAIL {e!r}"
    finally:
        socket.getaddrinfo = _real

print("retries=0:", go(0))  # reproduces login: FAIL ConnectError('[Errno 11002] getaddrinfo failed')
print("retries=3:", go(3))  # recovers: OK HTTP 401
```

Output:

```
retries=0: FAIL ConnectError('[Errno 11002] getaddrinfo failed')
retries=3: OK HTTP 401
```

(HTTP 401 without a token just confirms the connection was established.)

### Logs

```shell
hf auth login failed: Traceback (most recent call last):
  ...
  File ".../huggingface_hub/hf_api.py", line 2372, in _inner_whoami
    r = get_session().get(
        f"{self.endpoint}/api/whoami-v2",
        headers=self._build_hf_headers(token=token),
    )
  ...
  File ".../httpcore/_backends/sync.py", line 207, in connect_tcp
    with map_exceptions(exc_map):
  File ".../httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.ConnectError: [Errno 11002] getaddrinfo failed
```

### System info

```shell
- huggingface_hub version: 1.29.0
- Python version: 3.13.12
- Installation method: pip
- httpx: 0.28.1
- hf_xet: 1.6.0
- HF_HUB_OFFLINE: False
```

Alphaxiaoteng:
## Summary
Per maintainer guidance on #4796: do **not** add whoami retries/backoff. Instead, when `whoami` fails with a network/DNS error, raise a short actionable message:

> Couldn't reach https://huggingface.co (network/DNS). Check your connection and retry. Set HF_DEBUG=1 for the full traceback.

With `HF_DEBUG=1`, the original `httpx` exception is re-raised unchanged.

Fixes #4796

## Test plan
- [x] Local: patched `ConnectError` → clear `OSError` when `HF_DEBUG` is off
- [ ] CI unit tests for `HfApi.whoami` / auth

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/huggingface_hub/hf_api.py
+++ b/src/huggingface_hub/hf_api.py
@@ -2369,10 +2369,20 @@ class HfApi:
         return output
 
     def _inner_whoami(self, token: str) -> dict:
-        r = get_session().get(
-            f"{self.endpoint}/api/whoami-v2",
-            headers=self._build_hf_headers(token=token),
-        )
+        try:
+            r = get_session().get(
+                f"{self.endpoint}/api/whoami-v2",
+                headers=self._build_hf_headers(token=token),
+            )
+        except (httpx.ConnectError, httpx.TimeoutException, httpx.NetworkError) as err:
+            # Keep full traceback when debugging; otherwise surface a short actionable message.
+            # See https://github.com/huggingface/huggingface_hub/issues/4796
+            if constants.HF_DEBUG:
+                raise
+            raise OSError(
+                f"Couldn't reach {self.endpoint} (network/DNS). "
+                "Check your connection and retry. Set HF_DEBUG=1 for the full traceback."
+            ) from err
         try:
             hf_raise_for_status(r)
         except HfHubHTTPError as e:


Wauplin:
closing as per https://github.com/huggingface/huggingface_hub/issues/4796#issuecomment-5522734433

PR #4805, opened 2026-09-03
[Download] Tolerate missing HEAD Content-Length

### LINKED ISSUE #4741: Downloads fail with 'Distant resource does not have a Content-Length' when proxy strips HEAD Content-Length (HF_HUB_DISABLE_XET + mirror)
## Problem

`huggingface_hub` downloads fail with `FileMetadataError: Distant resource does not have a Content-Length.` on every **non-LFS metadata file** (README.md, config.json, tokenizer.json, manifests, ...) when the hub is accessed through a reverse proxy/mirror whose edge **strips the `Content-Length` header from HEAD responses**.

### Context (why we use a proxy)

We run a self-hosted Hugging Face reverse proxy (nginx → `huggingface.co`) behind Cloudflare to download models on a server in China where `huggingface.co` is not directly reachable:

```
HF_ENDPOINT=https://hf-mirror.example.com HF_HUB_DISABLE_XET=1 hf download <repo> --local-dir ./model
```

- `HF_HUB_DISABLE_XET=1` is mandatory: with Xet enabled, download bytes go to `cas-server.xethub.hf.co` over the Xet protocol and bypass the proxy entirely.
- With Xet disabled, every file needs a metadata HEAD call. The hub server answers `HEAD /resolve/...` with a `307` to `/api/resolve-cache/...`, which the client follows.

### Root cause

Cloudflare (and similar edge proxies) **strips the `Content-Length` header from HEAD responses** whose content looks compressible and is larger than ~1KB. We verified this exhaustively:

- `cache-control: no-transform` set by the origin → ignored.
- Compression Rules (`Compression: none` for the `/api/` path) → rule applies (no more gzip) but `Content-Length` is still stripped.
- Disabling every dynamic transformation setting (Automatic HTTPS Rewrites, Email Obfuscation, Rocket Loader, ...) → still stripped.
- Affected: HEAD responses > ~1KB with text-ish content. Small files (<1KB, e.g. a 665-byte `config.json`) keep their `Content-Length`; binary files (`*.safetensors`, served as `application/octet-stream` from the LFS CDN) keep it too.

The client then raises:

```
FileMetadataError: Distant resource does not have a Content-Length.
→ LocalEntryNotFoundError: An error happened while trying to locate the file on the Hub...
```

LFS weight files are unaffected because their `HEAD /resolve/...` returns a `302` carrying `x-linked-size`/`x-linked-etag`, so no follow-up HEAD is needed.

### Pre-existing (not caused by the tree cache)

Verified on `huggingface_hub` **v1.26.0** and **v1.23.0** (both predate the tree-listing cache introduced in #4394, v1.27.0): same error. So this is a pre-existing limitation of the metadata HEAD flow, not a regression from the tree cache.

### What we tried / proposed

1. PR #4700: use the cached `/tree` listing (which already contains `size`/`blob_id`/`lfs_sha256`/`lfs_size`/`xet_hash`) to rebuild the HEAD metadata without Xet. Maintainer preference: keep the pure-Python path simple, tree cache stays Xet-only as a pure optimization → PR closed without merging.
2. Workaround: `hf download --include "*.safetensors"` (weights only) + fetch metadata files with plain `curl` GETs (GET is unaffected). Works but is inconvenient for repos with many metadata files.

### Suggested mitigations (any would unblock proxy/mirror users)

- **Fallback to GET when the HEAD response lacks `Content-Length`** (or when the HEAD call fails) for the metadata lookup, e.g. using a ranged GET (`Range: bytes=0-0`) or a full GET with `Accept-Encoding: identity`, keeping the etag/size from the response. `x-linked-size`-style headers or the tree API could also provide the size.
- **Tolerate a missing `Content-Length` in HEAD metadata when an ETag is present**, and let the actual download (GET) validate size — the download itself works fine through the proxy.
- **Document** the proxy limitation and the `--include` workaround.

Happy to implement whichever direction maintainers prefer.

Wauplin:
What this PR does:

  - tolerate missing Content-Length in HEAD responses for regular HTTP downloads
  - validate the file size using the GET response when available
  - add regression tests for cache and local downloads

Fix https://github.com/huggingface/huggingface_hub/issues/4741.

---

> [!NOTE]
> **Medium Risk**
> Changes core download metadata and size validation in `file_download.py`, but behavior is narrowed (Xet still errors without size) and covered by new regression tests.
> 
> **Overview**
> **Regular HTTP downloads no longer fail when the Hub HEAD response omits `Content-Length`.** Metadata fetching still requires etag and commit hash, but missing size is allowed unless the file is an active Xet download (Xet still needs a known size up front).
> 
> During the GET, `http_get` can adopt size from response headers when HEAD did not provide it, and still runs the byte-count consistency check when a size is known (including from GET `Content-Length`). Progress totals continue to use HEAD `expected_size` when the streamed response is gzip/chunked without a usable length.
> 
> **CLI dry-run** reporting treats optional `DryRunFileInfo.file_size`: per-file sizes and totals show “unknown” when size is missing.
> 
> **Tests** cover cache vs `local_dir`, Xet enabled/disabled paths, and `http_get` validation when size comes only from the response.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2fb4f54ad9ea987324990fc838e12cee6e6ca6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/download.py
+++ b/src/huggingface_hub/cli/download.py
@@ -208,15 +208,18 @@ def download(
         if isinstance(result, DryRunFileInfo):
             result = [result]
         will_download = [r for r in result if r.will_download]
+        total_size = (
+            None if any(r.file_size is None for r in will_download) else sum(r.file_size or 0 for r in will_download)
+        )
         out.text(
             f"[dry-run] Will download {len(will_download)} files"
             f" (out of {len(result)})"
-            f" totalling {_format_size(sum(r.file_size for r in will_download))}."
+            f" totalling {_format_size(total_size) if total_size is not None else 'an unknown size'}."
         )
         items = [
             {
                 "file": info.filename,
-                "size": _format_size(info.file_size) if info.will_download else "-",
+                "size": _format_size(info.file_size) if info.will_download and info.file_size is not None else "-",
             }
             for info in sorted(result, key=lambda x: x.filename)
         ]
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -180,8 +180,9 @@ class DryRunFileInfo:
     Args:
         commit_hash (`str`):
             The commit_hash related to the file.
-        file_size (`int`):
-            Size of the file. In case of an LFS file, contains the size of the actual LFS file, not the pointer.
+        file_size (`int`, *optional*):
+            Size of the file, if known. In case of an LFS file, contains the size of the actual LFS file, not the
+            pointer.
         filename (`str`):
             Name of the file in the repo.
         is_cached (`bool`):
@@ -192,7 +193,7 @@ class DryRunFileInfo:
     """
 
     commit_hash: str
-    file_size: int
+    file_size: int | None
     filename: str
     local_path: str
     is_cached: bool
@@ -399,7 +400,9 @@ def http_get(
             resume_size = 0
 
         total: int | None = _get_file_length_from_http_response(response)
-        if total is None:
+        if expected_size is None:
+            expected_size = total
+        elif total is None:
             # Hub serves compressible text files (e.g. vocab.json) with `Content-Encoding: gzip` and
             # `Transfer-Encoding: chunked`, so the response carries no `Content-Length`. Fall back to the caller's
             # `expected_size` (always known from the metadata HEAD on the hf_hub path) so the progress bar, and any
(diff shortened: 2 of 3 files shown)


Wauplin (commented):
✔️

hanouticelina (approved):

PR #4810, opened 2026-09-04
[RepoCard] Read cards and templates as UTF-8, not the locale encoding

MohammedAlkindi:
`metadata_load()` reads the card with `Path(local_path).read_text()` and no `encoding=`, so it decodes with the locale default. The library's own writers pin UTF-8, so on a stock Windows install it writes UTF-8 and cannot read it back.

In the same file, `_content_to_file` (132), `RepoCard.load` (186), `RepoCard.save` (274) and both `metadata_update` handles (539, 547) all pass an explicit encoding. The three patched reads were the only ones that did not.

On cp1252, `metadata_load()` on a card whose YAML block holds a non-ASCII value raises rather than returning it:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 40: character maps to <undefined>
```

`RepoCard.from_template()` has the same problem for a custom `template_path`, which is why it is fixed alongside.

Two tests added, both failing before. On the offline part of `tests/test_repocard.py` the count goes 42 to 44 passed with nothing failing either side.

The Hub-backed tests are excluded from that comparison deliberately: they are non-deterministic against `hub-ci` from here, and two runs of unmodified `main` gave six failures each with three different names each way, so a diff across them would not mean anything.

---

> [!NOTE]
> **Low Risk**
> Narrow encoding fix aligned with existing UTF-8 I/O elsewhere in repocard; no auth or data-model changes.
> 
> **Overview**
> Fixes **locale-dependent decoding** on three repo-card reads that omitted `encoding="utf-8"`: `metadata_load()`, and both template file reads in `RepoCard.from_template()` (custom `template_path` and the default template). Writers and other load paths in the same module already use UTF-8, so on Windows (e.g. cp1252) UTF-8 cards with non-ASCII metadata or templates could raise `UnicodeDecodeError` on read.
> 
> Adds regression tests for **non-ASCII metadata round-trip** via `metadata_save`/`metadata_load` and for **non-ASCII custom template paths** in `from_template`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a3791d0ee3010487fb2360d024f7fa5e4e048a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/repocard.py
+++ b/src/huggingface_hub/repocard.py
@@ -325,9 +325,9 @@ class RepoCard:
         kwargs.update(template_kwargs)  # Template_kwargs have priority
 
         if template_path is not None:
-            template_str = Path(template_path).read_text()
+            template_str = Path(template_path).read_text(encoding="utf-8")
         if template_str is None:
-            template_str = Path(cls.default_template_path).read_text()
+            template_str = Path(cls.default_template_path).read_text(encoding="utf-8")
         template = jinja2.Template(template_str)
         content = template.render(card_data=card_data.to_yaml(), **kwargs)
         return cls(content)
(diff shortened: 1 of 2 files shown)


Wauplin:
Closing this issue as it feels low impact. Please let us know if you encounter a real issue around utf-8 encoding and we will figure out a solution.

PR #3030, opened 2025-04-25
Super-micro-tiny-PR to allow for direct copy-paste :)

fracapuano:
Ciao 👋 Super-tiny PR to enable direct copy paste of the new `pip install "huggingface_hub[hf_xet]"` on MacOS :))

PS: loving XET's speed 💨 ⚡

Issue #2988, opened 2025-04-08
Best practice for Unit Testing with Hugging Face models


davidmezzetti:
Hello,

With [txtai](https://github.com/neuml/txtai), I run a large number of unit tests with transformers models. Unfortunately, lately it's become untenable for me to run automated builds. I'm always running into 429 errors, file not found errors or other similar issues. Usually I just re-run the tests and it eventually works (which is probably even more wasteful of resources).

I've set the `HF_HUB_ETAG_TIMEOUT` and `HF_HUB_DOWNLOAD_TIMEOUT` values to be 100s. I'm open to other settings that would throttle the calls, so that the jobs would finish even if they take longer. I'm just not sure what the best practices are.

Please advise. This is definitely new behavior in the last couple of weeks.

julien-c:
do you pass a HF_TOKEN?

davidmezzetti:
I don't currently but if that is a way to work around this, I can add a GH secret for a read-only token.

hanouticelina:
Hi @davidmezzetti,
For testing, I suggest using [pytest-rerunfailures](https://pypi.org/project/pytest-rerunfailures/) package to be able to automatically retry specific tests that fail due to intermittent network issues. We currently use it to run our tests in `huggingface_hub`.
```bash
pytest --reruns 8 --reruns-delay 2 --only-rerun '(OSError|Timeout|HTTPError.*502|HTTPError.*504|not less than or equal to 0.01)' your_test_directory/
```
This basically tells pytest to retry failing tests up to 8 times with a 2-second delay between each retry attempt and we can target specific failures.

For rate limits (429 errors), unauthenticated requests have a lower threshold, whereas authenticated requests offer a higher limit. So, it’s worth testing while authenticated, as suggested by @julien-c. Another possible solution would be to setup some caching to minimize network calls.

davidmezzetti:
Thank you for the responses. I'll take a look at this. Though I don't use pytest for testing. Perhaps there is something similar for the standard Python unit testing framework.

julien-c:
Passing a `HF_TOKEN` should already help a lot

davidmezzetti:
Got it, I appreciate the guidance.

PR #4811, opened 2026-09-04
fix: atomic _create_symlink to prevent race under DDP

Swigler:
## Summary

- **Root cause of [vllm-project/llm-compressor#2984](https://github.com/vllm-project/llm-compressor/issues/2984):** `_create_symlink` does `os.remove(dst)` then `os.symlink(src, dst)` with ~40 lines of path computation between them. Any concurrent `os.path.isfile(dst)` call returns `False` during that window.
- **Impact:** Under DDP multi-GPU model loading (e.g. 131 safetensors shards × 8 GPUs), concurrent `from_pretrained` calls hit this window reliably, crashing with `OSError: does not appear to have a file named model-XXXXX.safetensors`. Two prior PRs to llm-compressor ([#2988](https://github.com/vllm-project/llm-compressor/issues/2988), [#3030](https://github.com/vllm-project/llm-compressor/issues/3030)) tried to fix it downstream with prefetch+barrier, but were rejected because the bug is here in huggingface_hub.
- **Fix:** Replace `os.remove()` + `os.symlink()` with atomic `os.symlink(src, tmp)` + `os.replace(tmp, dst)`. `os.replace()` is atomic on POSIX, so concurrent readers always see a valid symlink.

## Measured results

| Version | `os.path.isfile()` checks | Returned False | Rate |
|---|---|---|---|
| **Before** (os.remove → os.symlink) | 96,823 | 1,970 | **2.03%** |
| **After** (os.symlink tmp → os.replace) | 93,390 | 0 | **0.00%** |

Reproduced consistently across multiple runs on two different machines.

The existing comment on line 689 already acknowledged this window: *"exactly between `os.remove` and `os.symlink`"* — but only handled the `FileExistsError` case (two creates racing), not the missing-file case (a read during the gap).

## Test plan

- [x] Updated `test_create_symlink_concurrent_access` for new atomic behavior
- [x] Added `test_create_symlink_atomic_no_race` — thread-based test with 2 writers + 2 readers racing on 20 symlinks for 1 second, asserts zero `os.path.isfile()` misses
- [x] Existing `test_create_symlink_relative_src` passes unchanged
- [x] All 3 tests pass: `pytest tests/test_file_download.py::TestCreateSymlink -xvs`

---

> [!NOTE]
> **Medium Risk**
> Touches core cache pointer creation under concurrency; behavior change is intentional (atomic replace) but affects all symlink-based cache downloads on POSIX.
> 
> **Overview**
> Fixes a concurrency bug in **`_create_symlink`** when building Hub cache snapshot pointers. The old flow removed the destination symlink before creating a new one, so parallel loaders (e.g. DDP **`from_pretrained`**) could briefly see missing shard files and fail with “does not appear to have a file named …”.
> 
> Symlink updates now use a **temp symlink + `os.replace()`** (with mkdir retry on missing parent dirs), matching the library’s other atomic cache writes. The **`FileExistsError`** / wrong-target handling from remove-then-link is dropped in favor of overwrite via replace. **`PermissionError`** still falls through to copy/move, with temp symlink cleanup.
> 
> Tests refresh concurrent-access expectations and add **`test_create_symlink_atomic_no_race`** (multi-threaded writers/readers) to guard the regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c625965af939451f8ece963b8a5c160898ac6546. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/file_download.py
+++ b/src/huggingface_hub/file_download.py
@@ -640,11 +640,6 @@ def _create_symlink(src: str, dst: str, new_blob: bool = False) -> None:
     In case symlinks are not supported, a warning message is displayed to the user once when loading `huggingface_hub`.
     The warning message can be disabled with the `DISABLE_SYMLINKS_WARNING` environment variable.
     """
-    try:
-        os.remove(dst)
-    except OSError:
-        pass
-
     abs_src = os.path.abspath(os.path.expanduser(src))
     abs_dst = os.path.abspath(os.path.expanduser(dst))
     abs_dst_folder = os.path.dirname(abs_dst)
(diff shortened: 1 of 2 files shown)


Wauplin:
Hi @Swigler, thanks for the report and the proposed PR. This look like a very-old concurrency issue (introduced 4.5 years ago^^) happening when running `snapshot_download` concurrently from many processes using the same shared cache. After some (claude-led) investigation, it looks like your PR does not really fix the issue. I have opened https://github.com/huggingface/huggingface_hub/pull/4829 instead with some more details. Could you try it out and let me know what you think?

Swigler:
Closing in favor of #4829 which fixes the actual root cause (ref truncation in `snapshot_download`, not the symlink race). Thanks @Wauplin for the thorough investigation and repro.

PR #4817, opened 2026-09-05
[Download] Prevent cached file copies during local-dir dry runs

wakamex:
## Summary

- Check `dry_run` before copying a file from the Hub cache into `local_dir` and writing its download metadata.
- Add a regression test for both a missing destination file and an existing file with outdated contents, and verify that a normal download still copies the cached file.

I hit this while preparing a large model file for llama.cpp on a mechanical drive. I ran `hf download --dry-run --local-dir ...` to check the download, but the file was already in the Hub cache and the command started copying it into the destination. It appeared to hang, and stopping it left an incomplete model file there.

The cache-hit branch called `shutil.copyfile` before checking `dry_run`, so it could copy a large file or overwrite an existing destination even though the returned information said `will_download=False`. Moving the existing dry-run return ahead of the copy avoids that work.

For a small local reproduction, I seeded a temporary Hub cache with a 14-byte `weights.bin`, mocked only the remote metadata lookup, and called:

```python
info = hf_hub_download(
    "test/cached-model",
    "weights.bin",
    revision="a" * 40,
    cache_dir=root / "cache",
    local_dir=root / "local",
    dry_run=True,
)
```

Actual output using the original function and the patched function:

```text
Before: is_cached=True, will_download=False, destination_exists=True
After: is_cached=True, will_download=False, destination_exists=False
```

Both regression cases fail against the original function: the missing destination gets created, and the existing destination gets overwritten. Both pass with the fix. The focused regression, local-folder, symlink, and ETag tests report `63 passed, 4 skipped`. `make style` and `make quality` also pass.

I kept this focused on the cache-copy branch. Making every dry-run path avoid directory creation and metadata refresh would involve separate behavior changes and tests. There are no API changes, and the normal cache-copy path is preserved.

---

> [!NOTE]
> **Low Risk**
> Behavior-only fix for dry_run on an existing local_dir cache-hit path; normal downloads are unchanged.
> 
> **Overview**
> Fixes **`dry_run=True`** with **`local_dir`** when the file is already in the Hub cache: the cache-hit path used to **`shutil.copyfile`** into the destination and write download metadata **before** honoring dry run, so a “preview” could copy large files or overwrite an existing local file while reporting **`will_download=False`**.
> 
> The change **returns `DryRunFileInfo` first** when `dry_run` is set, and only then performs the locked copy and metadata write on a normal download. No API changes; non–dry-run cache copies behave as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4219849a361195e883331bb209fd92913249ac7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_file_download.py
+++ b/tests/test_file_download.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import hashlib
 import io
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


tests/test_file_download.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import hashlib
```
Wauplin:
```suggestion
```

tests/test_file_download.py
```diff
@@ -853,6 +854,54 @@ def test_passing_token_false_is_respected(self, mocker):
...
+        else:
+            assert target.read_bytes() == existing_content
+        assert not (local_dir / ".cache" / "huggingface" / "download" / f"{filename}.metadata").exists()
+
+        # An actual download must still copy the cached file into the destination.
+        result = hf_hub_download(repo_id, filename, revision=revision, cache_dir=cache_dir, local_dir=local_dir)
+        assert result == str(target)
+        assert target.read_bytes() == content
+
+
```
Wauplin:
```suggestion
```

Wauplin (commented):
Thanks for the PR! This is indeed a valid point. Looks good but let's not add a unit test for such a small corner case

Wauplin (approved):
✔️

PR #4823, opened 2026-09-07
[CLI] Don't let `hf update` self-upgrade a pip install on Windows

### LINKED ISSUE #4819: `hf update` on a Windows pip install fails with WinError 32 and can leave the installation broken
### Describe the bug

On Windows, `hf update` for a pip installation cannot succeed. The `pip` case of `_get_huggingface_hub_update_command()` (`cli/_cli_utils.py`) runs `pip install -U huggingface_hub` as a child of the running `hf.exe`, and Windows refuses to replace the launcher of a running process, so pip fails with WinError 32 on `Scripts\hf.exe`. By that point pip has begun uninstalling the old version, so the environment can end up with no `huggingface_hub` module and a dead `hf.exe`, as in the reproduction below. The same upgrade succeeds when run outside hf (`hfrepro\Scripts\python -m pip install -U huggingface_hub`) and repairs the broken state. Verified on 1.30.0; the `pip` branch on current main is identical and has no Windows handling, unlike the `hf_installer` branch beside it.

One possible fix, mirroring how pip treats its own self-upgrade on Windows: in the `pip` case on Windows, print the `python -m pip install -U huggingface_hub` command for the user to run instead of executing it from inside `hf.exe`. I can send a PR if useful.

### Reproduction

```
python -m venv hfrepro
hfrepro\Scripts\pip install huggingface_hub==1.29.0
hfrepro\Scripts\hf update
hfrepro\Scripts\hf version
```

### Logs

```shell
> hfrepro\Scripts\hf update
Current version: 1.29.0
Checking for updates to latest version...
Collecting huggingface_hub
[...]
Installing collected packages: huggingface_hub
  Attempting uninstall: huggingface_hub
    Found existing installation: huggingface_hub 1.29.0
    Uninstalling huggingface_hub-1.29.0:
ERROR: Could not install packages due to an OSError: [WinError 32] The process cannot access the file because it is being used by another process: 'c:\\users\\user\\appdata\\local\\temp\\hfrepro\\scripts\\hf.exe'
Check the permissions.
(exit code 1)

> hfrepro\Scripts\hf version
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\user\AppData\Local\Temp\hfrepro\Scripts\hf.exe\__main__.py", line 4, in <module>
ModuleNotFoundError: No module named 'huggingface_hub'
```

### System info

```shell
- huggingface_hub version: 1.30.0
- Python version: 3.12.10
- Running in iPython ?: No
- Running in notebook ?: No
- Token path ?: C:\Users\user\.cache\huggingface\token
- Has saved token ?: True
- Who am I ?: [redacted]
- Configured git credential helpers: manager
- Run by AI agent ?: Yes
- Installation method: pip
- httpx: 0.28.1
- hf_xet: 1.6.0
- gradio: N/A
- tensorboard: N/A
- HF_HUB_CACHE: C:\Users\user\.cache\huggingface\hub
- HF_ASSETS_CACHE: C:\Users\user\.cache\huggingface\assets
- HF_TOKEN_PATH: C:\Users\user\.cache\huggingface\token
- HF_STORED_TOKENS_PATH: C:\Users\user\.cache\huggingface\stored_tokens
- HF_HUB_OFFLINE: False
- HF_HUB_DISABLE_TELEMETRY: True
- HF_HUB_DISABLE_XET: False
- HF_HUB_ETAG_TIMEOUT: 10
- HF_HUB_DOWNLOAD_TIMEOUT: 10
```

Wauplin:
## Summary

On Windows, a running `.exe` can't be replaced, so `pip install -U huggingface_hub` launched from inside `hf.exe` blows up with `WinError 32` on `Scripts\hf.exe` — and it does so *after* pip has already uninstalled the current version, which leaves the user with a dead launcher and no `huggingface_hub` module at all (see the reproduction in #4819). The same upgrade works fine when pip isn't a child of `hf.exe`.

So `hf update` now simply refuses to do it: on Windows pip installs it prints the `python -m pip install -U huggingface_hub` command for the user to run themselves and exits non-zero, instead of half-breaking the install and reporting failure afterwards. This mirrors what pip itself does when asked to upgrade pip on Windows. Non-Windows pip installs, Homebrew and the standalone installer are untouched — the installer path already handles Windows on its own.

Simulated locally by faking `sys.platform`/installation method (no Windows box at hand), which gives:

```
> hf update
Current version: 1.30.0
Checking for updates to latest version...
Error: On Windows, a pip-installed `hf` cannot update itself: pip is not allowed to replace `hf.exe` while it is running. Run this command instead:
    C:\path\to\venv\Scripts\python.exe -m pip install -U huggingface_hub
```

The only test change is a `sys.platform` patch in `TestUpdateSkillOptOut`, so those tests keep exercising the normal update path when they run on Windows CI.

Fixes #4819. Alternative to #4820, which keys off `sys.argv[0]` to keep `python -m huggingface_hub.cli.hf update` working in-process; that path is safe but obscure enough that guarding the whole platform+install-method combination felt simpler.

---

> [!NOTE]
> **Low Risk**
> Narrow guard on Windows + pip only; avoids a known failure mode without changing update behavior elsewhere.
> 
> **Overview**
> **`hf update`** now **refuses to run pip in-process** when the CLI is a **pip install on Windows**, instead of risking a broken install where pip uninstalls `huggingface_hub` then fails to replace `hf.exe` (WinError 32).
> 
> In that case it prints the exact `python -m pip install -U huggingface_hub` command to run in another shell and exits with code 1, similar to pip’s own self-upgrade behavior. Other platforms and install methods (Homebrew, standalone installer) are unchanged.
> 
> Tests in **`TestUpdateSkillOptOut`** patch **`sys.platform`** to **`linux`** so they still exercise the normal update path on Windows CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22c9cca8c7801d778ef47cdb2bf0d17a6dce2810. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/system.py
+++ b/src/huggingface_hub/cli/system.py
@@ -20,6 +20,7 @@ import sys
 import click
 
 from huggingface_hub import __version__, constants
+from huggingface_hub.errors import CLIError
 
 from ..utils import dump_environment_info, installation_method
 from ._cli_utils import _fetch_latest_pypi_version, run_update
@@ -45,6 +46,17 @@ def update() -> None:
     if latest_version is not None and __version__ == latest_version:
         out.text(f"hf is up to date ({__version__})")
         return
+
+    # Windows refuses to replace the `hf.exe` launcher of a running process: pip would fail with a
+    # "file in use" error (WinError 32) *after* having uninstalled the current version, leaving a dead
+    # `hf.exe` and no `huggingface_hub` module behind. Print the command instead of running it, the same
+    # way pip refuses to upgrade itself on Windows.
+    if sys.platform == "win32" and installation_method() == "pip":
+        command = subprocess.list2cmdline([sys.executable, "-m", "pip", "install", "-U", "huggingface_hub"])
+        raise CLIError(
+            "On Windows, a pip-installed `hf` cannot update itself: pip is not allowed to replace `hf.exe` while it "
+            f"is running. Run this command instead:\n    {command}"
+        )
 
     # The standalone installer installs the `hf-cli` skill by default. If it's not installed at this
     # point, the user opted out (or removed it): tell the installer to leave it alone instead of
--- a/src/huggingface_hub/cli/system.py
+++ b/src/huggingface_hub/cli/system.py
@@ -20,7 +20,6 @@ import sys
 import click
 
 from huggingface_hub import __version__, constants
-from huggingface_hub.errors import CLIError
 
 from ..utils import dump_environment_info, installation_method
 from ._cli_utils import _fetch_latest_pypi_version, run_update
@@ -53,10 +52,11 @@ def update() -> None:
     # way pip refuses to upgrade itself on Windows.
     if sys.platform == "win32" and installation_method() == "pip":
         command = subprocess.list2cmdline([sys.executable, "-m", "pip", "install", "-U", "huggingface_hub"])
-        raise CLIError(
+        out.error(
             "On Windows, a pip-installed `hf` cannot update itself: pip is not allowed to replace `hf.exe` while it "
             f"is running. Run this command instead:\n    {command}"
         )
+        raise click.exceptions.Exit(code=1)
 
     # The standalone installer installs the `hf-cli` skill by default. If it's not installed at this
     # point, the user opted out (or removed it): tell the installer to leave it alone instead of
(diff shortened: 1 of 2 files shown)


Wauplin (commented):
A bit too bad but according to Claude and Codex there isn't really a simple solution for that so let's just print the hint to the user and that's it :heavy_check_mark:

hanouticelina (approved):
Thank you!

PR #4842, opened 2026-09-08
Fix dotenv parser truncating unquoted values containing '#'

BenYang12:
## Summary

- `load_dotenv` (in `utils/_dotenv.py`, used by `hf jobs --env` / `--env-file`) treats *any* `#` in an unquoted value as the start of an inline comment, so it silently truncates the value.
- By the usual dotenv convention (and `python-dotenv`), a `#` only starts an inline comment when it's **preceded by whitespace**. A `#` that's part of the value — common in passwords, tokens and URL fragments — should be kept.
- This makes unquoted values consistent with quoted ones (which already keep `#`) and with the whitespace-delimited comments the existing tests assume.

## Before / after

```python

# before my PR
load_dotenv("PASSWORD=p@ss#word")             # {'PASSWORD': 'p@ss'}      <- truncated
load_dotenv("TOKEN=abc#123")                  # {'TOKEN': 'abc'}          <- truncated
load_dotenv("URL=http://example.com/x#frag")  # {'URL': 'http://example.com/x'}

# after my PR
load_dotenv("PASSWORD=p@ss#word")             # {'PASSWORD': 'p@ss#word'}
load_dotenv("TOKEN=abc#123")                  # {'TOKEN': 'abc#123'}
load_dotenv("URL=http://example.com/x#frag")  # {'URL': 'http://example.com/x#frag'}
load_dotenv("KEY=value  # real comment")      # {'KEY': 'value'}          <- unchanged
```

## Fix

In the `line_pattern` regex, the unquoted-value branch now allows `#` (`[^#\n\r]+?` → `[^\n\r]*?`), and the inline comment must be preceded by horizontal whitespace (`[^\S\n]*(?:\#.*)?$` → `(?:[^\S\n]+\#[^\n\r]*)?$`). Quoted values and bare keys keep their own trailing-comment branch, since after a closing quote (or a bare key) a `#` starts a comment even with no whitespace in front of it.

No behaviour change for quoted values, whitespace-separated comments, `export`, empty values, or bare keys.

## One more behaviour change

Splitting the regex into explicit "assignment" / "bare key" branches also fixed a second bug: a line that is neither of those used to match as a bare key, pull the host value in, and clobber an explicit assignment above it. It is now simply ignored:

```python
load_dotenv("SECRET=explicit\nSECRET is documented above", environ={"SECRET": "HOST_ENV"})
# before: {'SECRET': 'HOST_ENV'}  <- explicit value overwritten by the host env
# after:  {'SECRET': 'explicit'}
```

## Notes

- One intentional divergence from `python-dotenv`: for `KEY= # comment` it returns `'# comment'`, while this PR keeps the current `huggingface_hub` behaviour and returns `''` — whitespace right after `=` separates an inline comment here too. A leaked comment silently becoming a job env var or secret seems clearly worse than the small inconsistency.
- Because of the above, an unquoted value still can't contain ` #` (whitespace + hash); quote it if you need that.

Added `test_hash_in_unquoted_value_is_kept` for the fix, plus `test_empty_value_with_inline_comment`, `test_comment_attached_to_closing_quote`, `test_bare_key_with_inline_comment` and `test_invalid_line_does_not_import_from_environ` to pin down the surrounding cases the new regex has to keep working. All existing `tests/test_utils_dotenv.py` still pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how env/secrets files are parsed for CLI jobs; fixes incorrect truncation but alters values for any `.env` lines that relied on the old `#`-anywhere comment rule.
> 
> **Overview**
> Fixes **`load_dotenv`** (used by **`hf jobs --env-file` / `--secrets-file`**) so unquoted values are no longer cut off at the first **`#`**. Inline comments now follow the usual rule: **`#` only starts a comment when preceded by whitespace**, so passwords, tokens, and URL fragments keep their full value.
> 
> The line-matching regex is reworked to cover related edge cases: **`KEY= # comment`** yields an empty value (comment text does not leak into env vars), **`#` immediately after a closing quote** still starts a comment, **bare keys** with trailing comments still resolve from **`environ`**, and lines like **`SECRET is documented above`** are ignored instead of being treated as bare keys that could overwrite an earlier explicit assignment.
> 
> New tests in **`test_utils_dotenv.py`** lock in these behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16e9f21afb867778acf1b0fc26c9ad453775572c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_dotenv.py
+++ b/src/huggingface_hub/utils/_dotenv.py
@@ -34,10 +34,10 @@ def load_dotenv(dotenv_str: str, environ: dict[str, str] | None = None) -> dict[
             (?:
                 '(?:\\'|[^'])*'           # single-quoted value
                 | \"(?:\\\"|[^\"])*\"     # double-quoted value
-                | [^#\n\r]+?              # unquoted value
+                | [^\n\r]*?               # unquoted value (may contain '#')
             )
         )?
-        [^\S\n]*(?:\#.*)?$                # optional inline comment
+        (?:[^\S\n]+\#[^\n\r]*)?$          # optional inline comment (must be preceded by whitespace)
     """,
         re.VERBOSE,
     )
--- a/src/huggingface_hub/utils/_dotenv.py
+++ b/src/huggingface_hub/utils/_dotenv.py
@@ -29,15 +29,24 @@ def load_dotenv(dotenv_str: str, environ: dict[str, str] | None = None) -> dict[
         ^\s*
         (?:export[^\S\n]+)?               # optional export
         ([A-Za-z_][A-Za-z0-9_]*)          # key
-        [^\S\n]*(=)?[^\S\n]*
-        (                                 # value group
+        [^\S\n]*
+        (?:
+            (=)                           # equal sign
             (?:
-                '(?:\\'|[^'])*'           # single-quoted value
-                | \"(?:\\\"|[^\"])*\"     # double-quoted value
-                | [^\n\r]*?               # unquoted value (may contain '#')
+                [^\S\n]*
+                (                         # quoted value
+                    '(?:\\'|[^'])*'       # single-quoted
+                    | \"(?:\\\"|[^\"])*\" # double-quoted
+                )
+                [^\S\n]*(?:\#[^\n\r]*)?   # inline comment (needs no preceding whitespace after a quote)
+                |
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Looks good, thank you!

PR #4843, opened 2026-09-09
[CLI] Reject negative and signed values in parse_size / parse_duration

BenYang12:
The main issue: the plain-number path in `parse_size` / `parse_duration` accepts values it shouldn't: `parse_size("-5")` returns `-5` and `parse_duration("-5")` returns `-5`, even though the unit form `"-5M"` is already rejected (there's a test for that) and sizes/durations can't be negative. `"+10"` and `"1_000"` slip through the same way.

## Summary

- `_parse_with_unit` tries `int(value)` before the unit regex. `int()` accepts a leading sign and PEP 515 underscores, so `-5`, `+10`, `1_000` are treated as valid, while the unit branch (`RE_NUMBER_WITH_UNIT` = `\d+[a-z]+`) rejects the unit equivalents like `-5M`.
- These parsers back `hf jobs run --timeout`, `hf spaces ... --timeout`, and the `scan-cache` / `delete-cache` size filters, so a negative value silently became a negative timeout instead of erroring.
- Chosen fix: take the bare-integer fast path only for non-negative digits (`re.fullmatch(r"\d+", stripped)`), matching the `\d+` the unit form already uses — same contract on both branches.

### Before

```python
>>> from huggingface_hub.utils._parsing import parse_size, parse_duration
>>> parse_size("-5")
-5
>>> parse_duration("-5")
-5
>>> parse_size("1_000")
1000
```

### After

```python
>>> parse_size("-5")
ValueError: Invalid value '-5'. Must match pattern '\d+[a-z]+' or be a plain number.
>>> parse_size("10"), parse_size("5M")
(10, 5000000)
```

Added `-5`, `+10`, `1_000` to the existing invalid-parametrize lists for both functions.

## Checks

`pytest tests/test_utils_parsing.py` → 57 passed. `ruff format` / `ruff check` clean.


hanouticelina:
Hi @BenYang12 , thanks for the PR! did you actually run into this in a real use case, or did you find it by reading the code? we actually prioritise fixing bugs that people hit in practice, so without a concrete report I'd rather not change this. closing for now, but happy to reopen if you can share an example where the current behavior caused a problem!

PR #4846, opened 2026-09-09
[CLI] Raise explicit error for shell-script extensions on Windows

hanouticelina:
Follow-up to Wauplin's suggestion in https://github.com/huggingface/huggingface_hub/pull/4764#issuecomment-5581909996.

## Summary

- Shell-script extensions never worked on Windows: the installer looked for `hf-<name>.exe`, didn't find it, fell back to a Python install and failed with a confusing uv/pip error (see https://github.com/huggingface/hf-claude/issues/5).
- `hf extensions install` now fails early with an explicit error when the repo ships a shell-script extension and the user is on Windows. Python extensions are unchanged.
- Added a note to the extensions guide.

Error shown on Windows:

```
'huggingface/hf-claude' is a shell-script extension, which is not supported on Windows. Only Python extensions can be installed on Windows.
```

---

> [!NOTE]
> **Low Risk**
> Targeted CLI install-path change for Windows plus clearer errors; Python extension installs and non-Windows behavior are unchanged aside from the binary URL probe fix.
> 
> **Overview**
> **Windows users** installing a repo that ships a root **`hf-<name>`** shell script no longer fall through to a Python/uv install and obscure pip errors. **`hf extensions install`** (and auto-install after confirmation) now **fails fast** with an explicit message that only Python extensions work on Windows.
> 
> The remote binary probe always requests **`hf-{short_name}`** from raw GitHub (not a **`.exe`** name on Windows), so shell-script repos are detected correctly before install logic runs.
> 
> **Auto-install** of official extensions no longer swallows install failures in a broad **`try/except`**, so real install errors surface after the user confirms.
> 
> The extensions guide adds a **warning** that shell-script extensions are unsupported on Windows. Shell-script install tests are **skipped on Windows** in CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8db1543de1a07fc14d74dcc06c5cdaeab0d44b77. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/en/guides/cli-extensions.md
+++ b/docs/source/en/guides/cli-extensions.md
@@ -39,6 +39,10 @@ falls back to installing the repo as a Python package.
 A shell script extension is the simplest type. You only need a GitHub repository with an executable file
 named `hf-<name>` at the root.
 
+> [!WARNING]
+> Shell script extensions are not supported on Windows. If your extension must work on Windows, make it a
+> [Python extension](#create-a-python-extension) instead.
+
 ### Minimal example
 
 Create a repository named `hf-hello` on GitHub with a single file:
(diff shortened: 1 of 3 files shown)


hanouticelina (commented):
✅

Wauplin (approved):
Thanks!

PR #4235, opened 2026-05-19
[CLI] Support hf:// URIs in cache rm

abhinavgautam01:
## Summary

- Add support for repo-level `hf://` URIs in `hf cache rm`
- Reuse the centralized `parse_hf_uri` helper introduced in #4189
- Keep revisions, filenames and buckets out of scope for now with clear errors
- Update CLI docs and generated CLI reference

Example:

```bash
hf cache rm hf://models/openai-community/gpt2 --dry-run
```

For now this only supports repo-level URIs like hf://models/openai-community/gpt2. Follow-up PRs can expand this to revisions and filenames...

Validation:
make style
make quality
PYTHONPATH=src python3 -m pytest tests/test_cli.py::TestCacheCommand
tests/test_cli.py::TestCacheCommand passes locally. 
Full tests/test_cli.py hits unrelated network/DNS failures in production-style tests.

---

> [!NOTE]
> **Medium Risk**
> Extends cache-deletion target parsing to accept `hf://` inputs; while scoped to repo-level URIs and covered by new tests, mistakes could cause unintended cache removals if URI-to-cache-id resolution is wrong.
> 
> **Overview**
> `hf cache rm` now accepts *repo-level* `hf://` URIs (e.g. `hf://models/openai-community/gpt2`) in addition to `model/...` IDs and revision hashes by resolving URIs via `parse_hf_uri`.
> 
> The command explicitly rejects non-repo URIs (e.g. buckets) and any `hf://` URI that includes a revision (`@...`) or a path, and the CLI docs/reference plus tests were updated to reflect and validate this behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08c0c4020c9f38909652e72fab5324dac97fdcbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4847, opened 2026-09-09
[CLI] Support file URIs in hf cache rm

### LINKED ISSUE #4124: hf cache rm -> ability to remove a specific quantization variant!

If you have a model in the cache with multiple quant variants, for example:
gemma-4-26B-A4B-it-GGUF
UD-IQ2_M
UD-IQ3_S
UD-IQ4_NL

There is no easy way to delete only one of them, you have to find the blobs yourself and delete them and the link to it.
Or delete everything, then redownload only what you need...

The hf cli should have a way to delete specific model and quant
It should ideally match exactly what was used to download it the first place, for example:
With llama.cpp, you do:
llama-server ...  -hf unsloth/gemma-4-26B-A4B-it-GGUF:UD-IQ4_NL
Then to remove this model, you should be able to do:
hf cache rm unsloth/gemma-4-26B-A4B-it-GGUF:UD-IQ4_NL

Option 1 (my current method): I have a script that first deletes the files I don't want under snapshots (based on what is in my llama-swap config.yaml), then a second pass look for any blobs that have no symbolic link and delete them (I don't know if this is the proper way, and I don't know if this will always work in the future, but it seems ok so far, I have more space on my disk and models I need still seem to cached and work)

Option 2: Delete the entire model with hf cache rm, then redownload the variants you need... this is obviously not great, this is very time consuming and bandwidth consuming (for huggingface as well I'm sure). 

**Additional context**

If you are constantly using (or testing) different models with different quant, you end up with a huge amount of data to store, and once you decided on your best quant for all your different models, there is no easy way to clean this up based on what I know and tried so far.

hanouticelina:
Follow-up to #4235. `hf cache rm` now accepts `hf://` file URIs to remove a single cached file, typically one GGUF quantization, without deleting the whole repo. Closes #4124.

## Summary

- `hf cache rm hf://models/<repo_id>/<path>` removes the file from every cached revision of the repo. The blob is deleted only if no other cached file still references it. Refs and snapshot folders are kept.
- Exact paths only: no folders, no globs. File targets can't be mixed with repo or revision targets in the same call. `@revision` in `hf://` URIs is still rejected, for repos and files alike, with a hint on what to do instead.
- New `HFCacheInfo.delete_files(*files)` next to `delete_revisions`, returning the usual `DeleteCacheStrategy`. The strategy gets a `files` field: snapshot entries that `execute()` unlinks first, before any blob is deleted.
- Blob accounting is a `blobs_to_unlink` map like in `delete_revisions`, so the shared blob store (#4498) can plug in later with a two-line change.

Manual test on a temp cache with two revisions sharing the Q4 blob:

```text
$ hf cache ls --revisions
ID              REVISION                             SIZE LAST_MODIFIED     REFS
--------------- ----------------------------------- ----- ----------------- ----
model/org/model aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa... 12.6M a few seconds ago main
model/org/model bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb...  4.2M a few seconds ago

$ hf cache rm hf://models/org/model/model-Q4_K_M.gguf --dry-run
About to delete 2 file(s) totalling 4.2M.
  - model/org/model@aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/model-Q4_K_M.gguf
  - model/org/model@bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb/model-Q4_K_M.gguf
✓ Dry run: no files were deleted.

$ hf cache rm hf://models/org/model/model-Q4_K_M.gguf -y
About to delete 2 file(s) totalling 4.2M.
  - model/org/model@aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/model-Q4_K_M.gguf
  - model/org/model@bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb/model-Q4_K_M.gguf
Cache deletion done. Saved 4.2M.
✓ Deleted 2 file(s); freed 4.2M.

$ hf cache ls --revisions
ID              REVISION                             SIZE LAST_MODIFIED     REFS
--------------- ------------------------------------ ---- ----------------- ----
model/org/model aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa... 8.4M a few seconds ago main
model/org/model bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb...  0.0 a few seconds ago

$ hf cache rm hf://models/org/model@main/model-Q8_0.gguf
Error: Revisions in hf:// URIs are not supported by `hf cache rm`. Pass a revision hash to delete a revision, or drop '@<revision>' to delete a file from all cached revisions.

$ hf cache rm hf://models/org/model/model-Q8_0.gguf model/org/model
Error: File targets cannot be mixed with repository or revision targets.
```

---

> [!NOTE]
> **Medium Risk**
> Changes local cache deletion semantics and filesystem unlink order; mistakes could remove wanted blobs, though confirmation/dry-run and shared-blob retention limit impact.
> 
> **Overview**
> **`hf cache rm`** can now target a single cached file via an `hf://` file URI (e.g. one GGUF quantization), removing that path from every cached revision while keeping refs and snapshot folders; blobs are dropped only when nothing else references them.
> 
> The CLI resolves file URIs separately from repo/revision targets (they cannot be mixed), rejects `@revision` on `hf://` inputs with clearer hints, and uses a new **`HFCacheInfo.delete_files`** path that returns **`DeleteCacheStrategy`** with a **`files`** set—`execute()` unlinks those snapshot entries before deleting unshared blobs. Docs and tests cover dry-run output, shared-blob behavior, and invalid target combinations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a7caa52e930f9631afa4a4e260361ed4d680c74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/cli/cache.py
+++ b/src/huggingface_hub/cli/cache.py
@@ -160,10 +160,12 @@ def _parse_hf_uri_target(target: str) -> HfUri:
     if not uri.is_repo:
         raise CLIError("Only repository hf:// URIs are supported by `hf cache rm`.")
     if uri.revision is not None:
-        raise CLIError(
-            "Revisions in hf:// URIs are not supported by `hf cache rm`. Pass a revision hash to delete a revision, "
-            "or drop '@<revision>' to delete a file from all cached revisions."
+        hint = (
+            "Drop '@<revision>' to delete the file from all cached revisions."
+            if uri.path_in_repo
+            else "Pass the revision hash instead, see 'hf cache ls --revisions'."
         )
+        raise CLIError(f"Revisions in hf:// URIs are not supported by `hf cache rm`. {hint}")
     return uri
 
 
--- a/src/huggingface_hub/cli/cache.py
+++ b/src/huggingface_hub/cli/cache.py
@@ -342,6 +342,7 @@ def compile_cache_sort(sort_expr: str) -> tuple[Callable[[CacheEntry], tuple[Any
 
 def _resolve_deletion_targets(hf_cache_info: HFCacheInfo, targets: list[str]) -> _DeletionResolution:
     """Resolve the deletion targets into a deletion resolution."""
+    targets = [stripped for target in targets if (stripped := target.strip())]
     file_uris: dict[str, HfUri] = {}
     for target in targets:
         if target.startswith("hf://") and (uri := _parse_hf_uri_target(target)).path_in_repo:
@@ -357,16 +358,13 @@ def _resolve_deletion_targets(hf_cache_info: HFCacheInfo, targets: list[str]) ->
     revisions: set[str] = set()
     missing: list[str] = []
 
-    for raw_target in targets:
-        target = raw_target.strip()
-        if not target:
-            continue
+    for target in targets:
         lowered = target.lower()
 
         if re.fullmatch(r"[0-9a-fA-F]{40}", lowered):
             match = revision_lookup.get(lowered)
             if match is None:
-                missing.append(raw_target)
+                missing.append(target)
                 continue
             repo, revision = match
             selected[repo].add(revision)
@@ -375,7 +373,7 @@ def _resolve_deletion_targets(hf_cache_info: HFCacheInfo, targets: list[str]) ->
 
         matched_repo = repo_lookup.get(_repo_cache_id_from_target(target).lower())
         if matched_repo is None:
-            missing.append(raw_target)
+            missing.append(target)
             continue
 
         for revision in matched_repo.revisions:
--- a/src/huggingface_hub/cli/cache.py
+++ b/src/huggingface_hub/cli/cache.py
@@ -167,11 +167,8 @@ def _parse_hf_uri_target(target: str) -> HfUri:
     return uri
 
 
-def _repo_cache_id_from_target(target: str) -> str:
-    """Return the cache id matching a repo target passed to `hf cache rm`."""
-    if not target.startswith("hf://"):
-        return target
-    uri = _parse_hf_uri_target(target)
+def _cache_id(uri: HfUri) -> str:
+    """Return the `type/id` cache id of a repo URI."""
     return f"{uri.type}/{uri.id}"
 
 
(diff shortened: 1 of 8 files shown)


hanouticelina (commented):
✅

hanouticelina:
bugbot run

src/huggingface_hub/cli/cache.py
```diff
@@ -331,31 +341,37 @@ def _sort_key(entry: CacheEntry) -> tuple[Any, ...]:
...
     repo_lookup, revision_lookup = build_cache_index(hf_cache_info)
 
     selected: dict[CachedRepoInfo, set[CachedRevisionInfo]] = defaultdict(set)
     revisions: set[str] = set()
     missing: list[str] = []
 
-    for raw_target in targets:
-        target = raw_target.strip()
-        if not target:
-            continue
```
Wauplin, line 343: any reason to remove the stripping here? if not I'd put it back even though it has low impact (can be useful if user copy pastes a list of target with some newlines in it?)

src/huggingface_hub/cli/cache.py
```diff
@@ -581,6 +620,28 @@ def rm(
         details = "\n".join(f"  - {entry}" for entry in resolution.missing)
         out.warning(f"Could not find in cache:\n{details}")
 
+    if resolution.files:
```
Wauplin, line 623: we consider that if someone delete files, they delete ONLY files, right? fine for me but I don't think we explicitly raise if someone passed files and revisions (meaning we might silently ignore one or the other)

Wauplin (commented):
LGTM, thank you for handling this! I have left two minor comments otherwise it looks good to me 👍

src/huggingface_hub/cli/cache.py
hanouticelina: actually this came from a bugbot comment, the stripping is still there, just movedt to L344, it strips every target and drops blank ones before we classify them. stripping inside the loop only covered repo ids and hashes. A file uri with a leading space skipped file detection, then got stripped in the repo branch and would have deleted the whole repo

src/huggingface_hub/cli/cache.py
hanouticelina: yes, only files. mixing is rejected explicitly in `cache.py L348`: if any target is a file URI, every target must be one, otherwise `CLIError("File targets cannot be mixed with repository or revision targets.")`. That covers repo ids, revision hashes and repo level `hf://` URIs.

Wauplin (approved):
Great, thank you!

PR #4868, opened 2026-09-10
[Http] Fix race in `get_session()` creating multiple clients

### LINKED ISSUE #4861: `get_session()` is not thread-safe
### Describe the bug

# `get_session()` is not thread-safe: racing threads each create and overwrite the global client, closing live connections

### Describe the bug

`get_session()` lazily creates the global client with broken double-checked locking — there is no re-check after acquiring the lock ([`utils/_http.py`](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/utils/_http.py)):

```python
def get_session() -> httpx.Client:
    global _GLOBAL_CLIENT
    if _GLOBAL_CLIENT is None:
        with _CLIENT_LOCK:
            _GLOBAL_CLIENT = _GLOBAL_CLIENT_FACTORY()   # <- no `if _GLOBAL_CLIENT is None:` re-check
    return _GLOBAL_CLIENT
```

Every thread that reads `None` before the first assignment queues on the lock and then unconditionally creates its own client, overwriting `_GLOBAL_CLIENT`. The overwritten clients are still in use by the threads that grabbed them; once garbage-collected, their pooled sockets are closed **under in-flight requests on other threads**, which then fail with connection errors that callers' retry layers often swallow silently.

Ironically, the v1.0 migration guide advertises the shared client with "httpx is thread-safe by design, allowing us to safely reuse the same client across multiple threads" — `httpx.Client` is, but the lazy initialization around it is not. (In v0.x this couldn't happen: `get_session()` returned one `requests.Session` per thread.)

### How we hit this

`datasets` `IterableDataset.decode(num_threads=N)` on a dataset whose `Audio` column references `hf://` paths: every decode thread opens files via `HfFileSystem`, which funnels into `get_session()` at cold start. `faulthandler` stack dumps showed worker threads queued in `get_session`, others in `socket._real_close` (the GC'd clients' sockets being torn down), and the resulting connection failures were absorbed by `datasets`' open-retry loop — so the symptom was simply "no progress, no errors". (To be fair, `decode(num_threads)`'s poor throughput has additional causes on the `datasets` side; this race is the huggingface_hub part, isolated by the repro above.)

### Expected behavior

`get_session()` creates the client exactly once:

```python
if _GLOBAL_CLIENT is None:
    with _CLIENT_LOCK:
        if _GLOBAL_CLIENT is None:
            _GLOBAL_CLIENT = _GLOBAL_CLIENT_FACTORY()
```

Two adjacent observations from the same debugging session, in case they're worth follow-ups:

- `HfFileSystem` pays a repo/bucket existence-check API round-trip (`repo_info`/`bucket_info` inside `resolve_path`) on **every file open** — for workloads opening many small files this doubles the API traffic; caching per (repo, revision) would help.
- `default_client_factory` sets `timeout=None`, so a thread stuck on a connection that was closed under it (e.g. by this race) waits forever instead of failing and retrying. This `timeout=None` is highly questionable.

### Reproduction

```python
import threading
from huggingface_hub import HfApi, set_client_factory
from huggingface_hub.utils._http import default_client_factory

creations = 0

def counting_factory():
    global creations
    creations += 1
    return default_client_factory()

set_client_factory(counting_factory)
api = HfApi()
barrier = threading.Barrier(16)

def probe():
    barrier.wait()
    api.dataset_info("datasets-examples/doc-audio-1")

threads = [threading.Thread(target=probe) for _ in range(16)]
[t.start() for t in threads]
[t.join() for t in threads]
print(f"clients created: {creations}")   # 16 — expected: 1
```
Result on our machine: **16 clients created for 16 threads** (and N for N threads generally). Monkeypatching the missing re-check into `get_session` brings it to exactly 1.

### Logs

```shell

```

### System info

```shell
huggingface_hub 1.9.2, httpx 0.28.1, httpcore 1.0.9, Python 3.14.0, Linux x86_64.
```

hanouticelina:
Fixes #4861.

## Summary
- `get_session()` checked `_GLOBAL_CLIENT is None` before taking the lock but not after, so N threads racing on the first call each created their own client, and only the last one survived as the shared client.
- Add the missing re-check inside the lock so the client is created exactly once.
- Add a regression test: 16 threads released by a barrier on a cold start.

Repro from the issue with a counting factory, main vs this PR:

    threads=16 clients_created=16   # main
    threads=64 clients_created=64
    threads=16 clients_created=1    # this PR
    threads=64 clients_created=1

---

> [!NOTE]
> **Low Risk**
> Small concurrency fix in shared HTTP client initialization with a targeted regression test; behavior is unchanged except eliminating duplicate client creation on parallel first access.
> 
> **Overview**
> Fixes a **double-checked locking** gap in `get_session()`: threads could all pass the outer `_GLOBAL_CLIENT is None` check on a cold start, enter the lock one after another, and each invoke `_GLOBAL_CLIENT_FACTORY()` before any assignment was visible—so many `httpx.Client` instances were created even though only the last one was kept as the global singleton.
> 
> The change adds the **second null check inside `_CLIENT_LOCK`** so only the first waiter creates the client; later threads reuse it.
> 
> A regression test runs **16 threads** synchronized with a barrier on first `get_session()` and asserts the counting factory runs **once** and every thread gets the **same** client instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c118ce7ac4b88a133ccd53f6f0d5e40b6b9054af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/utils/_http.py
+++ b/src/huggingface_hub/utils/_http.py
@@ -376,7 +376,8 @@ def get_session() -> httpx.Client:
     global _GLOBAL_CLIENT
     if _GLOBAL_CLIENT is None:
         with _CLIENT_LOCK:
-            _GLOBAL_CLIENT = _GLOBAL_CLIENT_FACTORY()
+            if _GLOBAL_CLIENT is None:
+                _GLOBAL_CLIENT = _GLOBAL_CLIENT_FACTORY()
     return _GLOBAL_CLIENT
 
 
--- a/tests/test_utils_http.py
+++ b/tests/test_utils_http.py
@@ -262,6 +262,30 @@ class TestConfigureSession:
             for j in range(N):
                 assert clients[i] is clients[j]
 
+    def test_get_session_concurrent_first_call_creates_single_client(self):
+        N = 16
+        created = []
+        barrier = threading.Barrier(N)
+        clients = [None] * N
+
+        def _factory() -> httpx.Client:
+            created.append(None)
+            return httpx.Client()
+
+        def _get_session_in_thread(index: int) -> None:
+            barrier.wait()
+            clients[index] = get_session()
+
+        set_client_factory(_factory)
+        threads = [threading.Thread(target=_get_session_in_thread, args=(index,)) for index in range(N)]
+        for th in threads:
+            th.start()
+        for th in threads:
+            th.join()
+
+        assert len(created) == 1
+        assert all(client is clients[0] for client in clients)
+
 
 class TestOfflineModeSession:
     def test_offline_mode(self):


hanouticelina (commented):
✅

Wauplin (approved):
Good call, thanks ! (though I'm not sure there was a problem before except a tiny performance issue on a dropped client)

PR #4854, opened 2026-09-10
Add `--container-registry-{username,password}`

alvarobartt:
## Description

This PR adds both `--container-registry-username` and `--container-registry-password` as custom containers might live in private registries e.g., a Hugging Face Space or a private AWS ECR, and a `username` and `password` might be required.

> [!NOTE]
> Note that this is just supported for endpoints created with custom containers, and that Inference Endpoints already allows those when creating the endpoint (see screenshot).
> Screenshot 2026-09-10 at 10 13 21

More information about the API spec at https://api.endpoints.huggingface.cloud/#post-/v2/endpoint/-namespace-

---

> [!NOTE]
> **Medium Risk**
> Registry passwords are accepted and forwarded in endpoint creation payloads; misuse or logging could expose secrets, though scope is limited to custom-container deploy with explicit validation.
> 
> **Overview**
> Adds **private registry authentication** when deploying Inference Endpoints with a **plain custom container** (`custom_image` with a top-level `url`, not managed `--engine` images).
> 
> `create_inference_endpoint` and `hf endpoints deploy` accept optional `container_registry_username` / `container_registry_password` (CLI: `--container-registry-username`, `--container-registry-password`). `_build_endpoint_image_payload` nests them under `model.image.custom.credentials` in the API payload; password is omitted when not provided. **Validation** rejects password without username, credentials without `custom_image`, and credentials combined with engine-keyed images or `--engine`.
> 
> Docs and CLI reference include Python and shell examples. Tests cover payload shaping and CLI/API error cases. `_xet.py` only picks up minor typing annotations unrelated to endpoints.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61d2f3a88874909c97691678a05e8d9e45664447. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4871, opened 2026-09-10
[CI] De-flake HTTP backoff tests + rerun on more transient hub-ci errors

Wauplin:
## Summary

Two small CI-flakiness fixes that came out of an investigation of the recent red runs on `main` (see the full write-up in the comment below). Neither of these is *the* main cause — most of the noise is on the hub-ci side — but both are cheap and ours to fix.

- **`tests/test_utils_http.py`: stop asserting on wall-clock time.** `test_backoff_on_429_uses_ratelimit_header` measured the real sleep with `time.time()` and asserted `round(t1 - t0, 1) == 2.0`. On a loaded runner it measures `2.1` and fails — that's exactly what killed `build-ubuntu (3.10, with hf_xet)` in #4854. Both this test and `test_backoff_sleep_time` now patch `time.sleep` and assert the durations that `http_backoff` *asked* for, which is what we actually care about. Bonus: the two tests went from ~4s of real sleeping to instant.

- **`--only-rerun`: cover the transient hub-ci errors we actually see.** The regex had `502` and `504` but not `503`, and 503 is by far the most common transient error in failing runs — one run on 2026-09-08 had **7484** occurrences of `Server error '503 Service Temporarily Unavailable' for url '.../api/repos/create'` and 274 distinct failing tests, none of which were retried. Also added `ConnectError` / `RemoteProtocolError`, which is how the Windows runners report `[WinError 10054] An existing connection was forcibly closed by the remote host` (seen in `test_user_overview`, `test_kernel_info`, `test_repo_list`, `test_get_repo_discussion_pagination`, ...).

```diff
-  --only-rerun '(OSError|FileNotFoundError|Timeout|HTTPError.*409|HTTPError.*502|HTTPError.*504|not less than or equal to 0.01)'
+  --only-rerun '(OSError|FileNotFoundError|Timeout|ConnectError|RemoteProtocolError|HTTPError.*409|HTTPError.*502|HTTPError.*503|HTTPError.*504|not less than or equal to 0.01)'
```

I deliberately did *not* add `500` (that would mask real bugs — see the bucket issue below) nor `AssertionError` (way too broad).

## Before / after locally

```
$ pytest tests/test_utils_http.py -k TestHttpBackoff -q
8 passed in 0.61s      # was ~4.5s, and 1-in-N flaky on `assert [2.1] == [2.0]`

$ pytest tests/test_utils_http.py -q
64 passed in 46.59s
```

Opening as draft: the test change is ready, but I'd like a second opinion on how aggressive the rerun regex should get.

---

> [!NOTE]
> **Low Risk**
> Test and CI configuration only; no changes to runtime HTTP or library behavior.
> 
> **Overview**
> Reduces CI flakiness in two places: **pytest reruns** and **`http_backoff` tests**.
> 
> The Ubuntu and Windows workflows widen `--only-rerun` so transient hub-ci failures get retried: **503** responses (in addition to 409/502/504) and **`ConnectError` / `RemoteProtocolError`** (e.g. Windows connection resets). Production library behavior is unchanged.
> 
> In `tests/test_utils_http.py`, exponential backoff and 429 ratelimit tests no longer measure real wall-clock delays. A **`mock_sleep`** fixture patches `huggingface_hub.utils._http.time.sleep`, and assertions check the **sleep durations `http_backoff` requests** (including the 2.0s wait for ratelimit header + margin), which removes timing flakes and avoids ~4s of real sleeping in those tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ec100245a97d33ba135500e61903212b97ec3be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -112,7 +112,7 @@ jobs:
         working-directory: ./src # For code coverage to work
         run: |
           source ../.venv/bin/activate
-          PYTEST="python -m pytest --cov=./huggingface_hub --cov-report=xml:../coverage.xml --vcr-record=none --reruns 8 --reruns-delay 2 --only-rerun '(OSError|FileNotFoundError|Timeout|HTTPError.*409|HTTPError.*502|HTTPError.*504|not less than or equal to 0.01)'"
+          PYTEST="python -m pytest --cov=./huggingface_hub --cov-report=xml:../coverage.xml --vcr-record=none --reruns 8 --reruns-delay 2 --only-rerun '(OSError|FileNotFoundError|Timeout|ConnectError|RemoteProtocolError|HTTPError.*409|HTTPError.*502|HTTPError.*503|HTTPError.*504|not less than or equal to 0.01)'"
 
           case "${{ matrix.test_name }}" in
 
@@ -213,7 +213,7 @@ jobs:
             "--vcr-record=none",
             "--reruns", "8",
             "--reruns-delay", "2",
-            "--only-rerun", "(OSError|FileNotFoundError|Timeout|HTTPError.*409|HTTPError.*502|HTTPError.*504|not less than or equal to 0.01)",
+            "--only-rerun", "(OSError|FileNotFoundError|Timeout|ConnectError|RemoteProtocolError|HTTPError.*409|HTTPError.*502|HTTPError.*503|HTTPError.*504|not less than or equal to 0.01)",
             "../tests"
           )
 
--- a/tests/test_utils_http.py
+++ b/tests/test_utils_http.py
@@ -51,6 +51,12 @@ class TestHttpBackoff:
         yield
         self.patcher.stop()
 
+    @pytest.fixture
+    def mock_sleep(self) -> Generator[Mock, None, None]:
+        """Patch out `time.sleep` so backoff durations are asserted instead of actually waited for."""
+        with patch("huggingface_hub.utils._http.time.sleep") as mock:
+            yield mock
+
     def test_backoff_no_errors(self) -> None:
         """Test normal usage of `http_backoff`."""
         data_mock = Mock()
@@ -135,26 +141,9 @@ class TestHttpBackoff:
         assert self.mock_request.call_count == 4
         assert response is mock_200
 
-    def test_backoff_sleep_time(self) -> None:
-        """Test `http_backoff` sleep time goes exponential until max limit.
-
-        Since timing between 2 requests is sleep duration + some other stuff, this test
-        can be unstable. However, sleep durations between 10ms and 50ms should be enough
-        to make the approximation that measured durations are the "sleep time" waited by
-        `http_backoff`. If this is not the case, just increase `base_wait_time`,
-        `max_wait_time` and `expected_sleep_times` with bigger values.
-        """
-        sleep_times = []
-
-        def _side_effect_timer() -> Generator[ConnectTimeout, None, None]:
-            t0 = time.time()
-            while True:
-                yield ConnectTimeout("Connection timeout")
-                t1 = time.time()
-                sleep_times.append(round(t1 - t0, 1))
-                t0 = t1
-
-        self.mock_request.side_effect = _side_effect_timer()
+    def test_backoff_sleep_time(self, mock_sleep: Mock) -> None:
+        """Test `http_backoff` sleep time goes exponential until max limit."""
+        self.mock_request.side_effect = ConnectTimeout("Connection timeout")
 
         with pytest.raises(ConnectTimeout):
             http_backoff("GET", URL, base_wait_time=0.1, max_wait_time=0.5, max_retries=5)
@@ -162,34 +151,23 @@ class TestHttpBackoff:
         assert self.mock_request.call_count == 6
 
         # Assert sleep times are exponential until plateau
-        expected_sleep_times = [0.1, 0.2, 0.4, 0.5, 0.5]
-        assert sleep_times == expected_sleep_times
+        assert mock_sleep.call_args_list == [call(0.1), call(0.2), call(0.4), call(0.5), call(0.5)]
 
-    def test_backoff_on_429_uses_ratelimit_header(self) -> None:
+    def test_backoff_on_429_uses_ratelimit_header(self, mock_sleep: Mock) -> None:
         """Test that 429 wait time uses full reset time from ratelimit header."""
-        sleep_times = []
-
-        def _side_effect_timer() -> Generator:
-            t0 = time.time()
-            mock_429 = Mock()
-            mock_429.status_code = 429
-            mock_429.headers = {"ratelimit": '"api";r=0;t=1'}  # Server says wait 1s
-            yield mock_429
-            t1 = time.time()
-            sleep_times.append(round(t1 - t0, 1))
-            t0 = t1
-            mock_200 = Mock()
-            mock_200.status_code = 200
-            yield mock_200
-
-        self.mock_request.side_effect = _side_effect_timer()
+        mock_429 = Mock()
+        mock_429.status_code = 429
+        mock_429.headers = {"ratelimit": '"api";r=0;t=1'}  # Server says wait 1s
+        mock_200 = Mock()
+        mock_200.status_code = 200
+        self.mock_request.side_effect = (mock_429, mock_200)
 
         response = http_backoff(
             "GET", URL, base_wait_time=0.1, max_wait_time=0.5, max_retries=3, retry_on_status_codes=429
         )
 
         assert self.mock_request.call_count == 2
-        assert sleep_times == [2.0]
+        assert mock_sleep.call_args_list == [call(2.0)]  # 1s from the header + 1s margin
         assert response.status_code == 200
 
 


Wauplin:
## CI flakiness investigation (started from #4854)

None of the failures on #4854 were caused by that PR. Summary of what I found, split by who owns the fix.

---

### 1. `hf_xet` jobs: every non-recursive bucket listing 500s on hub-ci — **hub-side, deterministic**

Run [34459590774](https://github.com/huggingface/huggingface_hub/actions/runs/34459590774) failed on exactly the 4 `with hf_xet` jobs (ubuntu 3.10/3.14, windows 3.10/3.14), 40 failures each, all the same root cause:

```
FAILED tests/test_buckets_hf_file_system.py::TestHfFileSystemBucketRO::test_read_text
  huggingface_hub.errors.HfHubHTTPError: Server error '500 Internal Server Error' for url
  'https://hub-ci.huggingface.co/api/buckets/__DUMMY_HUB_USER__/repo-.../tree/data?recursive=false'

  Hint matched multiple indexes, must hint by index name. Matched:
    kp: { bucketId: 1, parent: 1, path: 1 } name: 'bucket_parent_path'
    kp: { bucketId: 1, parent: 1, path: 1 } unique name: 'bucket_parent_path_unique'
```

…alternating randomly with:

```
  hint provided does not correspond to an existing index
```

Affected: all of `test_buckets_hf_file_system.py`, most of `test_buckets.py::test_copy_files_*`, `test_copy_files.py`, `test_buckets_cli.py::test_list_files_*`. Only the `with hf_xet` jobs because bucket tests are `@pytest.mark.xet`.

**Still broken right now.** Repro against hub-ci (fresh, empty bucket, 8 calls each):

| request | result |
|---|---|
| `GET /api/buckets/{id}/tree?recursive=false` | 5× 500 *"Hint matched multiple indexes"*, 3× 500 *"hint provided does not correspond to an existing index"* |
| `GET /api/buckets/{id}/tree?recursive=false&sort=uploadedAt` | 7× 500 *"…does not correspond to an existing index"*, 1× 504 |
| `GET /api/buckets/{id}/tree?recursive=true` | 7× 200, 1× 504 |

**Production is fine** — same two requests against `huggingface.co` on one of my own buckets return 200. So this is a hub-ci **database index state** problem, not a code bug in the listing itself.

**Why it started now.** `moon-landing` `b0830ef4d6` — *"refactor(buckets): drop the `bucketNonRecursiveListingWithDirectories` branching"* (#19826, 2026-09-09) — removed the feature flag that used to gate the hinted `bucket_directories` + `bucket_files` query path. The flag was `true` in prod since June, so prod has been on that path for months; on hub-ci the flag was presumably never set, so the hinted queries were never exercised there until the flag was deleted. First red huggingface_hub run right after: 2026-09-09 13:00 UTC.

**Where.** `server/app/bucketRoutes.ts` builds the hint as a **key pattern**, not a name:

```ts
// ~line 1404
const indexHint = sortByUploadedAt
    ? { bucketId: 1, parent: 1, uploadedAt: -1, path: 1 }
    : { bucketId: 1, parent: 1, path: 1 };
```

and passes it at lines 1468 (`bucket_directories`) and 1490 (`bucket_files`), both inside the same `Promise.all` — which is why the two error messages alternate: whichever query rejects first wins.

**What I'd ask the moon-landing / infra folks:**

1. Run `db.bucket_directories.getIndexes()` and `db.bucket_files.getIndexes()` on hub-ci's bucket cluster. Expected from `server/models/`:
   - `bucket_files` → `bucket_parent_path` (`{bucketId:1,parent:1,path:1}`) + `bucket_parent_uploadedAt_path`
   - `bucket_directories` → `bucket_parent_path_unique` (same key pattern, `unique`) + `bucket_path` + `bucketId_parent_hashed` + `bucket_parent_uploadedAt_path`

   The errors say one collection has **both** `bucket_parent_path` and `bucket_parent_path_unique`, and the other is **missing** its index entirely (including `bucket_parent_uploadedAt_path`, given the `sort=uploadedAt` result). Both need reconciling.

2. Consider why the drift happened without anyone noticing. `Database.ts` (`attach()`) does one `collection.createIndexes(indexesToCreate)` per collection and swallows any failure into `Pulse.alerts("db.createindex", …)`. Because it's a single batch call, one conflicting index aborts the whole array, so every index *after* the conflicting one silently never gets created — which matches `bucket_parent_uploadedAt_path` also being absent. Creating them individually, or at least surfacing the failure louder, would have caught this.

3. Defensive fix in the query itself: hint by **index name** rather than key pattern, the way line 1587 already does (`hint: "bucket_path_unique"`). That turns a hard 500 into a normal query even if a duplicate index shows up. Something like `hint: sortByUploadedAt ? "bucket_parent_uploadedAt_path" : {directories: "bucket_parent_path_unique", files: "bucket_parent_path"}`. (Named hints still 500 if the index is missing, so this is a complement to (1), not a replacement.)

---

### 2. hub-ci is badly overloaded — **infra-side, and this is the big one**

This is the dominant cause of "flaky lately", far more than any individual test. Measured from my laptop today:

**`POST /api/repos/create` — the single most-used endpoint in the test suite:**

```
15 sequential calls: 8 × 200, 7 × ReadTimeout (30s)
latencies (s): 2.53 2.92 3.26 3.68 3.88 11.11 15.46 18.54 30.27 30.28 30.29 30.30 30.30 30.34 30.38
```

~47% failure rate, and successful calls take 2.5–18s. A bit later the same day it was fully down for creates — 10 consecutive attempts failed (3 × `502 Bad Gateway`, then 7 × `503 Service Temporarily Unavailable`).

**`GET /api/whoami-v2` — a trivial auth check:**

```
30 sequential calls: 29 × 200, 1 × ReadTimeout
latencies (s): p50 ≈ 2.5, p90 ≈ 7.3, max 20.3
```

Multi-second p50 on `whoami` is the tell — this looks like saturation, not a slow endpoint.

**In CI logs.** Run [34243312023](https://github.com/huggingface/huggingface_hub/actions/runs/34243312023) (2026-09-08): **274 distinct failing tests**, with **7484** occurrences of `Server error '503 Service Temporarily Unavailable' for url 'https://hub-ci.huggingface.co/api/repos/create'` and 149 × `502`. Run 34244174970 the same afternoon: 255 failing tests, same signature. Those two runs alone account for most of the red on `main` last week.

Secondary effects visible in the logs, all consistent with an overloaded backend:
- `504 The request is taking longer than expected, please try again later.` on bucket listings
- Windows runners getting `httpx.ConnectError: [WinError 10054] An existing connection was forcibly closed by the remote host`, and `json.decoder.JSONDecodeError: Expecting value: line 1 column 1` (= empty body, connection dropped mid-response)
- `with hf_xet` jobs now take 28–31 minutes

**Ask:** can hub-ci get more headroom (or its own capacity/rate budget)? Right now a single `hf` CI run does thousands of repo create/delete cycles across 10 parallel jobs × 4 xdist workers, against an instance that can't reliably serve one create in 30s.

---

### 3. Read-after-write inconsistency on hub-ci — **hub-side, worth a conversation**

A steady family of assertion failures across almost every run, all "wrote, immediately read back, got the old value":

| test | failure |
|---|---|
| `test_repocard.py::TestRepocardMetadataUpdate::*` (many) | `assert {'license': 'cc-by-sa-4.0'} == {'license': …, 'tag': 'test'}`, `assert None == '__DUMMY_HUB_USER__/…'` |
| `test_cache_layout.py::TestReferenceUpdates::test_update_reference` | `assert '3327fc92…' != '3327fc92…'` (new commit, same sha returned) |
| `test_hf_api.py::TestHfApiPrivate::test_list_private_models` | `assert False` (just-created private repo missing from listing) |
| `test_file_download.py::TestHfHubDownloadToLocalDir::*` | `Expected '_download_to_tmp_and_move' to have been called once. Called 0 times`, `assert 'wrong_content' == 'content'` (stale etag) |
| `test_commit_scheduler.py::test_sync_and_squash_history` | `assert 'initial commit' == 'Super-squash…'` |
| `test_hf_api.py::TestCommitInBackground` | `assert 5 == 4` |
| `test_hf_file_system.py::TestHfFileSystemRepositoryRO::test_url` | `…/resolve/main/data` vs `…/tree/main/data` (`isdir()` saw a stale listing) |
| `test_buckets_cli.py::test_bucket_ls_alias`, `test_bucket_list_table` | `assert 1 == 0` |

For the bucket ones specifically there's a concrete explanation in `moon-landing/server/models/Database.ts`:

```ts
/// Use nearest for bucket operations - lowest latency, any member
export const BUCKET_READ_PREFERENCE: ReadPreferenceLike = "nearest";
export const BUCKET_READ_CONCERN = { level: "local" } as const;
```

`nearest` + `readConcern: local` means a bucket listing can be served by any replica-set member with no consistency guarantee, so upload-then-list is expected to occasionally return stale data. That's a deliberate latency tradeoff and probably fine for prod UI, but worth confirming it's intended for the API too — users doing `upload → list` in a script will hit it. On hub-ci with a small/lagging replica set it fires often enough to be a constant test flake.

The non-bucket ones (repocard, refs, private-model listing) point at the same thing on the main cluster. **Ask:** is hub-ci's replica lag monitored, and is it materially worse than prod's?

---

### 4. `400` on `.../info/lfs/objects/batch` for xet-enabled repos — **needs a hub-side look, unconfirmed**

Two tests fail together, in all 4 `with hf_xet` jobs, in 3 of the 15 runs I surveyed (34120723174, 34128275870, 34133937534):

```
FAILED tests/test_xet_upload.py::TestXetUpload::test_upload_file_with_bytesio
FAILED tests/test_xet_upload.py::TestXetUpload::test_fallback_to_lfs_when_xet_not_available
  httpx.HTTPStatusError: Client error '400 Bad Request' for url
  'https://hub-ci.huggingface.co/__DUMMY_HUB_USER__/model-…git/info/lfs/objects/batch'
  → huggingface_hub.errors.BadRequestError: (Request ID: …)
```

These are exactly the two tests that deliberately take the **LFS** path on a **xet-enabled** repo (BytesIO isn't xet-uploadable, so it falls back). The 400 comes back with an empty error body, which is why our exception message is just a request ID. The same signature also shows up as `BadRequestError: Bad request for commit endpoint:` in `test_hub_mixin_pytorch::test_push_to_hub`, `test_commit_scheduler::test_sync_local_folder`, `test_hf_api::TestCommitApi::test_pre_upload_before_commit`.

I could not reproduce it today — hub-ci wouldn't reliably create a repo (see §2); the one attempt that got through succeeded. **Ask:** could someone grep the hub-ci LFS-batch logs for one of those request IDs and see what the 400 actually is? And regardless — that endpoint returning a 400 with an empty body makes it undebuggable from the client side.

---

### 5. Two things on our side — fixed in this PR

- **`test_backoff_on_429_uses_ratelimit_header`** was asserting on wall-clock time (`round(t1 - t0, 1) == 2.0`) and measured `2.1` on a loaded runner. That's the *only* failure in #4854 that wasn't the bucket bug. Now patches `time.sleep` and asserts the requested durations.
- **`--only-rerun` didn't include `503`** — the single most common transient error in every red run (7484 hits in one run alone), while `502`/`504` were both already listed. Added `503` plus `ConnectError`/`RemoteProtocolError` for the Windows connection resets.

I did *not* add `500` (that would have hidden the bucket bug above) or `AssertionError` (too broad — and §3 shouldn't be papered over with retries).

Wauplin:
Small validation datapoint from this PR's own CI, which is red for exactly the reason described in §2 above.

`build-ubuntu (3.10, Python 3.10, torch_1.11)`:

```
= 1 failed, 45 passed, 2 warnings, 8 rerun in 239.81s (0:03:59) =
FAILED tests/test_hub_mixin_pytorch.py::TestPytorchHubMixin::test_push_to_hub
  huggingface_hub.errors.HfHubHTTPError: Server error '503 Service Temporarily Unavailable'
  for url 'https://hub-ci.huggingface.co/api/repos/create'
```

The `8 rerun` confirms the new `HTTPError.*503` in the regex is doing its job — on `main` that test would have failed immediately with 0 reruns. It still ended up red because hub-ci returned 503 for all 9 attempts spanning ~2m13s (15:26:48 → 15:29:01 UTC).

So: worth keeping, but no rerun budget fixes a multi-minute outage of `/api/repos/create`. I checked whether bumping `--reruns-delay` (currently 2s) would help and it wouldn't — the retry window was already ~2 minutes here because the failing requests were themselves timing out. §2 is the actual fix.

PR #4882, opened 2026-09-11
[CLI] Fix card data serialization

davanstrien:
## Summary

`hf models list` and `hf models info` currently serialize card metadata as an empty object. For `info`, this happens both in the default JSON output and when `cardData` is requested explicitly with `--expand`:

```console
$ hf models info microsoft/Phi-4-mini-instruct --expand cardData --json
{"id": "microsoft/Phi-4-mini-instruct", "card_data": {}}
```

The Hub response and Python API both contain the metadata. It gets lost in the CLI because `CardData` is decorated as a dataclass but stores its metadata dynamically in `__dict__`; `dataclasses.asdict()` therefore sees no declared fields and serializes it as `{}`. The same serialization path affects dataset and Space list/info output.

This patch keeps the existing `asdict()` behavior for the rest of each API object, then restores `CardData` using its existing `to_dict()` representation. `to_dict()` matches the public `cardData` shape returned by the Hub, including canonical keys such as `model-index` rather than the internal `eval_results` objects.

After the change, for example:

```console
$ hf models list --filter nl --apps vllm --sort downloads --limit 1 \
    --expand cardData,downloads --json \
  | jq '.[0] | {id, downloads, nl: (.card_data.language | contains(["nl"])), license: .card_data.license}'
{
  "id": "microsoft/Phi-4-mini-instruct",
  "downloads": 454378,
  "nl": true,
  "license": "mit"
}
```

I have kept this initial draft code-only and haven't included a regression test yet. Happy to add one once the approach looks right.

Related but out of scope: the same zero-field dataclass setup means any two `ModelCardData` instances compare equal through the generated `__eq__`. Happy to open a separate issue for that if useful.

Happy to convert this to an issue if that is preferred; a draft PR seemed the clearest way to show the problem and fix together.

---

> [!NOTE]
> **Low Risk**
> Narrow change to CLI JSON serialization for one dataclass type, with no auth, network, or persistence impact.
> 
> **Overview**
> **Fixes empty `card_data` in CLI JSON** for commands like `hf models list/info` (and the same path for datasets/spaces) when `--expand cardData` or default serialization runs.
> 
> `_dataclass_to_dict` still builds the payload with `dataclasses.asdict()`, but for any field whose value is a **`CardData`** instance it **replaces** that entry with **`CardData.to_dict()`**, so Hub-shaped keys (e.g. `model-index`, `license`, `language`) appear instead of `{}`.
> 
> The change is limited to CLI output serialization in `_output.py`; no API or Hub client behavior changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ea41b5581ba38e3680b510e3b87fd217d5dcb64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davanstrien:
cc @Wauplin @hanouticelina. Fine if you prefer to just take it as an issue and close the PR!

hanouticelina (approved):
reproduced the bug 👍 thank you!

PR #4884, opened 2026-09-11
[Core] Reject path traversal via embedded ".." segments in path_in_repo

bodapatisaikrishna:
## Summary

`_validate_path_in_repo()` (used by every `CommitOperationAdd`/`Copy`/`Delete`, so
`upload_file`, `upload_folder`, `create_commit`, `delete_file`, `delete_folder`, ...)
only rejects a `..` segment when it's the *first* path component:

```python
if path_in_repo == "." or path_in_repo == ".." or path_in_repo.startswith("../"):
    raise ValueError(...)
```

A `..` later in the path still escapes the repo root once resolved, and slips through
unchanged:

```python
>>> from huggingface_hub._commit_api import _validate_path_in_repo
>>> _validate_path_in_repo("a/../../etc/passwd")
'a/../../etc/passwd'   # not rejected, even though it resolves outside the repo root
>>> _validate_path_in_repo("a/b/../../../etc/passwd")
'a/b/../../../etc/passwd'   # same
```

This fixes it by rejecting any `..` segment anywhere in the path, not just a leading
one. A `..` token only tells you where the path lands once you resolve it, so the only
unambiguously safe answer is to reject it outright:

```python
if path_in_repo == "." or any(part == ".." for part in path_in_repo.split("/")):
    raise ValueError(...)
```

The single check subsumes the previous `path_in_repo == ".."` and
`path_in_repo.startswith("../")` cases, so those are gone.

This is stricter than the current behavior in one visible way: `"a/../file.txt"`
resolves back *inside* the repo, so it used to be accepted unchanged, and is now
rejected. That's intentional here  the point is that no `..` token survives to be
resolved, rather than trying to decide case by case whether each one escapes.

I don't know how the server itself handles a raw `..` in a commit path, so I can't say
whether this is exploitable end to end today  this closes a gap in the client side
guard either way, and the function's own comment says its purpose is exactly this
("prevent a server-side issue").

## Test plan

Added cases to `TestCommitOperationPathInRepo` in `tests/test_commit_api.py`:
`invalid_values` gets `"a/../file.txt"`, `"a/b/../../c/file.txt"`, `"a/../../file.txt"`
and `"a/b/../../../file.txt"`, covering both `..` that escapes the repo root and `..`
that resolves back inside it.

```
$ python -m pytest tests/test_commit_api.py tests/test_utils_paths.py -q
28 passed in 0.21s
```

`ruff check`/`ruff format`/`ty check` clean on the changed files.

Found this by reading `_commit_api.py` and testing `_validate_path_in_repo` directly
with a few traversal patterns, not from a reported issue.

---

> [!NOTE]
> **Low Risk**
> Narrow security hardening in path validation; behavior change only for paths containing `..` that were previously accepted (e.g. `a/../file.txt`).
> 
> **Overview**
> Tightens client side validation of `path_in_repo` for all commit operations (`CommitOperationAdd`/`Copy`/`Delete`, and thus upload/delete/commit APIs).
> 
> **`_validate_path_in_repo`** no longer only blocks `.`, `..`, and paths starting with `../`. It now rejects **any path segment equal to `..`**, so embedded traversal like `a/../../etc/passwd` or `a/../file.txt` raises `ValueError` before a commit is built.
> 
> That is stricter than before: paths such as `a/../file.txt` that would resolve inside the repo are rejected too, on the principle that `..` should not appear in commit paths at all.
> 
> Tests in `TestCommitOperationPathInRepo` add several embedded-`..` cases to `invalid_values`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 951d83ec543ca97eb0762680f0d84c84f6a47c07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/huggingface_hub/_commit_api.py
+++ b/src/huggingface_hub/_commit_api.py
@@ -315,6 +315,19 @@ def _validate_path_in_repo(path_in_repo: str) -> str:
         raise ValueError(f"Invalid `path_in_repo` in CommitOperation: '{path_in_repo}'")
     if path_in_repo.startswith("./"):
         path_in_repo = path_in_repo[2:]
+
+    # The checks above only catch a ".." segment at the very start of the path. A
+    # path like "a/../../etc/passwd" doesn't start with "../" but still escapes
+    # the repo root once its ".." segments are resolved, so walk the full path and
+    # track depth relative to the repo root to catch that case too.
+    depth = 0
+    for part in path_in_repo.split("/"):
+        if part in ("", "."):
+            continue
+        depth += -1 if part == ".." else 1
+        if depth < 0:
+            raise ValueError(f"Invalid `path_in_repo` in CommitOperation: '{path_in_repo}'")
+
     for forbidden in FORBIDDEN_FOLDERS:
         if any(part == forbidden for part in path_in_repo.split("/")):
             raise ValueError(
--- a/tests/test_commit_api.py
+++ b/tests/test_commit_api.py
@@ -38,8 +38,19 @@ class TestCommitOperationPathInRepo:
         ".file.txt": ".file.txt",
         "/file.txt": "file.txt",
         "./file.txt": "file.txt",
+        # ".." segments that resolve back into the repo (never escape the root) are fine
+        "a/../file.txt": "a/../file.txt",
+        "a/b/../../c/file.txt": "a/b/../../c/file.txt",
     }
-    invalid_values = [".", "..", "../file.txt"]
+    invalid_values = [
+        ".",
+        "..",
+        "../file.txt",
+        # a ".." segment anywhere in the path can still escape the repo root, not just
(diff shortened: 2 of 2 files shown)


Wauplin (commented):
FWIW I don't think `"a/../file.txt": "a/../file.txt",` should be considered as fine

bodapatisaikrishna:
Fair point , pushed 2c65555 dropping the depth tracking approach entirely and rejecting any \`..\` segment outright, wherever it appears, whether or not it would resolve back inside the repo root. Moved \`"a/../file.txt"\` and \`"a/b/../../c/file.txt"\` from the valid to the invalid test cases.

Wauplin (approved):
LGTM

PR #4892, opened 2026-09-13
[Core] Restrict repo_id validation regex to ASCII word characters

bodapatisaikrishna:
validate_repo_id()'s docstring documents the allowed charset as ASCII alphanumeric plus "-", "_", "." (matching the existing all ASCII VALID_VALUES test), but REPO_ID_REGEX used bare \w, which Python matches against the full Unicode word-character set by default. "café" and "文件夹/bar" both passed validation unchanged.

Add re.ASCII to the compiled regex's flags so \w/\b only match ASCII, matching the documented rules. No other behavior change.

---

> [!NOTE]
> **Low Risk**
> Stricter validation may reject previously accepted non ASCII repo IDs, but those were already invalid on the Hub and documented as ASCII only.
> 
> **Overview**
> Tightens client-side `repo_id` validation so it matches the documented **ASCII only** charset (alphanumeric plus `-`, `_`, `.`), aligning with Hub rules.
> 
> `REPO_ID_REGEX` now compiles with `re.ASCII`, so `\w` and `\b` no longer treat Unicode strings like `café` or `文件夹/bar` as valid. Tests add those cases to the invalid `repo_id` list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a6af5c11739e54dd75ad1569e2b1d3c5732a3ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Wauplin (approved):
Correct, thanks for flagging

PR #4896, opened 2026-09-13
[Download] Fix silently disabled tree cache on long Windows paths

### LINKED ISSUE #4895: [Windows] Tree cache disabled by long local_dir paths
### Describe the bug

On Windows with the default `LongPathsEnabled=0`, the on-disk tree cache (`_tree_cache.py`) is silently non-functional for a deep `local_dir` or `cache_dir`. `_tree_cache_path()` joins `<folder>/trees/<commit_hash>.json` without the `\\?\` extended-length prefix that every sibling path in the download stack already applies (`_local_folder.py`, `_snapshot_download.py`, `file_download.py`). When the resulting path exceeds 255 characters, `write_tree_cache` hits `WinError 206`, swallows it into the "Ignored error while writing tree cache file" warning, and the next `read_tree_cache` misses. Net effect: every `snapshot_download` into that directory re-fetches the tree listing. No crash, just the cache never working there. It is the perf-only sibling of the crash fixed in #4546.

I have a fix ready on a branch (prefix applied once in `_tree_cache_path()`, which both the reader and the writer go through, with `\\?\UNC\` handling for network shares and two Windows-only regression tests) and will open a PR referencing this issue.

### Reproduction

```python
import logging
import os
import tempfile

from huggingface_hub._tree_cache import (
    _IN_MEMORY_TREE_CACHE,
    TreeCacheEntry,
    read_tree_cache,
    tree_cache_folder_for_local_dir,
    write_tree_cache,
)

logging.basicConfig(level=logging.WARNING, format="%(levelname)s %(name)s: %(message)s")
base = tempfile.mkdtemp()
local_dir = os.path.join(base, "d" * (240 - len(base)))  # a deep download target, ~240 chars
os.makedirs("\\\\?\\" + os.path.abspath(local_dir), exist_ok=True)
folder = tree_cache_folder_for_local_dir(local_dir)
commit = "0123456789abcdef0123456789abcdef01234567"
entries = {"config.json": TreeCacheEntry(size=5, blob_id="blob-config")}

write_tree_cache(folder, commit, entries)
_IN_MEMORY_TREE_CACHE.clear()  # force the read to go to disk, as a fresh process would
cache_file = "\\\\?\\" + os.path.join(os.path.abspath(folder), "trees", commit + ".json")
print("cache file exists:", os.path.exists(cache_file))
print("read back:", read_tree_cache(folder, commit))
```

Expected: `cache file exists: True` and the entries read back. Actual on `main` (129bbb5c):

### Logs

```shell
WARNING huggingface_hub._tree_cache: Ignored error while writing tree cache file C:\Users\me\AppData\Local\Temp\tmpXXXXXX\ddd...ddd\.cache\huggingface\trees\0123456789abcdef0123456789abcdef01234567.json: [WinError 206] The filename or extension is too long: 'C:\\Users\\me\\AppData\\Local\\Temp\\tmpXXXXXX\\ddd...ddd\\.cache\\huggingface'
cache file exists: False
read back: None
```

With the prefix applied in `_tree_cache_path()`:

```shell
cache file exists: True
read back: {'config.json': TreeCacheEntry(size=5, blob_id='blob-config', lfs_sha256=None, lfs_size=None, xet_hash=None)}
```

### System info

```shell
- huggingface_hub version: main @ 129bbb5c (source checkout)
- Platform: Windows-10-10.0.26200-SP0 (Windows 11 Pro)
- Python version: 3.11.9
- LongPathsEnabled: not set (default 0)
```

askalf:
Fixes #4895.

## Summary

- On Windows, `_tree_cache_path()` now applies the extended-length `\\?\` prefix, so the tree cache is actually written and read under a deep `local_dir` or `cache_dir` instead of failing silently (swallowed `WinError 206`) and re-fetching the tree listing on every download. Perf-only sibling of #4546.
- That prefixing was already inlined 6 times across the download stack, so rather than add a 7th copy this PR moves it to a single `utils/_paths.as_extended_path()` helper and reuses it everywhere. Net **−110 / +56** lines.
- Long UNC paths are encoded as `\\?\UNC\server\share\...` (prefixing a UNC path verbatim gives the invalid `\\?\\\server\...`). The inlined copies all got this wrong, so they're fixed too.
- One Windows-only regression test: a write + read round-trip of the tree cache under a deep `local_dir`.

Scope of the bug: `snapshot_download` is the only writer, and it passes the raw `local_dir` / `storage_folder`. `hf_hub_download(local_dir=...)` was *not* affected — `_hf_hub_download_to_local_dir` already prefixes `local_dir` before deriving the tree cache folder, so its folder came out prefixed by accident.

Repro from the issue, on `main`:

```console
$ python repro.py
WARNING huggingface_hub._tree_cache: Ignored error while writing tree cache file C:\Users\me\...\ddd...ddd\.cache\huggingface\trees\0123456789abcdef0123456789abcdef01234567.json: [WinError 206] The filename or extension is too long: ...
cache file exists: False
read back: None
```

Same script on this branch:

```console
$ python repro.py
cache file exists: True
read back: {'config.json': TreeCacheEntry(size=5, blob_id='blob-config', lfs_sha256=None, lfs_size=None, xet_hash=None)}
```

## Decisions

- **One choke point, not per call site.** Both `read_tree_cache` and `write_tree_cache` go through `_tree_cache_path()`, so prefixing there keeps the on-disk path and the in-memory cache key in agreement, and the next caller can't forget.

- **One helper for the whole download stack.** `as_extended_path(path, max_length=255)` lives in `utils/_paths.py` and does the `os.name == "nt"` check, the abspath, the already-prefixed guard and the UNC form in one place. Call sites: `_tree_cache._tree_cache_path`, `_snapshot_download._local_file_exists`, `file_download` (lock path, blob path, `local_dir`), `_local_folder` (`incomplete_path`, `get_local_download_paths`, `get_local_upload_paths`, `_huggingface_dir`). `max_length` defaults to 255 (the file limit) and is passed as 247 (the directory limit, i.e. MAX_PATH minus room for an 8.3 name) where the caller creates the parent directories right after.

- **Two behavior deltas beyond the bug fix**, both in `_local_folder.py`:
  - `get_local_{download,upload}_paths` used an all-or-nothing rule ("prefix all three paths iff `lock_path` > 255"). It's now decided per path at the 247 directory limit. That closes a latent gap: a `local_dir` of 248–255 chars with a short filename left `file_path` unprefixed and its `mkdir` would fail.
  - `incomplete_path` now uses `abspath` instead of `resolve()` (no symlink resolution), matching every other call site.

- **One test, not a parametrized pair.** The round-trip under a deep `local_dir` is the case the issue is about; the docstring is just a link back to #4895.

No breaking changes. Short and POSIX paths are byte-for-byte unchanged.

## Testing

Author, before the shared-helper commit (Windows 11, Python 3.11.9, `LongPathsEnabled` unset):

```console
$ python -m pytest tests/test_tree_cache.py -k deep_folder -q      # base code + the new test
====================== 2 failed, 19 deselected in 0.37s ======================
$ python -m pytest tests/test_tree_cache.py -q                     # with the fix
============================= 22 passed in 0.34s ==============================
```

The shared-helper commit was written on Linux, so its Windows behavior has not been executed on a Windows box yet — `build-windows` on this PR is the check that matters. Locally: `tests/test_tree_cache.py`, `tests/test_local_folder.py` and `tests/test_utils_paths.py` pass, `make style` / `make quality` are clean, and the path-string logic (drive, UNC, already-prefixed, short, 247-vs-255 boundaries) was verified by driving `as_extended_path` through `ntpath`. Smoke test: `snapshot_download` writes the tree file and a second run with `HfApi.list_repo_tree` stubbed to raise is served entirely from disk.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and `make quality` were executed on my Windows machine and in this branch's CI. I'm responsible for the change and will handle review feedback.


Wauplin (approved):
Thank you! I took the liberty to update a bit the PR. Will merge it now :)

